Issue #17, opened 2019-02-06
Use rollout informer instead of API list


jessesuen:
In service.go:
```golang
func (c *Controller) getRolloutsForService(service *corev1.Service) ([]*v1alpha1.Rollout, error) {
	allROs, err := c.rolloutsclientset.ArgoprojV1alpha1().Rollouts(service.Namespace).List(metav1.ListOptions{})
	if err != nil {
		return nil, err
	}
```

We should use the informer instead of list API

dthomson25:
With this issue, we should investigate the code base and find other places that we can switch to informers in addition to the above example

dthomson25:
After investigating the code, the only place we use the API instead of the rollout informer is at https://github.com/argoproj/argo-rollouts/blob/master/controller/replicaset.go#L35.  In this use, we are required to use the API to get up to date information.

PR #45, opened 2019-03-25
Remove service informer

dthomson25:


dthomson25:
Address issues #45 by removing the Service informer and #17 by removing a call to the API server instead of the informer

Issue #32, opened 2019-03-04
Improve the kubernetes events emitted


dthomson25:
Argo Rollouts can do a better job of emitting Kubernetes events on important changes in the system.  For example, Argo Rollouts should emit an event when either the active or preview service is modified or when the stable replicaset is changed in a canary strategy.

dthomson25:
The events were made more clear with https://github.com/argoproj/argo-rollouts/pull/61

Issue #29, opened 2019-02-24
Expose Prometheus metrics on rollouts 


dthomson25:
Argo Rollouts should expose Prometheus metrics in order to provide better overall insight into the rollouts in a cluster. Some initial metrics to include could be:

- Number of rollouts and which strategies they are using
- When a rollout starts progressing from a new pod spec
- When a rollout finishes progressing from a new pod spec
- Every time the rollout increments a step

As a reference implement, we can probably use https://github.com/argoproj/argo/commit/b9cffe9cd7b347905a42cf3e217cc3b039bdfb3f as a baseline

dthomson25:
The metrics were improved with https://github.com/argoproj/argo-rollouts/pull/61

Issue #41, opened 2019-03-16
Rework Rollout Conditions


dthomson25:
Currently, there are `invalidSpec` and `available` conditions. The `invalidSpec` condition highlights invalid yaml, but the `available` condition needs to be fleshed out more. The `available` condition is not used for the Canary strategy and its use in the BlueGreen strategy does not indicate the status of a rollout well. Instead, the condition indicates that the active service is pointing at a replicaset and that replicaset is at the desired number of replicas.  As a result, the rollout could have an `available` condition set to true while the controller is still working on the rollout.

With the work required to rework the `available` condition, we will need to make sure that we are able to answer the following questions. What does it mean for a rollout to be available? Does that mean the rollout is in a steady-state? Does that mean the rollout is serving traffic to prod?

dthomson25:
The Progressing condition was introduced with the implementing ProgessDeadlineSeconds PR (#54) to indicates that a rollout is in a steady-state. The https://github.com/argoproj/argo-rollouts/pull/60 PR improves the available condition to make it more clear when a Rollout is available.

dthomson25:
Closing with #60 and #54 as the fixes

PR #47, opened 2019-03-25
Add Initial Prometheus Metrics

dthomson25:
Addresses https://github.com/argoproj/argo-rollouts/issues/29 by introducing a framework to expose Prothemeus metrics.  Adds metrics for the following:

- Reconciliation times for individual rollouts
- Reconciliation Error counters
- Invalid Spec counters
- Rollout creation time
- General Golang metrics

This PR did not include metrics on switching services and incrementing step index counters.  This will be incorporated with the work done with https://github.com/argoproj/argo-rollouts/issues/32

### DIFF
--- a/controller/bluegreen.go
+++ b/controller/bluegreen.go
@@ -27,7 +27,6 @@ func (c *Controller) rolloutBlueGreen(r *v1alpha1.Rollout, rsList []*appsv1.Repl
 		return err
 	}
 	allRSs := append(oldRSs, newRS)
-
 	// Scale up, if we can.
 	logCtx.Infof("Reconciling new ReplicaSet '%s'", newRS.Name)
 	scaledUp, err := c.reconcileNewReplicaSet(allRSs, newRS, r)
@@ -86,6 +85,7 @@ func (c *Controller) rolloutBlueGreen(r *v1alpha1.Rollout, rsList []*appsv1.Repl
 			return err
 		}
 	}
+
 	return c.syncRolloutStatusBlueGreen(allRSs, newRS, previewSvc, activeSvc, r, false)
 }
 
--- a/controller/sync.go
+++ b/controller/sync.go
@@ -347,6 +347,7 @@ func CreateTwoWayMergePatch(orig, new, dataStruct interface{}) ([]byte, bool, er
 
 // persistRolloutStatus persists updates to rollout status. If no changes were made, it is a no-op
 func (c *Controller) persistRolloutStatus(orig *v1alpha1.Rollout, newStatus *v1alpha1.RolloutStatus, newPause *bool) error {
+	c.metricsServer.IncPhase(orig, newStatus)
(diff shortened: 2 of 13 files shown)


dthomson25:
New CodeCov at https://codecov.io/github/argoproj/argo-rollouts/commit/b52044cdf0cd19c83d252166a615810ba5720169

controller/sync.go
```diff
@@ -346,7 +347,8 @@ func CreateTwoWayMergePatch(orig, new, dataStruct interface{}) ([]byte, bool, er
 }
 
 // persistRolloutStatus persists updates to rollout status. If no changes were made, it is a no-op
-func (c *Controller) persistRolloutStatus(orig *v1alpha1.Rollout, newStatus *v1alpha1.RolloutStatus, newPause *bool) error {
+func (c *Controller) persistRolloutStatus(orig *v1alpha1.Rollout, newStatus *v1alpha1.RolloutStatus, newPause *bool, phase metrics.ReconcilePhase) error {
+	c.metricsServer.IncPhase(orig, phase)
```
alexmt: I think `RolloutStatus` has or supposed to have all information which describes current state of rollout. Instead of imperatively set it in code we can infer the state within `persistRolloutStatus` method. Most likely you can check `InvalidSpec` and `Available` condition and make decision about current phase. What do you think?

dthomson25: I like this approach a lot more.  The only issue is that the `Available` condition needs to be cleaned up (as per #41), but I think it makes sense to go start with this implementation and work on that #41 afterward.

controller/bluegreen.go
```diff
@@ -20,73 +21,73 @@ func (c *Controller) rolloutBlueGreen(r *v1alpha1.Rollout, rsList []*appsv1.Repl
 	logCtx := logutil.WithRollout(r)
 	newRS, oldRSs, err := c.getAllReplicaSetsAndSyncRevision(r, rsList, true)
 	if err != nil {
-		return err
+		return c.metricsServer.IncError(r, err)
```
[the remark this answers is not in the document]
dthomson25: Yeah, I agree that makes sense.  One side effort of this approach is that I will need to add an extra counter for reconciliation errors because I don't have the rollout variable to get the required strategy label at https://github.com/argoproj/argo-rollouts/blob/master/controller/controller.go#L223.  I could use the informer to get the rollout, but that feels like the controller is doing too much work during the error handling. However, I don't think that side effort is bad though. The new counter will only be missing the strategy field, and in this case, the operator will need to check the controller's logs for more information. What do you think?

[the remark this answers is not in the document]
alexmt: I think both approaches are acceptable.  Please choose one which you like the best.

controller/sync.go
alexmt: Sure

PR #54, opened 2019-04-19
Add ProgressDeadlineSeconds

dthomson25:
Additionally, I simplified the testing framework to allow developers to test specific fields within the objects sent to the Kubernetes API instead of the entire object.

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -10,6 +10,7 @@ import (
 	patchtypes "k8s.io/apimachinery/pkg/types"
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	"github.com/argoproj/argo-rollouts/utils/conditions"
 	logutil "github.com/argoproj/argo-rollouts/utils/log"
 )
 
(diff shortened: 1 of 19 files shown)


controller/service.go
```diff
@@ -114,12 +115,15 @@ func (c *Controller) getPreviewAndActiveServices(r *v1alpha1.Rollout) (*corev1.S
...
 	if r.Spec.Strategy.BlueGreenStrategy.PreviewService != "" {
 		previewSvc, err = c.kubeclientset.CoreV1().Services(r.Namespace).Get(r.Spec.Strategy.BlueGreenStrategy.PreviewService, metav1.GetOptions{})
 		if err != nil {
 			if errors.IsNotFound(err) {
-				logCtx.Warnf("Service %v does not exist", r.Spec.Strategy.BlueGreenStrategy.PreviewService)
+				msg := fmt.Sprintf(conditions.ServiceNotFoundMessage, "preview", r.Spec.Strategy.BlueGreenStrategy.PreviewService)
+				c.recorder.Eventf(r, corev1.EventTypeWarning, conditions.ServiceNotFoundReason, msg)
+				cond := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionFalse, conditions.ServiceNotFoundReason, msg)
+				conditions.SetRolloutCondition(&r.Status, *cond)
+				c.rolloutsclientset.ArgoprojV1alpha1().Rollouts(r.Namespace).Update(r)
```
alexmt: I think we should use Patch to set conditions.

alexmt: Return Update/Patch error

controller/service.go
```diff
@@ -130,7 +134,11 @@ func (c *Controller) getPreviewAndActiveServices(r *v1alpha1.Rollout) (*corev1.S
 	activeSvc, err = c.kubeclientset.CoreV1().Services(r.Namespace).Get(r.Spec.Strategy.BlueGreenStrategy.ActiveService, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
-			logCtx.Warnf("Service %v does not exist", r.Spec.Strategy.BlueGreenStrategy.PreviewService)
+			msg := fmt.Sprintf(conditions.ServiceNotFoundMessage, "active", r.Spec.Strategy.BlueGreenStrategy.ActiveService)
```
alexmt: Same as for line 126

utils/conditions/conditions.go
```diff
@@ -40,6 +42,70 @@ const (
 	DuplicatedServicesMessage = "This rollout uses the same service for the active and preview services, but two different services are required."
 	// Available the reason to indicate that the rollout is serving traffic from the active service
 	Available = "Available"
+
+	// Reasons and Messages for rollout conditions
+
+	// Progressing:
```
alexmt: Delete this line?

controller/service.go
dthomson25: 👍

dthomson25: 👍

controller/service.go
dthomson25: 👍

utils/conditions/conditions.go
dthomson25: 👍

PR #73, opened 2019-05-15
Cleanup replicasets in canary deployment (#68)

### LINKED ISSUE #68: Rollout with Canary deployment does not cleanup replicasets
**Steps to reproduce**:

1. Create a rollout with canary deployment strategy and `revisionHistoryLimit` field set to e.g. 3
2. Deploy new version several times (more than number in `revisionHistoryLimit` )
3. See that replicasets are not getting deleted

**Expected behavior**

Max number of replicasets should be limited to number in `revisionHistoryLimit` field

**Version**: v0.3

alexmt:
Closes #68

### DIFF
--- a/controller/canary.go
+++ b/controller/canary.go
@@ -83,6 +83,12 @@ func (c *Controller) rolloutCanary(rollout *v1alpha1.Rollout, rsList []*appsv1.R
 		logCtx.Infof("Not finished reconciling Canary Pause")
 		return c.syncRolloutStatusCanary(oldRSs, newRS, stableRS, rollout)
 	}
+	if conditions.RolloutComplete(rollout, &rollout.Status) {
+		logCtx.Info("Cleaning up old replicasets")
+		if err := c.cleanupRollouts(oldRSs, rollout); err != nil {
+			return err
+		}
+	}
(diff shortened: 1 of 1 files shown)


controller/canary.go
```diff
@@ -83,6 +83,12 @@ func (c *Controller) rolloutCanary(rollout *v1alpha1.Rollout, rsList []*appsv1.R
 		logCtx.Infof("Not finished reconciling Canary Pause")
 		return c.syncRolloutStatusCanary(oldRSs, newRS, stableRS, rollout)
 	}
+	if conditions.RolloutComplete(rollout, &rollout.Status) {
+		logCtx.Info("Cleaning up old replicasets")
+		if err := c.cleanupRollouts(oldRSs, rollout); err != nil {
+			return err
+		}
+	}
```
jessesuen: This code occurs too late, and we will never hit this logic if, for example, the rollout canary strategy has pauses in it. It's possible to deploy many versions of a rollout with a pause step, and then reach a greater number of replicasets than what is controlled in `spec.revisionHistoryLimit`. The cleanup of rollouts needs to happen every reconciliation, and should be a no-op if there is nothing to clean up.

Issue #70, opened 2019-05-14
Rollout is not scaling down old replicasets properly


jessesuen:
Here is a rollout which is in a Suspended state, but with four replicasets scaled to two:

![image](https://user-images.githubusercontent.com/12677113/57736765-09818f80-765e-11e9-9509-6ced6bbc9cba.png)

The expectation is in a steady state (Suspended), there should only be two replicasets scaled higher than 0 (active and preview)

I ran a diff against the last three revisions (12, 11, 9) of the ReplicaSet. I'm not sure what happened to ReplicaSet revision 10. Notice that the only differences are in metadata and status. The replicaset spec is the same, which means the pod template is the same. However the bug is that the replicaset hash name is *not* the same.

```
$ diff rs-12 rs-11
6,8c6,7
<     rollout.argoproj.io/revision: '12'
<     rollout.argoproj.io/revision-history: '10'
<   creationTimestamp: '2019-05-14T21:54:49Z'
---
>     rollout.argoproj.io/revision: '11'
>   creationTimestamp: '2019-05-14T22:16:39Z'
16c15
<     rollouts-pod-template-hash: 65c456b799
---
>     rollouts-pod-template-hash: 7d58696fd9
18c17
<   name: web-service-integration-65c456b799
---
>   name: web-service-integration-7d58696fd9
27c26
<   resourceVersion: '95549146'
---
>   resourceVersion: '95556597'
29,30c28,29
<     /apis/apps/v1/namespaces/fdp-connectivity-web-service-integration-usw2-ppd-qal/replicasets/web-service-integration-65c456b799
<   uid: e584ea76-7692-11e9-9427-0a985b86565a
---
>     /apis/apps/v1/namespaces/fdp-connectivity-web-service-integration-usw2-ppd-qal/replicasets/web-service-integration-7d58696fd9
>   uid: f1c6de3d-7695-11e9-9427-0a985b86565a
36c35
<       rollouts-pod-template-hash: 65c456b799
---
>       rollouts-pod-template-hash: 7d58696fd9
57c56
<         rollouts-pod-template-hash: 65c456b799
---
>         rollouts-pod-template-hash: 7d58696fd9
```

```
$ diff rs-12 rs-9
6,7c6
<     rollout.argoproj.io/revision: '12'
<     rollout.argoproj.io/revision-history: '10'
---
>     rollout.argoproj.io/revision: '9'
16c15
<     rollouts-pod-template-hash: 65c456b799
---
[...]

PR #76, opened 2019-05-16
Scale down older RS on non-happy path

dthomson25:
Solves the second half of https://github.com/argoproj/argo-rollouts/issues/70 by scaling down and deleting old replicasets at any point of the rollout instead of just at the end.

### DIFF
--- a/controller/replicaset_test.go
+++ b/controller/replicaset_test.go
@@ -208,15 +208,6 @@ func TestReconcileOldReplicaSet(t *testing.T) {
 			scaleExpected:      false,
 		},
 		{
-			name:               "New ReplicaSet is not fully healthy",
-			rolloutReplicas:    10,
-			oldReplicas:        10,
-			newReplicas:        10,
-			readyPodsFromOldRS: 10,
-			readyPodsFromNewRS: 9,
-			scaleExpected:      false,
-		},
-		{
 			name:                "Clean up unhealthy pods",
 			rolloutReplicas:     10,
 			oldReplicas:         10,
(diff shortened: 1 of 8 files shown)


controller/replicaset_test.go
```diff
@@ -249,23 +240,23 @@ func TestReconcileOldReplicaSet(t *testing.T) {
...
 			rollout.Spec.Selector = &metav1.LabelSelector{MatchLabels: newSelector}
-			fake := fake.Clientset{}
-			k8sfake := k8sfake.Clientset{}
-			controller := &Controller{
-				rolloutsclientset: &fake,
-				kubeclientset:     &k8sfake,
-				recorder:          &record.FakeRecorder{},
-			}
-			scaled, err := controller.reconcileOldReplicaSets(allRSs, oldRSs, newRS, rollout)
+
```
dthomson25: Prior to changes, test was returning an empty replicaset instead of a modified rollout when the Update Rollout command was called. This caused a panic when the replicas pointer was dereferenced with the new changes. We needed to start the informers in order to have the reactor react on the update call and return the correct updated object.

controller/replicaset_test.go
```diff
@@ -207,15 +207,6 @@ func TestReconcileOldReplicaSet(t *testing.T) {
...
 		},
-		{
-			name:               "New ReplicaSet is not fully healthy",
-			rolloutReplicas:    10,
-			oldReplicas:        10,
-			newReplicas:        10,
-			readyPodsFromOldRS: 10,
-			readyPodsFromNewRS: 9,
-			scaleExpected:      false,
-		},
```
dthomson25: No longer needed since we should scale down regardless of the new rs state

Issue #81, opened 2019-05-21
Rollout does not scale preview service properly on changes


jessesuen:
We have a v0.3.1 Rollout in the following condition:

![image](https://user-images.githubusercontent.com/12677113/58073181-c9874480-7b56-11e9-9ff3-d6230e5cb460.png)

The bottom rollout is actually the desired spec as a preview, and has 3 replicas. However, the rollout spec has 2 desired replicas and currently has the following error condition:

`Rollout "partner-support-portal" has timed out progressing.`

jessesuen:
Simplified rollout object:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  annotations:
    rollout.argoproj.io/revision: '16'
  generation: 1
  labels:
    app: partner-support-portal
  name: partner-support-portal
spec:
  minReadySeconds: 30
  paused: false
  replicas: 2
  selector:
    matchLabels:
      app: partner-support-portal
  strategy:
    blueGreen:
      activeService: partner-support-portal-active
      previewService: partner-support-portal-preview
  template:
    metadata:
      labels:
        app: partner-support-portal
    spec:
      containers:
        - image: someimage:sometag
          name: partner-support-portal
status:
  HPAReplicas: 2
  availableReplicas: 5
  blueGreen:
    activeSelector: 5f59c4c4dc
    previewSelector: 7c67b5fd8d
  canary: {}
  conditions:
    - lastTransitionTime: '2019-05-20T07:22:41Z'
      lastUpdateTime: '2019-05-20T07:22:41Z'
      message: Rollout has minimum availability
      reason: AvailableReason
      status: 'True'
      type: Available
    - lastTransitionTime: '2019-05-20T11:02:00Z'
      lastUpdateTime: '2019-05-20T11:02:00Z'
      message: Rollout "partner-support-portal" has timed out progressing.
      reason: ProgressDeadlineExceeded
      status: 'False'
      type: Progressing
  currentPodHash: 6756d4c4c7
  observedGeneration: 5745f7fc9c
  readyReplicas: 5
  replicas: 5
  selector: 'app=partner-support-portal,rollouts-pod-template-hash=5f59c4c4dc'
```

Simplified replicaset
```yaml
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  annotations:
    rollout.argoproj.io/desired-replicas: '3'
    rollout.argoproj.io/revision: '16'
[...]

PR #82, opened 2019-05-21
Fix issue where preview replicaset was not scaled properly during a scaling event

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/81

Working on unit test for this

### DIFF
--- a/controller/bluegreen.go
+++ b/controller/bluegreen.go
@@ -248,6 +248,18 @@ func (c *Controller) scaleBlueGreen(rollout *v1alpha1.Rollout, newRS *appsv1.Rep
 		}
 	}
 
+	previewRS, _ := replicasetutil.GetReplicaSetByTemplateHash(allRS, rollout.Status.BlueGreen.PreviewSelector)
+	if previewRS != nil {
+		previewReplicas := rolloutReplicas
+		if rollout.Spec.Strategy.BlueGreenStrategy.PreviewReplicaCount != nil && !rollout.Status.BlueGreen.ScaleUpPreviewCheckPoint {
+			previewReplicas = *rollout.Spec.Strategy.BlueGreenStrategy.PreviewReplicaCount
+		}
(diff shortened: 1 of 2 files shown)


controller/bluegreen.go
```diff
@@ -248,6 +248,20 @@ func (c *Controller) scaleBlueGreen(rollout *v1alpha1.Rollout, newRS *appsv1.Rep
 		}
 	}
 
+	previewRS, _ := replicasetutil.GetReplicaSetByTemplateHash(allRS, rollout.Status.BlueGreen.PreviewSelector)
+	if previewRS != nil {
+		var previewReplicas int32
+		if rollout.Spec.Strategy.BlueGreenStrategy.PreviewReplicaCount != nil {
```
dthomson25: We need to take into account when the `status.blueGreenStatus.scaleUpPreviewCheckPoint` field is set to true. That field is used to indicates that the rollout has been unpaused and it should scale up the preview ReplicaSet to the full Replica count before switching the active service to it.

controller/bluegreen.go
```diff
@@ -248,6 +248,20 @@ func (c *Controller) scaleBlueGreen(rollout *v1alpha1.Rollout, newRS *appsv1.Rep
...
+			previewReplicas = *rollout.Spec.Strategy.BlueGreenStrategy.PreviewReplicaCount
+		} else {
+			previewReplicas = rolloutReplicas
```
dthomson25: Is there a reason that you aren't setting the previewReplicas to rolloutReplicas as the default and simplifying the if else statement to just an if statement?

controller/bluegreen.go
jessesuen: Thanks I thought about that but didn't know if there was a good way to know.

controller/bluegreen.go
jessesuen: Good suggestion.

Issue #80, opened 2019-05-20
Unintuitive automatic pause behavior for blue green strategy


jessesuen:
The current automatic pause behavior for the blue/green strategy is inconsistent, and hard to explain. Some examples:

* it does not pause during the initial rollout
* it does not pause if the preview service is unspecified, but will pause if it is.
* it does not pause during a fast rollback

These rules are not obvious or documented. I think we need to rework the behavior so that it is more easily predictable.

I also feel the default behavior should be switched to automatically perform a cutover instead of the current behavior of automatically pausing if there is a preview service. This would make it consistent behavior with rolling update, recreate, canary, where the default behavior is to perform the promotion automatically. Just because I don't specify a preview service, doesn't necessarily mean I don't want it to pause, and vice versa.

jessesuen:
Another problem is that there is no a way to automatically pause a blue green rollout before cutover, unless I specify a preview service.

jessesuen:
Fixed with spec.autoPromotionEnabled field

PR #99, opened 2019-06-05
Add autoPromotionEnabled with default value of true

dthomson25:
address https://github.com/argoproj/argo-rollouts/issues/80

### DIFF
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -80,6 +80,10 @@ type BlueGreenStrategy struct {
 	// resumed the new replicaset will be full scaled up before the switch occurs
 	// +optional
 	PreviewReplicaCount *int32 `json:"previewReplicaCount,omitempty"`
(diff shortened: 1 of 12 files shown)


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -80,6 +80,10 @@ type BlueGreenStrategy struct {
 	// resumed the new replicaset will be full scaled up before the switch occurs
 	// +optional
 	PreviewReplicaCount *int32 `json:"previewReplicaCount,omitempty"`
+	// AutoPromoteEnabled indicates if the rollout should automatically promote the new ReplicaSet
+	// to the active service or enter a paused state. If not specified, the default value is true.
+	// +optional
+	AutoPromoteEnabled *bool `json:"autoPromoteEnabled,omitempty"`
```
jessesuen: We are inconsistent: `AutoPromoteEnabled` vs. `AutoPromotionSeconds`. To be consistent it should be: `AutoPromotionEnabled`

dthomson25: fixed

Issue #88, opened 2019-05-23
Change in pod template hash after upgrading k8s dependencies


jessesuen:
I tried updating our k8s dependencies (e.g. k8s.io/client-go, k8s.io/kubernetes). However, after updating our dependencies, unit tests started failing because the computed hash of the pod template spec was different than from before the update.

Even worse, when starting the version of controller with the updated k8s libraries, it:
1. caused existing rollouts to become redeployed, since the controller detected the new pod template hash as a change in the spec.
2. somehow caused the active to service selector to select no valid pods, essentially causing downtime.

Currently we use the `k8s.io/kubernetes/pkg/controller` library to compute the hash:
```
controller.ComputeHash(&rollout.Spec.Template, rollout.Status.CollisionCount)
```

I suspect what is happening is that there are new fields in the pod template, which causing hashes to be different. Since rollouts does need to keep up with kubernetes libraries, we need to figure out a way to allow library updates to happen, but not trigger redeploys of rollouts.

jessesuen:
I think part of solution will have to be that whenever we are comparing pod templates, we can no longer perform a simple string comparison since we can't trust the hash that we store in the replicaset. The hashing function may have changed. Instead we will *always* have to compute the hash on both sides (the rollout's `spec.template.spec` vs. the replicaset's `spec.template.spec`) in order to have guaranteed true equivalence.

PR #100, opened 2019-06-07
Issue #91 - Ability to specify canaryService Service object to reach only canary pods

### LINKED ISSUE #91: Ability to specify canaryService Service object to reach only canary pods
Similar to blue-green's previewService, for users of the canary strategy, users would like to reach the pods of the canary through a service object, without hitting any of the old/stable pods. This proposal is to allow specifying a canary service like so:

```yaml
spec:
  strategy:
    canary:
      canaryService: guestbook-canary
```
If the user specifies a `canaryService` in the strategy, the controller will modify the specified service to add the `rollouts-pod-template-hash` as a label selector. 

For this to work, we will need to inject the `rollouts-pod-template-hash` label to the replicasets created by the canary strategy. This is something we already do for the blue-green strategy, but will need to do for canary to accomplish this feature.

alexmt:
Closes #91

### DIFF
--- a/controller/canary_test.go
+++ b/controller/canary_test.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"testing"
@@ -9,6 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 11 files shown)


controller/service.go
```diff
@@ -155,6 +156,35 @@ func (c *Controller) getPreviewAndActiveServices(r *v1alpha1.Rollout) (*corev1.S
...
+				c.persistRolloutStatus(r, newStatus, &r.Spec.Paused)
+			}
+			return err
+		}
+
+		if svc.Spec.Selector == nil {
+			svc.Spec.Selector = make(map[string]string)
+		}
+
+		hash := controller.ComputeHash(&r.Spec.Template, r.Status.CollisionCount)
```
dthomson25: Instead of calculating the hash from the template, we should use calculate the hash from the newRS podSpec to protect against the k8s dependency change (#88).

controller/service.go
```diff
@@ -155,6 +156,35 @@ func (c *Controller) getPreviewAndActiveServices(r *v1alpha1.Rollout) (*corev1.S
...
+		if svc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey] != hash {
+			patch := fmt.Sprintf(switchSelectorPatch, v1alpha1.DefaultRolloutUniqueLabelKey, hash)
```
dthomson25: Can you use the c.switchServiceSelector method to patch the service?

controller/canary_test.go
```diff
@@ -896,3 +898,52 @@ func TestCanaryRolloutStatusHPAStatusFields(t *testing.T) {
 	patch := f.getPatchedRollout(index)
 	assert.Equal(t, calculatePatch(r2, expectedPatchWithSub), patch)
 }
+
+func TestCanaryRolloutWithCanaryService(t *testing.T) {
+	f := newFixture(t)
+
+	canarySvc := newService("canary", 80, make(map[string]string))
```
dthomson25: What is there a reason that you instantiate the selector to nothing? If we don't, that will allow us to also test the case where the selector is nil.

controller/service.go
```diff
@@ -155,6 +156,35 @@ func (c *Controller) getPreviewAndActiveServices(r *v1alpha1.Rollout) (*corev1.S
...
+	if r.Spec.Strategy.CanaryStrategy != nil && r.Spec.Strategy.CanaryStrategy.CanaryService != "" {
+		svc, err := c.servicesLister.Services(r.Namespace).Get(r.Spec.Strategy.CanaryStrategy.CanaryService)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				msg := fmt.Sprintf(conditions.ServiceNotFoundMessage, r.Spec.Strategy.CanaryStrategy.CanaryService)
+				c.recorder.Event(r, corev1.EventTypeWarning, conditions.ServiceNotFoundReason, msg)
+				newStatus := r.Status.DeepCopy()
+				cond := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionFalse, conditions.ServiceNotFoundReason, msg)
+				conditions.SetRolloutCondition(newStatus, *cond)
+				c.persistRolloutStatus(r, newStatus, &r.Spec.Paused)
```
dthomson25: I'm wondering if we should abstract this logic into a separate method because we have this code repeated 3 times. Thoughts?

controller/service.go
```diff
@@ -234,6 +264,10 @@ func getRolloutServiceKeys(rollout *v1alpha1.Rollout) []string {
 		if rollout.Spec.Strategy.BlueGreenStrategy.PreviewService != "" {
 			servicesSet[fmt.Sprintf("%s/%s", rollout.Namespace, rollout.Spec.Strategy.BlueGreenStrategy.PreviewService)] = true
 		}
+	} else if rollout.Spec.Strategy.CanaryStrategy != nil {
```
dthomson25: Can we add another test to TestGetRolloutServiceKeysForCanary to cover when a canaryService specified?

controller/canary_test.go
alexmt: Good point. Changed `make(map[string]string)` to nil.

controller/service.go
alexmt: done

controller/service.go
alexmt: Done.

controller/service.go
```diff
@@ -155,6 +156,35 @@ func (c *Controller) getPreviewAndActiveServices(r *v1alpha1.Rollout) (*corev1.S
 	return previewSvc, activeSvc, nil
 }
 
+func (c *Controller) reconcileCanaryService(r *v1alpha1.Rollout, newRS *appsv1.ReplicaSet) error {
+	if r.Spec.Strategy.CanaryStrategy != nil && r.Spec.Strategy.CanaryStrategy.CanaryService != "" {
```
jessesuen: Style nit, we would save a block of indentation with:
```golang
if r.Spec.Strategy.CanaryStrategy == nil || r.Spec.Strategy.CanaryStrategy.CanaryService == "" {
    return nil
}
```

controller/service.go
alexmt: done

controller/service.go
alexmt: good catch! done

controller/service.go
alexmt: done

alexmt:
Addressed review comments. Also discovered that tests became flaky after merging https://github.com/argoproj/argo-rollouts/pull/95: additional service index requires run syncronize informers in test. The https://github.com/argoproj/argo-rollouts/pull/100/commits/9919549b147684e787414ddfb68c2e4f73f79941 fixes it

controller/service.go
```diff
@@ -3,17 +3,16 @@ package controller
 import (
 	"fmt"
 
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	"github.com/argoproj/argo-rollouts/utils/conditions"
+	logutil "github.com/argoproj/argo-rollouts/utils/log"
```
dthomson25: Style nix: can we move this back below the external lib imports

alexmt: done

PR #106, opened 2019-06-13
Adds CircleCI configuration

alexec:


.circleci/config.yml
```diff
@@ -0,0 +1,47 @@
+version: 2.1
+jobs:
+  build:
+    working_directory: /go/src/github.com/argoproj/argo-rollouts
+    docker:
+      - image: circleci/golang:1.12.6
```
alexec: @jessesuen can you comment on upgrading to 1.12.6 in this PR?

.circleci/config.yml
```diff
@@ -0,0 +1,47 @@
...
+    environment:
+      TEST_RESULTS: /tmp/test-results
+    steps:
+      - checkout
+      - run: mkdir -p $TEST_RESULTS
+      - restore_cache:
+          keys:
+            - v3-pkg-cache
+            - v2-pkg-cache
```
alexec: Delete this line.

.circleci/config.yml
jessesuen: I run with 1.12.5 so I think this should be fine

PR #117, opened 2019-07-03
Refactor controller package into multiple packages

dthomson25:


controller/controller.go
```diff
@@ -48,164 +37,101 @@ const (
 	DefaultMetricsPort = 8090
 )
 
-// Controller is the controller implementation for Rollout resources
-type Controller struct {
-	// rsControl is used for adopting/releasing replica sets.
-	replicaSetControl controller.RSControlInterface
+// Manager is the controller implementation for Rollout resources
+type Manager struct {
```
alexmt: For consistency struct should be renamed to ControllerManager

dthomson25: I agree with being consistent but I think it should be manager. The linter was complaining because of the package name is controller, which makes the struct controller.ControllerManager. This is redundant naming compare to controller.Manager

PR #120, opened 2019-07-11
Add experiment crd

dthomson25:


controller/controller.go
```diff
@@ -35,20 +36,32 @@ const (
 
 	// DefaultMetricsPort Default port to expose the metrics endpoint
 	DefaultMetricsPort = 8090
+
+	// DefaultRolloutThreads Default numnber of worker threads to start with the controller
+	DefaultRolloutThreads = 1
```
alexmt: I would bump defaults to e.g. 10. I think it is better to have real life default which allows to manage hundreds or thousands of rollouts. The `1` might cause delays.

dthomson25: 👍 Done

PR #124, opened 2019-07-17
Add ReplicaSet management for Experiments

dthomson25:


experiments/experiment.go
```diff
@@ -0,0 +1,133 @@
...
+			return false, nil
+		}
+	}
+	if templateReplicaCount != replicasetutil.GetAvailableReplicaCountForReplicaSets([]*appsv1.ReplicaSet{existingTemplateRS}) {
+		return false, nil
+	}
+	return true, nil
+}
+
+func (ec *ExperimentController) syncExperimentStatus(experiment *v1alpha1.Experiment, templateRSs map[string]*appsv1.ReplicaSet, allTemplatesAvailable bool) error {
```
alexmt: Looks like `allTemplatesAvailable` is not used. Please deleted unused parameter

experiments/replicaset.go
```diff
@@ -0,0 +1,200 @@
...
+		if matches > 1 {
+			return nil, fmt.Errorf("multiple ReplicaSets match single experiment template")
+		} else if matches == 1 {
+			templateToRS[template.Name] = templateRSs[0]
+		}
+	}
+	return templateToRS, nil
+}
+
+func (ec *ExperimentController) getNewReplicaSet(experiment *v1alpha1.Experiment, template v1alpha1.TemplateSpec) (*appsv1.ReplicaSet, error) {
```
alexmt: I think `getNewReplicaSet` is misleading since the method might create/update replicaset. Probably `reconcileReplicaSet` is better name.

experiments/replicaset.go
```diff
@@ -0,0 +1,200 @@
...
+		if statusIndex != nil {
+			experiment.Status.TemplateStatuses[*statusIndex] = templateStatus
+		} else {
+			experiment.Status.TemplateStatuses = append(experiment.Status.TemplateStatuses, templateStatus)
+		}
+
+		// Update the collisionCount for the Rollout and let it requeue by returning the original
+		// error.
+		_, roErr := ec.arogProjClientset.ArgoprojV1alpha1().Experiments(experiment.Namespace).Update(experiment)
+		if roErr == nil {
```
alexmt: I think we should at least log update error. 

Also is it possible to use patch in this case to reduce the possibility of concurrent update error?

Issue #119, opened 2019-07-10
Controller should default spec.replicas to 1, in order to work with HPA


jessesuen:
HPA requires `spec.replicas` to be defined, in order to function. Otherwise it fails with: `Internal error occurred: the spec replicas field ".spec.replicas" does not exist`

This means that when controller sees a rollout object with this field missing, it should default it to 1.

jessesuen:
The workaround is to manually `kubectl edit` the rollout object and set `spec.replicas` to be 1 (to act as if a controller defaulted it). Note that as with deployments, `spec.replicas` should be **_omitted_** from the applied manifest, otherwise it will fight against HPA.

jessesuen:
Fixed

PR #125, opened 2019-07-17
Add .Spec.Replicas if not set in rollouts

dthomson25:


dthomson25:
Solves https://github.com/argoproj/argo-rollouts/issues/119

Issue #110, opened 2019-06-18
Experiment CRD


jessesuen:
To drive canary analysis, we are proposing to introduce a mechanism to launch an ephemeral run of one or more ReplicaSets, typically gated by some time duration. To power this, we would introduce an "Experiment" CRD, which might look something like:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Experiment
name:
  name: guesbook-experiment
spec:
  durationSeconds: 600
  templates:
  - name: canary
    spec:
      containers:
      - name: guestbook
        image: guesbook:v2
  - name: baseline
    spec:
      containers:
      - name: guestbook
        image: guesbook:v1
```

This CRD would launch two replicasets, with the respective pod templates, for some time duration.

After the ReplicaSets completed the time duration, it would then be followed by an analysis which returns a score. 

```yaml
spec:
  durationSeconds: 600
  templates:
  - name: canary
...
  analysis:
    # syntax TBD????
    intervalSeconds: 60
    realtime: false
    prometheus:
      server: prometheus.default:9000
      query: grpc_server_handled_total{job="argocd-server-metrics",grpc_service="application.ApplicationService",grpc_code="Error"} > 0
```

Note that we would also need the ability to perform analysis of the Experiment in real-time, e.g. if the experiment is going horribly wrong and the score is below some threshold, the Experiment should be stopped.

```yaml
spec:
  durationSeconds: 600
  templates:
  - name: canary
...
  analysis:
    failureThreshold: 50
    failFast: true
status:
  score: 80
```

The way that this integrates with a Rollout, is by introducing a new step type into the canary strategy, which initiate the run of the experiment, and only proceed to the promotion, if the experiment was successful.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
name:
  name: guestbook
spec:
...
  strategy:
    canary:
      steps:
      - experiment: # syntax TBD
           templates:
           - name: baseline
             specFrom: baseline
           - name: canary
             specFrom: canary
      - setWeight: 50
      - pause: {}
```

edlee2121:
We should give some thoughts about how this would integrate with various frontend services such as LaunchDarkly and backend services such as Kayenta. The most general form would be to create a class of CRD's to interface with these services. Less general but perhaps simpler to use would be to add, ideally, generic options for integrating with these services.

mousavian:
Great stuff, I'm interested to have this. Do you think it'll be in 0.5.0? :D

[...]

PR #134, opened 2019-08-09
Add conditions and duration to experiments

dthomson25:
Implements second half of https://github.com/argoproj/argo-rollouts/issues/110


experiments/experiment.go
```diff
@@ -64,21 +71,49 @@ func (ec *ExperimentController) reconcileTemplate(experiment *v1alpha1.Experimen
 	return true, nil
 }
 
+func (ec *ExperimentController) checkEnqueueExperimentDuringRun(experiment *v1alpha1.Experiment) {
```
alexmt: Just a suggestion. It feels like `experiment.PassedDurations`  has `checkEnqueueExperimentDuringRun` has the shared logic. To get rid of duplication `experiment.PassedDurations` might return `true/false` and remaining duration which can be used in `checkEnqueueExperimentDuringRun`. So `checkEnqueueExperimentDuringRun` would look like:

```
func (ec *ExperimentController) checkEnqueueExperimentDuringRun(experiment *v1alpha1.Experiment) {
  if passed, timeRemaining := experiment.PassedDurations(experiment); !passed {
    logCtx.Infof("Enqueueing Experiment in %s seconds", timeRemaining.String())
    ec.enqueueExperimentAfter(experiment, timeRemaining)
  }
}
```

experiments/conditions.go
```diff
@@ -0,0 +1,54 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	"github.com/argoproj/argo-rollouts/utils/conditions"
+)
+
+func (ec *ExperimentController) calculateExperimentConditions(experiment *v1alpha1.Experiment, newStatus v1alpha1.ExperimentStatus, templateRSs map[string]*appsv1.ReplicaSet) v1alpha1.ExperimentStatus {
+	switch {
+	case conditions.ExperimentCompleted(newStatus):
```
alexmt: Just a suggestion.

Looks like `conditions.ExperimentCompleted`, `conditions. ExperimentProgressing` should never return true simultaneously for the same experiment. Only in this case it is safe to use these functions in switch expression. If, because of a bug, more than one return true than only first `case` got executed.

I would consider creating `ExperimentStatusCode` type which can take values `ExperimentCompleted`, `ExperimentProgressing` etc and single method `GetExperimentStatus` which can be used in `calculateExperimentConditions`.  You can even consider saving `statusCode`  in CRD status to simplify health assessment.

experiments/experiment.go
dthomson25: 👍

PR #153, opened 2019-09-06
Add InvalidSpec condition to Experiments

dthomson25:


dthomson25:
https://github.com/argoproj/argo-rollouts/issues/110

utils/conditions/experiments.go
```diff
@@ -170,3 +178,43 @@ func ExperimentRunning(experiment *v1alpha1.Experiment) bool {
...
+
+		if template.Name == "" {
+			message := fmt.Sprintf(ExperimentTemplateNameEmpty, experiment.Name)
+			return newInvalidSpecExperimentCondition(prevCond, InvalidSpecReason, message)
+		}
+		if ok := templateNameSet[template.Name]; ok {
+			message := fmt.Sprintf(ExperimentTemplateNameRepeatedMessage, experiment.Name, template.Name)
+			return newInvalidSpecExperimentCondition(prevCond, InvalidSpecReason, message)
+		}
+		templateNameSet[template.Name] = true
```
jessesuen: I think these messages should include the index of the template, otherwise it's hard to trace down which template has the issue.

dthomson25: done!

PR #156, opened 2019-09-10
ClusterRole missing patch privileges on replicasets

jessesuen:
Controller was hitting the following error:

```
E0910 23:45:08.441072       1 controller.go:287] [replicasets.apps "istio-rollout-67fb48cc58" is forbidden: User "system:serviceaccount:argo-rollouts:argo-rollouts" cannot patch resource "replicasets" in API group "apps" in the namespace "istio-rollout", replicasets.apps "istio-rollout-7bffb4d96b" is forbidden: Use
r "system:serviceaccount:argo-rollouts:argo-rollouts" cannot patch resource "replicasets" in API group "apps" in the namespace "istio-rollout"]
```

### DIFF
--- a/manifests/base/argo-rollouts-role.yaml
+++ b/manifests/base/argo-rollouts-role.yaml
@@ -8,36 +8,37 @@ metadata:
     app.kubernetes.io/part-of: argo-rollouts
 rules:
 - apiGroups:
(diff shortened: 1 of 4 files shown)


dthomson25:
Actually can you add this permission to the role too?

Issue #146, opened 2019-08-26
Infinitely loop on controller loop


dthomson25:
If the readiness probe starts to fail on a healthy rollout for longer than the deadline progressing seconds limit and the controller restarts, the controller will enter an infinite loop. The controller will constantly enqueue the degraded rollout. The controller should only enqueue the rollout once.

dthomson25:
Last PR before we can close https://github.com/argoproj/argo-rollouts/issues/110

PR #157, opened 2019-09-11
Fix infinite loop when Rollout should timeout

dthomson25:


dthomson25:
Closes #146

utils/conditions/conditions.go
```diff
@@ -379,20 +379,7 @@ func RolloutTimedOut(rollout *v1alpha1.Rollout, newStatus *v1alpha1.RolloutStatu
...
-	//
-	// * progressDeadlineSeconds is smaller than the difference between now and the time
-	//   the last rollout finished in the past.
-	// * the creation of a new ReplicaSet triggers a resync of the rollout prior to the
-	//   cached copy of the Rollout getting updated with the status.condition that indicates
-	//   the creation of the new ReplicaSet.
-	//
-	// The rollout will be resynced and eventually its Progressing condition will catch
-	// up with the state of the world.
-	if condition.Reason == NewRSAvailableReason {
```
dthomson25: This code is not needed because the RolloutComplete method will return true and prevent the RolloutTimeout from returning true.

https://github.com/argoproj/argo-rollouts/blob/master/rollout/sync.go#L407

Issue #92, opened 2019-05-31
Integration between Rollout with Experiment CRD


jessesuen:
Using the canary strategy, users would like to spin up new pods of the new desired pod spec, but without serving any production traffic. Additionally, users would like a way to reach the new canary pods through a service object. Essentially this is similar to blue-green's preview service, but for the canary strategy.

~The proposal is to accomplish this use case with the following spec definition:~
**_UPDATE_**: will solve this using experiment integration instead of a new preview step type.

```yaml
spec:
  strategy:
    canary:
      canaryService: guestbook-canary
      steps:
      - setWeight: 0
      - preview:
          replicas: 1
      - pause: {}
```

The second step would bring up a new ReplicaSet of the new desired spec, but would *only* have the pod-hash as a label to the pods. It would be missing all of the other pod labels in the spec (e.g. app=guestbook). Removing the other labels would allow us to prevent production traffic from hitting the preview pods.

NOTE: This assumes that the feature for canaryService (https://github.com/argoproj/argo-rollouts/issues/91), would modify the service to select **_only_** on the pod hash, and nothing else. In fact, we may need to raise an error condition if we detect any other user-defined label selectors in the canaryService (e.g. app=guestbook).

jessesuen:
Current direction is that this would be solved with experiments. So instead of introducing a `preview` step type, we would introduce the experiment integration between a Rollout and Experiment.

The ability to specify a `canaryService` was fixed in https://github.com/argoproj/argo-rollouts/issues/91.

jessesuen:
Experimentation may be applicable to blue-green, not just with canary. So removing canary specific label

jessesuen:
Fixed

PR #165, opened 2019-09-24
Add Initial Experiment Step

dthomson25:


dthomson25:
Initial PR to resolve https://github.com/argoproj/argo-rollouts/issues/92

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -144,6 +144,48 @@ type CanaryStrategy struct {
 	MaxSurge *intstr.IntOrString `json:"maxSurge,omitempty"`
 }
 
+// RolloutCanaryExperimentStep defines a template that is used to create a experiment for a step
+type RolloutCanaryExperimentStep struct {
```
jessesuen: I can imagine a future where ExperimentStep is not just being done as part of Canary, but BlueGreen. May want to name this just RolloutExperimentStep to not box ourselves.

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -144,6 +144,48 @@ type CanaryStrategy struct {
...
+	SpecRef ReplicaSetSpecRef `json:"specRef"`
+	// Replicas replica count for the template
+	Replicas int32 `json:"replicas"`
+	// AdditionalSelectors additional selectors to use for the RS created from the template
+	// +optional
+	Metadata RolloutExperimentAdditionalMetadata `json:"metadata,omitempty"`
+}
+
+//RolloutExperimentAdditionalMetadata extra labels to add to the template
+type RolloutExperimentAdditionalMetadata struct {
```
jessesuen: How about: `PodTemplateMetadata`. I would make clear that this metadata will go into the labels/annotations of the actual pod

pkg/apis/rollouts/v1alpha1/types.go
dthomson25: Done

pkg/apis/rollouts/v1alpha1/types.go
dthomson25: done

PR #170, opened 2019-09-27
Add initial provider and prometheus implementation

dthomson25:
I know that is PR will have some merge conflicts with #168, but they should be minor


pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -52,6 +52,13 @@ type AnalysisMetric struct {
 	MaxFailures *int32 `json:"maxFailures,omitempty"`
 	// FailFast will fail the entire analysis run prematurely
 	FailFast bool `json:"failFast,omitempty"`
+	// Provider configuration to the external system to use to verify the analysis
+	Provider AnalysisProvider `json:"provider"`
```
jessesuen: Nice.

providers/provider.go
```diff
@@ -0,0 +1,33 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	"github.com/argoproj/argo-rollouts/providers/prometheus"
+)
+
+// Provider methods to query a external systems and generate a measurement
+type Provider interface {
+	// Run start a new external system call for a measurement
+	// should be idoponent and do nothing is a call has been started
```
jessesuen: idempotent and do nothing if a call has been started

dthomson25: done

PR #171, opened 2019-09-28
Integrate analysis controller with provider interfaces

jessesuen:
NOTE: this switches status.metricResult to a list instead of map. The latter does not work well with constructing the json patch, and would error with: 

```
Error constructing AnalysisRun status patch: merging an object in json but data type is not struct, instead is: map
```


pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -61,6 +63,21 @@ type Metric struct {
...
 
+// EffectiveCount is the effective count based on whether or not count/interval is specified
+// If neither count or interval is specified, the effective count is 1
+// If only interval is specified, metric runs indefinitely and there is no effective count (nil)
+// Otherwise, it is the user specified value
+func (m *Metric) EffectiveCount() *int32 {
+	if m.Count == 0 {
+		if m.Interval == nil {
+			one := int32(1)
+			return &one
```
dthomson25: minor: you can use pointer.Int32Ptr(1) from the `	"k8s.io/utils/pointer"` library

jessesuen: I considered it, but wanted to reduce the import burden for anyone importing our types.

Makefile
```diff
@@ -77,6 +77,10 @@ lint:
 test:
 	go test -v -covermode=count -coverprofile=coverage.out `go list ./...`
 
+.PHONY: mocks
+mocks:
+	mockery -dir ./providers -name Provider -output ./providers/mocks
```
dthomson25: Can you add a step to the Contributing docs to install this?

PR #181, opened 2019-10-02
Add metric knob for maxInconclusive

jessesuen:
This adds a metric field `maxInconclusive`, which allows user to specify a maximum number of times a metric is allowed to measure Inconclusive, before making the entire metric Inconclusive.

### DIFF
--- a/utils/analysis/analysis.go
+++ b/utils/analysis/analysis.go
@@ -38,6 +38,9 @@ func ValidateMetric(metric v1alpha1.Metric) error {
 	if metric.MaxFailures < 0 {
 		return fmt.Errorf("maxFailures must be >= 0")
 	}
+	if metric.MaxInconclusive < 0 {
+		return fmt.Errorf("maxInconclusive must be >= 0")
+	}
 	if metric.MaxConsecutiveErrors != nil && *metric.MaxConsecutiveErrors < 0 {
 		return fmt.Errorf("maxConsecutiveErrors must be >= 0")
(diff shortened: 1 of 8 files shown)


utils/analysis/analysis.go
```diff
@@ -78,17 +81,16 @@ func IsWorse(current, new v1alpha1.AnalysisStatus) bool {
...
-	return run.Spec.Terminate || IsFailing(run)
-}
-
-// IsFailing returns whether or not any metric has measured Failed or Error, which will eventually
-// cause the entire run to fail.
-func IsFailing(run *v1alpha1.AnalysisRun) bool {
+	if run.Spec.Terminate {
+		return true
+	}
 	for _, res := range run.Status.MetricResults {
```
dthomson25: Can we add a nil check for run.status? This will be mainly helpful for the Rollout and Experiment implementation if the analysis hasn't started yet

utils/analysis/analysis.go
```diff
@@ -29,6 +29,9 @@ func ValidateMetric(metric v1alpha1.Metric) error {
 	if metric.Count < metric.MaxFailures {
 		return fmt.Errorf("count must be >= maxFailures")
 	}
+	if metric.Count < metric.MaxInconclusive {
+		return fmt.Errorf("count must be >= maxInconclusive")
```
dthomson25: Can you add to the TestValidateMetric test to cover this case?

PR #183, opened 2019-10-03
Rollout analysis plumbing

dthomson25:


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -196,8 +198,44 @@ type CanaryStep struct {
...
+// ArgumentValueFrom defines references to fields within resources to grab for the value (i.e. Pod Template Hash)
+type ArgumentValueFrom struct {
+	// PodTemplateHashValue gets the value from one of the children ReplicaSet's Pod Template Hash
+	PodTemplateHashValue *ValueFromPodTemplateHash `json:"podTemplateHashValue,omitempty"`
+}
+
+// ValueFromPodTemplateHash indicates which ReplicaSet pod template pod hash to use
+type ValueFromPodTemplateHash string
+
+const (
```
jessesuen: For consistency, our code words have had the first letter capitalized, so I think this should be `Stable` and `New`. 

Also, debating if the term should be `New`, `Canary`, `Current`?

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -196,8 +198,44 @@ type CanaryStep struct {
 	Pause *RolloutPause `json:"pause,omitempty"`
 	// Experiment defines the experiment object that should be created
 	Experiment *RolloutExperimentStep `json:"experiment,omitempty"`
+	// AnalysisRun defines the AnalysisRun that will run for a step
+	AnalysisRun *RolloutAnalysisRun `json:"analysisRun,omitempty"`
```
jessesuen: So we have the field name `parallelAnalysis` above and `analysisRun` here, which is inconsistent, despite reusing the same spec. I was thinking it should be just `analysis` in both cases. Keep in mind we'll also need a field for this in the Experiment spec.

```yaml
  strategy:
    canary: 
      analysis:
        templateName: success-rate
```

```yaml
  strategy:
    canary: 
      steps:
      - setWeight: 20
      - pause: {duration: 300}
      - analysis:
          templateName: success-rate
```

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -142,6 +142,8 @@ type CanaryStrategy struct {
 	// at any time during the update is atmost 130% of original pods.
 	// +optional
 	MaxSurge *intstr.IntOrString `json:"maxSurge,omitempty"`
+	// ParellelAnalysis runs a separate analysisRun while all the steps execute. This is intended to be a continuous validation of the new ReplicaSet
+	ParellelAnalysis *RolloutAnalysisRun `json:"parellelAnalysis,omitempty"`
```
jessesuen: parellel is misspelled, but i think the field should be renamed to just `analysis` anyways.

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -196,8 +198,44 @@ type CanaryStep struct {
...
 	// Experiment defines the experiment object that should be created
 	Experiment *RolloutExperimentStep `json:"experiment,omitempty"`
+	// AnalysisRun defines the AnalysisRun that will run for a step
+	AnalysisRun *RolloutAnalysisRun `json:"analysisRun,omitempty"`
 }
 
+// RolloutAnalysisRun defines a template that is used to create a analysisRun
+type RolloutAnalysisRun struct {
+	// TemplateName reference of the AnalysisTemplate name used by the Rollout to create the run
+	TemplateName string `json:"templateName,omitempty"`
```
jessesuen: Remove omitempty which should make it required in OpenAPI validation

pkg/apis/rollouts/v1alpha1/types.go
dthomson25: I like `Current` or `New` because of the BlueGreen case and we can reuse this code works.

dthomson25: After discussing with @jessesuen, we decided to go with `Latest`

PR #186, opened 2019-10-05
Implement job-based metric provider

jessesuen:
This PR adds a job as a metric provider. Success is measured by exit code. If the job completes with exit code zero, then the metric is successful, otherwise it failed.


analysis/analysis.go
```diff
@@ -175,9 +180,17 @@ func (c *AnalysisController) runMeasurements(run *v1alpha1.AnalysisRun, tasks []
...
 			} else {
 				if t.incompleteMeasurement == nil {
-					newMeasurement, err = provider.Run(t.metric, run.Spec.Arguments)
+					newMeasurement, err = provider.Run(run, t.metric, run.Spec.Arguments)
 				} else {
-					newMeasurement, err = provider.Resume(t.metric, run.Spec.Arguments, *t.incompleteMeasurement)
+					// metric is incomplete. either terminate or resume it
+					if terminating {
+						newMeasurement, err = provider.Terminate(run, t.metric, run.Spec.Arguments, *t.incompleteMeasurement)
+						if err == nil && newMeasurement.Status == v1alpha1.AnalysisStatusSuccessful {
```
dthomson25: It feels weird to call a terminated metric succesful, but Idk what else I would call it

dthomson25: Also, while this section doesn't have too complex code, I think we should probably have some unit tests here since this is one of the key sections of the analysis runs.

PR #188, opened 2019-10-08
Add AnalysisStep

dthomson25:


utils/analysis/filter.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+		if ar.Name == r.Status.Canary.CurrentParellelAnalysisRun {
+			return true
+		}
+		return false
+	})
+}
+
+// FilterAnalysisRunsByRolloutType returns a list of analysisRuns that have the rollout-type of the typeFilter
+func FilterAnalysisRunsByRolloutType(analysisRuns []*v1alpha1.AnalysisRun, typeFilter string) ([]*v1alpha1.AnalysisRun, []*v1alpha1.AnalysisRun) {
```
jessesuen: Based on the caller of this function, seems like we only care about the matching case, so I feel like we don't need to return the second return value containing the non-matching runs.

rollout/canary.go
```diff
@@ -291,30 +303,39 @@ func (c *RolloutController) syncRolloutStatusCanary(olderRSs []*appsv1.ReplicaSe
...
 		}
-		newStatus = c.calculateRolloutConditions(r, newStatus, allRSs, newRS, currExp)
+		newStatus = c.calculateRolloutConditions(r, newStatus, allRSs, newRS, currExp, currArs)
 		return c.persistRolloutStatus(r, &newStatus, pointer.BoolPtr(false))
 	}
 
-	if completedCurrentCanaryStep(olderRSs, newRS, stableRS, currExp, r) {
+	stepArs, _ := analysisutil.FilterAnalysisRunsByRolloutType(currArs, v1alpha1.RolloutTypeStepLabel)
+	var currStepAr *v1alpha1.AnalysisRun
+	if len(stepArs) == 1 {
```
jessesuen: This len == 1 check seems fragile? If it's expected to only have one value, then maybe the helper should be changed to just GetCurrentStepAnalysisRun() and it should return nil or the single run value.

utils/analysis/filter.go
dthomson25: Agreed

utils/analysis/filter.go
```diff
@@ -0,0 +1,75 @@
...
+		return ar.Name == name
+	})
+	if len(analysisRunsByName) == 1 {
+		return analysisRunsByName[0]
+	}
+	return nil
+}
+
+// FilterAnalysisRunsByStepIndex returns a list of analysisRuns that have the step index provided
+func FilterAnalysisRunsByStepIndex(analysisRuns []*v1alpha1.AnalysisRun, indexPtr *int32) ([]*v1alpha1.AnalysisRun, []*v1alpha1.AnalysisRun) {
```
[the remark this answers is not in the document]
dthomson25: If we have an inconclusive run and the analysis run is recreated after being unpaused, we can have multiple runs with the same step-index. With that being said, this method isn't being used

rollout/analysis.go
```diff
@@ -0,0 +1,144 @@
...
+	step, index := replicasetutil.GetCurrentCanaryStep(rollout)
+	currentAr := analysisutil.FilterAnalysisRunsByName(currentArs, rollout.Status.Canary.CurrentStepAnalysisRun)
+	if step == nil || step.Analysis == nil || index == nil {
+		_, err := c.cancelAnalysisRuns(rollout, []*v1alpha1.AnalysisRun{currentAr})
+		if err != nil {
+			return nil, err
+		}
+		return nil, err
+	}
+	if currentAr == nil && !rollout.Spec.Paused {
```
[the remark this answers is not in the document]
dthomson25: It was possible to reach this point, but the check could occur earlier. I moved the pause check into the beginning of the reconcileAnalysisRun to address this.

PR #189, opened 2019-10-08
Simplify provider interfaces to set error messages

jessesuen:
* Rename AnalysisProvider to MetricProvider
* Add message field to measurements
* Simplify provider interface so that it never returns error, and error messages are expected to be set in the measurement.


providers/provider.go
```diff
@@ -16,11 +16,11 @@ import (
 type Provider interface {
```
dthomson25: To your point earlier, do we want to rerun the package too?

dthomson25:
Do we not have a unit test for executing the resume provider?

https://codecov.io/gh/argoproj/argo-rollouts/pull/189/diff#D11-194

providers/provider.go
jessesuen: Sure. I will rename to metric_providers

analysis/analysis.go
```diff
@@ -213,18 +213,14 @@ func (c *AnalysisController) runMeasurements(run *v1alpha1.AnalysisRun, tasks []
 					metricResult.Count++
 				case v1alpha1.AnalysisStatusError:
 					metricResult.Error++
+					log.Warnf("metric errored: %s", newMeasurement.Message)
```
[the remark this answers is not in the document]
jessesuen: Sure.

PR #192, opened 2019-10-09
Update k8s library dependencies to v1.16

jessesuen:
Updates golang to v1.13
Update kubernetes to v1.16


manifests/crds/analysis-run-crd.yaml
```diff
@@ -754,11 +754,619 @@ spec:
...
+                                                      required:
+                                                      - port
+                                                      type: object
+                                                    timeoutSeconds:
+                                                      format: int32
+                                                      type: integer
+                                                  type: object
+                                                resources:
+                                                  properties:
+                                                    limits:
```
dthomson25: We need to remove the validation on the limits/requests of the ephemeral containers like we do for regular containers

jessesuen: Nice catch. We also need to do this for initContainers too.

PR #194, opened 2019-10-09
Use GenerateName when creating experiment

dthomson25:


rollout/experiment.go
```diff
@@ -33,7 +33,7 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
 	}
 	experiment := &v1alpha1.Experiment{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:            experimentutil.ExperimentNameFromRollout(r),
+			GenerateName:    experimentutil.ExperimentGeneratedNameFromRollout(r),
```
jessesuen: I looked at the implementation of ExperimentGeneratedNameFromRollout and it seems the generated name does not end in a dash. If we are using `metadata.generateName`, names should end in dashes, so that the random suffix is separate from the prefix (kubernetes will not add a dash).

PR #196, opened 2019-10-11
Add background analysis runs for rollouts

dthomson25:


rollout/analysis.go
```diff
@@ -66,45 +72,74 @@ func (c *RolloutController) reconcileAnalysisRuns(rollout *v1alpha1.Rollout, cur
...
+}
+
+func getPodHash(rollout *v1alpha1.Rollout, newRS *appsv1.ReplicaSet) string {
+	// Since the compute hash function is not guaranteed to be stable, we will use the podHash attached the newRS if possible.
+	if newRS != nil {
+		if podHash, ok := newRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]; ok {
+			return podHash
+		}
+	}
+	return controller.ComputeHash(&rollout.Spec.Template, rollout.Status.CollisionCount)
```
jessesuen: Do we ever expect to ever get to this call for controller.ComputeHash() ? I feel like it is an error to ever have newRS not have v1alpha1.DefaultRolloutUniqueLabelKey. I think if we don't have the hash in the labels of the newRS, I feel like we need to error so that this problem gets surfaced, otherwise we will be masking the problem.

Also we have replicasetutil.GetPodTemplateHash

dthomson25: I believe this case would only occur if a user edits the new RS and removes the label. I think it should be fine to just use the replicasetutil.GetPodTemplateHash and report an error if it's not there

PR #198, opened 2019-10-12
Fix regression where canary would not progress past a pause step

jessesuen:
If a rollout was paused, it would go into syncReplicasOnly() which would call syncRolloutStatusCanary() which had no code path to unpause it ever, even after the specified duration. This change will allow it to progress to the next step after the pause duration by advancing the current step index.

Note that this change allows metadata to be updated (e.g. advancing of currentStepIndex, marking the rollout complete, etc...). This should be safe to do since we never touch replicasets weights (aside from whatever happens from scaling events).

### DIFF
--- a/rollout/canary_test.go
+++ b/rollout/canary_test.go
@@ -1046,3 +1046,55 @@ func TestCanaryRolloutScaleWhilePaused(t *testing.T) {
 	expectedPatch := calculatePatch(r2, OnlyObservedGenerationPatch)
 	assert.Equal(t, expectedPatch, patch)
 }
+
+func TestResumeRolloutAfterPauseDuration(t *testing.T) {
+	f := newFixture(t)
+	defer f.Close()
+
+	steps := []v1alpha1.CanaryStep{
+		{
+			SetWeight: pointer.Int32Ptr(10),
+		},
+		{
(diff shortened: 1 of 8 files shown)


rollout/canary_test.go
```diff
@@ -1046,3 +1046,55 @@ func TestCanaryRolloutScaleWhilePaused(t *testing.T) {
...
+		{
+			SetWeight: pointer.Int32Ptr(20),
+		},
+	}
+	r1 := newCanaryRollout("foo", 1, nil, steps, pointer.Int32Ptr(1), intstr.FromInt(1), intstr.FromInt(1))
+	rs1 := newReplicaSetWithStatus(r1, 1, 1)
+	rs1PodHash := rs1.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
+	r1 = updateCanaryRolloutStatus(r1, rs1PodHash, 1, 1, 1, true)
+	r1.Spec.Paused = true
+	overAMinuteAgo := metav1.Time{time.Now().Add(-61 * time.Second)}
```
golangcibot: composites: `github.com/argoproj/argo-rollouts/vendor/k8s.io/apimachinery/pkg/apis/meta/v1.Time` composite literal uses unkeyed fields (from `govet`)

Issue #199, opened 2019-10-15
Implement abort logic


jessesuen:
During an abort we should:

1. set a new `status.aborted: true` field
2. improve the inferred weight to consider this field
3. set current step index to zero or nil
4. ensure we handle the case when there are no steps defined (normal rolling update)

PR #224, opened 2019-10-25
Add abort functionality

dthomson25:
Future work:
- add abort status into the list kubectl plugin command

### DIFF
--- a/rollout/analysis.go
+++ b/rollout/analysis.go
@@ -104,6 +104,9 @@ func (c *RolloutController) reconcileBackgroundAnalysisRun(roCtx *canaryContext)
 		}
 		return currentAr, err
 	}
+	if currentAr.Status != nil && currentAr.Status.Status == v1alpha1.AnalysisStatusInconclusive {
+		roCtx.PauseContext().AddPauseCondition(v1alpha1.PauseReasonInconclusiveAnalysis)
+	}
 	return currentAr, nil
 }
 
--- a/rollout/analysis.go
+++ b/rollout/analysis.go
@@ -41,19 +41,19 @@ func (c *RolloutController) getAnalysisRunsForRollout(rollout *v1alpha1.Rollout)
 	return ownedByRollout, nil
 }
 
-func (c *RolloutController) reconcileAnalysisRuns(roCtx *canaryContext) error {
+func (c *RolloutController) reconcileAnalysisRuns(roCtx *canaryContext) (bool, error) {
 	rollout := roCtx.Rollout()
 	otherArs := roCtx.OtherAnalysisRuns()
 	if rollout.Status.Abort {
 		allArs := append(roCtx.CurrentAnalysisRuns(), otherArs...)
-		return c.cancelAnalysisRuns(roCtx, allArs)
+		return false, c.cancelAnalysisRuns(roCtx, allArs)
 	}
 
 	newCurrentAnalysisRuns := []*v1alpha1.AnalysisRun{}
 
 	stepAnalysisRun, err := c.reconcileStepBasedAnalysisRun(roCtx)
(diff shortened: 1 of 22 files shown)


dthomson25:
Closes https://github.com/argoproj/argo-rollouts/issues/199 https://github.com/argoproj/argo-rollouts/issues/201 https://github.com/argoproj/argo-rollouts/issues/200

rollout/canary.go
```diff
@@ -64,7 +64,12 @@ func (c *RolloutController) rolloutCanary(rollout *v1alpha1.Rollout, rsList []*a
 	}
 
 	logCtx.Info("Reconciling AnalysisRun step")
-	if err = c.reconcileAnalysisRuns(roCtx); err != nil {
+	addPause, err := c.reconcileAnalysisRuns(roCtx)
+	if addPause {
```
jessesuen: Rather than surfacing a bool here, could we call

```golang
err := c.reconcileAnalysisRuns(roCtx)
if roCtx.PauseContext().HasAddPause() {
    logCtx.Info("Detected pause due to inconclusive AnalysisRun")
    return c.syncRolloutStatusCanary(roCtx)
}
if err != nil {
    return err
}
```

rollout/analysis.go
```diff
@@ -41,44 +41,49 @@ func (c *RolloutController) getAnalysisRunsForRollout(rollout *v1alpha1.Rollout)
 	return ownedByRollout, nil
 }
 
-func (c *RolloutController) reconcileAnalysisRuns(roCtx *canaryContext) error {
+func (c *RolloutController) reconcileAnalysisRuns(roCtx *canaryContext) (bool, error) {
```
jessesuen: See my other comment, I think instead of making sure we are returning true/false correctly in this function, I think the caller can check for `roCtx.PauseContext().HasAddPause()` after calling this function. With my suggested approach, it means we don't have to ensure we correctly return true/false everywhere, because there is a single source of truth.

rollout/canary.go
```diff
@@ -292,17 +297,40 @@ func (c *RolloutController) syncRolloutStatusCanary(roCtx *canaryContext) error
 	}
 
 	currBackgroundAr := analysisutil.GetCurrentBackgroundAnalysisRun(currArs)
-	if currBackgroundAr != nil {
-		if currBackgroundAr.Status == nil || !currBackgroundAr.Status.Status.Completed() || analysisutil.IsTerminating(currBackgroundAr) {
+	if currBackgroundAr != nil && !r.Status.Abort {
+		if currBackgroundAr.Status == nil || !currBackgroundAr.Status.Status.Completed() {
```
jessesuen: I'm not sure why I made AnalysisRun.Status a bool pointer. Seems like it just makes the code annoying to deal with.

rollout/analysis.go
dthomson25: Yeah I like that a lot more

dthomson25:
The clearing of abort is implicit because I do not set the newStatus.Abort to the old status value. As a result, I have a couple locations that add the abort:
https://github.com/argoproj/argo-rollouts/blob/6bba90cfcd30e446bc30a2f6fb5047f986b45ed1/rollout/canary.go#L304
https://github.com/argoproj/argo-rollouts/blob/6bba90cfcd30e446bc30a2f6fb5047f986b45ed1/rollout/canary.go#L317
https://github.com/argoproj/argo-rollouts/blob/6bba90cfcd30e446bc30a2f6fb5047f986b45ed1/rollout/canary.go#L312

rollout/canary.go
dthomson25: Yeah, it adds a lot of nil checks.

PR #236, opened 2019-10-29
Fix issue where a ReplicaSet name collision could cause hot loop

jessesuen:


experiments/replicaset.go
```diff
@@ -183,6 +167,45 @@ func (ec *experimentContext) createReplicaSet(template v1alpha1.TemplateSpec, co
...
+		Spec: appsv1.ReplicaSetSpec{
+			Replicas:        new(int32),
+			MinReadySeconds: template.MinReadySeconds,
+			Selector:        template.Selector,
+			Template:        newRSTemplate,
+		},
+	}
+}
+
+// isReplicaSetSemanticallyEqual checks to se if an existing ReplicaSet is semantically equal
```
dthomson25: nx: see

PR #238, opened 2019-10-29
Add analysis to RolloutExperimentStep

dthomson25:


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -152,7 +152,8 @@ type RolloutExperimentStep struct {
 	Duration int32 `json:"duration"`
 	// Templates what templates that should be added to the experiment. Should be non-nil
 	Templates []RolloutExperimentTemplate `json:"templates"`
-	//ProgressingDeadlineSeconds Is it ncessary?
+	//AnalysisTemplate what analysis to run with the experiment
+	Analysis *RolloutAnalysisStep `json:"analysis,omitempty"`
```
jessesuen: Shouldn't this be a list of []ExperimentAnalysisTemplateRef and be named `analyses`? This approach means we only support one analysis per experiment step.

dthomson25: I agree with having multiple analyses, but the ExperimentAnalysisTemplateRef doesn't work because it doesn't allow us to inject values into the analysis.

PR #239, opened 2019-10-30
Fix enqueue logic

jessesuen:
Enqueue logic had a few bugs in it:
* was not enqueuing properly after duration
* was unnecessarily enqueuing things that were not degraded


experiments/experiment.go
```diff
@@ -82,11 +82,12 @@ func (ec *experimentContext) reconcile() *v1alpha1.ExperimentStatus {
 		ec.reconcileAnalysisRun(analysis)
 	}
 
-	if duration := calculateEnqueueDuration(ec.ex); duration != nil {
```
dthomson25: Oh, we were only acting on the previous experiment... Yeah that would cause problems

PR #241, opened 2019-10-30
Make specifying replicas/duration optional in the experiment step

jessesuen:
This simplifies manifests by making replicas and duration as optional fields. Also it gets progressDeadlineSeconds from the rollout spec. 

Builds ontop of https://github.com/argoproj/argo-rollouts/pull/239, so only look at last commit.


jessesuen:
> LGTM, but quick question: If we have no duration, how does the experiment know when to stop and continue the rollout's steps? If the experiment has an analysis, then it should finish once the analysis completes. Otherwise, how will the experiment end and allow the rollout to continue? Is this a valid configuration? If not, we should add some validation to prevent the user from creating a never-ending experiment.

Sorry I forgot to answer this question. Never ending/indefinite experiments are a desired use case. To proceed to the next step of the rollout, a manual termination of the Experiment would need to be performed. 

>  If the experiment has an analysis, then it should finish once the analysis completes. 

Actually we may need to support both use cases. Currently stopping an experiment because all analysis completed, is not supported. In this scenario, the experiment will continue to run (to it's duration, or forever).

We will need new syntax to indicate that an Experiment should stop when analysis completes. This is a decision that might need to be made for individual analyses.

PR #246, opened 2019-10-30
Clean up Rollout Background AnalysisRuns

dthomson25:


utils/analysis/filter.go
```diff
@@ -115,6 +115,9 @@ func FilterAnalysisRunsToDelete(ars []*v1alpha1.AnalysisRun, olderRSs []*appsv1.
 	olderRsPodHashes := map[string]bool{}
 	for i := range olderRSs {
 		rs := olderRSs[i]
+		if rs == nil {
```
jessesuen: Just curious how this happens and if theres a bug somewhere, and we are addressing the symptom and not the problem. Worried that other parts of code expect olderRSs to always have valid ReplicaSets.

dthomson25: Discussed IRL, but the newRS can be nil if there is a scaling event and a spec change

Issue #240, opened 2019-10-30
ExperimentStep specRef does not support preview use case


jessesuen:
I was trying to come up with an example where an experiment could be used as part of a rollout step, to support a preview/staging replicaset. The requirement is that a preview replicaset should be able to run in the production namespace, but without receiving any traffic (because it does not have the same selector labels).

```yaml
  strategy:
    canary:
      steps:
      - experiment:
          duration: 30
          templates:
          - name: preview
            specRef: canary
```

The problem with the above example, is that `specRef: canary` causes the selector labels from the canary pod template spec to be carried forward. This means that it also includes labels like `app: guestbook` from the pod template, which means that traffic will be directed to the preview -- the opposite of what I'm trying to achieve.

While specRef works well in addressing the baseline vs. canary for mann-whitney analysis, it does not work well for the preview pod use case. We need to come up with syntax which allows for a canary replicaset to be brought up, but with a different set of selectors.

PR #249, opened 2019-10-31
Allow selectors to be overwritten when starting experiments

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/240

Also cleaned up and updated examples


rollout/experiment.go
```diff
@@ -62,9 +62,14 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
...
-		template.Selector = templateRS.Spec.Selector.DeepCopy()
+		if templateStep.Selector != nil {
+			template.Selector = templateStep.Selector.DeepCopy()
+		} else {
+			template.Selector = templateRS.Spec.Selector.DeepCopy()
+		}
+
 		if templateStep.Metadata.Labels != nil {
-			if templateStep.Metadata.Labels == nil {
+			if template.Template.ObjectMeta.Labels == nil {
```
jessesuen: This looked like a bug to me. Please confirm the intent.

dthomson25: Yeah that's a bug

Issue #253, opened 2019-11-01
Failed/Inconclusive Experiments do not abort/pause the rollout


jessesuen:
It seems that after my changes to add status to experiments, an Experiment does not seem to affect the rollout. Failed experiments should abort the rollout, and Inconclusive experiments should pause the rollout, just like analyses.

PR #255, opened 2019-11-01
Use deterministic names for experiments and analysis

jessesuen:
Experiments and analysis runs created from rollout objects will have deterministic names, preventing duplicates from being created.

Resolves https://github.com/argoproj/argo-rollouts/issues/243

PR #256, opened 2019-11-02
Pause and abort a rollout upon inconclusive and failed experiment

jessesuen:
Built ontop of https://github.com/argoproj/argo-rollouts/pull/255. 

View last commit only.

Resolves https://github.com/argoproj/argo-rollouts/issues/253

### DIFF
--- a/rollout/experiment.go
+++ b/rollout/experiment.go
@@ -139,6 +139,10 @@ func (c *RolloutController) reconcileExperiments(roCtx *canaryContext) error {
 		return c.cancelExperiments(roCtx, allExs)
 	}
 
+	if roCtx.PauseContext().GetPauseCondition(v1alpha1.PauseReasonInconclusiveAnalysis) != nil {
+		return nil
+	}
+
 	step, _ := replicasetutil.GetCurrentCanaryStep(rollout)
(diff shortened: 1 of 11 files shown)


rollout/experiment.go
```diff
@@ -167,6 +171,15 @@ func (c *RolloutController) reconcileExperiments(roCtx *canaryContext) error {
 		roCtx.SetCurrentExperiment(currentEx)
 	}
 
+	if currentEx != nil {
+		switch currentEx.Status.Status {
+		case v1alpha1.AnalysisStatusInconclusive:
+			roCtx.PauseContext().AddPauseCondition(v1alpha1.PauseReasonInconclusiveExperiment)
+		case v1alpha1.AnalysisStatusError, v1alpha1.AnalysisStatusFailed:
+			roCtx.PauseContext().AddAbort()
```
dthomson25: Do we still need the ExperimentFailed flag if we have an aborted field? The ExperimentFailed flag will only be set to true if the abort flag is true and I can't think of a reason to have that granularity

jessesuen: I don't think we need it. I was going to ask you the same question.

jessesuen: I'll remove it

Issue #254, opened 2019-11-01
Metric jobs need to be cleaned up


jessesuen:
I don't believe I implemented any clean up of jobs used as a metric.

PR #258, opened 2019-11-03
Use deterministic job names. Garbage collect old measurements

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/254

1. Use deterministic job names to ensure we do not create extra jobs as a result of operating on stale informer cache.
2. jobs created from analysisruns no longer incorporate the analysis run name into the job name, which can easily be > 63 characters. This is a problem since the k8s job controller uses the job name in the pod template labels, which when > 63 characters, caused jobs creation to fail.
3. Garbage collects old measurements

Built on top of https://github.com/argoproj/argo-rollouts/pull/256. See last commit only.


davidxia:
Are there any plans to add GC at the AnalysisRun level? Like keep only 10 most recent AnalysisRuns from a single Rollout?

PR #263, opened 2019-11-05
Add `get/retry experiment` commands. Support experiment retries

jessesuen:
image

* Add `kubectl argo rollouts get experiment NAME`
* Add `kubectl argo rollouts retry experiment NAME`

Support retrying experiments by using a collision counter for analysis runs in experiments.

PR #266, opened 2019-11-06
Rename AnalysisStatus to AnalysisPhase

jessesuen:
Built ontop of https://github.com/argoproj/argo-rollouts/pull/263. Only look at last commit


jessesuen:
I switched to Intellij just for this PR, so it wasn't so bad

PR #286, opened 2019-11-13
Use a duration string instead of int to represent duration

jessesuen:


test/e2e/functional/experiment-mann-whitney-u.yaml
```diff
@@ -3,7 +3,7 @@ kind: Experiment
 metadata:
   generateName: experiment-mann-whitney-u-
 spec:
-  duration: 3600
+  duration: 3h
```
dthomson25: I don't think it matters, but this was previous 1 hour not 3

PR #295, opened 2019-11-14
Introduce 'kubectl argo rollouts create' command

jessesuen:
1. Introduce `kubectl argo rollouts create -f file.yaml` with a `-w` watch option
2. Introduce `kubectl argo rollouts create analysisrun` with ` --from=templatename` or ` --from-file=template.yaml`
3. Made generating an AnalysisRun from a AnalysisTemplate into a helper
4. Always show the last job when displaying analysis jobs (not just the failed ones)

### DIFF
--- a/pkg/kubectl-argo-rollouts/cmd/create/create.go
+++ b/pkg/kubectl-argo-rollouts/cmd/create/create.go
@@ -1,0 +1,279 @@
+package create
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"strings"
+	"unicode"
+
+	"github.com/ghodss/yaml"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts"
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/cmd/get"
+	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/options"
+	analysisutil "github.com/argoproj/argo-rollouts/utils/analysis"
+)
+
+type CreateOptions struct {
+	get.GetOptions
+	options.ArgoRolloutsOptions
+
+	Files    []string
+	From     string
+	FromFile string
+}
+
+type CreateAnalysisRunOptions struct {
+	options.ArgoRolloutsOptions
+
+	Name         string
+	GenerateName string
+	ArgFlags     []string
+	From         string
+	FromFile     string
+}
+
+// NewCmdCreate returns a new instance of an `rollouts create` command
+func NewCmdCreate(o *options.ArgoRolloutsOptions) *cobra.Command {
+	createOptions := CreateOptions{
+		ArgoRolloutsOptions: *o,
+	}
+	var cmd = &cobra.Command{
+		Use:   "create -f my-experiment.yaml",
+		Short: "Create a rollout resource",
+		Example: o.Example(`
+  # Create a resource and watch it
+  %[1]s create -f my-experiment.yaml -w
+`),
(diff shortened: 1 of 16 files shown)


pkg/kubectl-argo-rollouts/cmd/create/create.go
```diff
@@ -0,0 +1,279 @@
...
+
+// isJSON detects if the byte array looks like json, based on the first non-whitespace character
+func isJSON(fileBytes []byte) bool {
+	for _, b := range fileBytes {
+		if !unicode.IsSpace(rune(b)) {
+			return b == '{'
+		}
+	}
+	return false
+}
```
dthomson25: I'm surprised there isn't an easier way of doing this with the libraries we imported.

dthomson25: That being said, I only do a quick search to see if I find one, but otherwise, I think it's fine

pkg/kubectl-argo-rollouts/cmd/create/create.go
```diff
@@ -0,0 +1,279 @@
...
+		SilenceUsage: true,
+		RunE: func(c *cobra.Command, args []string) error {
+			froms := 0
+			if createOptions.From != "" {
+				froms++
+			}
+			if createOptions.FromFile != "" {
+				froms++
+			}
+			if froms != 1 {
```
dthomson25: Is this code the same as
```
if createOptions.FromFile != "" &&  createOptions.From != "" {
```?

Comment (dthomson25) [reply pkg/kubectl-argo-rollouts/cmd/create/create.go]: Ah, it's not in the case where they are both empty. Nvm!

File: pkg/kubectl-argo-rollouts/cmd/create/create.go [hunk 3]
```diff
@@ -0,0 +1,279 @@
...
+				template = &tmpl
+			}
+			var name, generateName string
+			if createOptions.Name != "" {
+				name = createOptions.Name
+			} else if createOptions.GenerateName != "" {
+				generateName = createOptions.GenerateName
+			} else {
+				generateName = template.Name + "-"
+			}
```
dthomson25: Nix: I'm wondering if it would make the function easier to read if we make these lines into two helper functions:

```
template := getAnalysisTemplate(createOptions)
name, generateName := getAnalysisRunName(createOptions)
```

pkg/kubectl-argo-rollouts/cmd/create/create.go
```diff
@@ -0,0 +1,279 @@
...
+	o.AddKubectlFlags(cmd)
+	cmd.Flags().StringVar(&createOptions.Name, "name", "", "Use the specified name for the run")
+	cmd.Flags().StringVar(&createOptions.GenerateName, "generate-name", "", "Use the specified generateName for the run")
+	cmd.Flags().StringArrayVarP(&createOptions.ArgFlags, "argument", "a", []string{}, "Arguments to the parameter template")
+	cmd.Flags().StringVar(&createOptions.From, "from", "", "Create an AnalysisRun from an AnalysisTemplate in the cluster")
+	cmd.Flags().StringVar(&createOptions.FromFile, "from-file", "", "Create an AnalysisRun from an AnalysisTemplate in a local file")
+	return cmd
+}
+
+func (c *CreateAnalysisRunOptions) ParseArgs() ([]v1alpha1.Argument, error) {
```
dthomson25: Nix: I thought this function was parsing command line args instead of the args for the analysisRun when I saw it earlier in the code. Can we rename this to `ParseAnalysisArgs` to make it more clear?

pkg/kubectl-argo-rollouts/cmd/create/create.go
dthomson25: I have no strong opinions either way so feel free to dismiss this comment

pkg/kubectl-argo-rollouts/cmd/create/create.go
dthomson25: I guess the `templateArgs` in `templateArgs, err := createOptions.ParseArgs()` makes that more clear, but I didn't connect those dots when I looked at it the first time.

jessesuen: Actually it is parsing CLI args and returning template args, but I can rename

pkg/kubectl-argo-rollouts/cmd/create/create.go
jessesuen: sure.

PR #297, opened 2019-11-15
Introduce 'kubectl argo rollouts terminate' command

dthomson25:


pkg/kubectl-argo-rollouts/cmd/terminate/terminate.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	cmd.AddCommand(NewCmdTerminateAnalysisRun(o))
+	cmd.AddCommand(NewCmdTerminateExperiment(o))
+	return cmd
+}
+
+// NewCmdTerminateAnalysisRun returns a new instance of an `argo rollouts terminate analysisRun` command
+func NewCmdTerminateAnalysisRun(o *options.ArgoRolloutsOptions) *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:     "analysisRun ANALYSISRUN",
```
jessesuen: analysisRun should be all lowercase since this is consistent on how it will be used in kubectl (e.g. `kubectl get analysisruns`), and also because CLIs generally do not use camel case.

Issue #305, opened 2019-11-23
Rollout does not progress current stepIndex when "stable" RS is gone


jessesuen:
I intentionally borked a rollout by deleting the stable replicaset in the middle of an update.

The rollout ended up in the following condition:

```
Name:            canary-demo
Namespace:       jesse-test
Status:          ◌ Progressing
Strategy:        Canary
  Step:          0/2
  SetWeight:     1
  ActualWeight:  100
Images:          argoproj/rollouts-demo:yellow (canary)
Replicas:
  Desired:       2
  Current:       2
  Updated:       2
  Ready:         2
  Available:     2

NAME                                     KIND        STATUS         AGE    INFO
⟳ canary-demo                            Rollout     ◌ Progressing  170m
├──# revision:11
│  └──⧉ canary-demo-6758949f55           ReplicaSet  ✔ Healthy      2m33s  canary
│     ├──□ canary-demo-6758949f55-fdzm5  Pod         ✔ Running      2m33s  ready:1/1
│     └──□ canary-demo-6758949f55-kmmtr  Pod         ✔ Running      2m33s  ready:1/1
├──# revision:10
│  └──⧉ canary-demo-645d5dbc4c           ReplicaSet  • ScaledDown   17m
└──# revision:9
   └──⧉ canary-demo-877894d5b            ReplicaSet  • ScaledDown   18m
```

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"argoproj.io/v1alpha1","kind":"Rollout","metadata":{"annotations":{},"name":"canary-demo","namespace":"jesse-test"},"spec":{"replicas":2,"revisionHistoryLimit":2,"selector":{"matchLabels":{"app":"canary-demo"}},"strategy":{"canary":{"steps":[{"setWeight":1},{"pause":{}}]}},"template":{"metadata":{"labels":{"app":"canary-demo"}},"spec":{"containers":[{"image":"argoproj/rollouts-demo:blue","imagePullPolicy":"Always","name":"canary-demo","ports":[{"containerPort":8080,"name":"http","protocol":"TCP"}],"resources":{"requests":{"cpu":"5m","memory":"32Mi"}}}]}}}}
    rollout.argoproj.io/revision: "11"
  creationTimestamp: "2019-11-22T23:15:13Z"
  generation: 102
  name: canary-demo
[...]

Issue #304, opened 2019-11-23
Support instance ids for rollout controller segregation


jessesuen:
We sometimes need to debug a rollout in a namespace but do not wish to fight with the cluster scoped controller. 

Similar to how workflows uses an instance-id label to segregate controllers, the rollout controller could use this technique as well. This will also help with running new versions of the rollout controller for users who wish to opt-in to additional features.

PR #306, opened 2019-11-23
Set StableRS hash to current if replicaset does not actually exist

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/305
Will add a unit test

### DIFF
--- a/rollout/canary.go
+++ b/rollout/canary.go
@@ -1,6 +1,7 @@
 package rollout
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


dthomson25:
You mean that this PR  solves https://github.com/argoproj/argo-rollouts/issues/305 not #304 right?

dthomson25:
Closing in favor of https://github.com/argoproj/argo-rollouts/pull/320

Issue #301, opened 2019-11-20
troubles with previewReplicaCount in BlueGreen rollout strategy


vranystepan:
Hi folks,

When I'm experimenting with `previewReplicaCount`, I'm getting to the situation when it's creating and killing replicas in the preview RS:

```
⟳ rollout-bluegreen                            Rollout     ॥ Paused       35s  
├──# revision:2                                                                
│  └──⧉ rollout-bluegreen-5f49884f5c           ReplicaSet  ◌ Progressing  8s   preview
│     ├──□ rollout-bluegreen-5f49884f5c-qpvxn  Pod         ✔ Running      8s   ready:1/1
│     ├──□ rollout-bluegreen-5f49884f5c-z2snd  Pod         ✔ Running      8s   ready:1/1
│     ├──□ rollout-bluegreen-5f49884f5c-4b6vt  Pod         ◌ Terminating  0s   ready:0/1
│     ├──□ rollout-bluegreen-5f49884f5c-69zzs  Pod         ◌ Terminating  0s   ready:0/1
│     ├──□ rollout-bluegreen-5f49884f5c-6k69q  Pod         ◌ Terminating  0s   ready:0/1
│     ├──□ rollout-bluegreen-5f49884f5c-cdhmz  Pod         ◌ Terminating  0s   ready:0/1
│     ├──□ rollout-bluegreen-5f49884f5c-dk2kk  Pod         ◌ Terminating  0s   ready:0/1
│     ├──□ rollout-bluegreen-5f49884f5c-rxt9k  Pod         ◌ Terminating  0s   ready:0/1
│     ├──□ rollout-bluegreen-5f49884f5c-wvgqr  Pod         ◌ Terminating  0s   ready:0/1
│     └──□ rollout-bluegreen-5f49884f5c-x256k  Pod         ◌ Terminating  0s   ready:0/1
└──# revision:1                                                                
   └──⧉ rollout-bluegreen-7d6b6cb796           ReplicaSet  ✔ Healthy      34s  active
      ├──□ rollout-bluegreen-7d6b6cb796-2wkjr  Pod         ✔ Running      34s  ready:1/1
      ├──□ rollout-bluegreen-7d6b6cb796-5svlv  Pod         ✔ Running      34s  ready:1/1
      ├──□ rollout-bluegreen-7d6b6cb796-7lmcx  Pod         ✔ Running      34s  ready:1/1
[...]

PR #308, opened 2019-11-27
Fix Infinite loop with PreviewReplicaCount set

dthomson25:
Solves https://github.com/argoproj/argo-rollouts/issues/301 by removing the scaling preview ReplicaSet section in the`scaleBlueGreen` method. That section would scale up the preview ReplicaSet to the `.spec.Replicas` count without taking the rollout pause status into account (https://github.com/argoproj/argo-rollouts/compare/fix-infinite-loop?expand=1#diff-cfad87918ee0d3175ebd8e47f5efd159L309 vs https://github.com/argoproj/argo-rollouts/blob/master/utils/replicaset/replicaset.go#L89). The rollout should only scale to the `.spec.Replicas` count when the rollout is paused and the ScaleUpPreviewCheckPoint is set to true.

The `scaleBlueGreen` method also scales the new ReplicaSet. Since the new ReplicaSet is the same as the preview ReplicaSet, the controller scales it to the `.spec.strategy.bluegreen.previewReplicaCount` count (as the rollout is still paused). These two updates on the same ReplicaSet cause an infinite loop as they fight over the `.spec.Replica` value.

We did not hit this bug earlier since this code only happened while the scaling. With v0.6, the controller leverages that code path whenever the rollout is paused.

In addition, I added some tests to confirm the behavior of when to set the ScaleUpPreviewCheckPoint value.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
     docker:
       # CircleCI Go images available at: https://hub.docker.com/r/circleci/golang/
       - image: circleci/golang:1.13.1
-    resource_class: large
+    resource_class: xlarge
     environment:
       TEST_RESULTS: /tmp/test-results
     steps:
--- a/rollout/replicaset.go
+++ b/rollout/replicaset.go
@@ -77,18 +77,6 @@ func (c *RolloutController) reconcileNewReplicaSet(roCtx rolloutContext) (bool, 
 	}
 	roCtx.Log().Infof("Reconciling new ReplicaSet '%s'", newRS.Name)
 	allRSs := roCtx.AllRSs()
-	if rollout.Spec.Strategy.BlueGreen != nil {
-		rolloutReplicas := defaults.GetReplicasOrDefault(rollout.Spec.Replicas)
-		if *(newRS.Spec.Replicas) == rolloutReplicas {
-			// Scaling not required.
-			return false, nil
-		}
-		if *(newRS.Spec.Replicas) > rolloutReplicas {
-			// Scale down.
-			scaled, _, err := c.scaleReplicaSetAndRecordEvent(newRS, rolloutReplicas, rollout)
-			return scaled, err
-		}
-	}
 	newReplicasCount, err := replicasetutil.NewRSNewReplicas(rollout, allRSs, newRS)
 	if err != nil {
 		return false, err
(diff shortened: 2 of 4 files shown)


rollout/replicaset.go
```diff
@@ -77,18 +77,6 @@ func (c *RolloutController) reconcileNewReplicaSet(roCtx rolloutContext) (bool,
 	}
 	roCtx.Log().Infof("Reconciling new ReplicaSet '%s'", newRS.Name)
 	allRSs := roCtx.AllRSs()
-	if rollout.Spec.Strategy.BlueGreen != nil {
-		rolloutReplicas := defaults.GetReplicasOrDefault(rollout.Spec.Replicas)
-		if *(newRS.Spec.Replicas) == rolloutReplicas {
-			// Scaling not required.
-			return false, nil
```
dthomson25: We have the same if statement in scaleReplicaSetAndRecordEvent() (https://github.com/argoproj/argo-rollouts/blob/master/rollout/sync.go#L295) so this check is not needed.

rollout/replicaset.go
```diff
@@ -77,18 +77,6 @@ func (c *RolloutController) reconcileNewReplicaSet(roCtx rolloutContext) (bool,
...
-		}
-		if *(newRS.Spec.Replicas) > rolloutReplicas {
```
dthomson25: Technically, if someone set their previewReplicaCount to greater than the `.Spec.Replica`, the controller would get into an infinite loop as it jump between the two values as a result of this check

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     docker:
       # CircleCI Go images available at: https://hub.docker.com/r/circleci/golang/
       - image: circleci/golang:1.13.1
-    resource_class: large
```
dthomson25: I was getting repeated failed workflows with the following error
```
/usr/local/go/pkg/tool/linux_amd64/link: signal: killed
```

https://app.circleci.com/jobs/github/argoproj/argo-rollouts/651
https://app.circleci.com/jobs/github/argoproj/argo-rollouts/652
https://app.circleci.com/jobs/github/argoproj/argo-rollouts/650

This Stackoverflow suggested it was a resourcing issue: https://stackoverflow.com/questions/52129612/golang-circleci-2-0-test-failing-with-signal-killed

Issue #311, opened 2019-11-27
seg fault in argo-rollouts pod


amit-handda:
I am using minikube(1.4.0) and k8s(1.16.0) on mac(catalina). 
Installed argo-rollouts in the above cluster, as described in the instructions on the site. 
Created a rollout of nginx to test the blue-green strategy (attached).

Faced **controller seg-fault** when I update the rollout template to upgrade/downgrade nginx image version (1.15 -> 1.16 -> 1.17, eg). It applied 1.15->1.16 correctly but crashed while applying the second transition (1.16->1.17). It crashed similarly in another iteration while I was doing second transition.

I faced it twice, consistently . Hence, reporting it. is it a known issue ? 

```
time="2019-11-27T22:58:05Z" level=info msg="Creating event broadcaster"
time="2019-11-27T22:58:05Z" level=info msg="Setting up event handlers"
time="2019-11-27T22:58:05Z" level=info msg="Setting up experiments event handlers"
time="2019-11-27T22:58:05Z" level=info msg="Setting up analysis event handlers"
time="2019-11-27T22:58:05Z" level=info msg="Waiting for controller's informer caches to sync"
time="2019-11-27T22:58:05Z" level=info msg="Enqueuing parent of default/nginx-66ffbc8db7: Rollout default/nginx"
time="2019-11-27T22:58:05Z" level=info msg="Enqueuing parent of default/nginx-6d464c748b: Rollout default/nginx"
time="2019-11-27T22:58:05Z" level=info msg="Starting Controllers"
time="2019-11-27T22:58:05Z" level=info msg="Started controller"
time="2019-11-27T22:58:05Z" level=info msg="Starting Metric Server at 0.0.0.0:8090"
time="2019-11-27T22:58:05Z" level=info msg="Starting Service workers"
time="2019-11-27T22:58:05Z" level=info msg="Started Service workers"
time="2019-11-27T22:58:05Z" level=info msg="Successfully synced" namespace=kube-system service=kube-dns
time="2019-11-27T22:58:05Z" level=info msg="Successfully synced" namespace=default service=argo-rollouts-metrics
time="2019-11-27T22:58:05Z" level=info msg="Starting Experiment workers"
time="2019-11-27T22:58:05Z" level=info msg="Started Experiment workers"
time="2019-11-27T22:58:05Z" level=info msg="Successfully synced" namespace=default service=kubernetes
[...]

PR #314, opened 2019-11-29
Bluegreen: allow preview service/replica sets to be replaced and fix sg fault in syncReplicasOnly

mathetake:
## tl;dr

- solves https://github.com/argoproj/argo-rollouts/issues/311 .
- allow users to replace preview service's replica sets under Bluegreen type rollouts without prior abortion 

## motivation

In the course of bug fixing, I found that it is impossible for users to update preview service/replica with the new replica set of different template when the current preview service is not empt (i.e., rollout resource is paused). As of now, to achieve the replacement, users must set `status.abort` prior to try new templates.

To make matters worse, argo-rollouts controller would be panic if we do so (this is the same reason as https://github.com/argoproj/argo-rollouts/issues/311).

### DIFF
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -144,6 +144,10 @@ func (c *RolloutController) reconcileBlueGreenPause(activeSvc, previewSvc *corev
 	}
 
 	newRS := roCtx.NewRS()
+	if newRS == nil {
+		return false
+	}
+
 	newRSPodHash := newRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
 
 	if _, ok := activeSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey]; !ok {
@@ -223,6 +227,14 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 	oldRSs := roCtx.OlderRSs()
 	allRSs := roCtx.AllRSs()
 	newStatus := c.calculateBaseStatus(roCtx)
+
+	if replicasetutil.CheckPodSpecChange(r, newRS) {
+		roCtx.PauseContext().ClearPauseConditions()
+		roCtx.PauseContext().RemoveAbort()
+		newStatus = c.calculateRolloutConditions(roCtx, newStatus)
+		return c.persistRolloutStatus(roCtx, &newStatus)
+	}
+
 	newStatus.AvailableReplicas = replicasetutil.GetAvailableReplicaCountForReplicaSets([]*appsv1.ReplicaSet{newRS})
 	previewSelector, ok := serviceutil.GetRolloutSelectorLabel(previewSvc)
 	if !ok {
@@ -233,6 +245,7 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 	if !ok {
 		activeSelector = ""
 	}
+
 	newStatus.BlueGreen.ActiveSelector = activeSelector
 	if newStatus.BlueGreen.ActiveSelector != r.Status.BlueGreen.ActiveSelector {
 		previousActiveRS, _ := replicasetutil.GetReplicaSetByTemplateHash(oldRSs, r.Status.BlueGreen.ActiveSelector)
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -232,9 +232,6 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 		roCtx.PauseContext().ClearPauseConditions()
 		roCtx.PauseContext().RemoveAbort()
 		newStatus = c.calculateRolloutConditions(roCtx, newStatus)
-		if err := c.persistRolloutStatus(roCtx, &newStatus); err != nil {
-			return err
-		}
 	}
 
 	newStatus.AvailableReplicas = replicasetutil.GetAvailableReplicaCountForReplicaSets([]*appsv1.ReplicaSet{newRS})
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -231,7 +231,6 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 	if replicasetutil.CheckPodSpecChange(r, newRS) {
 		roCtx.PauseContext().ClearPauseConditions()
 		roCtx.PauseContext().RemoveAbort()
-		newStatus = c.calculateRolloutConditions(roCtx, newStatus)
 	}
 
 	newStatus.AvailableReplicas = replicasetutil.GetAvailableReplicaCountForReplicaSets([]*appsv1.ReplicaSet{newRS})
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -232,7 +232,9 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 		roCtx.PauseContext().ClearPauseConditions()
 		roCtx.PauseContext().RemoveAbort()
 		newStatus = c.calculateRolloutConditions(roCtx, newStatus)
-		return c.persistRolloutStatus(roCtx, &newStatus)
+		if err := c.persistRolloutStatus(roCtx, &newStatus); err != nil {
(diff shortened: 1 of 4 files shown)


rollout/sync.go
```diff
@@ -231,6 +246,11 @@ func (c *RolloutController) syncReplicasOnly(r *v1alpha1.Rollout, rsList []*apps
 	// The controller wants to use the rolloutCanary method to reconcile the rolllout if the rollout is not paused.
 	// If there are no scaling events, the rollout should only sync its status
 	if r.Spec.Strategy.Canary != nil {
+		newRS, oldRSs, err := c.getAllReplicaSetsAndSyncRevision(r, rsList, false)
```
golangcibot: ineffectual assignment to `oldRSs` (from `ineffassign`)

rollout/sync.go
```diff
@@ -209,17 +209,32 @@ func (c *RolloutController) syncReplicasOnly(r *v1alpha1.Rollout, rsList []*apps
...
+
 		roCtx := newBlueGreenCtx(r, newRS, oldRSs)
 		previewSvc, activeSvc, err := c.getPreviewAndActiveServices(r)
 		if err != nil {
-			return nil
+			return err
+		}
+
+		// replace preview service's selector if the new replica creation happens
+		if previewSvc != nil && previewSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey] != newRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]{
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		if previewSvc != nil && previewSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey] != newRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey] {
```

dthomson25:
Hey @mathetake, thanks for taking the time to make a PR! I see this PR is addressing three different issues:
- Rollout getting stuck if it is paused and the PodSpec has changed
- The panic when the new ReplicaSet doesn't exist and the rollout is paused
- Aborting rollouts using the BlueGreen strategy

For the first issue, I have a concern that this approach will cause the controller to switch the active service's selector too early for the latest ReplicaSet if the rollout is using the `.spec.strategy.blueGreen.autoPromotionSeconds` field. The latest ReplicaSet will not wait for the entire auto-promotion seconds since the controller will use the old pause condition instead of resetting it. To address this, I think we should reset the rollout object by removing the pause condition and abort status when there is a PodSpec change.

In the canary case, the controller handles this issue by resetting the rollout in the `syncRolloutStatusCanary` method, which the controller calls in every reconciliation loop. The controller first checks if the Pod Template or Steps Changed (https://github.com/argoproj/argo-rollouts/blob/master/rollout/canary.go#L266) and then removes the pause condition and abort field (https://github.com/argoproj/argo-rollouts/blob/master/rollout/canary.go#L276-L276). 

We should have the same behavior for the BlueGreen strategy in the `syncRolloutStatusBlueGreen` method (https://github.com/argoproj/argo-rollouts/blob/master/rollout/bluegreen.go#L220) by checking for podSpec changes. We can leverage the `checkPodSpecChange` method (https://github.com/argoproj/argo-rollouts/blob/master/utils/replicaset/replicaset.go#L274) to achieve that behavior.

However, this does not necessarily resolve the panic in `reconcileBlueGreenPause`. To do so, we should return false if the new ReplicaSet is nil.

For aborting Bluegreen rollouts, the controller does not use the abort field for rollouts using the BlueGreen Strategy at the moment. I filed this [issue](https://github.com/argoproj/argo-rollouts/issues/316) to address this gap.

mathetake:
Hi @dthomson25 , thanks for looking into my PR!

> We should have the same behavior for the BlueGreen strategy in the syncRolloutStatusBlueGreen method (https://github.com/argoproj/argo-rollouts/blob/master/rollout/bluegreen.go#L220) by checking for podSpec changes. We can leverage the checkPodSpecChange method (https://github.com/argoproj/argo-rollouts/blob/master/utils/replicaset/replicaset.go#L274) to achieve that behavior.

This is exactly what i was thinking since my changes seem a little bit strange due to its discrepancy between canary and bluegreen's behavior.

give me some time to implement your idea

mathetake:
@dthomson25  in https://github.com/argoproj/argo-rollouts/pull/314/commits/fafa26daf0741ea8c575057a4398e1980f575a27 , i changed the patch according to your idea above.

スクリーンショット 2019-12-03 12 17 06
スクリーンショット 2019-12-03 12 17 27

it seems to work correctly except the handling `abort` field mentioned in https://github.com/argoproj/argo-rollouts/issues/316 .

I would appreciate if you cloud take a look

rollout/bluegreen.go
```diff
@@ -223,6 +227,14 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 	oldRSs := roCtx.OlderRSs()
 	allRSs := roCtx.AllRSs()
 	newStatus := c.calculateBaseStatus(roCtx)
+
+	if replicasetutil.CheckPodSpecChange(r, newRS) {
+		roCtx.PauseContext().ClearPauseConditions()
+		roCtx.PauseContext().RemoveAbort()
+		newStatus = c.calculateRolloutConditions(roCtx, newStatus)
+		return c.persistRolloutStatus(roCtx, &newStatus)
```
dthomson25: Actually, I don't think we should return early here since we definitely want to set the ActiveSelector, HPAReplicas, and HPAReplicas values. I think the previewService is okay to set since the next reconciliation will change it when the newRS is created (because the rollout is now unpaused). Lastly, the calculateScaleUpPreviewCheckPoint function will return false since it checks for a Pod Spec change which is what we would want for ScaleUpPreviewCheckPoint.

carlosdoordash: Has the expected selector state of the previewService changed between 0.5.0 and 0.6.0?

- In 0.5.0, the previewService will no longer select green (newRS) nor blue once unpaused.
- In 0.6.0, the previewService will continue to select green once promoted.

Along the same lines, what is the expected behavior of the previewService on rollback?

- Is it true in 0.5.0 that the previewService will no longer select green (newRS) nor blue on rollback?
- I assume in >0.6.0, the previewService will continue to select green on rollback???

rollout/bluegreen.go
```diff
@@ -223,6 +227,13 @@ func (c *RolloutController) syncRolloutStatusBlueGreen(previewSvc *corev1.Servic
 	oldRSs := roCtx.OlderRSs()
 	allRSs := roCtx.AllRSs()
 	newStatus := c.calculateBaseStatus(roCtx)
+
+	if replicasetutil.CheckPodSpecChange(r, newRS) {
+		roCtx.PauseContext().ClearPauseConditions()
+		roCtx.PauseContext().RemoveAbort()
+		newStatus = c.calculateRolloutConditions(roCtx, newStatus)
```
dthomson25: We also can remove the calculateRolloutConditions method call here since we call it at https://github.com/argoproj/argo-rollouts/pull/314/files#diff-cfad87918ee0d3175ebd8e47f5efd159R270 before the persist command.

dthomson25: After this change, we are good to merge!

rollout/bluegreen.go
dthomson25: @carlosdoordash, your first two bullets points describe the behaviors for the two versions correctly. 

Before I answer your two questions, I am assuming that by green, you mean the new version that needs to be rolled back, and blue is the healthy version. Please let me know if that assumption is incorrect. 

```
Is it true in 0.5.0 that the previewService will no longer select green (newRS) nor blue on rollback?
```

Yes, this was the behavior for v0.5 as the blue version (now becomes the newRS) is automatically promoted to active service, and the preview service's selector points at nothing.

```
I assume in >0.6.0, the previewService will continue to select green on rollback???
```

For rollback in v0.6, the blue version (now becomes the newRS) is automatically promoted to the active service and the preview service points at the ReplicaSet that matches the current Rollout Spec (which is blue in this case).

Please let me know if I should clarify more!

rollout/bluegreen.go
mathetake: fixed in https://github.com/argoproj/argo-rollouts/pull/314/commits/4a076172096e501afa982e683ee83bcae892294d !

Issue #177, opened 2019-10-02
Web metric provider


jessesuen:
Instead of prometheus, a generic ~webhook~ HTTP URL may be desired as a metric provider. This might look something like:

```yaml
metrics:
- name: my-webhook
  provider:
    web:
      url: http://my-website/api/my-metric
      successCondition: result.data.values[1] > 99
```

dthomson25:
One question for the implementation is how can users include something like an API key with the webhook if that API key is a secret?

jessesuen:
I think it should, but not necessarily for MVP.

jessesuen:
I don't think `webhook` is appropriate terminology for this.  Webhooks are used to POST event information payloads to external sources. Whereas this is really just performing an HTTP GET against some URL and interpreting the payload. 

I think this should be renamed to HTTP metric provider.

Furthermore, we also want reserve the webhook feature/concept for events like when analysis fails, to perform some other HTTP request. (e.g. if analysis fails, then perform an HTTP post call to slack)

allcentury:
@jessesuen - Amazing talk at kubecon this year, thanks again for showing off this great project.  We would also be interested in this as we use Datadog as our metrics provider.  I'd be happy to be involved in this implementation, especially if you already have ideas for it.

tomsanbear:
Hey ArgoRollouts team, thanks for the awesome work so far.

I've just drafted a PR for implementing this, and would appreciate some feedback as this is a feature I'd love to get implemented as soon as possible. I read the feedback above and included some basic options for basic GET calls to a metrics Rest API.

jessesuen:
Copying discussion from slack:
https://argoproj.slack.com/archives/CKTN88XFH/p1579644561025100 

cabrinha:
> Looks like the webmetric provider got implemented ... but is there any chance of datadog being implemented as a metric provider? https://github.com/argoproj/argo-rollouts/issues/177

tomsanbear:
> Implementing a more native ‘datadog’ metric provider wouldn’t be much work actually, if you use the prometheus as a reference for implementation.
Datadog APIs (for retrieving metrics) afaik also operate with GET calls, so the WebMetric implementation should be sufficient for a standin solution till someone contributes a native datadog integration.

jesse:
I agree with @tomsanbear that in theory, Data Dog should be now supported using Argo Rollouts with the imminent v0.7 release using the web metric provider, with the caveat that we currently don’t yet support referencing secrets as part of HTTP (auth) headers:
[...]

PR #318, opened 2019-12-03
feat: Web metric provider

tomsanbear:
Today I ran into a situation where it'd be nice to link external metrics to Argo Rollouts, but unfortunately my platform is not supported. I notice the issue open for the generic WebMetric and created a basic implementation (Issue #177)

I would appreciate some feedback before I spend any more time on it. If there are glaring implementation issues, or schema changes you might suggest that would be great (I'm new-ish to Golang).

Some notes on my implementation:
- I used JsonPath for specifying where to find the metric in the response body
- I assume it's a float at all times (didn't investigate it for now)
- Made timeout configurable for the http client, but did not restrict the context runtime

Cheers!


metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,203 @@
...
+		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Could not parse JSON body: %v", err)
+	}
+
+	buf := new(bytes.Buffer)
+	err = p.jsonParser.Execute(buf, data)
+	if err != nil {
+		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Could not find JSONPath in body: %s", err)
+	}
+	out := buf.String()
+	outAsFloat, err := strconv.ParseFloat(buf.String(), 64)
```
dthomson25: I'm concerned that only accepting float64 as a result does not provide enough flexibility for the web metrics. For example, if the endpoint pinged returned a struct like ```{
      "status": "success"
}```, the web provider couldn't check if the status field is equal to success.  To achieve both goals, the jsonParser should parse the response and pass that parsed object into the evaluate condition method. That should achieve both outcomes.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -267,3 +268,16 @@ type ScopeDetail struct {
 	Start  string `json:"start"`
 	End    string `json:"end"`
 }
+
+// I don't like this repetition, but it fits the pattern...
+type WebMetricMetric struct {
+	Url      string            `json:"url"`
+	Headers  []WebMetricHeader `json:"headers,omitempty"`
+	Timeout  int               `json:"timeout,omitempty"`
+	JsonPath string            `json:"jsonPath"`
```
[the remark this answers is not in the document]
dthomson25: Great idea using the JSON Path to parse the HTTP response!

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -267,3 +268,16 @@ type ScopeDetail struct {
 	Start  string `json:"start"`
 	End    string `json:"end"`
 }
+
+// I don't like this repetition, but it fits the pattern...
+type WebMetricMetric struct {
+	Url      string            `json:"url"`
+	Headers  []WebMetricHeader `json:"headers,omitempty"`
+	Timeout  int               `json:"timeout,omitempty"`
```
dthomson25: I'm debating if the Timeout field should be a string (i.e. 1s vs 1m)

metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"k8s.io/client-go/util/jsonpath"
+
+	"github.com/argoproj/argo-rollouts/utils/evaluate"
+	metricutil "github.com/argoproj/argo-rollouts/utils/metric"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	log "github.com/sirupsen/logrus"
+)
```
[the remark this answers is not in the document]
tomsanbear: No problem with this, will run the linter this time around.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -97,6 +97,7 @@ type MetricProvider struct {
 	// Prometheus specifies the prometheus metric to query
 	Prometheus *PrometheusMetric `json:"prometheus,omitempty"`
 	Kayenta    *KayentaMetric    `json:"kayenta,omitempty"`
+	WebMetric  *WebMetricMetric  `json:"webmetric,omitempty"`
```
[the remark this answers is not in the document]
tomsanbear: Thanks for letting me know about running the codegen 👍

metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,203 @@
...
+	startTime := metav1.Now()
+
+	// Measurement to pass back
+	measurement := v1alpha1.Measurement{
+		StartedAt: &startTime,
+	}
+
+	// Create request
+	request := &http.Request{
+		Method: "GET",
```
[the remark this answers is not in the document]
tomsanbear: My thinking on this one was more for an initial implementation, if we extend this to include other types of Method calls, we should in turn support submitting bodies. LMK your thoughts.

metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,203 @@
...
+		return metricutil.MarkMeasurementError(measurement, err)
+	}
+
+	for _, header := range metric.Provider.WebMetric.Headers {
+		request.Header.Set(header.Key, header.Value)
+	}
+
+	// Send Request
+	response, err := p.client.Do(request)
+	if err != nil || response.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
tomsanbear: Makes sense to me 👍

pkg/apis/rollouts/v1alpha1/analysis_types.go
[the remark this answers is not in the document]
tomsanbear: Yeah I think the JSON Path will be perfect for this use case, if in the future someone has a need for a non-JSON response type, can also be configured in the manifest 👍 For initial implementation I'm assuming JSON will be sufficient.

pkg/apis/rollouts/v1alpha1/analysis_types.go
tomsanbear: I'll leave that one up to you, For all my HTTP work in the past I've always had my timeouts in seconds (rarely in millis), if we are OK with that, then we can be a little more specific with the field name. Otherwise adding in a simple time string parse is fairly trivial. LMKYT

metricproviders/webmetric/webmetric.go
tomsanbear: Yeah I had this in mind as well... I'll dig into the evaluation code a bit more to get a more elegant solution here.

tomsanbear:
Thanks for the comments @dthomson25 
I'll push some changes in with what I've commented, and what you've mentioned this week. 

Cheers!

tomsanbear:
Alright, found some time today to fit this in. Got a smarter parser in there for interpreting between ints, floats, bools and strings.

Haven't bothered yet with adding in the alternate HTTP methods, I'll wait to hear back on the discussion.

dthomson25:
Also, can you rebase on master? We made a CI change in another commit that this PR can't use until rebasing with master. The change will make sure the Circle has enough resources to run the tests, and there shouldn't be any conflicts

metricproviders/webmetric/webmetric.go
```diff
@@ -52,21 +50,21 @@ func (p *Provider) Run(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric) v1alph
...
-	for _, header := range metric.Provider.WebMetric.Headers {
+	for _, header := range metric.Provider.Web.Headers {
 		request.Header.Set(header.Key, header.Value)
 	}
 
 	// Send Request
 	response, err := p.client.Do(request)
-	if err != nil || response.StatusCode != http.StatusOK {
+	if err != nil || response.StatusCode < 200 || response.StatusCode >= 300 {
 		return metricutil.MarkMeasurementError(measurement, err)
```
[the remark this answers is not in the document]
tomsanbear: Ah yes, my bad, will split the if statement.

metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Send Request
+	response, err := p.client.Do(request)
+	if err != nil || response.StatusCode < 200 || response.StatusCode >= 300 {
+		return metricutil.MarkMeasurementError(measurement, err)
+	}
+
+	value, status, err := p.parseResponse(metric, response)
+	if err != nil || response.StatusCode != http.StatusOK {
```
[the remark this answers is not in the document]
tomsanbear: No need for that now, will remove and just check for error with the response parsing.

metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,229 @@
...
+	measurement := v1alpha1.Measurement{
+		StartedAt: &startTime,
+	}
+
+	// Create request
+	request := &http.Request{
+		Method: "GET", // TODO maybe make this configurable....also implies we will need body templates
+	}
+
+	request.URL, err = url.Parse(metric.Provider.Web.URL)
```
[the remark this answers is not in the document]
tomsanbear: This one I actually did since we can't have a ':=' assignment combined with the field  assignment I have there. I could swap that out with a new var that then gets assigned to request.URL, I'll push that in the next commit, and we can nitpick that if it's a problem also.

metricproviders/webmetric/webmetric.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Bool conversion
+	inAsBool, err := strconv.ParseBool(in)
+	if err == nil {
+		return inAsBool
+	}
+
+	// Else
+	return in
```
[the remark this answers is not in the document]
tomsanbear: Ok, I'll take a look into this, thanks for the feedback on this portion!

metricproviders/webmetric/webmetric_test.go
```diff
@@ -0,0 +1,296 @@
...
+			webServerResponse: `{"key": [{"key2": {"value": 1}}]}`,
+			metric: v1alpha1.Metric{
+				Name:             "foo",
+				SuccessCondition: "result > 0",
+				FailureCondition: "result <= 0",
+				Provider: v1alpha1.MetricProvider{
+					Web: &v1alpha1.WebMetric{
+						// URL:      server.URL,
+						JSONPath: "{$.key[0].key2.value}",
+						Headers:  []v1alpha1.WebMetricHeader{v1alpha1.WebMetricHeader{Key: "key", Value: "value"}},
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
						Headers:  []v1alpha1.WebMetricHeader{{Key: "key", Value: "value"}},
```

tomsanbear:
I made the easy changes and added some additional testing. But will have to revisit the evaluation logic as discussed above when I have some time this week. 👍

dthomson25:
Hey @tomsanbear, I wanted to check in and see if you had a chance to get those last changes in. I totally understand if not with the holidays and everything the past couple of weeks. Please let me know if there's anything I can do to help!

tomsanbear:
> Hey @tomsanbear, I wanted to check in and see if you had a chance to get those last changes in. I totally understand if not with the holidays and everything the past couple of weeks. Please let me know if there's anything I can do to help!

Hey @dthomson25 it's as you said holidays have been particularly busy for me. I think this weekend at the latest I should be able to get this last bit finished up. I took a look at the code for the evaluation mechanism during some downtime and it seems fairly straight forward to add for this PR. Thanks for reaching out!

tomsanbear:
@dthomson25 So I did a quick code of what we discussed... not sure if it's along the same idea as what you were thinking. Just a few of my thoughts:
1. Is this a little too verbose? The way jsonpath handles this logic is mostly the problem here :thinking: 
2. Any thoughts on the error handling for the eval engine?

If you had a different approach in mind I'll be interested to hear it.

jessesuen:
We were trying this out, and `utils/analysis/factory.go` needs to be updated with this:

```golang
	if metric.Provider.Web != nil {
		numProviders++
	}
```

utils/evaluate/evaluate.go
```diff
@@ -1,13 +1,29 @@
...
+			}
+			panic(err)
+		},
+		"asFloat": func(in string) float64 {
+			inAsFloat, err := strconv.ParseFloat(in, 64)
+			if err == nil {
+				return inAsFloat
+			}
+			panic(err)
+		},
```
dthomson25: This is definitely in line with what I was thinking! For the error handling, we should add a recover at line 28 to catch these panics. From there, we can return an error saying `Unable to convert to Float` and capture that in the status. Otherwise, the panic will break the rest of the analysis run. I think it will make the comparisons more explicit for the users and gives them flexibility if needed.

dthomson25: This should also be a good candidate for some unit tests using the assert.Panic function

tomsanbear: Sounds good! I'll add that in later today 👍

mkdocs.yml
```diff
@@ -24,10 +24,14 @@ nav:
...
     - BlueGreen: features/bluegreen.md
     - Canary: features/canary.md
+<<<<<<< HEAD
+    - Traffic Management: features/traffic-management/index.md
+=======
     - Traffic Management: 
       - Overview: features/traffic-management/index.md
       - Istio: features/traffic-management/istio.md 
       - NGINX: features/traffic-management/nginx.md 
+>>>>>>> c79494b99e33e54ab9734f885c196d1d81e8f1f6
```
dthomson25: Looks like a merge conflict got through

utils/evaluate/evaluate.go
tomsanbear: So, I added tests, and I did also add the recover clause. But it may be unecessary after playing around with the evaluation library. From the looks of it, it already wraps panics from internal functions as errors with nice pretty print debug messages. 
```
got expected error: strconv.ParseFloat: parsing "NotANum": invalid syntax (1:9)
        	            	 | asFloat(result) == 1.1
        	            	 | ........^
```

mkdocs.yml
tomsanbear: Fixed :+1:

tomsanbear:
Nope! I am happy with this for a first implementation as this achieves what I was looking to have for pulling other metrics and stats.

PR #319, opened 2019-12-03
Update dependencies for istio/client-go

dthomson25:
To introduce the istio/client-go library, I needed to upgrade the K8s libraries to v1.17. As a result, I needed to:
- Change the query method on the Prometheus API interface and update the mock to match the changes in the interface
- Add the `listType` go tag to all the types.go to start preparing for server-side apply. Check out https://github.com/kubernetes/kube-openapi/issues/175 for the API violations that would occur without these tags and https://godoc.org/k8s.io/kube-openapi/pkg/idl for what tags are available.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,12 +2,12 @@
 
 
 [[projects]]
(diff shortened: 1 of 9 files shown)


dthomson25:
Closing for now

PR #331, opened 2019-12-15
chore: Migrate from dep to go modules

lcostea:
Migrate from `dep` to go modules:

- Removed the files used by `dep` to install dependencies
- Moved to using `go.mod` 
- Updated documentation
- Updated ci jobs to stop using dep


lcostea:
@dthomson25 Thanks, I started looking at other projects on how they handle the code-gen and `mod vendor` is used by all, I couldn't see an alternative. So I will give that a try.

dthomson25:
Hey @lcostea, I wanted to check in to see if you had a chance to try to make those changes. If not worries! I know the end of the year can get busy at times!

lcostea:
@dthomson25 Just got back today from the New Year's vacation, I will get back on track with everything in the following days.

cmd/rollouts-controller/main.go
```diff
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
+	_ "k8s.io/code-generator/cmd/client-gen/generators"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

tools.go
```diff
@@ -0,0 +1,8 @@
+// This package contains code generation and build utilities
+// This package imports things required by build scripts, to force `go mod` to see them as dependencies
+package tools
+
+import (
+	_ "k8s.io/code-generator/cmd/client-gen/generators"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

lcostea:
@dthomson25 Re the codegen I found a solution that doesn't need vendoring, inspired from https://github.com/weaveworks/flagger/blob/master/hack/update-codegen.sh

dthomson25:
Awesome! That's great that we don't need to vendor the folder!

lcostea:
@dthomson25 only now I saw that all my CircleCI pipelines have been failing.
Let me work on fixing those first, no need for any review yet

dthomson25:
I took a look at your failed pipeline and I think the Prometheus library is getting upgraded and that version introduces a new method to the API interface which the mock doesn't implement.

lcostea:
Thanks, passed that step by installing `golangci-lint` binary directly and not as a dependency. That `go get` was modifying the `go.mod`, while before with `dep` it didn't have any effect.
It looks ok now, I still have failures on modules installs and builds, but it looks like an OOM issue: I get a lot of `/usr/local/go/pkg/tool/linux_amd64/link: signal: killed` and locally it passes when I run the same commands in the circleci container.
So it looks ready to be reviewed

dthomson25:
Hey @lcostea, I pushed your branch to the main repo so the code could take advantage of the increased resources. Looks like it passed!

I noticed two issues when I ran through the Makefile:
1. I got `line 10: realpath: command not found` when I ran ./hack/update-codegen.sh. When I added the first answer from [this stackoverflow](https://stackoverflow.com/questions/3572030/bash-script-absolute-path-with-os-x/18443300), I was able to get it to work.


2. I think the version of the open-api gen is different as I see
```
modified:   pkg/apis/api-rules/violation_exceptions.list
modified:   pkg/apis/rollouts/v1alpha1/openapi_generated.go
```
Do you have this behavior too?

dthomson25:
Hey @lcostea, I talked with @jessesuen, and we decided that we will move forward to the new openapi gen version and address the changes in another PR. So, we can merge this once we get the realpath fix in and fix this last merge conflict.

lcostea:
@dthomson25 fixed the conflict
regarding `realpath` I see the command is available on the ci and I also checked alpine, debian, ubuntu, centos images and they all have it
On mac I see it is available on `coreutils` formula, so what about if I add this to the docs: you need to install coreutils with brew? I wouldn't want to use a dirty replacement, when the original one works on all major distros. Otherwise I will look for a standard replacement.

jessesuen:
> I wouldn't want to use a dirty replacement, when the original one works on all major distros. Otherwise I will look for a standard replacement.

`DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"`

Is a pretty standard bash trick to get the real directory of a script -- insofar as I wouldn't even call it that dirty. I prefer using something like this which is platform agnostic, versus increasing the barrier to contribution.

lcostea:
@jessesuen @dthomson25 using now what you suggested for getting the script directory

lcostea:
no other changes, please go ahead

PR #336, opened 2019-12-23
docs: Add Istio docs

dthomson25:
This PR adds the initial documentation for integrating Istio and Argo Rollouts. It talks at a high level about how the controller modifies Istio resources to achieve traffic splitting with configuration from a Rollout object.

### DIFF
--- a/docs/features/traffic-management/istio.md
+++ b/docs/features/traffic-management/istio.md
@@ -1,3 +1,88 @@
 # Istio
 
-TBD
+NOTE: This functionality will be available in the v0.7 version, and everything described below is subject to change.
+
+[Istio](https://istio.io/) is one of the most popular service mesh in the community and offers a rich feature-set to control the flow of traffic. Istio offers this functionality through a set of CRDs, and the Argo Rollouts controller modifies these resources to manipulate the traffic routing into the desired state. However, the Argo Rollouts controller modifies the Istio resources minimally to gives the developer flexibility while configuring their resources.
+
+## Istio and Rollouts
+
+The Argo Rollout controller achieves traffic shaping by manipulating the Virtual Service. A Virtual Service provides the flexibility to define how to route traffic when a host address is hit.  The Argo Rollouts controller manipulates the Virtual Service by using the following Rollout configuration:
+
+- Canary Service name
+- Stable Service name
+- Virtual Service Name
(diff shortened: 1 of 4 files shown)


docs/features/traffic-management/istio.md
```diff
@@ -1,3 +1,88 @@
...
+Additionally, if a GitOps tool does an apply after the Argo Rollouts controller changes the Virtual Service's weight, the apply would revert the weight to the percentage stored in the Git repo. At best, the user can specify the desired weight of 100% to the stable service and 0% to the canary service. In this case, the Virtual Service is synced with the Git repo when the Rollout completed all the steps. 
+
+Argo CD has an [open issue here](https://github.com/argoproj/argo-cd/issues/2913) to address this problem. The solution is that Argo CD supports adding an annotation called `argo-cd.argoproj.io/respect-existing-fields` to the Virtual Service. This annotation indicates that Argo CD should respect the current weights listed in this virtual service and rely on the controller to honor the values of the weights within the CR.
```
jessesuen: Let's not get into any specifics just yet. Please reword that a proposed solution is to use an argo-cd specific annotation which prevents overwriting the weights during a sync operation. Feel free to rephrase.

docs/features/traffic-management/istio.md
```diff
@@ -1,3 +1,88 @@
...
+  ...
+  strategy:
+    canary:
+      steps:
+      - setWeight: 5
+      - pause:
+          duration: 600
+      canaryService: canary-svc # required
+      stableService: stable-svc  # required
+      networking:
```
jessesuen: Not totally sold on the `networking` field name. Probably needs the word traffic in it. But we can discuss later.

jessesuen: Maybe `trafficSplit`:

```yaml
spec:
  strategy:
    canary:
      trafficSplit:
        smi: {}
```


```yaml
spec:
  strategy:
    canary:
      trafficSplit:
        istio:
...
```

dthomson25: Yeah, I agree that we can find something better than `networking`. I'm wondering if `trafficSplit` is too specific since this configuration will be also be used for traffic mirroring and header-based routing

dthomson25: What about `trafficRouting`?

Issue #248, opened 2019-10-31
Provide a way to allow analyses to complete the experiment 


jessesuen:
Currently Experiments either run forever, or complete after a specified duration. A new mode should be that a user can decide that an Experiment should run until one or more analyses complete.

We will probably need a way to mark individual analyses as required for completion. e.g.:

```yaml
spec:
  analyses:
  # success-rate is background analysis that we want running until experiment completes
  - name: success-rate
    templateName: success-rate

  # test1 and test2 are tests which we want to complete and terminate the experiment after
  - name: test1
    templateName: test1
    requiredForCompletion: true
  - name: test2
    templateName: test2
    requiredForCompletion: true
```

PR #345, opened 2020-01-06
feat: Allow AnalysisRun to complete an experiment

dthomson25:
closes https://github.com/argoproj/argo-rollouts/issues/248


pkg/apis/rollouts/v1alpha1/experiment_types.go
```diff
@@ -201,4 +203,6 @@ type ExperimentAnalysisRunStatus struct {
 	Phase AnalysisPhase `json:"phase"`
 	// Message is a message explaining the current status
 	Message string `json:"message,omitempty"`
+	// RequiredForCompletion indicates that experiment should complete after analysis finishes
+	RequiredForCompletion bool `json:"requiredForCompletion,omitempty"`
```
jessesuen: I feel we don't need to duplicate `RequiredForCompletion` in the status for this structure. Can't this be inferred from the `ExperimentAnalysisTemplateRef`?

experiments/experiment.go
```diff
@@ -438,13 +446,22 @@ func (ec *experimentContext) assessTemplates() (v1alpha1.AnalysisPhase, string)
 func (ec *experimentContext) assessAnalysisRuns() (v1alpha1.AnalysisPhase, string) {
 	worstStatus := v1alpha1.AnalysisPhaseSuccessful
 	message := ""
+
+	requiredAnalysisRunCompleted := false
 	for _, a := range ec.ex.Spec.Analyses {
 		as := experimentutil.GetAnalysisRunStatus(*ec.newStatus, a.Name)
+		if as.RequiredForCompletion && as.Phase == v1alpha1.AnalysisPhaseSuccessful {
+			requiredAnalysisRunCompleted = true
```
jessesuen: I think theres a bug here. This will set `requiredAnalysisRunCompleted` to be true as soon as **_any one_** of the required runs completes, but not all. I think this should be reversed to set it to true and set it to false as soon as one required run is still not complete.

utils/experiment/experiment.go
```diff
@@ -45,6 +45,9 @@ func IsTerminating(experiment *v1alpha1.Experiment) bool {
 		case v1alpha1.AnalysisPhaseFailed, v1alpha1.AnalysisPhaseError, v1alpha1.AnalysisPhaseInconclusive:
 			return true
 		}
+		if run.Phase == v1alpha1.AnalysisPhaseSuccessful && run.RequiredForCompletion {
+			return true
+		}
```
jessesuen: This doesn't seem correct. This logic states that if **_any one_** of the N number of analyses runs which required for completion is successful, then we are saying that the entire experiment is about to finish. It should only do this if all of them are completed and successful.

experiments/experiment.go
```diff
@@ -380,6 +385,9 @@ func (ec *experimentContext) calculateStatus() *v1alpha1.ExperimentStatus {
 				// We will now fail the experiment.
 				ec.newStatus.Phase = analysesStatus
 				ec.newStatus.Message = analysesMessage
+			} else if analysesMessage == requiredAnalysisCompletedMessage {
```
jessesuen: I think comparing the message against a well known message seems too magical. Is the point of this code to bubble up a better reason for a running experiment?

pkg/apis/rollouts/v1alpha1/experiment_types.go
dthomson25: Yeah, fair point.

utils/experiment/experiment.go
dthomson25: fixed

experiments/experiment.go
dthomson25: fixed

experiments/experiment.go
dthomson25: That's a fair point. No, the purpose of this code is to provide a way to mark an experiment successful if the required analysis runs are completed. Since we shouldn't mark the experiment successful if the analysis is successful but not required, the magic string makes it easy to determine that case. Can you take another look?

utils/experiment/experiment.go
```diff
@@ -46,7 +46,27 @@ func IsTerminating(experiment *v1alpha1.Experiment) bool {
 			return true
 		}
 	}
-	return false
+	return CompletedAllRequiredAnalysisRuns(experiment, &experiment.Status)
+}
+
+// CompletedAllRequiredAnalysisRuns has at least one required for completition analysis run
+// and it completed successfully
```
jessesuen: Can we rename this to: `RequiredAnalysisRunsSuccessful()` ? I didn't realize this would only only return when they were successful and not just completed.

jessesuen:
LGTM, but suggest to rename method

Issue #291, opened 2019-11-13
Add configuration to delay background analysis


dthomson25:
In the case where an application needs some time to send metrics to the provider, the background analysis for rollouts needs a way to delay the AnalysisRun. Otherwise, the AnalysisRun will be inconclusive or possibly failure.

PR #350, opened 2020-01-09
feat: Allow controller to delay analysis

dthomson25:
Closes https://github.com/argoproj/argo-rollouts/issues/291

The controller can delay AnalysisRuns by:
* Adding a delayStart field for a metric in the AnalysisTemplate. This causes the AnalysisRun to wait the amount of time listed in the field before starting
* For Rollouts using background analysis, a user can specify StartAtStep which tells the rollout to start the AnalysisRun once the rollout reaches that step.

TODO:
* ~~Add documentation~~


analysis/analysis.go
```diff
@@ -127,6 +127,20 @@ func generateMetricTasks(run *v1alpha1.AnalysisRun) []metricTask {
...
 		}
 		if lastMeasurement == nil {
+			if metric.StartDelay != "" {
+				if run.Status.StartedAt == nil {
+					continue
+				}
+				duration, err := metric.StartDelay.Duration()
+				if err != nil {
+					logCtx.Warnf("failed to parse duration: %v", err)
+					continue
```
jessesuen: This is fine here, but can we also have this fail validation of the rollout itself (before reconciliation)?

dthomson25: I already have it a part of validation here: https://github.com/argoproj/argo-rollouts/pull/350/files?file-filters%5B%5D=.go#diff-11d6da907459856384573d1855393e33R99-R103. Is this what you mean?

jessesuen: You're right. Ignore this.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -48,6 +49,8 @@ type Metric struct {
 	// Interval defines an interval string (e.g. 30s, 5m, 1h) between each measurement.
 	// If omitted, will perform a single measurement
 	Interval DurationString `json:"interval,omitempty"`
+	// InitialDelay how long the AnalysisRun should wait before starting this metric
+	InitialDelay DurationString `json:"startDelay,omitempty"`
```
jessesuen: API violation since json field name `startDelay` != golang field name `initialDelay`

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -248,6 +248,14 @@ type CanaryStep struct {
 	Analysis *RolloutAnalysisStep `json:"analysis,omitempty"`
 }
 
+// RolloutAnalysisBackground defines a template that is used to create a background analysisRun
+type RolloutAnalysisBackground struct {
+	RolloutAnalysisStep `json:",inline"`
+	// StartAtStep indicates which step the background analysis should start on
+	// If not listed, controller defaults to 0
+	StartAtStep *int32 `json:"startAtStep,omitempty"`
```
jessesuen: What do you think of:`startingStep` ? Might be easier to talk about it conversationally.

PR #354, opened 2020-01-14
feat: Implement watch for Istio resources

dthomson25:
To ensure the controller maintains the desired state of the virtual service, this PR adds a watch on virtual services and enqueues rollout that reference those virtual services.


utils/controller/controller.go
```diff
@@ -18,6 +23,43 @@ import (
...
+		Enqueue(objsToEnqueue[i], queue)
+	}
+}
+
+// WatchResource is a long-running function that will continually call the
+// processNextWatchObj function in order to watch changes on a resource kind
+// and enqueue a different resources kind that interact with them
+func WatchResource(client dynamic.Interface, gvk schema.GroupVersionResource, queue workqueue.RateLimitingInterface, indexer cache.Indexer, index string) {
+	log.Infof("Starting watch on resource '%s'", gvk.Resource)
+	watch, err := client.Resource(gvk).Watch(metav1.ListOptions{})
```
jessesuen: This Watch() is performing a cluster-scoped watch. Which means that a controller configured to only watch a namespace, will not work. We need to honor when the rollout controller is operating only at a namespace scope (the `--namespace` flag).

utils/controller/controller.go
```diff
@@ -18,6 +23,43 @@ import (
...
+		log.Infof("Watch for resource '%s' failed: resource not found", gvk.Resource)
+		return
+	}
+	if err != nil {
+		panic(err)
+	}
+	for watchEvent := range watch.ResultChan() {
+		processNextWatchObj(watchEvent, queue, indexer, index)
+	}
+}
```
jessesuen: The problem with this function, is that it returns indiscriminately if the resource GVK doesn't exist,  vs. normal failures that happen during the watch. The difference is that when the GVK doesn't exist, we want to backoff longer (e.g. 5 minutes). However if the watch fails simply because the stream got closed, we probably want to retry immediately. Otherwise, we will incur a 1 minute delay every time the watch is closed (which can actually be frequent).

rollout/controller.go
```diff
@@ -197,6 +213,12 @@ func (c *RolloutController) Run(threadiness int, stopCh <-chan struct{}) error {
 		}, time.Second, stopCh)
 	}
 	log.Info("Started Rollout workers")
+
+	gvk := schema.ParseGroupResource("virtualservices.networking.istio.io").WithVersion(c.defaultIstioVersion)
+	go wait.Until(func() {
+		controllerutil.WatchResource(c.dynamicclientset, gvk, c.rolloutWorkqueue, c.rolloutsIndexer, virtualServiceIndexName)
+	}, time.Minute, stopCh)
```
jessesuen: Related to my other comment. The use of `wait.Until` with an indiscriminate 1 minute retry is too simplistic. I would switch this to use exponential backoff, which will work well in all circumstances. (e.g. if we discover the resource isn't in the cluster, then we could eventually backoff to only check every 5 min, reducing API call rate).

utils/controller/controller.go
```diff
@@ -18,6 +23,43 @@ import (
...
+// and enqueue a different resources kind that interact with them
+func WatchResource(client dynamic.Interface, gvk schema.GroupVersionResource, queue workqueue.RateLimitingInterface, indexer cache.Indexer, index string) {
+	log.Infof("Starting watch on resource '%s'", gvk.Resource)
+	watch, err := client.Resource(gvk).Watch(metav1.ListOptions{})
+	if k8serrors.IsNotFound(err) {
+		log.Infof("Watch for resource '%s' failed: resource not found", gvk.Resource)
+		return
+	}
+	if err != nil {
+		panic(err)
```
jessesuen: If we panic here, does this stop the `wait.Until()` goroutine in the caller? 

I think WatchResource() might be better off surfacing errors so that the caller can decide how to handle the various errors.

rollout/trafficrouting/istio/istio.go
```diff
@@ -31,6 +31,20 @@ func NewReconciler(r *v1alpha1.Rollout, client dynamic.Interface, recorder recor
...
+	if rollout.Spec.Strategy.Canary == nil {
+		return []string{}
+	}
+	if rollout.Spec.Strategy.Canary.TrafficRouting == nil {
+		return []string{}
+	}
+	if rollout.Spec.Strategy.Canary.TrafficRouting.Istio == nil {
+		return []string{}
+	}
+	return []string{fmt.Sprintf("%s/%s", rollout.Namespace, rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualService.Name)}
```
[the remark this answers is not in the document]
dthomson25: Yes we do

utils/controller/controller.go
```diff
@@ -18,6 +23,72 @@ import (
...
+	}
+	return nil
+}
+
+// WatchResourceWithExponentialBackoff creates a watch for the gvk provided. If there are any error,
+// the function will rety again using exponetial backoff. It starts at 1 second wait, and wait afterwards
+// increases by a factor of 2 and caps at 5 minutes.
+func WatchResourceWithExponentialBackoff(stopCh <-chan struct{}, client dynamic.Interface, namespace string, gvk schema.GroupVersionResource, queue workqueue.RateLimitingInterface, indexer cache.Indexer, index string) {
+	backoff := wait.Backoff{
+		Duration: time.Duration(1 * time.Second),
```
golangcibot: unnecessary conversion (from `unconvert`)

utils/controller/controller.go
```diff
@@ -18,6 +23,72 @@ import (
...
+		Cap:      time.Duration(5 * time.Minute),
```
golangcibot: unnecessary conversion (from `unconvert`)

utils/controller/controller.go
```diff
@@ -18,6 +23,72 @@ import (
...
+	}
+
+	if err != nil {
+		log.Errorf("Error with watch: %v", err)
+		return err
+	}
+	for watchEvent := range watchI.ResultChan() {
+		processNextWatchObj(watchEvent, queue, indexer, index)
+	}
+	return nil
```
jessesuen: There should be a `watchI.Stop()` call which will clean up resources held by the watch.

utils/controller/controller.go
```diff
@@ -18,6 +23,72 @@ import (
...
+	}
+	for {
+		select {
+		case <-stopCh:
+			return
+		default:
+		}
+		err := WatchResource(client, namespace, gvk, queue, indexer, index)
+		if err == nil {
+			continue
```
jessesuen: Minor, but If there is no error, I would suggest that we "reset" the exponential backoff by re-initializing the backoff variable.

utils/controller/controller.go
```diff
@@ -18,6 +23,72 @@ import (
...
 )
 
+// processNextWatchObj will process a single object from the watch by seeing if
+// that object is in an index and enqueueing the value object from the indexer
+func processNextWatchObj(watchEvent watch.Event, queue workqueue.RateLimitingInterface, indexer cache.Indexer, index string) {
+	obj := watchEvent.Object
+	acc, err := meta.Accessor(obj)
+	if err != nil {
+		log.Errorf("Error processing object from watch: %v", err)
+		return
```
jessesuen: Is there a reason you don't surface these errors to the caller? I'm afraid that we might get into an error hotloop which we will not back off from with the exponential backoff

PR #355, opened 2020-01-14
feat: Add validation to istio virtual services

dthomson25:
Checks that the Virtual Service modified by the rollout has:
1. All the routes listed in Rollout
2. Each route only has two hosts and they are the stable and canary services

### DIFF
--- a/rollout/trafficrouting/istio/istio.go
+++ b/rollout/trafficrouting/istio/istio.go
@@ -133,6 +133,10 @@ func (r *Reconciler) reconcileVirtualService(obj *unstructured.Unstructured, des
 		return nil, false, err
 	}
 
+	if err := validateHTTPRoutes(r.rollout, httpRoutes); err != nil {
+		return nil, false, err
+	}
+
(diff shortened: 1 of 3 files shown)


rollout/trafficrouting/istio/istio.go
```diff
@@ -133,6 +133,10 @@ func (r *Reconciler) reconcileVirtualService(obj *unstructured.Unstructured, des
 		return nil, false, err
 	}
 
+	if err := validateHTTPRoutes(r.rollout, httpRoutes); err != nil {
```
jessesuen: I feel like this should be happening when validating the entire rollout, not just when we are reconciling virtual services. Note that we will eventually have an admission web hook that needs to validate a rollout, so there needs to be a helper to statically lint a rollout.

jessesuen: You may disregard this comment. Performing validation on referenced resources can/should happen during reconciliation time, and not via validating admission webhook.

PR #365, opened 2020-01-17
fix: Fix premature scaledown

dthomson25:


rollout/bluegreen.go
```diff
@@ -333,13 +325,30 @@ func (c *RolloutController) scaleBlueGreen(rollout *v1alpha1.Rollout, newRS *app
...
-	// Old replica sets should be fully scaled down if they aren't receiving traffic from the active or
-	// preview service. This case handles replica set adoption during a saturated new replica set.
-	for _, old := range controller.FilterActiveReplicaSets(oldRSs) {
-		oldLabel, ok := old.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
-		if !ok || (oldLabel != activeSelector && oldLabel != previewSelector) {
-			if _, _, err := c.scaleReplicaSetAndRecordEvent(old, 0, rollout); err != nil {
-				return err
+	annotationedRSs := int32(0)
+	logCtx := logutil.WithRollout(rollout)
+	for _, targetRS := range oldRSs {
```
jessesuen: As far as I can tell, we are not sorting the old replicas, which means we will scale down in an inappropriate order.

I think you need this line:

```golang
sort.Sort(sort.Reverse(replicasetutil.ReplicaSetsByRevisionNumber(oldRSs)))
```

rollout/bluegreen.go
```diff
@@ -333,13 +325,30 @@ func (c *RolloutController) scaleBlueGreen(rollout *v1alpha1.Rollout, newRS *app
...
+				logCtx.Info("At ScaleDownDelayRevisionLimit and scaling down the rest")
+			} else {
+				now := metav1.Now()
+				scaleDownAt := metav1.NewTime(scaleDownAtTime)
+				if scaleDownAt.After(now.Time) {
+					newDesiredReplicas = rolloutReplicas
+				}
+			}
+
+			if *(targetRS.Spec.Replicas) != newDesiredReplicas {
```
jessesuen: The logic in `scaleDownOldReplicaSetsForBlueGreen` has not yet been updated to adjust the scale of demoted replicasets to match the active service.

Issue #225, opened 2019-10-25
Add patchStrategy:"merge" patchMergeKey:"name" 


jessesuen:
Anywhere we need to do strategic merge patching should have these struct tags.

`patchStrategy:"merge" patchMergeKey:"name" `

PR #386, opened 2020-01-27
feat: Add patchMergeKey and patchStrategy struct tags and comments

dthomson25:
Closes https://github.com/argoproj/argo-rollouts/issues/225


dthomson25:
I removed the patchMergeKey and patchStrategy from the status fields as they are affecting the tests and users don't need to do strategy merge patches of the status fields.

Issue #395, opened 2020-02-05
Kustomize label configuration for rollouts has wrong api group


cehoffman:
See: https://github.com/argoproj/argo-rollouts/blob/afa3e0b485a6ca5b196d9cd8e3bde5c57324c559/docs/features/kustomize/rollout-transform.yaml#L70

The label paths concerning affinity and antiaffinity have the wrong group set. The group should be removed entirely to match the other label selectors.

Issue #364, opened 2020-01-17
Update kustomize transformers with new service references


jessesuen:
In v0.7 we added some additional references to services. This should be reflected in rollout-transform.yaml

https://github.com/argoproj/argo-rollouts/blob/master/docs/features/kustomize/rollout-transform.yaml

I think we should also add the Istio VirtualService to `nameReference:`

jessesuen:
Fixed

PR #399, opened 2020-02-07
fix: rollout transformer for pod affinity. add new v0.7 name references and testing

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/395 and https://github.com/argoproj/argo-rollouts/issues/364


dthomson25:
Looks like the CI is failing because the CI container doesn't have kustomize installed:
```
./test/kustomize/test.sh: line 7: kustomize: command not found
```

Issue #348, opened 2020-01-09
Allow Analysis for BlueGreen


dthomson25:
Rollouts should be able to leverage Analysis Runs for Bluegreen in addition to canary. For the Bluegreen strategy, users should be able to start an analysis while running a preview service and after switching over to the new RS to the active stack.

PR #415, opened 2020-02-23
feat: Add Pre Promotion Bluegreen Analysis

dthomson25:
The first half of closing https://github.com/argoproj/argo-rollouts/issues/348. 

Add the ability for a BlueGreen Rollout to launch an AnalysisRun before the active Service's selector is switched. As a part of this change, I also had to implement support for aborting BlueGreen Rollouts. In this case, an aborted BlueGreen Rollout tries to switch the active service's selector to the most recent running ReplicaSet.


rollout/analysis_test.go
```diff
@@ -1146,3 +1152,275 @@ func TestDoNotCreateBackgroundAnalysisRunOnNewRollout(t *testing.T) {
...
+			"abort": true,
+			"pauseConditions": null,
+			"controllerPause":null,
+			"blueGreen": {
+				"currentPrePromotionAnalysisRun": null
+			}
+		}
+	}`
+	assert.Equal(t, calculatePatch(r2, expectedPatch), patch)
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

rollout/bluegreen_test.go
```diff
@@ -1111,5 +1155,49 @@ func TestScaleDownLimit(t *testing.T) {
...
+			"blueGreen": {
+				"activeSelector": "%s"
+			},
+			"conditions": %s,
+			"selector": "foo=bar,rollouts-pod-template-hash=%s"
+		}	
+	}`, rs1PodHash, expectedConditions, rs1PodHash)
+	patch := f.getPatchedRollout(patchIndex)
+	assert.Equal(t, calculatePatch(r2, expectedPatch), patch)
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

utils/analysis/filter_test.go
```diff
@@ -84,19 +57,35 @@ func TestFilterCurrentRolloutAnalysisRuns(t *testing.T) {
...
+		assert.Len(t, nonCurrentArs, 1)
+		assert.Contains(t, currentArs, ars[0])
+		assert.Contains(t, currentArs, ars[1])
+
+	})
+	t.Run("BlueGreen", func(t *testing.T) {
+		r := &v1alpha1.Rollout{
+			Status: v1alpha1.RolloutStatus{
+				BlueGreen: v1alpha1.BlueGreenStatus{
+					CurrentPrePromotionAnalysisRun:       "foo",
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
					CurrentPrePromotionAnalysisRun: "foo",
```

rollout/analysis_test.go
```diff
@@ -1146,3 +1152,430 @@ func TestDoNotCreateBackgroundAnalysisRunOnNewRollout(t *testing.T) {
...
+	f.run(getKey(r2, t))
+	patch := f.getPatchedRollout(patchIndex)
+	expectedPatch := fmt.Sprintf(`{
+		"status": {
+			"blueGreen": {
+				"activeSelector": "%s"
+			},
+			"selector":"foo=bar,rollouts-pod-template-hash=%s"
+		}
+	}`,rs2PodHash,rs2PodHash)
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	}`, rs2PodHash, rs2PodHash)
```

duboisf:
Been looking forward to being able to use smoke tests for blue green deployments :tada: 

I played with this PR locally and I think it's not handling the case when `autoPromotionSeconds` isn't defined. I tested creating a Rollout that used a `prePromotionAnalysis` (using the `pass` `AnalysisTemplate` defined in this repo). The result was that the rollout ended up in the pause state, forever because a `BlueGreenPause` pause condition gets added during the rollout and nothing removes it, except if I promote the rollout manually with `kubectl argo rollouts promote my_rollout`.

This seems to happen because I didn't define the `autoPromotionSeconds`, because AFAIK the only place that the `BlueGreenPause` pause condition get's removed is here: https://github.com/argoproj/argo-rollouts/blob/dee1ce2edc98fc9b12e475efb91e20faf41bd327/rollout/bluegreen.go#L178-L186

dthomson25:
Hey @duboisf, what value do you have for autoPromotionEnabled field? If that field is set to true, the rollout will not promote until the pause condition is manually removed by the user (which `kubectl argo rollouts promote my_rollout` does).

duboisf:
@dthomson25 `autoPromotionEnabled` was absent, forgot to show the configuration:

```
  strategy:
    blueGreen:
      activeService: cddemo
      prePromotionAnalysis:
        templateName: pass
      previewService: cddemo-preview
```

I also tried with `autoPromotionEnabled: false` and got the same result:

```
  strategy:
    blueGreen:
      activeService: cddemo
      autoPromotionEnabled: false
      prePromotionAnalysis:
        templateName: pass
      previewService: cddemo-preview
```

dthomson25:
🤦‍♂, sorry @duboisf, I mixed up `autoPromotionEnabled`. If `autoPromotionEnabled` is set to **false**, the rollout will not continue until manual intervention. However, the default value on an unset field is true and the rollout should auto promote. So, I believe the second case you provided is working as expected. For the first case, can you confirm that the AnalysisRun's status is not running and has completed successfully? I know that the `pass` AnalysisTemplate example doesn't complete since it does not have a `count` under the metric field. I created the following [issue](https://github.com/argoproj/argo-rollouts/issues/427) to prevent Rollouts and Experiments from getting stuck from infinite running AnalysisRuns.

duboisf:
@dthomson25 checked and the `pass` analysis run succeeds, in this case the one that ran was `cddemo-7988d8bf59-49-pass`:

```
$ kc get ar -w
NAME                        STATUS
cddemo-7549644796-48-pass   Successful
cddemo-dc685cf85-47-pass    Successful
cddemo-7988d8bf59-49-pass
cddemo-7988d8bf59-49-pass   Running
cddemo-7988d8bf59-49-pass   Successful
```

Pod that was created was `244d8c37-ffdb-4b46-bb3f-eaaf90d13393.pass.1-dcjp9`:
```
NAME                                                READY   STATUS      RESTARTS   AGE
239c41e7-e760-4728-896a-fa69cc5956a4.pass.1-9mrdn   0/1     Completed   0          5h18m
244d8c37-ffdb-4b46-bb3f-eaaf90d13393.pass.1-dcjp9   0/1     Completed   0          2m57s
cddemo-7988d8bf59-4frnp                             1/1     Running     0          3m5s
cddemo-dc685cf85-ghb95                              1/1     Running     1          5h19m
d4b05b40-549a-44c5-a7db-850310848ee4.pass.1-tzdkd   0/1     Completed   0          5h19m
```

Looking at the analysis run I did tweak it a bit, I do a loop just so that it takes a couple of seconds (10 ish) to complete:

```
kind: AnalysisRun
metadata:
  annotations:
    rollout.argoproj.io/revision: "49"
  creationTimestamp: "2020-03-03T19:29:11Z"
  generation: 3
  labels:
    rollout-type: PrePromotion
    rollouts-pod-template-hash: 7988d8bf59
  name: cddemo-7988d8bf59-49-pass
  namespace: default
  ownerReferences:
  - apiVersion: argoproj.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Rollout
    name: cddemo
    uid: 4c94e002-6d8c-42de-acea-7f37b0aee875
  resourceVersion: "260328"
  selfLink: /apis/argoproj.io/v1alpha1/namespaces/default/analysisruns/cddemo-7988d8bf59-49-pass
  uid: 244d8c37-ffdb-4b46-bb3f-eaaf90d13393
spec:
  metrics:
  - failureLimit: 1
    name: pass
    provider:
      job:
        metadata:
          creationTimestamp: null
        spec:
          backoffLimit: 0
          template:
            metadata:
              creationTimestamp: null
            spec:
              containers:
              - args:
                - -c
                - for i in $(seq 10); do echo $i; sleep 1; done
                command:
                - /bin/sh
                image: alpine:3.8
                name: sleep
                resources: {}
              restartPolicy: Never
status:
  metricResults:
  - count: 1
    measurements:
    - finishedAt: "2020-03-03T19:29:22Z"
      metadata:
        job-name: 244d8c37-ffdb-4b46-bb3f-eaaf90d13393.pass.1
      phase: Successful
      startedAt: "2020-03-03T19:29:11Z"
    name: pass
    phase: Successful
    successful: 1
  phase: Successful
  startedAt: "2020-03-03T19:29:11Z"
```

I might be doing something wrong, wouldn't be the first time :grin:

dthomson25:
Can you try 
```
spec:
  metrics:
  - failureLimit: 1
    count: 1
    name: pass
    provider:
      job:
        metadata:
          creationTimestamp: null
        spec:
          backoffLimit: 0
          template:
            metadata:
              creationTimestamp: null
            spec:
              containers:
              - args:
                - -c
                - for i in $(seq 10); do echo $i; sleep 1; done
                command:
                - /bin/sh
                image: alpine:3.8
                name: sleep
                resources: {}
              restartPolicy: Never
```
and see if that works? I added a count of 1 to the job metric.

rollout/service.go
```diff
@@ -75,9 +74,24 @@ func (c *RolloutController) reconcileActiveService(roCtx *blueGreenContext, prev
 	}
 
 	newPodHash := activeSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey]
-	if skipPause(roCtx, activeSvc) || roCtx.PauseContext().CompletedBlueGreenPause() {
+	//
```
jessesuen: did you mean to add a comment?

duboisf:
@dthomson25 tried adding `count: 1` in the analysis template and there is no change in the behavior, the rollout stays paused

dthomson25:
Hey @duboisf, I believe I found the issue. Previously, a rollout with the `autoPromotionEnabled`  set to true (either explicitly or defaulted) never allowed itself to become paused. With the Analysis changes, the AnalysisRun forces the Rollout to pause. However, the controller never had any way to unpause the rollout when the autoPromotionEnabled was true causing the rollout to become stuck. Can you try the new version with this fix?

duboisf:
@dthomson25 it worked! :tada:

Updating the image of the rollout triggered the new replicaset to be created, then the `pass` analysis run ran successfully and finally the rollout switched the prod service to the new replicaset.

I even pushed my luck and removed the `count: 1` in the `pass` analysis template :grin:

It's now behaving as I expected it to. FYI I removed the `autoPromotionEnabled`, which defaults to `true` if I'm reading the code correctly.

So the fix was removing the pause condition if `skipPause == true` in the `reconcileBlueGreenPause` method?

dthomson25:
@duboisf Yep that was the fix! The Rollout shouldn't be in a paused state if the active service is pointing at the new ReplicaSet.

PR #423, opened 2020-03-01
feat: pause duration as string with time unit

### LINKED ISSUE #292: Make Pause duration use a string (5s) instead of int
Since this is a breaking change, we will need to make it a intOrString for a release, before migrating it to a string.

cronik:
Support declaring pause duration as string with optional time
unit (i.e. "s", "m", "h"). If no time unit or value is an integer,
seconds are inferred.

closes #292

### DIFF
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -1,6 +1,9 @@
 package v1alpha1
 
 import (
+	"strconv"
+	"time"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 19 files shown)


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -321,7 +324,42 @@ const (
...
+				d, err := time.ParseDuration(p.Duration.StrVal)
+				if err != nil {
+					return -1
+				}
+				return int32(d.Seconds())
+			}
+			// special case where no unit was specified
+			return int32(s)
+		}
+		return int32(p.Duration.IntVal)
```
golangcibot: unnecessary conversion (from `unconvert`)

dthomson25:
Overall, I think this looks great! I like the test in types.go, but I don't like how it brings our code coverage down by 30 percent since the autogenerated code is now being included. I think we can get around this by changing our test target in the Makefile to run two go test commands. The first command only runs the tests for v1alpha without coverage, and the second command runs all the other tests with coverage

cronik:
I haven't used codecov.io before but it looks like they support an `ignore` paths configuration which could be another option if it does what it sounds like.

https://docs.codecov.io/docs/ignoring-paths

```yaml
# .codecov.yml
ignore:
  - "pkg/apis/rollouts/v1alpha1"
```

I did play around with modifying the test command to exclude v1alpha with the following:

```make
.PHONY: test
test: test-kustomize
	go test -failfast -covermode=count -coverprofile=coverage.out -coverpkg $(shell go list ./... | grep -v github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1 | xargs | tr ' ' ',') ${TEST_TARGET}
```
It works but the console output is not ideal (it prints the huge package list multiple times). 

Maybe if the covecov config does not work I could just move types_test.go to a different package?

dthomson25:
I definitely think adding ignore field to the .codedev.yaml is the ideal solution. Can you try adding it and seeing if that changes the coverage? If not, I'll open a separate PR to fix this.

cronik:
Looks like the codecov config worked!

cronik:
I'm good, thanks!

Issue #421, opened 2020-02-29
ALB Ingress Weighted Target Group Support


jessesuen:
ALB Ingress Controller recently added support for [AWS Weighted Target Groups](https://aws.amazon.com/blogs/aws/new-application-load-balancer-simplifies-deployment-with-weighted-target-groups/) as of their most recent [v1.1.5 release](https://github.com/kubernetes-sigs/aws-alb-ingress-controller/releases/tag/v1.1.5).

https://github.com/kubernetes-sigs/aws-alb-ingress-controller/pull/1088

The ingress annotation for using weighted target groups look like:

```yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: simple-ingress-example
  annotations:
    alb.ingress.kubernetes.io/actions.weighted-routing: '{"Type":"forward","ForwardConfig":{"TargetGroupStickinessConfig":{"Enabled":"True","DurationSeconds":100},"TargetGroups":[{"Weight":20,"ServiceName":"Service1","ServicePort":"80"},{"Weight":80,"TargetGroupArn":"arn:aws:******/external-legecy-tg"}]}}'
spec:
  rules:
  - host: m00nf1sh.example.com
    http:
      paths:
      - path: /foo
        backend:
          serviceName: weighted-routing
          servicePort: use-annotation
```

The one-line json for 'weighted-routing' above expands to:

```json
{ 
   "Type":"forward",
   "ForwardConfig":{ 
      "TargetGroupStickinessConfig":{ 
         "Enabled":"True",
         "DurationSeconds":100
      },
      "TargetGroups":[ 
         { 
            "Weight":20,
            "ServiceName":"service1",
            "ServicePort":"80"
         },
         { 
            "Weight":80,
            "TargetGroupArn":"arn:aws:******/external-legecy-tg"
         }
      ]
   }
}
```

The above splits traffic between a Kubernetes Service, and an **_external-provisioned target group_**, `arn:aws:******/external-legecy-tg`.

I'm unsure if it would be possible to split between two Kubernetes services. If I interpret the following message correctly, it may not be possible:

> The TargetGroups within ForwardConfig have been extended to allow use both Kubernetes Service and TargetGroupArn. Customers can configure backend with Kubernetes Service only / TargetGroupArn only / mix of two(for migration use case)

jessesuen:
Confirmed from implementor that it can target 2+ services:

jessesuen:argo: Today at 11:28 AM
[...]

PR #426, opened 2020-03-03
feat: Nginx canary traffic management

### LINKED ISSUE #401: Feature: ingress-nginx traffic management
We intend on contributing ingress-nginx traffic management. Opening this issue to discuss the proposed design.

I propose we extend the [proposed spec](https://github.com/argoproj/argo-rollouts/blob/737d9cd537303ccf40c29056925da730a9763f07/docs/features/traffic-management/nginx.md#integration-with-argo-rollouts) with an additional `canaryIngress` field. this means the user has to create **two** `extensions/v1beta1.Ingress` resources, and pass both to the rollouts controller.
We should also rename `primaryIngress` to `stableIngress` to match `stableService`.

```yaml
kind: Rollout
spec:
  ...
  strategy:
    canary:
      canaryService: canary-service # required
      stableService: stable-service  # required
      trafficRouting:
        nginx:
           annotationPrefix: example.nginx.com/ # optional
           stableIngress: stable-ingress  # required
```

The controller will then perform the following actions:

1. Validate the `Ingress` referenced by `stableIngress` is pointed at the `stableService ` in all host rules (`/rules[*]/http/paths[*]/backend/serviceName`)
2. Create a copy of the `stableIngress` (append `name` with `-canary`). In this copy, add the  `nginx.ingress.kubernetes.io/canary: "true"` annotation.
4. During reconcile, update the `nginx.ingress.kubernetes.io/canary-weight` annotation on the canary ingress to match the desired weight. Set to `0` once we reach `100`, similar as is done with the current Istio traffic management reconciler.

moensch:
Implements nginx canary traffic management (Resolves #401).


rollout/trafficrouting/nginx/nginx.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// Reconciler holds required fields to reconcile Nginx resources
+type Reconciler struct {
+	rollout           *v1alpha1.Rollout
+	log               *logrus.Entry
+	client            kubernetes.Interface
+	recorder          record.EventRecorder
+	controllerKind    schema.GroupVersionKind
+	defaultAPIVersion string
```
golangcibot: `defaultAPIVersion` is unused (from `structcheck`)

rollout/controller.go
```diff
@@ -64,6 +66,7 @@ type RolloutController struct {
 	rolloutsSynced         cache.InformerSynced
 	rolloutsIndexer        cache.Indexer
 	servicesLister         v1.ServiceLister
+	ingressesLister        extensionslisters.IngressLister
```
golangcibot: `ingressesLister` is unused (from `structcheck`)

rollout/controller.go
```diff
@@ -81,6 +84,7 @@ type RolloutController struct {
 	// simultaneously in two different workers.
 	rolloutWorkqueue workqueue.RateLimitingInterface
 	serviceWorkqueue workqueue.RateLimitingInterface
+	ingressWorkqueue workqueue.RateLimitingInterface
```
golangcibot: `ingressWorkqueue` is unused (from `structcheck`)

pkg/apis/rollouts/v1alpha1/openapi_generated.go
```diff
@@ -1663,6 +1664,48 @@ func schema_pkg_apis_rollouts_v1alpha1_MetricResult(ref common.ReferenceCallback
 	}
 }
 
+func schema_pkg_apis_rollouts_v1alpha1_NginxTrafficRouting(ref common.ReferenceCallback) common.OpenAPIDefinition {
+	return common.OpenAPIDefinition{
+		Schema: spec.Schema{
+			SchemaProps: spec.SchemaProps{
+				Description: "NginxTrafficRouting configuration for Nginx ingress controller to controll traffic routing",
```
golangcibot: `controll` is a misspelling of `control` (from `misspell`)
```suggestion

```

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -160,6 +160,20 @@ type CanaryStrategy struct {
 type RolloutTrafficRouting struct {
 	// Istio holds Istio specific configuration to route traffic
 	Istio *IstioTrafficRouting `json:"istio,omitempty"`
+
+	// Nginx holds Istio specific configuration to route traffic
+	Nginx *NginxTrafficRouting `json:"nginx,omitempty"`
+}
+
+// NginxTrafficRouting configuration for Nginx ingress controller to controll traffic routing
```
golangcibot: `controll` is a misspelling of `control` (from `misspell`)
```suggestion

```

rollout/trafficrouting/nginx/nginx_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+func ingress(name string, port int, serviceName string) *extensionsv1beta1.Ingress {
+	return &extensionsv1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: metav1.NamespaceDefault,
+		},
+		Spec: extensionsv1beta1.IngressSpec{
+			Rules: []extensionsv1beta1.IngressRule{
+				extensionsv1beta1.IngressRule{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				{
```

ingress/ingress_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+func newIngress(name string, port int, serviceName string) *extensionsv1beta1.Ingress {
+	return &extensionsv1beta1.Ingress{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: metav1.NamespaceDefault,
+		},
+		Spec: extensionsv1beta1.IngressSpec{
+			Rules: []extensionsv1beta1.IngressRule{
+				extensionsv1beta1.IngressRule{
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
				{
```

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+
+// NewIngressController returns a new ingress controller
+func NewIngressController(
+	kubeclientset kubernetes.Interface,
+	ingressInformer extensionsinformers.IngressInformer,
+	rolloutsInformer informers.RolloutInformer,
+	resyncPeriod time.Duration,
+	rolloutWorkQueue workqueue.RateLimitingInterface,
+	ingressWorkQueue workqueue.RateLimitingInterface,
+	metricServer *metrics.MetricsServer) *IngressController {
```
dtjm: Not sure how argo project feels about this, but my personal preference is to use a struct arg in cases like this where there are more than 2 or 3 parameters, especially when you have 2 parameters of the same type next to each other which makes them easy to swap on accident. 

It's easier to read because it's a little more self-documenting when you make the call the function. 

Something like this:

```
type IngressControllerConfig struct {
  KubeClientSet kubernetes.Interface
  IngressInformer extensionsinformers.IngressInformer
  // etc 
}

// I like to conceptualize structs in 4 different sections:
// config/dependencies, state, and signaling 
type IngressController struct {
  // config 
  cfg IngressControllerConfig 
  
  // signaling 
  enqueueRollout func(obj interface{})
}

func NewIngressController(cfg IngressControllerConfig) *IngressController {
  controller := &IngressController{
    cfg: cfg,
  }
  // ...
} 
```

then it could easily be called like 

```
NewIngressController(IngressControllerConfig{
  KubeClientSet: x,
  IngressInformer: y,
})
```

docs/features/traffic-management/nginx.md
```diff
@@ -24,12 +26,15 @@ spec:
       stableService: stable-service  # required
       trafficRouting:
         nginx:
-           primaryIngress: primary-ingress  # required
-           annotationPrefix: example.nginx.com/ # optional
+           stableIngress: primary-ingress  # required
+           annotationPrefix: customingress.nginx.ingress.kubernetes.io # optional
+           additionalIngressAnnotations: #optional
+             canary-by-header: X-Canary
+             canary-by-header-value: iwantsit
```
dtjm: Might be nice to link to https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary, when I got here I was thinking "where did these come from?"

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+	controllerutil "github.com/argoproj/argo-rollouts/utils/controller"
+	ingressutil "github.com/argoproj/argo-rollouts/utils/ingress"
+	logutil "github.com/argoproj/argo-rollouts/utils/log"
+)
+
+const (
+	ingressIndexName = "byIngress"
+)
+
+type IngressController struct {
```
dtjm: as a complete noob to the project, I would expect some documentation on the type, what are its responsibilities, is it implementing some interface, is it some sort of plugin, how does it fit into the larger project, etc etc

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/argoproj/argo-rollouts/controller/metrics"
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	informers "github.com/argoproj/argo-rollouts/pkg/client/informers/externalversions/rollouts/v1alpha1"
+	controllerutil "github.com/argoproj/argo-rollouts/utils/controller"
+	ingressutil "github.com/argoproj/argo-rollouts/utils/ingress"
+	logutil "github.com/argoproj/argo-rollouts/utils/log"
+)
+
+const (
+	ingressIndexName = "byIngress"
```
dtjm: maybe document what this is for and how it's used?

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+		ingressSynced:   ingressInformer.Informer().HasSynced,
+
+		rolloutWorkqueue: rolloutWorkQueue,
+		ingressWorkqueue: ingressWorkQueue,
+		resyncPeriod:     resyncPeriod,
+		metricServer:     metricServer,
+	}
+
+	util.CheckErr(rolloutsInformer.Informer().AddIndexers(cache.Indexers{
+		ingressIndexName: func(obj interface{}) (strings []string, e error) {
```
dtjm: I don't think the named parameters are necessary since you're not using them, and generally I find them to be confusing 

```suggestion
		ingressIndexName: func(obj interface{}) ([]string, error) {
```

controller/controller.go
```diff
@@ -51,6 +53,9 @@ const (
 
 	// DefaultServiceThreads Default number of service worker threads to start with the controller
 	DefaultServiceThreads = 10
+
+	// DefaultIngressThreads Default number of ingress worker threads to start with the controller
```
dtjm:
```suggestion
	// DefaultIngressThreads is the default number of ingress worker threads to start with the controller
```

otherwise it will look weird in Godoc

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+	})
+	controller.enqueueRollout = func(obj interface{}) {
+		controllerutil.EnqueueRateLimited(obj, rolloutWorkQueue)
+	}
+
+	return controller
+}
+
+func (c *IngressController) Run(threadiness int, stopCh <-chan struct{}) error {
+	log.Info("Starting Ingress workers")
```
dtjm: would this be appropriate use for the logging context?

https://github.com/argoproj/argo-rollouts/blob/3b4b31da62893f333b8245d7a78eca8b9b00e071/utils/log/log.go#L23

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+	for i := 0; i < threadiness; i++ {
+		go wait.Until(func() {
+			controllerutil.RunWorker(c.ingressWorkqueue, logutil.ServiceKey, c.syncIngress, c.metricServer)
+		}, time.Second, stopCh)
+	}
+
+	log.Info("Started Ingress workers")
+	<-stopCh
+	log.Info("Shutting down Ingress workers")
```
dtjm: where does the shutting down actually happen?  You don't seem to be doing anything here. Might be worth documenting.

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return controller
+}
+
+func (c *IngressController) Run(threadiness int, stopCh <-chan struct{}) error {
+	log.Info("Starting Ingress workers")
+	for i := 0; i < threadiness; i++ {
+		go wait.Until(func() {
+			controllerutil.RunWorker(c.ingressWorkqueue, logutil.ServiceKey, c.syncIngress, c.metricServer)
```
dtjm: is there a chance that the goroutine leaks if  `RunWorker`  doesn't terminate?

ingress/ingress.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (c *IngressController) syncIngress(key string) error {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+	_, err = c.ingressLister.Ingresses(namespace).Get(name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			if !strings.HasSuffix(name, "-canary") {
```
dtjm: this `-canary` suffix smells like a user-facing thing.  Is it documented somewhere?

utils/ingress/ingress.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+)
+
+// GetRolloutIngressKeys returns ingresses keys (namespace/ingressName) which are referenced by specified rollout
+func GetRolloutIngressKeys(rollout *v1alpha1.Rollout) []string {
+	ingressesSet := make(map[string]bool)
```
dtjm: why do we need this map? Couldn't we just append to the slice directly?

docs/features/traffic-management/nginx.md
moensch: That should already be there (first paragraph).

dtjm: oops 👍

utils/ingress/ingress.go
moensch: It's a quick and easy way to de-dupe in case multiple `Rollouts` reference the same `StableIngress`. On second thought, I wonder whether this should be an error condition and whether we should add a check to ensure no two Rollouts reference the same Ingress.

ingress/ingress.go
moensch: Working through this I found a couple of places where more inline comments would have benefited me. I might leave it to @dthomson25 to see if I should do this in my PR or on a follow-up (if I comment this new code, I'd also go through other related similar sections in other files and do the same).

ingress/ingress.go
moensch: The IngressController runs outside of the context of a single Rollout. It's a whole different thread listening for changes to `Ingress` resources referenced in any of the observed Rollout resources.

ingress/ingress.go
moensch: I agree. When we decided the controller should make the canaryIngress (rather than have the user supply it - see https://github.com/argoproj/argo-rollouts/issues/401#issuecomment-583649829) I needed to come up with some convention on what to name the canary copy. I opted for the `-canary` suffix, but this thing feels unclean to me.
What I am trying to do in this code is:
If something external is deleting the referenced stableIngress, do nothing (it'll just yield errors). If something external is deleting the canaryIngress, reconcile again to ensure it is recreated.
I would have preferred to introduce a custom annotation to check whether it's the stable or the canary ingress, but I don't have access to the full Ingress resource in this function, only namespace and name.

utils/ingress/ingress.go
dtjm: I'm not clear how you'd ever catch a dupe because it seems like you're adding `$x` and `$x-canary` to the map

ingress/ingress.go
dtjm: 👍

ingress/ingress.go
moensch: Big +1 on that - it was difficult working through that. I'm down for changing this across several constructors if the maintainers are on board (`controller/controller.go#NewManagers()` suffers of the same).

ingress/ingress.go
dtjm: I would be afraid that someone on the other side of the cluster is putting `-canary` on their name without knowing about the behavior on this side.  Is there any way this could affect something like that unintentionally?

ingress/ingress.go
jessesuen: > but my personal preference is to use a struct arg in cases like this where there are more than 2 or 3 parameters, especially when you have 2 parameters of the same type next to each other which makes them easy to swap on accident.

I agree

ingress/ingress.go
jessesuen: It's a valid concern, but I think the behavior of cloning the reference ingress it with`-canary` can be documented away. The ingress clone name needs to be deterministic (since the controller needs to handle handle retries of create properly in the event reconciliation fails), so randomly generated names using `generateName` is not an option. 

> I would be afraid that someone on the other side of the cluster is putting -canary on their name without knowing about the behavior on this side

To protect ourselves from operating on an ingress which was not created by the rollout, we'll need to verify the ownership reference of the ingress clone, matches the UID of the rollout. If it doesn't, it means it was created by someone else, and the rollout should have an error condition

ingress/ingress.go
jessesuen: Pretty sure this was copied from the other controllers, so he was just following our lead. Shutdown would only happen when controller is being gracefully terminated.

README.md
```diff
@@ -16,6 +16,7 @@ To learn more about Argo Rollouts go to the [complete documentation](https://arg
 Organizations below are **officially** using Argo Rollouts. Please send a PR with your organization name if you are using Argo Rollouts.
 
 1. [Intuit](https://www.intuit.com/)
+1. [Twilio SendGrid](https://sendgrid.com)
```
jessesuen: 🎉 😄 👍

ingress/ingress.go
moensch: Which way would you like me to handle this?
1. Move to struct args in this PR
2. Leave this open, convert to struct args in a different branch, rebase this afterwards
3. Complete this PR as is and so the conversation later

jessesuen:
So we will be eventually adding support for ALB weighted target groups in alb-ingress-controller (https://github.com/argoproj/argo-rollouts/issues/421).

ALB ingress controller also requires a reference to an ingress. I'm thinking that we should move `stableIngress` up one level:

```yaml
  strategy:
    canary:
      canaryService: canary-service  # required for nginx/alb ingress
      stableService: stable-service    # required for nginx/alb ingress
      stableIngress: primary-ingress  # required for nginx/alb ingress
      trafficRouting:
        nginx:
           annotationPrefix: example.nginx.com/ # optional
```

@dthomson25 what do you think?

ingress/ingress.go
moensch: I'll add a check to validate the ownerRef if the canaryIngress already exists and error if it doesn't match expected.

moensch:
> ALB ingress controller also requires a reference to an ingress. I'm thinking that we should move `stableIngress` up one level:

Since `stableIngress` is the only required field right now for nginx, it would mean folks wanting to enable nginx traffic routing would need to specify this here. It's not pretty, but it'll work I guess?

```yaml
strategy:
  canary:
    trafficRouting:
      nginx: {}
```

jessesuen:
> It's not pretty, but it'll work I guess?

I think it'll be better than the alternative, which is to duplicate the field in both alb and nginx specs:

```yaml
strategy:
  canary:
    trafficRouting:
      nginx:
        stableIngress: my-ingress
```

```yaml
strategy:
  canary:
    trafficRouting:
      alb:
        stableIngress: my-ingress
```

dthomson25:
I'm in the duplicate field camp because the istio implementation doesn't use the stableIngress at all and I think that would be confusing to other users. As an alternative, we could do
```
strategy:
  canary:
    trafficRouting:
      ingress:
        type: alb # or nginx
        stableIngress: my-ingress
```

utils/ingress/ingress.go
moensch: I see now how what I said yesterday was plain wrong - fixed :)

moensch:
@dthomson25 @jessesuen Did you come to an agreement on how to lay out the ingress CRD? I would love to get this wrapped up today or tomorrow so I can put it in front of my devs.

jessesuen:
@dthomson25 convinced me that original proposal is a better approach:

```yaml
strategy:
  canary:
    trafficRouting:
      nginx:
        stableIngress: my-ingress
```

Rationale is:
1.  the field: `strategy.canary.stableIngress` is meaningless unless it is used in the context of Nginx or ALB ingress, unlike `strategy.canary.canaryService` or `strategy.canary.stableService` which are useful in all circumstances.
2. Rollout spec validation can happen at create/update time, since we can mark `strategy.canary.nginx.stableIngress` as a *required* field in OpenAPI validation. If it is under `strategy.canary.stableIngress`, it would need to be validated *after* it enters the system (or with a validating webhook).

moensch:
So I think this is ready for the final review then? I hit some snags with codecov though. I don’t understand why in the last two commits coverage went down when I actually added a new test ;)

dthomson25:
Yeah, something weird is happening with codecov. For example, the coverage in the service.go shouldn't change at all. I'll see if I can fix it, but I'll take a look now.

rollout/trafficrouting/nginx/nginx.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// Check if stable ingress exists, error if it does not
+	stableIngress, err := r.cfg.Client.ExtensionsV1beta1().Ingresses(r.cfg.Rollout.Namespace).Get(stableIngressName, metav1.GetOptions{})
+	if err != nil {
+		r.log.WithField(logutil.IngressKey, stableIngressName).Errorf("Error retrieving ingress: %v", err)
+		return err
+	}
+
+	// Check if canary ingress exists, determines whether we later call Create() or Update()
+	canaryIngress, err := r.cfg.Client.ExtensionsV1beta1().Ingresses(r.cfg.Rollout.Namespace).Get(canaryIngressName, metav1.GetOptions{})
```
dthomson25: I'd rather avoid making an API server call on each reconciliation if possible. I think we can achieve this with the following approach:

- Get Canary Ingress from IngressLister
-  If canary Ingress not found
   - Create Canary Ingress
   - If successful
     - return nil
   - If isAlreadyExists
     - Get Canary Ingress from API server
- Compare desired and current Canary Ingress and apply Patch

dthomson25: While I wouldn't say it's a one to one comparison, we take a similar approach with ReplicaSets: https://github.com/argoproj/argo-rollouts/blob/master/rollout/sync.go#L56-L207

rollout/controller.go
```diff
@@ -14,9 +14,11 @@ import (
 	"k8s.io/client-go/dynamic"
 	appsinformers "k8s.io/client-go/informers/apps/v1"
 	coreinformers "k8s.io/client-go/informers/core/v1"
+	extensionsinformers "k8s.io/client-go/informers/extensions/v1beta1"
 	"k8s.io/client-go/kubernetes"
 	appslisters "k8s.io/client-go/listers/apps/v1"
 	v1 "k8s.io/client-go/listers/core/v1"
+	extensionslisters "k8s.io/client-go/listers/extensions/v1beta1"
```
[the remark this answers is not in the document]
jessesuen: Agree. extensions/v1beta1 will no longer be served in the next v1.18 Kubernetes release. 

See [v1.14 changelog](https://v1-14.docs.kubernetes.io/docs/setup/release/notes/#deprecations)

  - Ingress resources are now available via `networking.k8s.io/v1beta1`. Ingress resources in `extensions/v1beta1` are deprecated and will no longer be served in v1.18. Existing persisted data is available via the new API group/version  ([#74057](https://github.com/kubernetes/kubernetes/pull/74057), [@liggitt](https://github.com/liggitt))

I think we can simply use the `networking.k8s.io/v1beta1` informer. We probably don't need to go so far as we did with with Istio VirtualService with the CLI flag to specify apiVerision, since the go-client might drop that package eventually.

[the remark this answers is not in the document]
jessesuen: Caveat: if we don't provide the CLI flag, it means we do not support K8s v1.13.

rollout/trafficrouting/nginx/nginx.go
```diff
@@ -0,0 +1,205 @@
...
+	// Update ingress name
+	desiredCanaryIngress.SetName(canaryIngressName)
+
+	// Remove Argo CD instance label to avoid the canaryIngress being pruned by Argo CD
+	// TODO: This will not work as intended if `application.instanceLabelKey` was changed from the
+	// default value.
+	delete(desiredCanaryIngress.Labels, "app.kubernetes.io/instance")
+
+	// Delete other annotations we never want
+	delete(desiredCanaryIngress.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
```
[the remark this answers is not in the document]
moensch: I agree, making ObjectMeta from scratch makes this cleaner - but what about Labels? Some folks might rely on the prom metrics exporter for nginx ingress and as such might want to preserve labels.

rollout/controller.go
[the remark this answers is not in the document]
dthomson25: Yeah, `networking.k8s.io/v1beta` should be the default

rollout/trafficrouting/nginx/nginx.go
```diff
@@ -0,0 +1,216 @@
...
+	// See: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary
+	if val, ok := stableIngress.Annotations["kubernetes.io/ingress.class"]; ok {
+		desiredCanaryIngress.Annotations["kubernetes.io/ingress.class"] = val
+	}
+
+	// Ensure canaryIngress is owned by this Rollout for cleanup
+	desiredCanaryIngress.SetOwnerReferences([]metav1.OwnerReference{*metav1.NewControllerRef(r.cfg.Rollout, r.cfg.ControllerKind)})
+
+	// Copy only the rules which reference the stableService from the stableIngress to the canaryIngress
+	// and change service backend to canaryService. Rules **not** referencing the stableIngress will be ignored.
```
dthomson25: Correct me if I am misunderstanding, but don't we want to keep the other rules that do not reference the stable service? Does the controller send a portion of **all** traffic to the config in the canary ingress or to only the routes with changes? I'm concerned that traffic will be routed to the canary config, but there is no route for it.

rollout/controller.go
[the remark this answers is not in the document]
jessesuen: Sounds like a plan. I hope that the k8s v1.18 go-client SDK will still have the type definitions/informers libraries for `extensions/v1beta` even if the API server no longer serves it.

rollout/trafficrouting/nginx/nginx.go
moensch: I have validated that canaries are on a per host rule level.

You can have you main Ingress like this:
```yaml
spec:
  rules:
    - host: podinfo.example.com
      http:
        paths:
          - backend:
              serviceName: podinfo
              servicePort: 80
    - host: podinfo2.example.com
      http:
        paths:
          - backend:
              serviceName: podinfo2
              servicePort: 80
```

And then two separate canaries like this:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/canary: "true"
    nginx.ingress.kubernetes.io/canary-weight: "10"
  name: podinfo-canary
spec:
  rules:
  - host: podinfo.example.com
    http:
      paths:
      - backend:
          serviceName: podinfo-canary
          servicePort: 80
```

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/canary: "true"
    nginx.ingress.kubernetes.io/canary-weight: "30"
  name: podinfo2-canary
spec:
  rules:
  - host: podinfo2.example.com
    http:
      paths:
      - backend:
          serviceName: podinfo2-canary
          servicePort: 80
```

So for the first IngressRule, we'd sent 10% to its canary, for the second one we'd send 30% to its canary.

And setting the code up the way I did allows for this scenario. Because someone might have a different `Rollout` using the stable `stableIngress`, but operating on a different `IngressRule` (read: Essentially two rollouts sharing an Ingress but different routes).

rollout/trafficrouting/nginx/nginx.go
moensch: I think one of the issues I'm hitting here is that the canaryIngress isn't actually reconciled in the Ingress controller, but instead in the Rollouts controller. How would you propose I proceed? I could move all the canary ingress creation and reconciliation out of the Rollouts controller and into the ingress controller (and have the rollouts controller basically work off the assumption that the canaryIngress exists, and only update the annotation).

rollout/controller.go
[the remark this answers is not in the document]
moensch: tl;dr should we hold this back for kube 1.18 support?

I’ll see how this can be achieved without too much code duplication. I am still heavily leaning towards just supporting `extensions/v1beta1`. I tried to find a decent break down of kube versions in the wild but the only thing I found is that EKS just started supporting 1.14 a short while ago. So at least for AWS, 1.18 seems quite far out.
I know Argo projects probably don’t yet advertise a support matrix but maybe we should start?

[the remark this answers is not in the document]
cronik: Support for `extensions/v1beta1` would be much appreciated. I'm in the same boat as @moensch being stuck with a 1.13 k8s distribution for the foreseeable future and have a number of teams looking forward for traffic shaped canary rollouts.

rollout/trafficrouting/nginx/nginx.go
dthomson25: That's a great point that I didn't think about! Okay, I'm good with this approach then. Thanks for testing confirming this behavior.

rollout/trafficrouting/nginx/nginx.go
dthomson25: I'm still leaning towards having the Ingress creation in the Rollouts controller otherwise we will have to coordinate the Ingress and Rollout controllers. For example, if the rollout controller doesn't find the canary Ingress, did an error occur when the Ingress was created or does the rollout controller need to wait a little longer? That's hard for the Rollout controller to know without information passed to it or the controller stores information from the previous reconciliation. I'm worried that is going to increase the scope of this PR significantly.

cabrinha: Why not require users to create a secondary Ingress on their own? Users are already required to create a secondary Service for their canary.

moensch: This (reducing API calls) should be addressed in https://github.com/argoproj/argo-rollouts/pull/426/commits/86e1169913e54d714a1dd32a4866f18af3ebb237

rollout/controller.go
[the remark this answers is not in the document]
moensch: To close the loop here from the slack conversation: Maintainers agreed to add kube 1.18 support in a later release.

moensch:
I was wondering: Right now, any errors that occur during trafficmanagement reconcile do not affect the resource health, they only issue the warning kube events.

Should we maybe introduce a new `RolloutConditionType` for something like `TrafficroutingFailure` so that argo-cd's health check Lua script can key off that for a `Degraded` state? I'd add this directly to `trafficrouting.go`.

rollout/trafficrouting/nginx/nginx.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// Check if canary ingress exists (from lister which has a cache), determines whether we later call Create() or Update()
+	canaryIngress, err := r.cfg.IngressLister.Ingresses(r.cfg.Rollout.Namespace).Get(canaryIngressName)
+
+	canaryIngressExists := true
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			// An error other than "not found" occurred
+			r.log.WithField(logutil.IngressKey, canaryIngressName).WithField("err", err.Error()).Error("error retrieving canary ingress")
+			return errors.New(fmt.Sprintf("error retrieving canary ingress `%s` from cache: %v", canaryIngressName, err))
```
dthomson25: Feel free to ignore this, but what do you think about `fmt.Errorf` instead of `errors.New(fmt.Sprintf` to format errors?

rollout/trafficrouting/nginx/nginx_test.go
```diff
@@ -0,0 +1,463 @@
...
+	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	kubeinformers "k8s.io/client-go/informers"
+	fake "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/record"
```
dthomson25: style nix: Can we organize the imports so they are in the following format?
import (
standard library packages

external packages

internal packages
)

utils/ingress/ingress.go
```diff
@@ -0,0 +1,47 @@
...
+			fmt.Sprintf("%s/%s", rollout.Namespace, GetCanaryIngressName(rollout)),
+		}
+	}
+
+	return nil
+}
+
+// GetCanaryIngressName constructs the name to use for the canary ingress resource from a given Rollout
+func GetCanaryIngressName(rollout *v1alpha1.Rollout) string {
+	// names limited to 253 characters
```
dthomson25: Great catch!

rollout/trafficrouting/nginx/nginx.go
moensch: For some dumb reason I keep forgetting this exists - added.

PR #434, opened 2020-03-09
improvement: Surface failure reasons for Rollouts/AnalysisRuns

### LINKED ISSUE #404: Analysis failure / rollout aborted reason should be surfaced
When a rollout fails analysis metrics, we don't currently surface the failure reason (e.g. the value did not meet the success or failure expression/condition). This message should be surfaced to AnalysisRun status.message and eventually Rollout, so that the Argo CD health check could extract the failure message as the health message.

khhirani:
Resolves #404


analysis/analysis.go
```diff
@@ -483,6 +482,37 @@ func assessMetricStatus(metric v1alpha1.Metric, result v1alpha1.MetricResult, te
...
+		phase = v1alpha1.AnalysisPhaseInconclusive
+		msg1 = fmt.Sprintf("assessed %s", phase)
+		msg2 = fmt.Sprintf("inconclusive (%d) > inconclusiveLimit (%d)", result.Inconclusive, metric.InconclusiveLimit)
+		message := []string{msg1, msg2}
+		return phase, message
+	}
+	consecutiveErrorLimit := DefaultConsecutiveErrorLimit
+	if metric.ConsecutiveErrorLimit != nil {
+		consecutiveErrorLimit = *metric.ConsecutiveErrorLimit
+	}
```
dthomson25: I know this was here before your change, but can you give this into the utils/defaults package and create a method called `GetConsecutiveErrorLimitOrDefault` with this logic?

analysis/analysis.go
```diff
@@ -483,6 +482,37 @@ func assessMetricStatus(metric v1alpha1.Metric, result v1alpha1.MetricResult, te
 	return v1alpha1.AnalysisPhaseRunning
 }
 
+func assessMetricFailureInconclusiveOrError(metric v1alpha1.Metric, result v1alpha1.MetricResult) (v1alpha1.AnalysisPhase, []string) {
+	var msg1, msg2 string
```
dthomson25: I don't think we need msg1 since it always consists of `assessed ` plus the phase.  Methods that call `assessMetricFailureInconclusiveOrError` already have that phase information.

analysis/analysis.go
```diff
@@ -413,20 +415,24 @@ func (c *AnalysisController) asssessRunStatus(run *v1alpha1.AnalysisRun) v1alpha
 				everythingCompleted = false
 			} else {
 				// otherwise, remember the worst status of all completed metric results
-				if worstStatus == "" {
+				if worstStatus == "" || analysisutil.IsWorse(worstStatus, metricStatus) {
 					worstStatus = metricStatus
-				} else {
-					if analysisutil.IsWorse(worstStatus, metricStatus) {
-						worstStatus = metricStatus
+					_, message := assessMetricFailureInconclusiveOrError(metric, *result)
```
dthomson25: Instead of calling `assessMetricFailureInconclusiveOrError` again to get the message, we can pass the message back through `assessMetricStatus` as another return value.

analysis/analysis.go
```diff
@@ -483,6 +482,37 @@ func assessMetricStatus(metric v1alpha1.Metric, result v1alpha1.MetricResult, te
 	return v1alpha1.AnalysisPhaseRunning
 }
 
+func assessMetricFailureInconclusiveOrError(metric v1alpha1.Metric, result v1alpha1.MetricResult) (v1alpha1.AnalysisPhase, []string) {
```
dthomson25: If we add the message as another return value for `assessMetricStatus`, I don't think we will need this function since its functionality will be covered by `assessMetricStatus`. I think having all the assessment of the metric Status makes it easier for developers to easier to see how the analysis run is evaluated.

analysis/analysis_test.go
```diff
@@ -424,7 +424,8 @@ func TestAssessRunStatus(t *testing.T) {
 				},
 			},
 		}
-		assert.Equal(t, v1alpha1.AnalysisPhaseRunning, c.asssessRunStatus(run))
+		status, _ := c.assessRunStatus(run)
+		assert.Equal(t, v1alpha1.AnalysisPhaseRunning, status)
```
dthomson25: Let's assert the message is empty string for these tests

analysis/analysis_test.go
```diff
@@ -1380,3 +1383,109 @@ func TestKeyNotInSecret(t *testing.T) {
...
+func TestAssessMetricFailureInconclusiveOrError(t *testing.T) {
+	metric := v1alpha1.Metric{}
+	result := v1alpha1.MetricResult{
+		Failed: 1,
+		Measurements: []v1alpha1.Measurement{{
+			Phase: v1alpha1.AnalysisPhaseFailed,
+		}},
+	}
+	phase, msg := assessMetricFailureInconclusiveOrError(metric, result)
+	msg0, msg1 := fmt.Sprintf("assessed %s", phase), fmt.Sprintf("failed (%d) > failureLimit (%d)", result.Failed, metric.FailureLimit)
```
dthomson25: I think we should break this up into two lines since you call multiple functions in a declaration

analysis/analysis.go
khhirani: Passing a message through assessMetricStatus caused some unexpected behavior, so it's easier to have a separate method instead.

analysis/analysis.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	analysisutil "github.com/argoproj/argo-rollouts/utils/analysis"
+	util "github.com/argoproj/argo-rollouts/utils/defaults"
```
dthomson25: we should just use the package name default instead of util

utils/defaults/defaults_test.go
```diff
@@ -164,3 +164,10 @@ func TestGetExperimentProgressDeadlineSecondsOrDefault(t *testing.T) {
 	defaultValue := &v1alpha1.Experiment{}
 	assert.Equal(t, DefaultProgressDeadlineSeconds, GetExperimentProgressDeadlineSecondsOrDefault(defaultValue))
 }
+
+func TestGetConsecutiveErrorLimitOrDefault(t *testing.T) {
+	errorLimit := (*int32)(nil)
```
dthomson25: I think you could rewrite this as `var errorLimit *int32`

utils/defaults/defaults_test.go
```diff
@@ -164,3 +164,10 @@ func TestGetExperimentProgressDeadlineSecondsOrDefault(t *testing.T) {
...
+	metricNonDefaultValue := v1alpha1.Metric{}
+	assert.Equal(t, errorLimit, metricNonDefaultValue.ConsecutiveErrorLimit)
```
dthomson25: I don't think we need to have this assertion since the `metricNonDefaultValue` declaration above implies that the `ConsecutiveErrorLimit` is nil

utils/defaults/defaults_test.go
```diff
@@ -164,3 +164,10 @@ func TestGetExperimentProgressDeadlineSecondsOrDefault(t *testing.T) {
 	defaultValue := &v1alpha1.Experiment{}
 	assert.Equal(t, DefaultProgressDeadlineSeconds, GetExperimentProgressDeadlineSecondsOrDefault(defaultValue))
 }
+
+func TestGetConsecutiveErrorLimitOrDefault(t *testing.T) {
```
dthomson25: Can you add a list when the value is found?

analysis/analysis_test.go
```diff
@@ -493,7 +499,7 @@ func TestAssessRunStatusUpdateResult(t *testing.T) {
 			},
 		},
 	}
-	status := c.asssessRunStatus(run)
+	status, _ := c.assessRunStatus(run)
```
dthomson25: Let's confirm that the returned message is ""

analysis/analysis_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/argoproj/argo-rollouts/utils/defaults"
+
```
dthomson25: Please move this import to the third group of imports

PR #439, opened 2020-03-12
fix: Update Role/ClusterRole for Ingress access

moensch:
And ran `make manifests` to update consolidated manifests.


moensch:
Tested the changes in my local setup and they're good to go.

PR #444, opened 2020-03-19
feat: Add ALB Ingress controller support

dthomson25:
Closes https://github.com/argoproj/argo-rollouts/issues/421

Check out the alb.md for more information on the change.


utils/json/json_test.go
```diff
@@ -0,0 +1,30 @@
...
+type I interface {
+	Foo() string
+}
+
+func TestMustMarshalPanics(t *testing.T) {
+	test := map[string]interface{}{
+		"foo": make(chan int),
+	}
+	assert.Panics(t, func() { MustMarshal(test) })
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
}
```

utils/ingress/ingress_test.go
```diff
@@ -88,3 +92,62 @@ func TestGetCanaryIngressName(t *testing.T) {
...
+		assert.Contains(t, m.String(), "ro1:alb.ingress.kubernetes.io/actions.svc1")
+		assert.Contains(t, m.String(), "ro2:alb.ingress.kubernetes.io/actions.svc2")
+	})
+
+}
+
+func TestALBActionAnnotationKey(t *testing.T) {
+	expected := "alb.ingress.kubernetes.io/actions.svc"
+	assert.Equal(t, expected, ALBActionAnnotationKey("svc"))
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
}
```

ingress/alb_test.go
```diff
@@ -0,0 +1,169 @@
...
+				"Weight": 15
+			}
+		]
+	}
+}`
+
+func newALBIngress(name string, port int, serviceName string, rollout string) *extensionsv1beta1.Ingress {
+	canaryService := fmt.Sprintf("%s-canary", serviceName)
+	albActionKey := ingressutil.ALBActionAnnotationKey(serviceName)
+	managedBy := fmt.Sprintf("%s:%s",rollout, albActionKey)
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	managedBy := fmt.Sprintf("%s:%s", rollout, albActionKey)
```

ingress/alb.go
```diff
@@ -0,0 +1,91 @@
+package ingress
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/sirupsen/logrus"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

utils/json/json_test.go
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

ingress/ingress_test.go
```diff
@@ -142,3 +152,35 @@ func TestSyncIngressReferencedByRollout(t *testing.T) {
...
+	}
+
+	ctrl, kubeclient, enqueuedObjects := newFakeIngressController(ing, rollout)
+
+	err := ctrl.syncIngress("default/test-stable-ingress")
+	assert.NoError(t, err)
+	actions := kubeclient.Actions()
+	assert.Len(t, actions, 0)
+	assert.Len(t, enqueuedObjects, 0)
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

ingress/ingress.go
```diff
@@ -121,12 +125,29 @@ func (c *Controller) syncIngress(key string) error {
...
 
-	if rollouts, err := c.getRolloutsByIngress(namespace, name); err == nil {
-		for i := range rollouts {
-			// reconciling the Rollout will ensure the canaryIngress is updated or created
-			c.enqueueRollout(rollouts[i])
-		}
+	switch ingress.Annotations["kubernetes.io/ingress.class"] {
+	case "aws-alb":
+		return c.syncALBIngress(ingress, rollouts)
+	case "nginx":
```
cronik: Does this mean that all Ingress configs must have the kubernetes.io/ingress.class annotation to work with traffic shaped rollouts? How would this work for clusters with multiple nginx ingresses with different names or don't use the name?

In our large shared cluster we actually have multiple nginx ingress for different envs and service requirements, with this restriction all those would be excluded from using this feature.

dthomson25: Hey, @cronik that's a great point! I'm going to add some command-line flags that allow you to specify which values of kubernetes.io/ingress.class are used for nginx and albs

dthomson25: I also just saw that users can change their annotation prefix (i.e. `alb.ingress.kubernetes.io`) if they want so I'll add that functionality too

dthomson25: Okay, I added the `--alb-ingress-classes` and `--nginx-ingress-classes` flags to address these usecases.

cronik: would it also make sense to add an option like `--default-ingress-class` for cases where it is not annotated explicitly?

dthomson25: I.E if there's no "kubernetes.io/ingress.class" annotation, have the flag dicate that the controller should consider it an NGINX or ALB ingress? Do the ingress controller support that behavior?

cronik: Yeah, I don't know about ALB, but nginx-ingress has this behavior by default unless the ingress class name is explicitly set on the ingress controller.

https://kubernetes.github.io/ingress-nginx/user-guide/multiple-ingress/#multiple-ingress-nginx-controllers
https://kubernetes.github.io/ingress-nginx/user-guide/cli-arguments/

| arg | Description |
| --- | ------------ |
| `--ingress-class string` | Name of the ingress class this controller satisfies. The class of an Ingress object is set using the annotation "kubernetes.io/ingress.class". All ingress classes are satisfied if this parameter is left empty. |

Basically when `--ingress-class` is not set it acts as the default controller for all ingress.
Our cluster has multiple nginx ingress controllers, most ingress don't specify the ingress class annotation and get picked up by the default controller, when an ingress needs a specific class then the annotation is added to restrict to a specific one.

docs/features/traffic-management/alb.md
```diff
@@ -0,0 +1,67 @@
...
+
+Since the ALB Ingress controller allows users to configure the annotation prefix used by the Ingress controller, Rollouts can specify the optional `annotationPrefix` field. The Ingress uses that prefix instead of the default `alb.ingress.kubernetes.io` if the field set.
+
+The Rollout adds another annotation called `rollouts.argoproj.io/managed-alb-actions` to the Ingress to help the controller manage the Ingresses. This annotation indicates which actions are being managed by Rollout objects (since multiple Rollouts can reference one Ingress). If a Rollout is deleted, the Argo Rollouts controller uses this annotation to see that this action is no longer managed, and it is reset to only the stable service with 100 weight.
```
jessesuen: > If a Rollout is deleted, the Argo Rollouts controller uses this annotation to see that this action is no longer managed, and it is reset to only the stable service with 100 weight.

I wanted to call out that this behavior is such that we would be leaving around our annotation when the original ingress object did not have it in the beginning. 

I **_think_** this is the correct behavior, since it is safer than deleting the annotation, which could cause downtime, but just wanted to point out that we could be leaving around leftover cruft.

rollout/trafficrouting/alb/alb.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *Reconciler) Reconcile(desiredWeight int32) error {
+	rollout := r.cfg.Rollout
+	ingressName := rollout.Spec.Strategy.Canary.TrafficRouting.ALB.Ingress
+	ingress, err := r.cfg.IngressLister.Ingresses(rollout.Namespace).Get(ingressName)
+	if err != nil {
+		return err
+	}
+	stableService := r.cfg.Rollout.Spec.Strategy.Canary.StableService
+	port := r.cfg.Rollout.Spec.Strategy.Canary.TrafficRouting.ALB.ServicePort
+	if !ingressutil.HasRuleWithService(ingress, stableService) {
```
jessesuen: We error here if there is no rule mentioning the service. But how does this work initially, when the ingress doesn't have the annotation yet?

cmd/rollouts-controller/main.go
```diff
@@ -147,6 +154,8 @@ func newCommand() *cobra.Command {
 	command.Flags().IntVar(&serviceThreads, "service-threads", controller.DefaultServiceThreads, "Set the number of worker threads for the Service controller")
 	command.Flags().IntVar(&ingressThreads, "ingress-threads", controller.DefaultIngressThreads, "Set the number of worker threads for the Ingress controller")
 	command.Flags().StringVar(&istioVersion, "istio-api-version", defaultIstioVersion, "Set the default Istio apiVersion that controller should look when manipulating VirtualServices.")
+	command.Flags().StringArrayVar(&albIngressClasses, "alb-ingress-classes", defaultALBIngressClass, "Set the default Istio apiVersion that controller should look when manipulating VirtualServices.")
+	command.Flags().StringArrayVar(&nginxIngressClasses, "nginx-ingress-classes", defaultNGINXIngressClass, "Set the default Istio apiVersion that controller should look when manipulating VirtualServices.")
```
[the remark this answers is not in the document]
jessesuen: For @cronik's request regarding `--default-ingress-class`, I agree we need to solve that problem.

However, my question is: will this now solved if someone supplies the empty string? e.g.:

```
--nginx-ingress-classes nginx --nginx-ingress-classes ''
```

ingress/ingress.go
dthomson25: I tested Jesse's solution made in another comment successful (`--nginx-ingress-classes nginx --nginx-ingress-classes '')`. Since there is no good way to pick a default value, I think it is better to leave it unset, and allow users to specify it themselves if needed.

Issue #144, opened 2019-08-21
Ability to restart pods while skipping pause/steps


jessesuen:
Similar to deployment's ability to restart pods using an annotation (https://github.com/kubernetes/kubernetes/pull/76062), Rollouts needs to support this as well.

Technically, Rollouts already supports this, as long as:
* there is no pause before blue-green (`spec.strategy.blueGreen.autoPromotionEnabled: false`)
* there are no canary steps (`spec.strategy.canary.steps`)

But because Rollouts has the ability/concept to pause & run steps, which the Deployment resource does not have, it has a problem where users want to trigger a pod restart but without going through the pause/steps.

jessesuen:
I think the solution to this may be the following: in order to signal to the controller that pods should be restarted while skipping canary steps & blue-green pause, the user action should be to set a timestamp annotation at both the template level and at the rollout level. i.e.

```yaml
metadata:
  annotations:
    rollout.argoproj.io/restartedAt: 2019-10-11T01:02:03
 spec:
  template:
    metadata:
      annotations:
        rollout.argoproj.io/restartedAt: 2019-10-11T01:02:03
```

The logic in the controller would be:
1. if current pod template hash is ever different than the replicaset's pod template hash, we know we need to redeploy (no change to current logic)
2. when deciding whether or not to skip canary steps and the blue-green pause, first examine `rollout.argoproj.io/restartedAt` at both `metadata.annotations` and `spec.template.metadata.annotations`
3. if the values of those two are identical, then this is detected as a pod restart event and we will decide to skip steps or the blue-green pause.
4. when creating the new replicaset for the pod restart, inject a permanent indicator `rollout.argoproj.io/autopromote: "true"` into the replicaset annotations.
5. for every reconcile going forward, if we see ever see `rollout.argoproj.io/autopromote: "true"` in the annotations of a replicaset, we will always skip steps and the blue-green pause

carlosdoordash:
Using this solution means modifying the rollout spec. Doesn't that mean that we lose the ability to quickly rollback to a non-scaled-down replicaset by just patching the container image?

jessesuen:
If we are going back to a blue-green replicaset within its scale-down delay, or back to the "stable" hash of a canary replicaset, this is considered a fast-tracked rollback and the steps are skipped. This ability will not be lost with this feature.

jessesuen:
[...]

PR #453, opened 2020-04-01
feat: Add ability to restart Pods

dthomson25:
Closes https://github.com/argoproj/argo-rollouts/issues/144

Allows users to restart the pods of a Rollout by setting the Rollout's spec.restartedAt to a specific time. We accomplish this behavior by killing one pod at a time. Additionally, this allows Rollouts to have a scheduled restart by setting the spec.restartedAt to a future time. The pod restarts only occur if the current time is past the set restart time.

TODO:
- [x] Add docs
- [x] Add delete pod permissions to rbac


rollout/bluegreen.go
```diff
@@ -35,6 +35,12 @@ func (c *RolloutController) rolloutBlueGreen(r *v1alpha1.Rollout, rsList []*apps
 
 	roCtx := newBlueGreenCtx(r, newRS, oldRSs, arList)
 	logCtx := roCtx.Log()
+
+	err = c.podRestarter.Reconcile(roCtx)
```
jessesuen, line 39: With this here, it seems like we still perform restarts over pod template changes. Isn't it desired that if we ever change the pod spec, we should cancel the restart request (since restarting pods is useless if we are about to move to a new spec)?

rollout/restart.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/argoproj/argo-rollouts/utils/replicaset"
+)
+
+const (
+	// restartPodCheckTime prevents the Rollout from not making any progress with restarting Pods. When pods can be restarted
+	// faster than the old pods can be scaled down, the parent's ReplicaSet's availableReplicas does not change. A rollout
+	// uses changes to the availableReplicas of the ReplicaSet to detect when the controller should requeue and continue
+	// deleting pods. In this situation, the rollout does not renqueue and wont make any more progress restarting pods until
+	// the resync period passes or another change is made to the Rollout. The controller requeue Rollouts with deleted
+	// Pods every 30 seconds to make sure the rollout is not stuck.
```
jessesuen, line 20: s/Pods/Polls/

rollout/restart.go
```diff
@@ -0,0 +1,163 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/argoproj/argo-rollouts/utils/replicaset"
+)
+
+const (
+	// restartPodCheckTime prevents the Rollout from not making any progress with restarting Pods. When pods can be restarted
+	// faster than the old pods can be scaled down, the parent's ReplicaSet's availableReplicas does not change. A rollout
+	// uses changes to the availableReplicas of the ReplicaSet to detect when the controller should requeue and continue
```
jessesuen, line 17: s/uses/relies on/

rollout/restart.go
```diff
@@ -0,0 +1,163 @@
...
+		if reconciledReplicaSet {
+			return nil
+		}
+	}
+	logCtx.Info("all pods have been restarted and setting restartedAt status")
+	roCtx.SetRestartedAt()
+	return nil
+}
+
+func (p RolloutPodRestarter) reconcilePodsInReplicaSet(roCtx rolloutContext, rs *appsv1.ReplicaSet) (bool, error) {
```
jessesuen, line 79: Could you add comment what the purpose of this method is, and the meaning of a true/false return value.

I think we should rename this to indicate what it's doing since `reconcilePodsInReplicaSet` is nondescript (e.g. `restartReplicaSetPod`). I actually initially thought the `reconciledReplicaSet` boolean value meant that all pods are restarted properly, but it actually means the opposite.

rollout/restart.go
```diff
@@ -0,0 +1,163 @@
...
+	logCtx := roCtx.Log().WithField("Reconciler", "PodRestarter")
+	restartedAt := roCtx.Rollout().Spec.RestartAt
+	pods, err := p.client.CoreV1().Pods(rs.Namespace).List(metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(rs.Spec.Selector),
```
jessesuen, line 83: This is selecting **all** pods with the label selector. However, there could be overlap with pods from another deployment/rollout. I think this needs to also filter pods not owned by the replicaset, lest we terminate pods not owned by the replicaset.

rollout/bluegreen.go
dthomson25: I moved the reconcile calls to after the pod template check and made sure to set the restartedat time if the pod template has changed.

dthomson25:
If you restart a rollout whose replicas are scaled to zero, the rollout is considered to be restarted and it sets the `status.restartedAt` time

PR #462, opened 2020-04-03
feat: adds nodeselector to deployment

rparsonsbb:
Adds note about not using install.yaml from master
adds blackbook to official rollout user
adds OS nodeselector to all deployments

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,15 @@ Deployments resources offer two strategies to deploy changes: `RollingUpdate` an
 ## Documentation
 To learn more about Argo Rollouts go to the [complete documentation](https://argoproj.github.io/argo-rollouts/).
 
+> Do not install argo-rollouts from the master branch install.yaml.  This manifest uses the `latest` tag for the image which may be unstable.  Make sure to use the `stable` branch for all installations unless testing the latest iteration.
+
 ## Who uses Argo Rollouts?
 Organizations below are **officially** using Argo Rollouts. Please send a PR with your organization name if you are using Argo Rollouts.
 
 1. [ADP](https://www.adp.com)
 1. [Intuit](https://www.intuit.com/)
 1. [Twilio SendGrid](https://sendgrid.com)
+1. [Blackbook](https://blackbook.com)
 
 ## Community Blogs and Presentations
 * [How Intuit Does Canary and Blue Green Deployments](https://www.youtube.com/watch?v=yeVkTTO9nOA)
--- a/manifests/base/argo-rollouts-deployment.yaml
+++ b/manifests/base/argo-rollouts-deployment.yaml
@@ -15,6 +15,8 @@ spec:
       labels:
         app.kubernetes.io/name: argo-rollouts
     spec:
+      nodeSelector:
+        beta.kubernetes.io/os: linux
       serviceAccountName: argo-rollouts
       containers:
       - command:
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -12,12 +12,15 @@ Deployments resources offer two strategies to deploy changes: `RollingUpdate` an
 ## Documentation
 To learn more about Argo Rollouts go to the [complete documentation](https://argoproj.github.io/argo-rollouts/).
 
+> Do not install argo-rollouts from the master branch install.yaml.  This manifest uses the `latest` tag for the image which may be unstable.  Make sure to use the `stable` branch for all installations unless testing the latest iteration.
```
jessesuen, line 15: We shouldn't need to call this out. In place of this comment, I think the GitHub README.md should have a Quickstart section (the same to the documentation), like:

## Quickstart

Argo Rollouts can be installed by running the following commands:
```
$ kubectl create namespace argo-rollouts
$ kubectl apply -n argo-rollouts -f https://raw.githubusercontent.com/argoproj/argo-rollouts/stable/manifests/install.yaml
```

README.md
```diff
@@ -12,12 +12,15 @@ Deployments resources offer two strategies to deploy changes: `RollingUpdate` an
...
+
 ## Who uses Argo Rollouts?
 Organizations below are **officially** using Argo Rollouts. Please send a PR with your organization name if you are using Argo Rollouts.
 
 1. [ADP](https://www.adp.com)
 1. [Intuit](https://www.intuit.com/)
 1. [Twilio SendGrid](https://sendgrid.com)
+1. [Blackbook](https://blackbook.com)
```
jessesuen, line 23: 👍 🎉

manifests/base/argo-rollouts-deployment.yaml
```diff
@@ -15,6 +15,8 @@ spec:
       labels:
         app.kubernetes.io/name: argo-rollouts
     spec:
+      nodeSelector:
+        beta.kubernetes.io/os: linux
```
jessesuen, line 19: We can't not enforce these labels here:
1. this specific label has already been deprecated in favor of `kubernetes.io/os`
2. this label is too opinionated may not be consistent across kubernetes installations

rparsonsbb: Its a tough one for sure and I use kubernetes.io/os in my own version, without this though you'll find that this is broken for any mixed os cluster.

If this isn't added then much like the getting started in the readme there should be a note for mixed OS clusters manually adding the nodeSelector on their own?

README.md
rparsonsbb: Cool

manifests/base/argo-rollouts-deployment.yaml
cronik: [kustomize](https://kustomize.io/) is probably what you are looking for to apply your additional requirements to the base install.

dthomson25:
Hey @rparsonsbb, I wanted to circle back on this PR. Could we add a comment to the installation guide that this version runs on Linux and you will need to add a node selector like the following if you have a cluster with multiple OSs?
```yaml
spec:
  template:
    spec:
      nodeSelector:
        beta.kubernetes.io/os: linux
```

ivan-santos-eq:
Thoughts on closing this one in favor of using kustomize instead?

PR #445, opened 2020-03-19
Create Rollout Strategy Anti-Affinity

khhirani:
WIP

PR #463, opened 2020-04-07
docs: Anti-Affinity Documentation

khhirani:
For #445


docs/features/anti-affinity/anti-affinity.md
```diff
@@ -0,0 +1,60 @@
+# Anti Affinity
+
+Depending on a cluster's configuration, a Blue Green Rollout (or a Canary rollout that uses traffic management) can cause newly created pods to restart after deploying a new version. This can be problematic, especially for applications that cannot startup quickly or do not gracefully exit.
+
+This behavior occurs because cluster auto-scaler wants to scale down the extra capacity which was created to support a Rollout
+running in double capacity. It usually happens if a Rollout has its own dedicated instance group since a Rollout has a greater effect on
```
dthomson25, line 6: We should add a quick comment connecting this back to the restarting pods. I.E. If the scaled-down node had a pod owned the Rollout, that pod would be deleted and recreated.

docs/features/anti-affinity/anti-affinity.md
```diff
@@ -0,0 +1,60 @@
...
+![ Original Rollout is running, spread across two nodes](images/step-0.png)
+
+When the `spec.template` of the Rollout changes, the controller creates a new ReplicaSet with the spec update and the total number of pods doubles. In this case, the number of pods increases
+to 16.
+
+Since each node can only hold 6 pods, the cluster autoscaler must increase the node count to 3 to accommodate
+all 16 pods. The resulting distribution of pods across nodes is shown here:
+![ Rollout running at double capacity](images/step-1.png)
+
+Once the Rollout finishes progressing, the old version is scaled down. This leaves the cluster with more nodes than necessary, thus wasting resources (as shown below).
```
dthomson25, line 19:
```suggestion
Once the Rollout finishes progressing, the old ReplicaSet is scaled down. This leaves the cluster with more nodes than necessary, thus wasting resources (as shown below).
```

docs/features/anti-affinity/anti-affinity.md
```diff
@@ -0,0 +1,60 @@
...
+![ Original Rollout is running, spread across two nodes](images/step-2.png)
+
+The cluster auto-scaler terminates the extra node and the pods are rescheduled on the remaining 2 nodes.
+![ Original Rollout is running, spread across two nodes](images/step-3.png)
+
+To reduce the chance of this behavior, a rollout can inject anti-affinity. This prevents new pods on nodes which have the previous version's pods.
```
dthomson25, line 25: I think we should specify where the anti-affinity is injected (i.e. into the ReplicaSet)

dthomson25: Can we say `new pods from running on nodes` instead of `new pods on nodes`?

docs/features/anti-affinity/anti-affinity.md
```diff
@@ -0,0 +1,60 @@
...
+
+You can learn more about anti-affinity [here](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/).
```
dthomson25, line 27: Can we change the link to https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity to make it point at the anti-affinity section?

docs/features/anti-affinity/anti-affinity.md
```diff
@@ -0,0 +1,60 @@
...
+
+Repeating the above example with anti-affinity enabled, here is what happens when the `.spec.template` of the Rollout changes. Due to anti-affinity, the new pods cannot be scheduled on nodes which run the old ReplicaSet's pods.
+As a result, cluster auto-scaler must create 2 nodes to host the new ReplicaSet's pods.
```
dthomson25, line 30: We should add one last comment saying that we wont restart the pods in this case because the nodes scaled down are guaranteed to not have the new Pods.

docs/features/anti-affinity/anti-affinity.md
```diff
@@ -0,0 +1,60 @@
...
+
+```yaml
+strategy:
+    canary:
+      antiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            weight: 1 # Between 1 - 100
+```
+
+The main downside to this approach is that deployments can take longer because new nodes are more likely to be created in order to schedule pods with respect to anti-affinity rules. This delay most frequently occurs when a rollout has its own dedicated instance group,
```
dthomson25, line 59: We should add an important header here like [the following](https://github.com/argoproj/argo-rollouts/blame/master/docs/getting-started.md#L10) to help highlight this paragraph

dthomson25: here's how the example I gave looks when compiled: https://argoproj.github.io/argo-rollouts/getting-started/

PR #465, opened 2020-04-08
feat: Improve wavefront provider

### LINKED ISSUE #452: More information on failing wavefront queries
When a wavefront query returns no results, it is difficult to debug why it failed. Below is an example of a failed query:
```
  metricResults:
  - consecutiveError: 3
    error: 3
    measurements:
    - finishedAt: "2020-03-31T17:50:04Z"
      message: No TimeSeries found in response from Wavefront
      phase: Error
      startedAt: "2020-03-31T17:50:03Z"
    - finishedAt: "2020-03-31T17:51:04Z"
      message: No TimeSeries found in response from Wavefront
      phase: Error
      startedAt: "2020-03-31T17:51:04Z"
    - finishedAt: "2020-03-31T17:52:04Z"
      message: No TimeSeries found in response from Wavefront
      phase: Error
      startedAt: "2020-03-31T17:52:04Z"
    name: success-rate
    phase: Running
  phase: Running
```
This query doesn't have any information on why it failed. Was it a valid query? Was one of the tags not found?

To address this, we should print the full query in the metrics results. Additionally, the wavefront UI offers hints as to why a query failed. We should see if those hints are returned in the Query API and store them in the results.
image

Note: It is likely we can do something similar with Prometheus metrics.

dthomson25:
Closes #452 by
* Grabbing the datapoint closest to the startTime and records which Epoch was used in the metadata
* Adds an Endtime equal to the StartTime so that the query is easily reproducible (where before the EndTime was null which wavefront treats as now)
* Adds the warning if the response includes them

This PR also changes the service controller to use a merge patch instead of a JSON patch.

### DIFF
--- a/metricproviders/wavefront/wavefront.go
+++ b/metricproviders/wavefront/wavefront.go
@@ -11,7 +11,7 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
-	wavefront_api "github.com/spaceapegames/go-wavefront"
+	wavefrontapi "github.com/spaceapegames/go-wavefront"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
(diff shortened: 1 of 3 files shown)


metricproviders/wavefront/wavefront.go
```diff
@@ -37,55 +37,67 @@ func (p *Provider) Type() string {
...
+	result, err := p.processResponse(metric, response, startTime)
 	if err != nil {
 		return metricutil.MarkMeasurementError(newMeasurement, err)
 
 	}
-	newMeasurement.Value = newValue
-	newMeasurement.Phase = newStatus
+	newMeasurement.Value = result.newValue
+	newMeasurement.Phase = result.newStatus
+	newMeasurement.Metadata["epochs-used"] = result.epochsUsed
```
jessesuen, line 100: 1. "epochs" is the plural of epoch, which in this case doesn't make sense, since it is a single value
2. i think we can just call this `time`, which is the terminology prometheus and wavefront seem to use.

dthomson25: We discussed offline and decided on timestamps since there can be multiple timestamps if there are multiple timeseries

PR #477, opened 2020-04-16
chore: Consistently use Config struct for controller constructors

moensch:
Starting in #426 we changed the constructors for the new ingress controller, and the one for the `service` controller to accept a config struct to avoid potential issues with out of order parameters.

This extends that change to the remaining controllers and also removes "stuttering" (`analysis.AnalysisController -> analysis.Controller`)

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -8,7 +8,7 @@ set -o pipefail
 #   create a temporary directory, use this as an output base, and copy everything back once generated.
 export GOPATH=$(go env GOPATH) # export gopath so it's available to generate scripts
 SCRIPT_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null 2>&1 && pwd )"
-CODEGEN_VERSION=$(go list -m k8s.io/code-generator | awk '{print $2}' | head -1)
+CODEGEN_VERSION=$(go list -m k8s.io/code-generator | awk '{print $NF}' | head -1)
 CODEGEN_PKG="${GOPATH}/pkg/mod/k8s.io/code-generator@${CODEGEN_VERSION}"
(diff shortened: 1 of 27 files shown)


moensch:
I think that's enough renaming and shuffling for one PR :)

hack/update-codegen.sh
```diff
@@ -8,7 +8,7 @@ set -o pipefail
 #   create a temporary directory, use this as an output base, and copy everything back once generated.
 export GOPATH=$(go env GOPATH) # export gopath so it's available to generate scripts
 SCRIPT_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null 2>&1 && pwd )"
-CODEGEN_VERSION=$(go list -m k8s.io/code-generator | awk '{print $2}' | head -1)
+CODEGEN_VERSION=$(go list -m k8s.io/code-generator | awk '{print $NF}' | head -1)
```
moensch, line 11: Oh I should have highlighted this change in the summary. I think this change I added will make this more reliable since on my machine I ended up with this:

```
$>  go list -m k8s.io/code-generator
k8s.io/code-generator v0.17.3 => k8s.io/code-generator v0.17.4-beta.0

$> ls -ld $GOPATH/pkg/mod/k8s.io/code-generator*
dr-x------  20 xx  xx  640 Feb 27 21:41 /Users/xx/go/pkg/mod/k8s.io/code-generator@v0.16.4
dr-x------  20 xx  xx  640 Mar 10 23:40 /Users/xx/go/pkg/mod/k8s.io/code-generator@v0.17.4-beta.0
```

dthomson25: I just confirmed it on my machine, so we should be good to go!

Issue #359, opened 2020-01-15
SMI (Service Mesh Interface) TrafficSplit Support


jessesuen:
For canary traffic routing, In addition to:
* istio supoprt (merged)
* nginx-ingress ([proposed](https://github.com/argoproj/argo-rollouts/blob/master/docs/features/traffic-management/nginx.md))

We also need a generic SMI traffic routing provider to support meshes like Linkerd, or even Istio using the [SMI adapter for Istio](https://github.com/deislabs/smi-adapter-istio).

This implementation should be much simpler than Istio provider since the `TrafficSplit` resource could be completely owned by the `Rollout` object itself, as opposed to referenced like the Istio provider.

wmorgan:
Cool, please let us know how we can help from the Linkerd side

ilevine:
Would you consider to add support for Gloo, envoy based API gateway? we would like to help
https://github.com/solo-io/gloo

jessesuen:
Thanks for the offers of support! I think we might ask for some feedback on the proposal and API specification. We've already discussed what this might look like, internally but will formalize it into a proposal and send it out to the community. That said, the `TrafficSplit` interface is so simple that I can't imagine the spec to be any more complicated than:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
spec:
  ...
  strategy:
    canary:
      canaryService: canary-service # required
      stableService: stable-service  # required
      trafficRouting:
        smi: {}
      steps:
      - setWeight: 10
      - pause: {duration: 1h}
      - setWeight: 20
...
```

@ilevine yes, the goal for this is for rollouts to be extensible to support different service mesh / ingress controllers. The SMI provider support (this issue) will give us the common denominator for basic weight based traffic splits. For any advanced routing features which only available to specific mesh providers (e.g. istio header based traffic splitting), we would need have mesh specific fields in the specification.

wmorgan:
cc @grampelberg

grampelberg:
This is awesome! I'd love to do anything that I can to help out =)

jessesuen:
I filed a second issue https://github.com/argoproj/argo-rollouts/issues/406, which can/should/will be implemented independently, to support SMI TrafficMetrics as an additional metric provider to drive canary analysis and promotion.

This issue is only aimed to achieve weighted traffic split between the canary vs. stable Service, through the manipulation of the SMI `TrafficSplit` resource.

[...]

PR #487, opened 2020-04-24
docs: SMI Proposal

dthomson25:
This PR introduces a proposal to support SMI and is the first part of solving https://github.com/argoproj/argo-rollouts/issues/359. 

At a high level, the proposal suggests add the following an SMI section to the trafficRouting section of the Rollout:
```yaml
  strategy:
    canary:
      steps:
      - setWeight: 5
      - pause:
          duration: 600
      canaryService: canary-svc # required
      stableService: stable-svc # required
      trafficRouting:
        # New fields added
        smi:
         rootService: root-svc # required
         trafficSplitName: rollout-example-traffic-split # optional
```

which then creates a TrafficSplit with the following yaml:
```yaml
apiVersion: split.smi-spec.io/v1alpha1
kind: TrafficSplit
metadata:
  name: rollout-example-traffic-split
spec:
  service: root-svc
  backends:
  - service: stable-svc
    weight: 5
  - service: canary-svc
    weight: 95
```


docs/features/traffic-management/smi.md
```diff
@@ -0,0 +1,58 @@
...
+    canary:
+      steps:
+      - setWeight: 5
+      - pause:
+          duration: 600
+      canaryService: canary-svc # required
+      stableService: stable-svc # required
+      trafficRouting:
+        smi:
+         rootService: root-svc # required
```
grampelberg, line 33: Any reason to expose the `root-svc` to users? You should be able to treat `stable-svc` the same way, and then this all works for clients that don't have traffic split support (eg. anything unmeshed).

docs/features/traffic-management/smi.md
```diff
@@ -0,0 +1,58 @@
...
+spec:
+  service: root-svc
+  backends:
+  - service: stable-svc
+    weight: 5
+  - service: canary-svc
+    weight: 95
+```
+
+As a Rollout progresses through all its steps, the controller updates the TrafficSplit's backend weights to reflect the current weight of the Rollout. When the Rollout has successfully finished executing all the steps, the controller modifies the TrafficSplit to send 100% of traffic to the stable Service.
```
grampelberg, line 55: You might want to talk through what happens with the labels and how traffic is actually handled under the covers. This is a great overview but I find the picture super helpful for explaining how the system fits together.

docs/features/traffic-management/smi.md
dthomson25: Are you suggesting something like the following?
```yaml
apiVersion: split.smi-spec.io/v1alpha1
kind: TrafficSplit
metadata:
  name: rollout-example-traffic-split
spec:
  service: stable-svc
  backends:
  - service: stable-svc
    weight: 95
  - service: canary-svc
    weight: 5
```
Wouldn't that violate the fourth tradeoff listed in the [SMI TrafficSplit spec](https://github.com/servicemeshinterface/smi-spec/blob/master/traffic-split.md#tradeoffs):
```
* TrafficSplits cannot be self-referential
```

If having a self-referential TrafficSplit is allowed, then the rootService could be optional and it should reference the stableService if anything.

docs/features/traffic-management/smi.md
dthomson25: Great suggestion! I added a picture to make it a little more clear what components are involved. Let me know what you think!

docs/features/traffic-management/smi.md
grampelberg: > * TrafficSplits cannot be self-referential

This is specifically related to recursion, you can't have a traffic split that forwards to a traffic split. In retrospect, I probably could have worded that a little bit better.

> If having a self-referential TrafficSplit is allowed, then the rootService could be optional and it should reference the stableService if anything.

I think you're right. I'll need to double check out Flagger's doing it, but I believe that's the way they're making it all work under the covers.

docs/features/traffic-management/smi.md
grampelberg: Cool! That's super helpful!

docs/features/traffic-management/smi.md
dthomson25: Great! Just pushed the change to make the rootService optional

dthomson25:
@grampelberg, unless you have any other feedback, I'll merge the proposal in and @khhirani will do the implementation! Thanks again for taking the time to review!

grampelberg:
@dthomson25 go for it, thank you for involving me in the process! I'd love to help out during implementation any way I can =0

PR #489, opened 2020-05-04
docs: Add FAQ about controller behavior with new rollout

dthomson25:


docs/FAQ.md
```diff
@@ -14,6 +14,9 @@ Argo Rollouts does not require a service mesh or ingress controller to be used.
 ### Which deployment strategies does Argo Rollouts support?
 Argo Rollouts supports BlueGreen, Canary, and Rolling Update. Additionally, Progressive Delivery features can be enabled on top of the blue-green/canary update, which further provides advanced deployment such as automated analysis and rollback.
 
+## Does the Rollout object follow the provided strategy when it is first created?
+No, when the Rollout is first created, the controller does not use the provided strategy to bring up the first ReplicaSet. The controller tries to get the Rollout into a steady state as fast as possible by creating a fully scaled up ReplicaSet from the provided `.spec.template`. Once the Rollout has a stable ReplicaSet to transition from, the controller starts using the provided strategy to tranistion the previous ReplicaSet to the desired ReplicaSet.
```
jessesuen, line 18: It is true that Deployments also don't follow max-surge, etc.. during the initial deploy. So I would add a sentence compare this to deployments:

> ~No, when the Rollout is first created, the controller does not use the provided strategy to bring up the first ReplicaSet~

As with Deployments, Rollouts does not follow the strategy parameters on the initial deploy. Instead, the controller tries to get the Rollout into a steady state as fast as possible...

Issue #476, opened 2020-04-16
Argo Rollout 0.8.0 cannot be installed on GKE 1.14 


gendolf3d:
I have a GKE cluster with Kubernetes 1.14. Installation of Argo-rollouts 0.8.0 failed due to:
```
error: error validating "https://raw.githubusercontent.com/argoproj/argo-rollouts/v0.8.0/manifests/install.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.spec.properties.strategy.properties.canary.properties.maxSurge): unknown field "x-kubernetes-int-or-string" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.spec.properties.strategy.properties.canary.properties.maxUnavailable): unknown field "x-kubernetes-int-or-string" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps]; if you choose to ignore these errors, turn validation off with --validate=false
```

moensch:
I wouldn't consider this an issue - on Kubernetes versions below 1.16 you have to apply the CRD with `--validate=false`.

gendolf3d:
In this case Documentation should be updated.

dthomson25:
I created a PR to make this change, and I'll update the release to include this too.

PR #494, opened 2020-05-05
docs: Update install caveat to include v1.14

dthomson25:
Closes https://github.com/argoproj/argo-rollouts/issues/476. This issue occurs because Argo Rollouts v0.8 is starting to use the `x-kubernetes-int-or-string` in the validation, which is not available before v1.15.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -8,9 +8,7 @@
 Argo Rollouts can be installed at a cluster or namespace level. 
 
 !!! important
-    Since the API server rejects CRD manifests with unknown fields, the Argo Rollout's CRDs do not work with **v1.13.x** 
-    because they have a new field from the v1.14. Adding the `--validate=false` to the commands below allows the cluster to 
-    accept the Argo Rollouts CRDs.
+    When installing Argo Rollouts on Kubernetes v1.14 or lower, the CRD manifests must be kubectl applied with the --validate=false option. This is caused by use of new CRD fields introduced in v1.15, which are rejected by default in lower API servers.
 
 ### Cluster-Level installation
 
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -8,8 +8,8 @@
 Argo Rollouts can be installed at a cluster or namespace level. 
 
 !!! important
-    Since the API server rejects CRD manifests with unknown fields, the Argo Rollout's CRDs do not work with **v1.13.x** 
-    because they have a new field from the v1.14. Adding the `--validate=false` to the commands below allows the cluster to 
+    Since the API server rejects CRD manifests with unknown fields, the Argo Rollout's CRDs do not work with **v1.14.x**
+    because they have a new field from the v1.15. Adding the `--validate=false` to the commands below allows the cluster to
```
jessesuen, line 12: The way this is worded makes it seem we don't work with v1.14. Lets rephrase it as a workaround:

When installing Argo Rollouts on Kubernetes v1.14 or lower, the CRD manifests must be kubectl applied with the `--validate=false` option. This is caused by use of new CRD fields introduced in v1.15, which are rejected by default in lower API servers.

caoyang001: in kubernetes v1.14 the Rollout Controller found this err ，does that mean the Argo Rollout's CRDs do not work with v1.14?

```
ERRO[2021-05-10T17:07:05+08:00] rollout syncHandler error: the server could not find the requested resource (put rollouts.argoproj.io apiserver-demo)  namespace=default rollout=rollout-demo
E0510 17:07:05.309393     421 controller.go:172] the server could not find the requested resource (put rollouts.argoproj.io rollout-demo)
```

this may be relevant...:https://github.com/operator-framework/operator-sdk/issues/1124#issuecomment-499165754

PR #498, opened 2020-05-07
feat: add security context to run as non-root

duboisf:
Use numeric user in Dockerfile. This allows the use of a security context in kubernetes with `runAsNonRoot: true`, because kubernetes can inspect the image and see that the USER is != 0.

Removed user and group creation, it isn't necessary to get a working docker image. Also when running in openshift, the user in the docker image is usually ignored and a random user is used to run the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,12 +51,6 @@ ARG MAKE_TARGET="controller plugin-linux plugin-darwin"
 RUN make ${MAKE_TARGET}
 
 
-RUN groupadd -g 999 argo-rollouts && \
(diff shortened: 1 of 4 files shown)


duboisf:
Converted PR to draft as I want to run this version for a bit in all my clusters.

duboisf:
Ok it's working in our openshift and kops (in AWS) clusters

duboisf:
I did have to do a tweak to the `argo-rollouts-clusterrole` in openshift, might follow up with another PR for that. Was getting an error regarding blockOwnerDeletion and finalizers.

PR #510, opened 2020-05-15
fix: Add finalizers

johnkost:


dthomson25:
I just merged in https://github.com/argoproj/argo-rollouts/pull/511 so I think this PR is no longer needed. Thank you though for taking the time to make a PR!

PR #520, opened 2020-05-27
feat: SMI TrafficSplit Support for Canary

### LINKED ISSUE #359: SMI (Service Mesh Interface) TrafficSplit Support

khhirani:
Closes #359


rollout/trafficrouting/smi/smi_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	expectedTS := trafficSplit("traffic-split-name", rollout, 10)
+
+	assert.Equal(t, expectedTS.TypeMeta, ts.TypeMeta)
+	assert.Equal(t, expectedTS.ObjectMeta, ts.ObjectMeta)
+	assert.Equal(t, expectedTS.Spec.Service, ts.Spec.Service)
+	assert.Equal(t, expectedTS.Spec.Backends[0].Service, ts.Spec.Backends[0].Service)
+	assert.Equal(t, expectedTS.Spec.Backends[0].Weight.Value(), ts.Spec.Backends[0].Weight.Value())
+	assert.Equal(t, expectedTS.Spec.Backends[1].Service, ts.Spec.Backends[1].Service)
+	assert.Equal(t, expectedTS.Spec.Backends[1].Weight.Value(), ts.Spec.Backends[1].Weight.Value())
```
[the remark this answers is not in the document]
khhirani: It wasn't working because the Traffic Split spec contains Resource.Quantity pointers. (https://github.com/servicemeshinterface/smi-sdk-go/blob/master/pkg/apis/split/v1alpha1/traffic_split.go#L38)

rollout/trafficrouting/smi/smi.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// Type indicates this reconciler is an SMI reconciler
+func (r *Reconciler) Type() string {
+	return Type
+}
+
+// TODO: Make code compatible with multiple TrafficSplit versions
+func (r *Reconciler) Reconcile(desiredWeight int32) error {
+	trafficSplitName := r.cfg.Rollout.Spec.Strategy.Canary.TrafficRouting.SMI.TrafficSplitName
```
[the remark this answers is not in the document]
khhirani: The default right now is an empty string. Should we change it to something else?

rollout/trafficrouting_test.go
```diff
@@ -228,4 +228,18 @@ func TestNewTrafficRoutingReconciler(t *testing.T) {
...
+			SMI: &v1alpha1.SMITrafficRouting{},
+		}
+		roCtx := &canaryContext{
+			rollout: r,
+			log:     logutil.WithRollout(r),
+		}
+		rc.defaultTrafficSplitVersion = "v1alpha1"
+		networkReconciler := rc.NewTrafficRoutingReconciler(roCtx)
+		assert.NotNil(t, networkReconciler)
+		assert.Equal(t, smi.Type, networkReconciler.Type())
```
dthomson25, line 243: we should add another test for an unsupported apiVersion

rollout/trafficrouting_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting/smi"
```
dthomson25, line 7: Style nix: import ordering

rollout/trafficrouting_test.go
```diff
@@ -238,7 +257,8 @@ func TestNewTrafficRoutingReconciler(t *testing.T) {
 			log:     logutil.WithRollout(r),
 		}
 		rc.defaultTrafficSplitVersion = "v1alpha1"
```
dthomson25, line 259: Minor: In general, we have to be very careful about modifying structs that used in other tests otherwise they may have unintended consequences. For example, if this test was moved above the test that verifies the Unsupported TrafficSplit API Version error, that test would fail. As a result, we should declare our own controller object just for this test to prevent the global controller variable doesn't affect other tests. The tests are passing now, but it may cause a problems if these tests are refactored in the future and the developer doesn't realize that change.

PR #524, opened 2020-06-03
docs: Example of using with latest version of kustomize

mhaddon:


mhaddon:
As far as I am concerned this is ready to be merged.
Unfortunately there appears to be a spotty test which rejects the PR. 
Since I do not seem to have the ability to make it run another test, I will leave this PR open for someone else to merge when the tests work again.

dthomson25:
Rerun the tests and they are passing now. I'll take a look into the flaky test with this [issue](https://github.com/argoproj/argo-rollouts/issues/529). Thanks again for contributing!

PR #527, opened 2020-06-04
docs: Fix backoffLimit in wrong place

int32bit:
The property `backoffLimit` should be in `.job.spec.backoffLimit` rather than `.job.backoffLimit`.


jessesuen:
Thanks!

PR #548, opened 2020-06-24
docs: add istio getting started guide

jessesuen:
Guide on setting up istio on minikube, as well as getting started using Rollouts w/ Istio


docs/getting-started/setup/index.md
```diff
@@ -14,45 +14,194 @@ helm repo update
...
+[Istio getting started guide](../istio/index.md), the examples use the "default" istio 
+ingress gateway, which we can obtain from kubectl:
+
+```shell
+$ kubectl get svc -n istio-system istio-ingressgateway
+NAME                   TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                            AGE
+istio-ingressgateway   LoadBalancer   10.100.136.45   10.100.136.45   15020:31711/TCP,80:31298/TCP....   7d22h
+```
+
+In the above example, the hostname `rollouts-demo.local` should be configured to resolve to the
```
dthomson25, line 141: We should indicate that we are going to explain below

docs/getting-started/setup/index.md
```diff
@@ -14,45 +14,194 @@ helm repo update
...
+Ingress object itself, using `kubectl`:
 
 ```shell
 $ kubectl get ing rollouts-demo-stable
 NAME                   CLASS    HOSTS                 ADDRESS        PORTS   AGE
 rollouts-demo-stable   <none>   rollouts-demo.local   192.168.64.2   80      80m
 ```
 
-From the above output, the correct entry in `/etc/hosts` would be:
+In the example above, the hostname `rollouts-demo.local` should be configured to resolve to the
```
dthomson25, line 121: We should indicate that we are going to explain how to resolve that hostname correctly below

Issue #358, opened 2020-01-15
Rollout spec validation


jessesuen:
There are some *static* validation checks we can do to prevent updates of rollouts in an invalid way.

We need two things:
1. an *optional* [validating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) to prevent these rollouts from entering the system.
2. if the cluster does not support or is not configured to use admission webhooks, for the rollout to have an "InvalidSpec" condition, which is surfaced to the conditions, and subsequently to tools like Argo CD.

Some things to be aware of:
1.  We have to be cognizant that any referenced resources (e.g. Services and VirtualServices) could become invalid out-of-band. So for those class of errors, we **do not** to need to prevent the rollout admission from happening.
2. Nothing should prevent the controller from making status updates. For example even if the spec is invalid, the controller should still be able to update status. The validating webhook should not prevent this.
3. The order in which resources are applied, should not affect the ability to admit rollouts to the system. For example, if a rollout references a non-existant service, this might just be because the service object has not yet been `kubectl applied` to the system simply because it came lower in the YAML file.

jessesuen:
Just to elaborate. The checks performed by the validating admission webhook, need to actually be less strict than what is required when validating a rollout as part of reconciliation, which will be stricter and deeper. 

For example, the validating admission webhook should *not* deny admission of a rollout when the rollout references a non-existent service (imagine the scenario when the service is deleted out-of-band). This could cause the controller to stop updating status of the rollout. Instead, during reconciliation, the rollout will have an error condition which identifies the problem of the non-existent service.

jessesuen:
Examples of things that would need to be in the validating webhook and static validation during reconciliation:
* volumeMounts are referencing things that are listed in volumes.

dthomson25:
Another example would be preventing a rollout that creates an experiment that never terminates

jessesuen:
Experiment templates names should be lowercased since we will be having reserved words when referencing the `podTemplateHash` during experiment/analysis steps.

jessesuen:
Closing as fixed. Will punt on the validating admission webhook since we now have the tolerant informer preventing users from borking the controller with an invalid spec.

PR #549, opened 2020-06-24
feat: Controller Validation

khhirani:
Create static validation checks for Rollout spec, as described in #358


rollout/controller.go
```diff
@@ -315,8 +319,13 @@ func (c *Controller) syncHandler(key string) error {
 	}()
 
 	prevCond := conditions.GetRolloutCondition(rollout.Status, v1alpha1.InvalidSpec)
-	invalidSpecCond := conditions.VerifyRolloutSpec(r, prevCond)
-	if invalidSpecCond != nil {
+	rolloutValidationErrors := validation.ValidateRollout(rollout) //invalidSpecCond := conditions.VerifyRolloutSpec(r, prevCond)
```
dthomson25, line 322: Style Nix: Please remove //invalidSpecCond := conditions.VerifyRolloutSpec(r, prevCond)

rollout/controller_test.go
```diff
@@ -1055,15 +1060,15 @@ func TestPodTemplateHashEquivalence(t *testing.T) {
...
 
 	r1 := newBlueGreenRollout("guestbook", 1, nil, "active", "")
 	r1Resources := `
 limits:
-  cpu: 150m
+  cpu: 2000m
   memory: 8192M
 requests:
-  cpu: 2000m
+  cpu: 150m
```
dthomson25, line 1071: Why do you change the limits and requests here and below?

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+	ScaleDownLimitLargerThanRevisionLimit = "This rollout's revision history limit can not be smaller than the rollout's scale down limit"
+	// AvailableReason the reason to indicate that the rollout is serving traffic from the active service
+	AvailableReason = "AvailableReason"
+	// NotAvailableMessage the message to indicate that the Rollout does not have min availability
+	NotAvailableMessage = "Rollout does not have minimum availability"
+	// AvailableMessage the message to indicate that the Rollout does have min availability
+	AvailableMessage = "Rollout has minimum availability"
+)
+
+// called in strategy.go -> create.go
```
dthomson25, line 58: Can you elaborate on this comment more?

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, ValidateRolloutSpec(rollout, field.NewPath("spec"))...)
+	return allErrs
+}
+
+// ValidateRolloutSpec checks for a valid spec otherwise returns a list of erros.
+func ValidateRolloutSpec(rollout *v1alpha1.Rollout, fldPath *field.Path) field.ErrorList {
+	spec := rollout.Spec
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(*spec.Replicas), fldPath.Child("replicas"))...)
```
dthomson25, line 69: Always assuming that the spec.Replicas is set will always be true for the controller, but a linter or a webhook that may not be the case as they look at the rollout before passing it in.

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(spec.Selector)
+	if err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("selector"), spec.Selector, "invalid label selector"))
+	} else {
+		data, _ := json.Marshal(&spec.Template)
+		var template core.PodTemplateSpec
+		_ = json.Unmarshal(data, &template)
```
dthomson25, line 87: We should get the err and add it to the error list if it fails instead of silently failing.

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+		allErrs = append(allErrs, apivalidation.ValidateNonnegativeField(int64(*spec.ProgressDeadlineSeconds), fldPath.Child("progressDeadlineSeconds"))...)
+		if *spec.ProgressDeadlineSeconds <= spec.MinReadySeconds {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("progressDeadlineSeconds"), spec.ProgressDeadlineSeconds, "must be greater than minReadySeconds"))
+		}
+	}
+
+	// TODO: Same as MinReadySeconds check above?
+	//if rollout.Spec.MinReadySeconds > defaults.GetProgressDeadlineSecondsOrDefault(rollout) {
+	//	return newInvalidSpecRolloutCondition(prevCond, InvalidSpecReason, RolloutMinReadyLongerThanDeadlineMessage)
+	//}
```
dthomson25, line 107: Yes, it's the same check but we need it here otherwise the check won't be ready in the case where progressingDeadlineSeconds is not nil and the rollout.Spec.MinReadySeconds is greater than the default.

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+
+	//allErrs = append(allErrs, ValidateRolloutStrategy(spec.Strategy, fldPath.Child("strategy"))...)
```
dthomson25, line 109: Style Nix: please remove comment

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+
+func ValidateRolloutStrategy(rollout *v1alpha1.Rollout, fldPath *field.Path) field.ErrorList {
+	//func ValidateRolloutStrategy(strategy v1alpha1.RolloutStrategy, fldPath *field.Path) field.ErrorList {
+	strategy := rollout.Spec.Strategy
+	allErrs := field.ErrorList{}
+	if strategy.BlueGreen == nil && strategy.Canary == nil {
+
+	}
+	if strategy.BlueGreen != nil && strategy.Canary != nil {
+
```
dthomson25, line 123: Please add the return errors statement here.

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,268 @@
...
+	//	return newInvalidSpecRolloutCondition(prevCond, InvalidSpecReason, InvalidStrategyMessage)
+	//}
+	//
+	//if rollout.Spec.Strategy.BlueGreen != nil {
+	//	allErrs = append(allErrs, ValidateRolloutStrategyBlueGreen(rollout)...)
+	//}
+	//
+	//if rollout.Spec.Strategy.Canary != nil {
+	//	allErrs = append(allErrs, ValidateRolloutStrategyCanary(rollout)...)
+	//}
```
dthomson25, line 149: Style Nix: Please remove unused commented code

utils/conditions/conditions.go
```diff
@@ -23,32 +18,12 @@ import (
...
 
 const (
-	// Verify Spec constants
-
 	// InvalidSpecReason indicates that the spec is invalid
 	InvalidSpecReason = "InvalidSpec"
 	// MissingFieldMessage the message to indicate rollout is missing a field
 	MissingFieldMessage = "Rollout has missing field '%s'"
 	// RolloutSelectAllMessage the message to indicate that the rollout has an empty selector
 	RolloutSelectAllMessage = "This rollout is selecting all pods. A non-empty selector is required."
```
dthomson25, line 26: I'm wondering if we still need this consts since we move these checks to validation

rollout/controller_test.go
khhirani: The CPU request must be less than the limit in order to pass the Rollout Spec Validation.

pkg/apis/rollouts/validation/validation.go
khhirani: These are just my personal notes about how ValidateDeployment is used in the Kubernetes library.

pkg/apis/rollouts/validation/validation.go
```diff
@@ -0,0 +1,237 @@
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("selector"), spec.Selector, "invalid label selector"))
+	} else {
+		data, structConvertErr := json.Marshal(&spec.Template)
+		if structConvertErr != nil {
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("template"), structConvertErr))
+		}
+		var template core.PodTemplateSpec
+		structConvertErr = json.Unmarshal(data, &template)
+		if structConvertErr != nil {
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("template"), structConvertErr))
```
dthomson25, line 92: If the Marshal returns an error, then the Unmarshal and ValidatePodTemplateSpecForReplicaSet will almost certainly return errors since the marshal failed. I think we should stop validating the ValidatePodTemplateSpecForReplicaSet at that point since the errors from ValidatePodTemplateSpecForReplicaSet would be misleading (i.e. saying an image is required but the yaml does have an image). This makes the error more straightforward to debug.

rollout/controller_test.go
```diff
@@ -1072,10 +1077,10 @@ requests:
 	r2 := newBlueGreenRollout("guestbook", 1, nil, "active", "")
 	r2Resources := `
   limits:
-    cpu: 0.15
+    cpu: 2
```
dthomson25, line 1080: can you make 2 a string to keep it as similar as the previous test as possible?

dthomson25: same with 0.15

pkg/apis/rollouts/validation/validation_test.go
```diff
@@ -0,0 +1,190 @@
...
+	ro.Spec.Strategy.Canary.Steps[0].SetWeight = &setWeight
+	allErrs = ValidateRolloutStrategyCanary(ro, field.NewPath(""))
+	assert.Equal(t, InvalidSetWeightMessage, allErrs[0].Detail)
+
+	ro.Spec.Strategy.Canary.Steps[0].Pause = &v1alpha1.RolloutPause{
+		Duration: &pauseDuration,
+	}
+	ro.Spec.Strategy.Canary.Steps[0].SetWeight = nil
+	allErrs = ValidateRolloutStrategyCanary(ro, field.NewPath(""))
+	assert.Equal(t, InvalidDurationMessage, allErrs[0].Detail)
```
dthomson25, line 132: I think the readability of this test can be improved by using t.Run to describe what check you are running and for each t.Run make a copy of the original rollout and modify the copied rollout. This helps isolate the tests because we can easily see what changes are required to the rollout for the test. It's definitely more verbose, but it helps future readers.

dthomson25: For example, this block (line 127 to 132) can be rewritten as the following:
```golang
t.Run("invalid duration set in paused step", func(t *testing.T) {
	pauseDuration := intstr.FromInt(-1)
        invalidRo := ro.DeepCopy()
  	invalidRo.Spec.Strategy.Canary.Steps[0].Pause = &v1alpha1.RolloutPause{
		Duration: &pauseDuration,
	}
	allErrs = ValidateRolloutStrategyCanary(invalidRo, field.NewPath(""))
	assert.Equal(t, InvalidDurationMessage, allErrs[0].Detail)
})
```

Issue #555, opened 2020-06-26
Rollout stuck after switching from bluegreen to Canary


dthomson25:
When a Rollout switches from a bluegreen to a canary strategy, the Rollout should skip all the steps of the canary and set the currentStepIndex to the number of canary steps. If a rollout is in a steady-state when the switch occurs, the rollout becomes stuck with a currentStepIndex of 0 and makes no progress.

PR #558, opened 2020-06-26
fix: Set the currentStepIndex to max after bg to canary

dthomson25:
closes https://github.com/argoproj/argo-rollouts/issues/555.

@jessesuen do you know why we defaulted checkStepHashChange to return false if the currentStepHash is empty?

### DIFF
--- a/utils/replicaset/replicaset.go
+++ b/utils/replicaset/replicaset.go
@@ -407,10 +407,10 @@ func MaxSurge(rollout *v1alpha1.Rollout) int32 {
 }
 
 // checkStepHashChange indicates if the rollout's step for the strategy have changed. This causes the rollout to reset the
-// currentStepIndex to zero. If there is no previous pod spec to compare to the function defaults to false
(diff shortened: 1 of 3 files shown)


utils/replicaset/replicaset.go
```diff
@@ -407,10 +407,10 @@ func MaxSurge(rollout *v1alpha1.Rollout) int32 {
 }
 
 // checkStepHashChange indicates if the rollout's step for the strategy have changed. This causes the rollout to reset the
-// currentStepIndex to zero. If there is no previous pod spec to compare to the function defaults to false
+// currentStepIndex to zero. If there is no previous pod spec to compare to the function defaults to true
```
jessesuen, line 410: Replace: If there is no previous pod spec to compare to

With: If there was no previously recorded step hash

dthomson25:
The controller nulled all the canary status field when the user switches the Rollout from canary back to bluegreen

Issue #373, opened 2020-01-22
Global AnalysisTemplates (cross namespace references)


jessesuen:
Currently AnalysisTemplates are namespaced scoped objects, which limits their reusability within a cluster. 

We always envisioned the ability/use case where AnalysisTemplates could be defined someplace centrally in the cluster, and referenced by Rollouts & Experiments in different namespaces. This enables operators to install, and come up with a set of AnalysisTemplate definitions which are common to all users in their cluster, so that app developers would not need to understand how to write templates, or know the details of thing like prometheus hostnames, metric names, etc...

It would also help enable our original vision where AnalysisTemplates are shareable entities even across different companies and communities -- where a library of common patterns and templates could be contributed by different communities (e.g. AnalysisTemplates specific to the prometheus metrics produced by istio, linkerd, nginx-ingress, etc...).

To support centralized AnalysisTemplates, this proposal is to introduce an optional `namespace:` field to the analysis reference, like so:

```yaml
  strategy:
    canary:
      analysis:
        templateName: success-rate
        namespace: shared-rollouts-templates
```

For security purposes, the referenced AnalysisTemplates would need to "opt-in" on being able to be referenced globally in this manner. 

```yaml
kind: AnalysisTemplate
apiVersion: argoproj.io/v1alpha1
metadata:
  name: success-rate
spec:
  global: true  // indicates that this template can be referenced across namespaces
  metrics:
  - name: success-rate
```

One open issue is how to handle the case where an AnalysisTemplate references a secret as part of https://github.com/argoproj/argo-rollouts/issues/372.

jdamour98:
+1

mclarke47:
I'm going to pick this up!

dthomson25:
This change is going to consist of three steps: Updating the spec, passing the AnalysisTemplate namespace from the Rollout to the Experimented created by the Rollout, and grabbing the global AnalysisTemplate when creating an AnalysisRun

### Updating the Spec:
1. Modifying the [AnalysisTemplateSpec](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/v1alpha1/analysis_types.go#L30) in AnalysisTemplate object to have a global field in the analysis.go.
[...]

PR #560, opened 2020-06-27
feat: ClusterAnalysisTemplates (Cluster scoped AnalysisTemplates)

mclarke47:
Add CRD ClusterAnalysisTemplate.

It works like an AnalysisTemplate but is cluster wide. 
The resulting job runs in the namespace of the Rollout.

Local testing on minikube:
Screen Shot 2020-06-29 at 14 02 54

Fixes https://github.com/argoproj/argo-rollouts/issues/373


rollout/analysis.go
```diff
@@ -343,16 +343,33 @@ func (c *Controller) newAnalysisRunFromRollout(roCtx rolloutContext, rolloutAnal
...
+					return nil, err
+				}
+				templates = append(templates, template)
 			}
-			templates = append(templates, template)
+
 		}
+		// TODO should we create a separate run?
+		//      or flatten the cluster templates into the namespaced templates?
+		//      or disallow mixing and matching namespaced and cluster templates (I'd prefer not to)
```
mclarke47, line 372: Some questions

mclarke47: Figured it out

manifests/crds/analysis-run-crd.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.5
+    controller-gen.kubebuilder.io/version: (devel)
```
davidxia, line 5: Does this need to be a specific version or is `(devel)` replaced before a release?

pkg/apis/rollouts/v1alpha1/experiment_types.go
```diff
@@ -186,11 +186,16 @@ type ExperimentList struct {
...
 }
 
+// One of TemplateName or ClusterTemplateName must be specified
 type ExperimentAnalysisTemplateRef struct {
 	// Name is the name of the analysis
 	Name string `json:"name"`
 	// TemplateName reference of the AnalysisTemplate name used by the Experiment to create the run
-	TemplateName string `json:"templateName"`
+	// +optional
+	TemplateName string `json:"templateName,omitempty"`
```
davidxia, line 195: Is omitempty on both this and ClusterTemplateName OK for now or do we need validation that exactly one exists?

mclarke47: I think so but I'm not sure where to add this.

manifests/crds/analysis-run-crd.yaml
mclarke47: I this so but I'm not sure how to

mclarke47:
I think the CI is broken on master, which is causing the build failure

pkg/apis/rollouts/v1alpha1/experiment_types.go
jessesuen: In workflows, we took a different approach and introduced a `clusterScope: true` boolean.

```yaml
  strategy:
    canary:
      analysis:
        templateName: success-rate
        clusterScope: true
```

In this way, templateName was never optional. I'm thinking we should make templateName mandatory with a separate boolean to indicate its a ClusterAnalysisTemplate.

mclarke47: Ok, I will make the changes

mclarke47: I have made the changes

docs/features/analysis.md
```diff
@@ -170,6 +171,59 @@ Multiple measurements can be performed over a longer duration period, by specify
         query: ...
 ```
 
+## Cluster analysis templates
```
dthomson25, line 174: Can we add a note saying not available until v0.9 is released?

examples/rollout-canary-with-analysis.yaml
```diff
@@ -0,0 +1,45 @@
+# This example demonstrates a Rollout using the canary update strategy with a customized rollout
```
dthomson25, line 1: I think we can modify the existing rollout-analysis-step.yaml example to use both AnalysisTemplates and ClusterAnalysisTemplates instead of adding another example

pkg/apis/rollouts/v1alpha1/experiment_types.go
dthomson25: Awesome! if you remove the omitempty from the TemplateName json tag, it should be good to go!

dthomson25:
I filed a fix for the CI here: https://github.com/argoproj/argo-rollouts/pull/579. Once it is merged in, you can rebase and that should fix CI

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -338,9 +341,14 @@ type RolloutAnalysisBackground struct {
...
 type RolloutAnalysis struct {
+	// Whether to look for the templateName at cluster scope or namespace scope
+	// Deprecated and will be removed in v0.9
+	// +optional
+	ClusterScope bool `json:"clusterScope,omitempty"`
 	// TemplateName reference of the AnalysisTemplate name used by the Rollout to create the run
 	// Deprecated and will be removed in v0.9
-	TemplateName string `json:"templateName,omitempty"`
+	// +optional
+	TemplateName string `json:"templateName"`
```
dthomson25, line 351: This TemplateName needs the omitempty flag because the rollout is letting the user pick between using the TemplateName and Templates (with the goal being eliminating TemplateName in the future)

mclarke47:
@dthomson25 Done! Let me know if there is anything else to fix :)

mclarke47:
Nope! Thanks for reviewing!

PR #578, opened 2020-07-08
feat: Uplevel AnalysisRun status to Rollout status

### LINKED ISSUE #559: Uplevel Status Information about Analysis to Rollout
In order to make it easier to build automation on top of the Rollout, the controller should pass some metadata from the running AnalysisRun to the Rollout status.  Currently, the Rollout only shows that the Rollout is aborted with no reasons on why it failed. A user or automation would need to check the AnalysisRun to glean information on what went wrong. For CD tools that manage a Rollout, building automation can get tricky because they sometimes only have the Rollout object to evaluate or it's expensive to get the AnalysisRun

To simplify the experience, the Rollout should carry some duplicate information on the AnalysisRun that it created. It can track information on the status of the AnalysisRun. Here are my proposed fields to add to the Rollout object:

```golang
type RolloutAnalysisRunStatus {
    name string
    status AnalysisRunStatus // Successful, Processing, Error, etc
    message string // Grabbed from the .status.message of the AnalysisRun
}
```
The Rollout should have this struct for the BlueGreenStatus for pre and post analysis and Canary strategy for step analysis and background analysis. That allows us to track this information for all the different analysis occurring.

Additionally, The controller can omit events on the Rollout indicating status changes of the AnalysisRun. If an AnalysisRun transitions from Running to Error, we can add events to help highlight that change in the Rollout.

dthomson25:
Closes #559


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -465,6 +465,11 @@ type RolloutStatus struct {
 	PauseConditions []PauseCondition `json:"pauseConditions,omitempty"`
 	//ControllerPause indicates the controller has paused the rollout
 	ControllerPause bool `json:"controllerPause,omitempty"`
+	// ReconciledAbort indicates the controller reconciled an aborted rollout. The controller uses this to understand if
+	// the controller needs to do some specific work when a Rollout is aborted. For example, the reconcileAbort is used
+	// to indicate if the Rollout should enter an aborted state when the latest AnalysisRun is a failure, or the controller
+	// has already put the Rollout into an aborted and should create a new AnalysisRun.
+	ReconciledAbort bool `json:"reconciledAbort,omitempty"`
```
jessesuen, line 472: Instead of `ReconciledAbort bool`, can we have `AbortedAt *metav1.Time` which gets cleared out whenever Status.Abort = true

PR #580, opened 2020-07-09
fix: Populate .spec.template with default values before Rollout Validation

### LINKED ISSUE #576: Examples don't work with master controller
I think the validation is being too strict on fields that aren't present on the pod spec:

For example `terminationMessagePolicy` has a default value of File according to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core

Screen Shot 2020-07-08 at 11 17 13
Screen Shot 2020-07-08 at 11 15 14

Replication steps:

Run master controller against minikube, apply a manifest from the examples directory.

khhirani:
Closes #576


pkg/apis/rollouts/validation/validation.go
```diff
@@ -77,7 +79,15 @@ func ValidateRolloutSpec(rollout *v1alpha1.Rollout, fldPath *field.Path) field.E
 	if err != nil {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("selector"), spec.Selector, "invalid label selector"))
 	} else {
-		data, structConvertErr := json.Marshal(&spec.Template)
+		// Set default values in PodTemplate to address issue https://github.com/argoproj/argo-rollouts/issues/576
```
dthomson25, line 82: Maybe add something like:

The upstream K8s validation we are using expects the default values of a PodSpec to be set otherwise throwing a validation error. However, the Rollout does not need to have them set since the ReplicaSet it creates will have the default values set. As a result, the controller sets the default values before validation to prevent the validation errors due to the lack of these default fields. See https://github.com/argoproj/argo-rollouts/issues/576 for more info.

Issue #568, opened 2020-07-05
The traffic ActualWeight's calculation is wrong


henrywangx:
The actualWeight should be 5% for the[ official example](https://argoproj.github.io/argo-rollouts/getting-started/nginx/). But it's 50%.
I think the actualWeight's [calculation](https://github.com/argoproj/argo-rollouts/blob/master/pkg/kubectl-argo-rollouts/info/rollout_info.go#L73) method is not correct.
It should use the actual ingress/smi's weight to calculate the actual weight.

henrywangx:
sorry, I found the **actualWeight** indicates the pod's weight not traffic's weight. Yes it misleads my mind.

jessesuen:
I think this is a valid bug. The output from the kubectl plugin was written before mesh and ingress traffic routing support was implemented, and it was only displaying pod weights. The plugin output should be updated to understand when traffic routing is used, and have `actualWeight` be equal to the `setWeight`.

jessesuen:
Fixed by https://github.com/argoproj/argo-rollouts/pull/584

PR #584, opened 2020-07-12
fix: Fix ActualWeight for canary

shuheiktgw:
https://github.com/argoproj/argo-rollouts/issues/568

As described in the issue, the `ActualWeight` for canary was wrong, so I've fixed it to show the right weight.

### DIFF
--- a/pkg/kubectl-argo-rollouts/info/rollout_info.go
+++ b/pkg/kubectl-argo-rollouts/info/rollout_info.go
@@ -65,13 +65,18 @@ func NewRolloutInfo(
 
 		roInfo.ActualWeight = "0"
 		currentStep, _ := replicasetutil.GetCurrentCanaryStep(ro)
+
 		if currentStep == nil {
 			roInfo.ActualWeight = "100"
 		} else if ro.Status.AvailableReplicas > 0 {
-			for _, rs := range roInfo.ReplicaSets {
-				if rs.Canary {
-					roInfo.ActualWeight = fmt.Sprintf("%d", (rs.Available*100)/ro.Status.AvailableReplicas)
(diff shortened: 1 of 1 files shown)


pkg/kubectl-argo-rollouts/info/rollout_info.go
```diff
@@ -69,8 +69,8 @@ func NewRolloutInfo(
 			roInfo.ActualWeight = "100"
 		} else if ro.Status.AvailableReplicas > 0 {
 			for _, rs := range roInfo.ReplicaSets {
-				if rs.Canary {
-					roInfo.ActualWeight = fmt.Sprintf("%d", (rs.Available*100)/ro.Status.AvailableReplicas)
+				if rs.Canary && rs.Status == "Healthy" {
```
dthomson25, line 72: I think we are going to need to have 2 cases here: one for traffic management and one without. 

The old code covers the non-traffic management case, where the traffic is split evenly amount all the available replicas. Since no tools are used to split the traffic split, the actual percentage of traffic going to the new version is the number of available replicas in the new ReplicSet over the total number of replicas in all the ReplicaSets of that Rollout.

In the traffic management case, we should only use SetWeight info. The traffic shaping tools will ensure the percentage instead of relying on replicas count. You can detect the trafficManagement by checking that `.spec.strategy.canary.trafficManagement != nil`.

shuheiktgw: @dthomson25 Thank you for the review! Let me fix it quickly 👍

shuheiktgw: @dthomson25 I've fixed it, so would you mind reviewing the change?

shuheiktgw:
Thank you for your review!

PR #585, opened 2020-07-13
test: BDD Test for e2e testing

khhirani:
Create a BDD test for Blue-Green deployments for e2e testing.


test/bdd/bluegreen_test.go
```diff
@@ -0,0 +1,193 @@
...
+	numAttempts := 4
+	isSuccess, err := retry(numAttempts, func() (bool, error) {
+		rolloutExists, err := argoexec.RunCommand("kubectl", argoexec.CmdOpts{}, "get", "rollout", c.RolloutName)
+		return rolloutExists != "", err
+	})
+	if err != nil {
+		return err
+	}
+	if !isSuccess {
+		return fmt.Errorf("Rollout not created")
```
dthomson25, line 44: When this error (or any other error) returns, is it obvious which test is failing? If not, then this error won't make it easier for the developer to troubleshoot the issue.

dthomson25:
Also, it looks like you have more listing and merge issues. Please fix both of those before merging

khhirani:
> Also, it looks like you have more listing and merge issues. Please fix both of those before merging

`make lint` is not compatible with godog formatting.

PR #600, opened 2020-07-31
feat: Controller Validation for objects referenced by Rollout

khhirani:
WIP


pkg/apis/rollouts/validation/validation_references.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/api/extensions/v1beta1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting/istio"
+)
+// Controller will validate references in reconciliation
+
+type ReferencedResources struct {
+	AnalysisTemplates []v1alpha1.AnalysisTemplate
```
jessesuen, line 14: It doesn't look like we support ClusterAnalysisTemplate

rollout/controller.go
```diff
@@ -346,6 +347,81 @@ func (c *Controller) syncHandler(key string) error {
...
+	referencedResources := validation.ReferencedResources{}
+
+	if r.Spec.Strategy.BlueGreen != nil {
+		blueGreen := r.Spec.Strategy.BlueGreen
+		_, _, err := c.getPreviewAndActiveServices(r)
+		if err != nil {
+			return err
+		}
+		for _, template := range blueGreen.PrePromotionAnalysis.Templates {
+			analysisTemplate, err := c.argoprojclientset.ArgoprojV1alpha1().AnalysisTemplates(r.Namespace).Get(template.TemplateName, metav1.GetOptions{})
```
jessesuen, line 359: I think we should use an analysistemplate lister instead of the API client. The former does not incur a API cost

pkg/apis/rollouts/validation/validation_references_test.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+var svc = ServiceWithType{
+	Service: &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "stable-service-name",
+		},
+	},
+	Type: StableService,
+}
```
dthomson25, line 139: I think these should be helper functions instead of vars. Vars can be modified by tests and potentially have side effects on other tests, while helper functions can introduce instances that are scoped to that test

pkg/apis/rollouts/validation/validation_references_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+func strToUnstructured(yamlStr string) *unstructured.Unstructured {
```
dthomson25, line 141: You should be able to use https://github.com/argoproj/argo-rollouts/blob/master/utils/unstructured/unstructured.go#L8 instead of reimplementing this method.

pkg/apis/rollouts/validation/validation_references_test.go
```diff
@@ -0,0 +1,271 @@
...
+func TestValidateAnalysisTemplateWithType(t *testing.T) {
+	// Success - specify count so that metric runs deterministically
+	allErrs := ValidateAnalysisTemplateWithType(analysisTemplateWithType)
+	assert.Empty(t, allErrs)
+
+	// Fail case - AnalysisTemplate runs indefinitely
+	analysisTemplate.Spec.Metrics[0].Count = 0
+	allErrs = ValidateAnalysisTemplateWithType(analysisTemplateWithType)
+	assert.Len(t, allErrs, 1)
+	assert.Equal(t, "spec.strategy.canary.steps[0].analysis.templates[0].templateName: Invalid value: \"analysis-template-name\": AnalysisTemplate analysis-template-name has metric metric-name which runs indefinitely", allErrs[0].Error())
```
dthomson25, line 171: This may be easier to read if you create an field.Error struct with all the fields filled out and then do an err.Error() to get the string.

pkg/apis/rollouts/validation/validation_references_test.go
```diff
@@ -0,0 +1,271 @@
...
+	allErrs := ValidateRolloutReferencedResources(ro, refResources)
+	assert.Empty(t, allErrs)
+}
+
+func TestValidateAnalysisTemplateWithType(t *testing.T) {
+	// Success - specify count so that metric runs deterministically
+	allErrs := ValidateAnalysisTemplateWithType(analysisTemplateWithType)
+	assert.Empty(t, allErrs)
+
+	// Fail case - AnalysisTemplate runs indefinitely
```
dthomson25, line 167: Since this is defining a "new" test, I recommend wrapping it in a t.Run("Fail case - AnalysisTemplate runs indefinitely") to make it clear it's a new test case

rollout/controller.go
```diff
@@ -376,6 +366,272 @@ func (c *Controller) syncHandler(key string) error {
...
+		if blueGreen.PostPromotionAnalysis != nil {
+			// CanaryStepIndex will be ignored
+			templates, err := c.getReferencedAnalysisTemplates(rollout, blueGreen.PostPromotionAnalysis, validation.PostPromotionAnalysis, 0)
+			if err != nil {
+				return nil, err
+			}
+			analysisTemplates = append(analysisTemplates, templates...)
+		}
+	} else if rollout.Spec.Strategy.Canary != nil {
+		canary := rollout.Spec.Strategy.Canary
```
dthomson25, line 526: It's probably a good idea to add a comment about why we don't need to do anything with the background analysis

rollout/controller_test.go
```diff
@@ -335,12 +337,10 @@ func newReplicaSet(r *v1alpha1.Rollout, replicas int) *appsv1.ReplicaSet {
 func calculatePatch(ro *v1alpha1.Rollout, patch string) string {
 	origBytes, err := json.Marshal(ro)
 	if err != nil {
-		fmt.Println(patch)
 		panic(err)
 	}
 	newBytes, err := strategicpatch.StrategicMergePatch(origBytes, []byte(patch), v1alpha1.Rollout{})
 	if err != nil {
-		fmt.Println(patch)
```
dthomson25, line 343: We should keep the patch println since that helps users debug broken tests

rollout/controller_test.go
```diff
@@ -1292,3 +1292,125 @@ func TestSwitchBlueGreenToCanary(t *testing.T) {
...
+			Ingress: "alb-ingress-name",
+		},
+	}
+	r.Namespace = metav1.NamespaceDefault
+	defer f.Close()
+
+	// Fail case - cannot find ALB Ingress
+	c, _, _ := f.newController(noResyncPeriodFunc)
+	_, err := c.getReferencedIngresses(r)
+	assert.Equal(t, "spec.strategy.canary.trafficRouting.alb.ingress: Invalid value: \"alb-ingress-name\": ingress.extensions \"alb-ingress-name\" not found", err.Error())
```
dthomson25, line 1353: Creating a field.Error and then using err.Error() may help with readability

rollout/service_test.go
```diff
@@ -91,16 +91,9 @@ func TestGetPreviewAndActiveServices(t *testing.T) {
 		noActiveSvcRollout.Spec.Strategy.BlueGreen.ActiveService = ""
 		_, _, err := c.getPreviewAndActiveServices(noActiveSvcRollout)
 		assert.NotNil(t, err)
-		assert.EqualError(t, err, "Invalid Spec: Rollout missing field ActiveService")
+		//assert.EqualError(t, err, "Invalid Spec: Rollout missing field ActiveService")
```
dthomson25, line 94: Please remove the comment

PR #650, opened 2020-08-19
feat: Add option to specify the namespace of istio virtual service

sakajunquality:
Sometimes virtual services are not in the same namespace as workloads.
So I think I would be nice if namespace can be specified along with the name of virtual service.

### DIFF
--- a/manifests/crds/rollout-crd.yaml
+++ b/manifests/crds/rollout-crd.yaml
@@ -421,6 +421,8 @@ spec:
                               properties:
                                 name:
                                   type: string
+                                namespace:
+                                  type: string
                                 routes:
                                   items:
                                     type: string
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -11372,6 +11372,8 @@ spec:
                               properties:
                                 name:
                                   type: string
+                                namespace:
+                                  type: string
                                 routes:
                                   items:
                                     type: string
--- a/manifests/namespace-install.yaml
+++ b/manifests/namespace-install.yaml
@@ -11372,6 +11372,8 @@ spec:
                               properties:
                                 name:
                                   type: string
+                                namespace:
+                                  type: string
                                 routes:
                                   items:
                                     type: string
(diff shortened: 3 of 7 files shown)


jessesuen:
I didn't realize VirtualServices could reference services across namespaces. But in order for this VirtualService to work, does it then need to reference the stable-svc and canary-svc in the namespace of the Rollout? e.g.:


```yaml
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: rollout-vsvc
spec:
  gateways:
    - istio-rollout-gateway
  hosts:
    - istio-rollout.dev.argoproj.io
  http:
    - name: primary
      route:
        - destination:
            host: stable-svc  # <<<< how does istio reference this service when it exists in a different namespace?
          weight: 100
        - destination:
            host: canary-svc
          weight: 0
```

sakajunquality:
@jessesuen Hi,

As documented [here](https://istio.io/latest/docs/reference/config/networking/virtual-service/#Destination), fully qualified domain names, which includes namespaces, can be specified.
e.g. `reviews.default.svc.cluster.local` instead of `reviews` 
That way you can route traffic to other namesapces.



For example

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: reviews-route
  namespace: foo
spec:
  hosts:
  - review.abc.com
  http:
  - match:
    - uri:
        prefix: "/wpcatalog"
    - uri:
        prefix: "/consumercatalog"
    rewrite:
      uri: "/newcatalog"
    route:
    - destination:
        host: reviews # interpreted as reviews.foo.svc.cluster.local
        subset: v2
  - route:
    - destination:
        host: reviews # interpreted as reviews.foo.svc.cluster.local
        subset: v1
```

if reviews are in bar namespace

```
```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: reviews-route
  namespace: foo
spec:
  hosts:
  - review.abc.com
  http:
  - match:
    - uri:
        prefix: "/wpcatalog"
    - uri:
        prefix: "/consumercatalog"
    rewrite:
      uri: "/newcatalog"
    route:
    - destination:
        host: reviews # interpreted as reviews.bar.svc.cluster.local
        subset: v2
  - route:
    - destination:
        host: reviews # interpreted as reviews.bar.svc.cluster.local
        subset: v1
```

sakajunquality:
Thank you for your review. I though finding virtual services is enough at first. But looks like finding route is also necessary.
I will fix and test again.

jessesuen (changes requested):
Per feedback, this need to work with selecting the FQDN of the service name, and validation should be updated to support this.

sakajunquality:
rebased to fix conflict

sakajunquality:
I will reopn this when it becomes ready.

PR #656, opened 2020-08-22
perf: Create IstioVirtualServiceLister

khhirani:
Create an Istio Virtual Service lister so that we can avoid making unnecessary API calls to get Istio Virtual Services.

### DIFF
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -144,7 +144,7 @@ func newCommand() *cobra.Command {
 			jobInformerFactory.Start(stopCh)
 
 			// Check if Istio installed on cluster before starting dynamicInformerFactory
-			if istioutil.DoesIstioExist(dynamicClient, istioVersion) {
+			if istioutil.DoesIstioExist(dynamicClient, namespace, istioVersion) {
 				dynamicInformerFactory.Start(stopCh)
 			}
 
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -7,14 +7,12 @@ import (
 	"strconv"
 	"time"
 
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/dynamic/dynamicinformer"
-
 	smiclientset "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/split/clientset/versioned"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -29,6 +27,7 @@ import (
 	informers "github.com/argoproj/argo-rollouts/pkg/client/informers/externalversions"
 	"github.com/argoproj/argo-rollouts/pkg/signals"
 	controllerutil "github.com/argoproj/argo-rollouts/utils/controller"
+	istioutil "github.com/argoproj/argo-rollouts/utils/istio"
 	kubeclientmetrics "github.com/argoproj/argo-rollouts/utils/kubeclientmetrics"
 )
 
@@ -111,7 +110,7 @@ func newCommand() *cobra.Command {
 				kubeinformers.WithTweakListOptions(func(options *metav1.ListOptions) {
 					options.LabelSelector = jobprovider.AnalysisRunUIDLabelKey
 				}))
-			gvk := schema.ParseGroupResource("virtualservices.networking.istio.io").WithVersion(defaultIstioVersion)
+			istioGVR := istioutil.GetIstioGVR(istioVersion)
 			dynamicInformerFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 0, namespace, nil)
 			cm := controller.NewManager(
 				namespace,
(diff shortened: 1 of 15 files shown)


cmd/rollouts-controller/main.go
```diff
@@ -108,6 +111,8 @@ func newCommand() *cobra.Command {
 				kubeinformers.WithTweakListOptions(func(options *metav1.ListOptions) {
 					options.LabelSelector = jobprovider.AnalysisRunUIDLabelKey
 				}))
+			gvk := schema.ParseGroupResource("virtualservices.networking.istio.io").WithVersion(defaultIstioVersion)
+			dynamicInformerFactory := dynamicinformer.NewDynamicSharedInformerFactory(dynamicClient, 0)
```
jessesuen, line 115: dynamicinformer.NewFilteredDynamicSharedInformerFactory(dynamicClient, 0, namespace, nil)

rollout/controller.go
```diff
@@ -251,9 +259,32 @@ func NewController(cfg ControllerConfig) *Controller {
...
+			controller.processNextEnqueuedObj(obj)
+		},
+	})
+
 	return controller
 }
 
+func (c *Controller) DoesIstioExist() bool {
+	gvk := schema.ParseGroupResource("virtualservices.networking.istio.io").WithVersion(c.defaultIstioVersion)
+	_, err := c.dynamicclientset.Resource(gvk).List(metav1.ListOptions{})
```
jessesuen, line 281: This list should respect the `--namespace` flag to the controller. So I think this should be:

```golang
_, err := c.dynamicclientset.Resource(gvk).Namespace(c.namespace).List(metav1.ListOptions{})

```

cmd/rollouts-controller/main.go
```diff
@@ -132,13 +138,18 @@ func newCommand() *cobra.Command {
...
 				albIngressClasses)
-
 			// notice that there is no need to run Start methods in a separate goroutine. (i.e. go kubeInformerFactory.Start(stopCh)
 			// Start method is non-blocking and runs all registered informers in a dedicated goroutine.
 			kubeInformerFactory.Start(stopCh)
 			argoRolloutsInformerFactory.Start(stopCh)
 			jobInformerFactory.Start(stopCh)
 
+			// Check if Istio installed on cluster before starting dynamicInformerFactory
+			_, err = dynamicClient.Resource(gvk).List(metav1.ListOptions{})
```
jessesuen, line 148: Since this code is repeated, I think we should have a DoesIstioExist(client) standalone helper utility function that takes a dynamic client.

Also, to limit the number of objects return, we should say `metav1.ListOptions{Limit:1}`. Otherwise we are listing all VirtualServices in the cluster, which could be expensive, simply for the purposes of checking if istio exists.

rollout/controller.go
```diff
@@ -148,6 +151,8 @@ func NewController(cfg ControllerConfig) *Controller {
 		},
 	}
 
+	gvk := schema.ParseGroupResource("virtualservices.networking.istio.io").WithVersion(cfg.DefaultIstioVersion)
```
jessesuen, line 154: 1. The gvk variable is the wrong acronym since its a GroupVersionResource and not GroupVersionKind, so it would be gvr.

2. Consider constructing the GVR explicitly without parsing. i.e.

```golang
	gvr := schema.GroupVersionResource{
		Group:    "networking.istio.io",
		Version:  cfg.DefaultIstioVersion,
		Resource: "virtualservices",
	}
```

rollout/controller.go
```diff
@@ -267,15 +288,44 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
...
+	// If it is added, then run the istioVirtualServiceInformer so that we can use istioVirtualServiceLister
+	go c.runVirtualServiceInformer(stopCh)
 
 	<-stopCh
 	log.Info("Shutting down workers")
 
 	return nil
 }
 
+func (c *Controller) runVirtualServiceInformer(stopCh <-chan struct{}) {
```
jessesuen, line 301: It appears that this channel cannot be stopped. Since it's forever in a loop which does not honor the stopCh.

rollout/controller.go
```diff
@@ -267,15 +288,44 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
...
+		if !istioutil.DoesIstioExist(c.dynamicclientset, c.defaultIstioVersion) {
+			time.Sleep(10 * time.Minute)
+		} else {
+			c.istioVirtualServiceInformer.Run(stopCh)
+			cache.WaitForCacheSync(stopCh, c.istioVirtualServiceInformer.HasSynced)
+		}
+	}
+}
+
+func (c *Controller) processNextEnqueuedObj(obj interface{}) {
```
jessesuen, line 313: The method is too generically named. Controller.processNextEnqueuedObj is only dealing with istio objects, and should be named as such.

rollout/controller.go
```diff
@@ -267,15 +288,44 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
...
+	acc, err := meta.Accessor(obj)
+	if err != nil {
+		log.Errorf("Error processing object from watch: %v", err)
```
jessesuen, line 316: I would maybe include the object in the message in addition to the error.

utils/istio/istio.go
```diff
@@ -0,0 +1,23 @@
+package istio
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+func DoesIstioExist(dynamicClient dynamic.Interface, version string) bool {
+	_, err := dynamicClient.Resource(GetIstioGVR(version)).List(metav1.ListOptions{})
```
jessesuen, line 10: ListOptions should have count of 1

PR #677, opened 2020-08-26
feature(controller): Allow setting canary weight without side-effects. Fixes #556

### LINKED ISSUE #556: Scale canary stack even when weight is 0 (with traffic routing)
Consider the following canary steps **AND the rollout is configured with traffic routing**:

```yaml
steps:
- setWeight: 0
- pause: {}
- setWeight: 10
- pause: { duration: 1h }
- setWeight: 20
- pause: { duration: 1h }
- setWeight: 30
- pause: { duration: 1h }
...
```

When the rollout is at the initial pause (the second step), the canary stack is scaled to zero. As soon as it reaches the `- setWeight: 10` step, the canary stack is scaled to 10%.  The fact that the rollout is scaled to zero at the initial pause,  makes some use cases of rollouts not possible. For example, some users want perform some tests on the canary stack, **_before_** the stack receives any production traffic, but still do a gradual release. Something like this:

```yaml
steps:
- setWeight: 0
- analysis:
    templates:
    - templateName: preview-test
- setWeight: 10
- pause: { duration: 1h }
- setWeight: 20
- pause: { duration: 1h }
- setWeight: 30
- pause: { duration: 1h }
...
```

I think we should allow users to control scaling the canary stack even when weight is zero to enable this use case. Of course, if a rollout is aborted, we should always scale the canary back to 0 replicas.

EDIT: corrected description to indicate that we match the scale of the canary stack to match the current weight (and not 100% as previously mistated)

aweis89:
Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/556) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] I've signed the CLA.


utils/replicaset/canary.go
```diff
@@ -83,7 +83,15 @@ func DesiredReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS, stableRS *a
 // For more examples, check the TestCalculateReplicaCountsForCanary test in canary/canary_test.go
 func CalculateReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS *appsv1.ReplicaSet, stableRS *appsv1.ReplicaSet, oldRSs []*appsv1.ReplicaSet) (int32, int32) {
 	rolloutSpecReplica := defaults.GetReplicasOrDefault(rollout.Spec.Replicas)
-	setWeight := GetCurrentSetWeight(rollout)
+	setWeight := int32(0)
+	if scs, use := UseSetCanaryScale(rollout); use == true {
+		if scs.Replicas != nil {
+			return *scs.Replicas, rolloutSpecReplica
+		}
+		setWeight = *scs.Weight
```
jessesuen, line 91: I feel there is potential for nil dereference. I would suggest to a little more defensive and do something like:

```
		if scs.Replicas != nil {
			return *scs.Replicas, rolloutSpecReplica
		} else if scs.Weight != nil {
			setWeight = *scs.Weight
		} else {
			setWeight = GetCurrentSetWeight(rollout)
		} 
```

utils/replicaset/canary.go
```diff
@@ -279,6 +287,25 @@ func GetCurrentSetWeight(rollout *v1alpha1.Rollout) int32 {
 	return 0
 }
 
+func UseSetCanaryScale(rollout *v1alpha1.Rollout) (v1alpha1.SetCanaryScale, bool) {
```
jessesuen, line 290: 1. Style: instead of return value of `(v1alpha1.SetCanaryScale, bool)`, which means we would sometimes return an empty struct, we could instead return a pointer (*v1alpha1.SetCanaryScale), where nil obviously representing that it should be used.

2. This method should return false (or nil per my previous suggestion) if the rollout spec.strategy.canary.trafficRouting is nil. This is because the `setCanaryScale` should not be used with the regular replica-weighted canary, since that strategy can only use matchWeight.

manifests/crds/rollout-crd.yaml
```diff
@@ -15,13 +15,11 @@ spec:
     name: Current
     type: integer
   - JSONPath: .status.updatedReplicas
-    description: Total number of non-terminated pods targeted by this rollout that
-      have the desired template spec
+    description: Total number of non-terminated pods targeted by this rollout that have the desired template spec
```
aweis89, line 18: Seemingly `make manifests` removes the added line. Let me know if you'd like me to revert this nonetheless

aweis89:
Not ready for merge. After doing some testing, it seems like this PR causes the controller to think the desired state hasn't been reached when there's no setWeight or if setWeight is not being used for scaling.

aweis89:
> Not ready for merge. After doing some testing, it seems like this PR causes the controller to think the desired state hasn't been reached when there's no setWeight or if setWeight is not being used for scaling. This causes the state to go into Progressing until some deadline is reached after which the state is Degraded.

It seems like the issue here: https://github.com/argoproj/argo-rollouts/blob/master/pkg/kubectl-argo-rollouts/info/rollout_info.go#L125
Apparently `UpdatedReplicas` refers to the `newRS` but `Replicas` refers to `newRS` + `oldRS`.
I'm not sure I'm clear on why this issue is popping up for this PR specifically as it would seem like during a canary, the above condition should always be true, provided `setWeight > 0` 🤔

aweis89:
> Not ready for merge. After doing some testing, it seems like this PR causes the controller to think the desired state hasn't been reached when there's no setWeight or if setWeight is not being used for scaling. This causes the state to go into Progressing until some deadline is reached after which the state is Degraded. The last condition is:
> 
> ```
>   - lastTransitionTime: "2020-08-27T00:02:49Z"
>     lastUpdateTime: "2020-08-27T00:02:49Z"
>     message: ReplicaSet "game-play-service-6758864447" has timed out progressing.
>     reason: ProgressDeadlineExceeded
>     status: "False"
>     type: Progressing
> ```
> 
> Notably, the referenced replicaset is healthy and has reached the target replicas as specified by `setCanaryWeight`

This is now resolved. Tested out the `setCanaryScale.weight` feature, and seems to work like a charm (tested setting %200, %100, %50, %10). Also did a test using `setCanaryScale.replicas` features which workers as well. We intend to use these features pretty heavily, so they should be getting lots of testing in the upcoming weeks.

aweis89:
We found another issue today when using `setCanaryScale.replicas` (as the first step), and when deploying a new Rollout, it seems to get stuck at setting the replicas to that first step's count even though there's no canary due to being the first deploy. Interestingly the controller sets the `desired-replicas` annotation to the correct count as determined by Rollout's `spec.replicas`, it just won't set that on ReplicaSet's `spec.replicas` for some reason. There doesn't seem to be a similar issue with `setCanaryScale.weight` field

jessesuen:
Added an E2E test for this:

https://github.com/argoproj/argo-rollouts/blob/c3f409ad5029407ef1bbd5784ac3b9a27c19b41f/test/e2e/canary_test.go#L17-L72

aweis89:
Notably `matchTrafficWeight` currently takes effect immediately, so `setWeight` should be set in a prior step. If this is not desired, let me know and I will change as per: https://github.com/argoproj/argo-rollouts/issues/556#issuecomment-683212582

PR #680, opened 2020-08-26
fix: remove clusteranalysistemplates from argo-rollouts-role

davidxia:
This entry does not do anything and can be misleading.
I believe only ClusterRoles can grant access to cluster-scoped
resources. See https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole.

Checklist:

* [x] I've signed the CLA.

### DIFF
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -14101,7 +14101,6 @@ rules:
   - argoproj.io
   resources:
   - analysistemplates
-  - clusteranalysistemplates
   verbs:
   - get
   - list


jessesuen (changes requested):
I don't believe this change is correct. This privilege is allowing the rollouts controller the ability to list/get clusteranalysistemplates. Which it needs to do in order to spawn AnalysisRuns from the definition of ClusterAnalysisTemplate.

Also, BTW, install.yaml is an autogenerated file, so manifest changes need to be made in the kustomize bases, and then `make manifests` should be run to generate the install.yaml

davidxia:
Thanks for reviewing. The diff here lacks context since it's truncated.

I removed `clusteranalysistemplates` from the Role here 1 since I think only ClusterRoles can grant access to cluster-scoped resources. The ClusterRole already has `clusteranalysistemplates` 2 which should be enough, right?

1: https://github.com/argoproj/argo-rollouts/blob/f4955ef8ffaab2f6d1453881903cf45b46b44830/manifests/install.yaml#L14115
2:

PR #704, opened 2020-09-07
refactor: move reconciliation logic into rolloutContext

jessesuen:
The Rollout Controller struct is currently overloaded. This change does the following:

1. removes all reconciliation logic in the Controller struct, and moves it into the existing rolloutContext object.
2. the rolloutContext object is no longer passed around from method to method, since all reconciliation functions are now member functions of the rolloutContext and can access the rollout fields/objects directly.
3. when a rolloutContext is instantiated, it gathers all objects associated with the rollout from the listers at the point of instantiation, rather than during the reconciliation
4. eliminates canaryCtx and blueGreenCtx objects, since they can be superseded by rolloutContext

There's probably more simplifications that can be done, but was already a huge refactor.


khhirani (approved):
LGTM

PR #705, opened 2020-09-07
feat: Add datadog metric provider. Fixes #702

### LINKED ISSUE #702: Datadog metric provider
# Summary

In addition to WaveFront / Kayenta I'd like to suggest implementing a [Datadog ](https://www.datadoghq.com/) provider. 

# Use Cases

Due to limitations raised [here ](https://github.com/argoproj/argo-rollouts/issues/405) it's not possible to use the webmetric provider with Datadog. Additionally by providing an explicit provider makes configuring the analysis template simpler for the user.

---

stephen-harris:
Checklist:

* [x] I've signed the CLA.


jessesuen:
I think we need to treat DataDog API Keys and App Keys more first class in the metric provider spec. For reference, this is how we do something similar wavefront:
https://argoproj.github.io/argo-rollouts/features/analysis/#wavefront-metrics

I've read through https://docs.datadoghq.com/account_management/api-app-keys/, but I still have some questions on where would be an appropriate place to configure API keys and app keys.

With Wavefront, a single API token can be specified for an entire domain, e.g. `example1.wavefront.com`. For this reason, API tokens are configured at the rollout controller level, in a secret in the same namespace of argo-rollouts controller. The rollout end users do not need to specify API keys when using the wavefront metric provider, since this is configured at the controller.

Since I don't understand DataDog, can you answer some questions about it:

1. Does it make more sense to configure DataDog API keys and/or App keys at a controller level? With wavefront, it made sense to configure it at a controller level, since the same token could be used for the entire wavefront address. By having centrally configured keys, it relieved the burden from developers having to configure and duplicate keys in every one of their application's namespace. This would be a pain to manage, especially with a lot of namespaces. 

2. What is the scope of API keys and App keys? In a large organization, who (account administrators, developers, operators) typically generates API keys vs. App keys? How many of either of these keys typically exist? I see from their documentation, "Your org must have at least one API key and at most five API keys."  I'm not sure how large an "org" is in this context, but the statement leads me to believe that API keys are quite limited, and would lend itself to be centrally configured vs. configured at a namespace secret.

metricproviders/datadog/datadog.go
```diff
@@ -0,0 +1,147 @@
...
+		endpoint = metric.Provider.Datadog.URL + "/api/v1/query"
+	}
+
+	url, _ := url.Parse(endpoint)
+
+	now := unixNow()
+	q := url.Query()
+	q.Set("query", metric.Provider.Datadog.Query)
+	q.Set("from", strconv.FormatInt(now-300, 10))
+	q.Set("to", strconv.FormatInt(now, 10))
```
jessesuen, line 59: The 300 value (5 minutes) used in the `from` is a hard wired value that probably won't work for everyone. I think the interval needs to be user specified in the spec as an `interval` duration string field. e.g.:

```yaml
    provider:
      datadog:
        interval: 5m
        URL: https://api.datadoghq.com
        query: |
          sum:requests.error.count{service:{{args.service-name}}} /
          sum:requests.request.count{service:{{args.service-name}}}

```

metricproviders/datadog/datadog_test.go
```diff
@@ -0,0 +1,189 @@
...
+		expectedErrorMessage string
+	}{
+		// When last value of time series matches condition then succeed.
+		{
+			webServerStatus:   200,
+			webServerResponse: `{"status":"ok","series":[{"pointlist":[[1598867910000,0.0020008318672513122],[1598867925000,0.0003332881882246533]]}]}`,
+			metric: v1alpha1.Metric{
+				Name:             "foo",
+				SuccessCondition: "asFloat(result) < 0.001",
+				FailureCondition: "asFloat(result) >= 0.001",
```
jessesuen, line 38: Why is `asFloat` necessary when the return values already appear as floats?

jessesuen (changes requested):

stephen-harris:
Hi @jessesuen I'll make those changes over the weekend, but to answer your questions around DataDog:

I wasn't aware of the 5-API key limit. Our organisation is a heavy datadog user and we appear to have had that limit waivered as we have over 200 API keys (each team has a key per environment, and some are using different keys for different servcies).  In our org each team is encouraged to create their own API & App keys, but not through DataDog directly (presumably because only admins can create API keys)

Application keys are tied to a user, so I imagine even in other organisations they are creating app keys per service.

Just checking the issues, I think @johnkost & @ozooxo might be using DataDog - I don't know if they want to weigh in here.

johnkost:
We use DataDog as well and haven't had any issues with the 5-API key limit (roughly same setup as @stephen-harris).

I think it makes sense to have this setup similarly at the controller level as Argo-Rollouts is an application, just like any of the other apps we have.

metricproviders/datadog/datadog_test.go
stephen-harris: It's using jsonpath - same as the web metric to extract the result, so it comes out as a string. The metric provider could convert it back to a float automatically. I can't imagine they'd ever be a query that returns something that isn't a float (but as it is we avoid that assumption anyway).

stephen-harris:
If anyone wants to try this out, I've published an image which forks off of 0.9.0: https://hub.docker.com/layers/stephenharris13/argo-rollouts/0.9.0-datadog-bbf9c9b/images/sha256-8151ef6208e792eacd19d476f238c0ed605a8c250f75faea1e4d766c993d91bc?context=explore

metricproviders/datadog/datadog_test.go
jessesuen: Hmmm, I didn't realize that about the jsonpath library. I think we better take a different approach than the webmetric, and avoid using `k8s.io/client-go/util/jsonpath` if we know that it has the limitation of always returning strings.

Flagger appears to make the assumption that the pointlist returns a float. I think we can do the same.
https://github.com/weaveworks/flagger/blob/ffbbc2ca3384509f71791bc3aa476e0cc51845eb/pkg/metrics/providers/datadog.go#L42-L45

Can you refactor such that, `asFloat()` will not be necessary?

BTW, I've filed https://github.com/argoproj/argo-rollouts/issues/728 so that web metric preserves the original format, and `asFloat()` would only be necessary if the return value was a string.

docs/features/analysis.md
```diff
@@ -873,3 +873,43 @@ For success conditions that need to evaluate a numeric return value the `asInt`
...
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: wavefront-api-tokens
+type: Opaque
+data:
+  api-key: <datadog-api-key>
+  app-key: <datadog-app-key>
```
jessesuen, line 914: 1. docs still say "wavefront-api-tokens". I think this should be `datadog-auth-keys`
2. don't we need to allow multiple API/APP keys to be configured?

stephen-harris: (1) - oops! It should be `datadog`. Fixed now

(2) - I don't thinks so. All key pairs are assigned to the organisation, and using them you can pull any metrics from your organisation. Our organisation encourages teams to own their own keys, but team a can use its keys to access metrics reported by team b. We could make the secret name configurable, and you specify it as an AnalysisTemplate parameter, but I don't see the value in doing that.

jessesuen: OK I'll take your word for it, but I'll also pose the question in #argo-rollouts slack in case there's users who may need to configure multiple, in case there's something we're missing.

jessesuen: Just to be clear:

With the proposed spec, we allow the DataDog `address` to be specified in the metric (so it can vary between AnalysisTemplates), but API/APP keys are configured centrally in the controller. This presumes that the same API/APP key can be used for different DataDog URLs. This may very well be possible with DataDog (I don't know the answer to this), but I wanted to call that out as a potential flaw.

stephen-harris: That's a good point. No API/APP keys are not cross-site. In fact I believe an organisation is in either the EU or US site. Although not a secret, for ease of configuration we could move these to the `datadog` secret (alongside `app_key` and `api_key`). What do you think?

metricproviders/datadog/datadog.go
```diff
@@ -0,0 +1,197 @@
...
+	measurement.FinishedAt = &finishedTime
+
+	return measurement
+}
+
+func (p *Provider) parseResponse(metric v1alpha1.Metric, response *http.Response) (string, v1alpha1.AnalysisPhase, error) {
+	var data interface{}
+
+	jsonParser := jsonpath.New("metrics")
+	jsonParser.Parse("{.series[0].pointlist[-1:][1]}")
```
jessesuen, line 111: Instead of using jsonParser to access the float array we are interested in, could we just unmarshal the body into a data type which we expect it to be in? e.g.:

```
type datadogResponse struct {
	Series []struct {
		Pointlist [][]float64 `json:"pointlist"`
	}
}
```

I think theres a lot of overhead of using jsonpath/reflection approach currently.

docs/features/analysis.md
jessesuen: Yes. I agree we should make the address and api/app keys centrally configured, and remove the `address` field from the metric. I'm thinking the Secret should look something like:

```yaml
data:
  default: |
    address: https://api.datadoghq.com
    api-key: igq7imbxlm97vnlw61bo
    app-key: 35bu8ofvc0osyjmqxtgi
```

This also leaves the door open so if in the future, multiple addresses and/or app keys could to be specified in the same cluster. e.g.:

```yaml
data:
  us-app1: |
    address: https://api.datadoghq.com
    api-key: igq7imbxlm97vnlw61bo
    app-key: 35bu8ofvc0osyjmqxtgi
  us-app2: |
    address: https://api.datadoghq.com
    api-key: igq7imbxlm97vnlw61bo
    app-key: cb7ehxo1jwagyceth8v8
  eu: |
    address: https://api.datadoghq.eu
    api-key: cb7ehxo1jwagyceth8v8
    app-key: timk0do15pafuvmfwimg
```

And the names `us-app1` and `us-app2` could be referenced in the metric.

But for MVP, for simplicity, we can go out with a single configured DataDog region and single api/app key combination.

metricproviders/datadog/datadog.go
stephenharris: Refactored this so its no longer use jsonpath

docs/features/analysis.md
stephenharris: I've implemented this, though I'm starting to think perhaps we should have just one level (e.g. a secret per app/region), and then reference the name of the secret in the metric.

```
apiVersion: v1
kind: Secret
metadata:
   name: datadog
data:
  address: https://api.datadoghq.com
  api-key: igq7imbxlm97vnlw61bo
  app-key: 35bu8ofvc0osyjmqxtgi
---
apiVersion: v1
kind: Secret
metadata:
   name: datadog-app2
data:
  address: https://api.datadoghq.eu
  api-key: cb7ehxo1jwagyceth8v8
  app-key: timk0do15pafuvmfwimg
```

Not too fussed though; I think in most use cases organisations will only have one region, and will only have key pair used by rollouts.

jessesuen: > though I'm starting to think perhaps we should have just one level (e.g. a secret per app/region), and then reference the name of the secret in the metric.

Actually I agree with you. A secret per app/region sounds better as it doesn't put a special datastructure in the secret payload. Then things like kustomize patching the secret will just work. Let's go with your suggestion, but we don't have to do anything yet with regards to referencing until someone has a need to reference multiple regions.

metricproviders/datadog/datadog.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// GarbageCollect is a no-op for the Datadog provider
+func (p *Provider) GarbageCollect(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric, limit int) error {
+	return nil
+}
+
+func NewDatadogProvider(logCtx log.Entry, kubeclientset kubernetes.Interface) (*Provider, error) {
+	ns := Namespace()
+	secret, err := kubeclientset.CoreV1().Secrets(ns).Get(DatadogTokensSecretName, metav1.GetOptions{})
```
jessesuen, line 171: FYI I updated the k8s libraries from underneath you, which now require context objects passed into every call. You can just use context.TODO() for now.

jessesuen (approved):
Awesome work!

PR #708, opened 2020-09-08
chore: Set preserveUnknownFields: false to prevent mistake

mtsmfm:
Setting `preserveUnknownFields: false` makes validation strict and can prevent nesting mistake like the following:

```yaml
strategy:
  # <= canary is missing!
  steps: [...]
```

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

Checklist:

  * I think this is a chore
* [x] I've signed the CLA.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ all: controller image
 codegen: mocks
 	./hack/update-codegen.sh
 	./hack/update-openapigen.sh
-	PATH=$$DIST_DIR:$$PATH go run ./hack/gen-crd-spec/main.go
+	PATH=${DIST_DIR}:$$PATH go run ./hack/gen-crd-spec/main.go
 
 .PHONY: controller
 controller: clean-debug
--- a/hack/gen-crd-spec/main.go
+++ b/hack/gen-crd-spec/main.go
@@ -57,6 +57,7 @@ func NewCustomResourceDefinition() []*extensionsobj.CustomResourceDefinition {
 		"controller-gen",
 		"paths=./pkg/apis/rollouts/...",
 		"crd:trivialVersions=true",
+		"crd:preserveUnknownFields=false",
 		"output:crd:stdout",
 	).Output()
 	if err != nil {
--- a/hack/gen-crd-spec/main.go
+++ b/hack/gen-crd-spec/main.go
@@ -72,6 +72,11 @@ func NewCustomResourceDefinition() []*extensionsobj.CustomResourceDefinition {
 	// clean up stuff left by controller-gen
 	deleteFile("config/webhook/manifests.yaml")
 	deleteFile("config/webhook")
+	deleteFile("config/argoproj.io_analysisruns.yaml")
+	deleteFile("config/argoproj.io_analysistemplates.yaml")
+	deleteFile("config/argoproj.io_clusteranalysistemplates.yaml")
+	deleteFile("config/argoproj.io_experiments.yaml")
+	deleteFile("config/argoproj.io_rollouts.yaml")
 	deleteFile("config")
 
 	crds := []*extensionsobj.CustomResourceDefinition{}
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -1,3 +1,5 @@
+#! /usr/bin/env bash
+
 set -o errexit
 set -o nounset
 set -o pipefail
--- a/manifests/crds/analysis-run-crd.yaml
+++ b/manifests/crds/analysis-run-crd.yaml
@@ -18,6 +18,7 @@ spec:
     shortNames:
     - ar
     singular: analysisrun
+  preserveUnknownFields: false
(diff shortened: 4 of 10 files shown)


mtsmfm:
> * I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.

Is there any existing tests for CRD validation?
Sorry I'm not sure which test code I need to change.

mtsmfm:
> * My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

Hmm, I can't find USERS.md. Should I create the file?

jessesuen (changes requested):
Hi @mtsmfm the files you changed are auto-generated and shouldn't be modified directly since it will be lost on the next call to generate manifests. The right way would be to send the flag to the controller-gen tool here: https://github.com/argoproj/argo-rollouts/blob/master/hack/gen-crd-spec/main.go#L56-L60, and then re-run `make manifests`

mtsmfm:
@jessesuen Thank you for pointing out.
I've noticed `make codegen` doesn't work well.
I committed both codegen script fix and changing `preserveUnknownFields`.
Please feel free to ask me to split this PR if you want.

jessesuen (approved):

jessesuen:
Thanks for fixing the makefile too!

jessesuen:
Unfortunately, this change needs to be reverted:

The problem is, we had to remove validation of resource limits in the schema, because OpenAPI validation does not have proper support for things like ResourceQuantity.  When this is coupled with preserveUnknownFields, it has the side effect of dropping the limits.

mtsmfm:
@jessesuen 

> because OpenAPI validation does not have proper support for things like ResourceQuantity

Sorry, I'm not sure what `ResourceQuantity` means.
Can you give me example yaml?

jessesuen:
The following will longer work by setting `preserveUnknownFields: false`:

```yaml
kind: Rollout
metadata:
  name: container-resource-formats
spec:
  aaaaa: aaaa
  selector:
    matchLabels:
      app: container-resource-formats
  template:
    metadata:
      labels:
        app: container-resource-formats
    spec:
      containers:
      - name: a
        image: nginx:1.19-alpine
        command: [sleep, infinity]
        resources:
          requests:
            cpu: 0.001
      - name: b
        image: nginx:1.19-alpine
        command: [sleep, infinity]
        resources:
          requests:
            cpu: '0.001'
      - name: c
        image: nginx:1.19-alpine
        command: [sleep, infinity]
        resources:
          requests:
            cpu: 1m
  strategy:
    canary: {}
```

Because we remove validation for these fields here:
https://github.com/argoproj/argo-rollouts/blob/master/hack/gen-crd-spec/main.go#L230-L261

The reason why we remove validation for it is because the generated OpenAPI schema validation cannot support the various `cpu:` formats.

jessesuen:
FYI, [the recent switch to use tolerant dynamic informers](https://github.com/argoproj/argo-rollouts/pull/666) actually drops unknown fields, so preserveUnknownFields is largely unnecessary in argo-rollouts as of v0.9.1. The controller will actually drop the fields when it persists the rollout.

mtsmfm:
> The controller will actually drop the fields when it persists the rollout.

Let's say we have the following YAML:

```yaml
strategy:
  canary: {}
  steps: # <= accidentally nested under strategy, not canary
    - setWeight: 50
    - pause: {}
```

In this case, `steps` is just dropped silently and the change will be rolled out 100%, right?

I'm proposing to set `preserveUnknownFields: false` to return an error from a server without applying the change in such a case.

jessesuen:
> In this case, steps is just dropped silently and the change will be rolled out 100%, right?

Correct, that is what would happen without `preserveUnknownFields: false`, so it's not the best solution.

> I'm proposing to set preserveUnknownFields: false to return an error from a server without applying the change in such a case.

I understand. If there was a way to do that while still allowing the example I pasted to work, then we can certainly add back `preserveUnknownFields: false`, but at the moment, `preserveUnknownFields: false` is not allow container resources to be set properly.

mtsmfm:
Thank you for your kindly explanation.
Now I understand we need to remove the hack (probably not only `removeK8S118Fields` but also `removeNestedItems`?)
and it's because of OpenAPI (or k8s CRD) limitation.

jessesuen:
> Now I understand we need to remove the hack (probably not only removeK8S118Fields but also removeNestedItems?)

The hack was added for a good reason. In PR https://github.com/argoproj/argo-rollouts/pull/752, I actually updated controller-gen to latest version to see if underlying issue was addressed in v0.4.1, and even attempted to remove the hack workarounds, including removing `removeK8S118Fields`, but it was still not working. The CRD generated by controller-gen v0.4.1 would not accept container resources in various formats.

You're welcome to try but I did not have any luck.

Issue #707, opened 2020-09-07
docs: I can't see any docs about kubectl plugin commands


shibataka000:
## Summary 

I can't see any docs about kubectl plugin commands in https://argoproj.github.io/argo-rollouts/ .

For example, following pages are broken.

- https://argoproj.github.io/argo-rollouts/generated/kubectl-argo-rollouts/kubectl-argo-rollouts.md
- https://argoproj.github.io/argo-rollouts/generated/kubectl-argo-rollouts/kubectl-argo-rollouts_abort.md
- https://argoproj.github.io/argo-rollouts/generated/kubectl-argo-rollouts/kubectl-argo-rollouts_create.md
- etc

And nav bar say `None` , `None` , `None` ...

![image](https://user-images.githubusercontent.com/1219666/92389638-6b6bdf80-f154-11ea-8c8f-54e3a1219e82.png)

I guess these pages are generated by `make plugin-docs` .

## Diagnostics

None

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Fixed

PR #711, opened 2020-09-16
[fix] Adds dependency to generate plugin-docs when serving locally (#707)

ignasi35:
Checklist:

* [x] I've signed the CLA.
* [x] (n/a) I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] (n/a) My builds are green. Try syncing with master if they are not. 
* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md). <-- this file is missing (is this step obsolete?)

I'm keeping this as a DRAFT PR because this is _just_ a change I did to run the docs locally (while the site is not fixed, see #707). I understand it's not ideal since it adds extra load on every `make serve-docs`. Feel free to close if you think it's the wrong approach. This is my first time using `mkdocs` or navigating the Argo project repositories so I'm very unfamiliar with the tooling and how it's used here.

This is related to #707 but I don't think it's the fix. I suspect the fix requires extra improvements on the `release-docs` task in the makefile but I'd not sure how to test the contents produced are the expected output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ release-docs: plugin-docs
 
 # convenience target to run `mkdocs serve` using a docker container
 .PHONY: serve-docs
-serve-docs:
+serve-docs: plugin-docs
 	docker run --rm -it -p 8000:8000 -v ${CURRENT_DIR}:/docs squidfunk/mkdocs-material serve -a 0.0.0.0:8000
 
 .PHONY: release-precheck


jessesuen (approved):

jessesuen:
LMK when it's ready for review

ignasi35:
> LMK when it's ready for review

For me it's ready. :-)

Issue #715, opened 2020-09-17
Panic after switching to tolerant informers


jessesuen:
## Summary 

This code is incorrect because it's casting the object to a rollout but we now return unstructured.Unstructured with the new tolerant informer

```
	cfg.AnalysisRunInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc: func(obj interface{}) {
			controllerutil.EnqueueParentObject(obj, register.RolloutKind, controller.enqueueRollout)
		},
		UpdateFunc: func(old, new interface{}) {
			newAR := new.(*v1alpha1.AnalysisRun)
			oldAR := old.(*v1alpha1.AnalysisRun)
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #719, opened 2020-09-19
fix: panic after switching to tolerant informers

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/715

The controller was panicking when dealing with AnalysisRuns and Experiments because the switch to tolerant informer did not also switch the informer handlers to deal with unstructured.Unstructured objects

The new e2e tests successfully reproduce the problem before the fix. It also reproduces a problem where an analysisrun is created when a rollout is initially deployed, but that will be addressed in a different PR.

Also pork-barreled into this PR are:
* fix for unit test which did not like golang 1.15's changes to error message
* a change to logging to only log relevant information when reconciling canary steps


khhirani (approved):
LGTM

Issue #714, opened 2020-09-17
Rollout is creating analysisRun on initial deploy


jessesuen:
## Summary 

In v0.9.0, when deploying a freshly created Rollout, an analysis run is being created when it shouldn't be. Am able to reproduce in v0.9.0 but not in v0.8.3

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
This seems to be happening only when trafficRouting is used

PR #722, opened 2020-09-19
fix: do not create background analysisruns with initial deploy

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/714

The check for detecting an initial deploy and conditionally creating a background analysisrun was incomplete. It did not catch a scenario where `Rollout.Status.StableRS == ""`. Not sure why this is true in v0.9. but not v0.8

This is stacked ontop of another PR so only look at https://github.com/argoproj/argo-rollouts/pull/722/commits/9d1424e117b1a8a400cad885c76760f200b96235


khhirani (approved):
LGTM

Issue #723, opened 2020-09-19
analysisrun in a reconciliation hotloop


jessesuen:
## Summary 

I have an analysisrun in a reconciliation hotloop

```
INFO[2020-09-19T03:27:29-07:00] Started syncing Analysis at (2020-09-19 03:27:29.462115 -0700 PDT m=+4.259801488)  analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] skipping measurement: run is terminating      analysisrun=rollouts-demo-594bbbb9bf-1 metric=success-rate namespace=default
INFO[2020-09-19T03:27:29-07:00] taking 0 measurements                         analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] enqueuing analysis after 0s                   analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] No status changes. Skipping patch             analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] Reconciliation completed                      analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default time_ms=2.1602650000000003
INFO[2020-09-19T03:27:29-07:00] Started syncing Analysis at (2020-09-19 03:27:29.464552 -0700 PDT m=+4.262238246)  analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] skipping measurement: run is terminating      analysisrun=rollouts-demo-594bbbb9bf-1 metric=success-rate namespace=default
INFO[2020-09-19T03:27:29-07:00] taking 0 measurements                         analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] enqueuing analysis after 0s                   analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] No status changes. Skipping patch             analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default
INFO[2020-09-19T03:27:29-07:00] Reconciliation completed                      analysisrun=rollouts-demo-594bbbb9bf-1 namespace=default time_ms=0.640943
[...]

PR #724, opened 2020-09-19
fix: analysis controller could get into a hotloop with terminated run

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/723


khhirani (approved):
LGTM

Issue #721, opened 2020-09-19
AnalysisRun are not being associated with correct revision


jessesuen:
## Summary 

The analysis run is incorrectly being associated with the wrong ReplicaSet revision:

```
Name:            rollout-background-analysis
Namespace:       default
Status:          ✔ Healthy
Strategy:        Canary
  Step:          2/2
  SetWeight:     100
  ActualWeight:  100
Images:          argoproj/rollouts-demo:yellow (stable)
Replicas:
  Desired:       1
  Current:       1
  Updated:       1
  Ready:         1
  Available:     1

NAME                                                    KIND         STATUS        AGE   INFO
⟳ rollout-background-analysis                           Rollout      ✔ Healthy     113s
├──# revision:2
│  └──⧉ rollout-background-analysis-cdd685668           ReplicaSet   ✔ Healthy     85s   stable
│     └──□ rollout-background-analysis-cdd685668-q2s4c  Pod          ✔ Running     85s   ready:1/1
└──# revision:1
   ├──⧉ rollout-background-analysis-747669bc9b          ReplicaSet   • ScaledDown  113s
   └──α rollout-background-analysis-cdd685668-1         AnalysisRun  ✔ Successful  85s   ⚠ 2      <<<<< this should be associated with revision 2, not under 1
```

## Diagnostics

What version of Argo Rollouts are you running?

```
$ kro version
kubectl-argo-rollouts: v0.10.0+e64c941.dirty
  BuildDate: 2020-09-02T07:27:11Z
  GitCommit: e64c94177371e49e02a0d336dbeb1a98130ac586
  GitTreeState: dirty
  GoVersion: go1.15
  Compiler: gc
  Platform: darwin/amd64
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
This seems to be caused by recent change to rolloutContext

PR #725, opened 2020-09-20
fix: revision annotation was off-by-one for canary rollouts

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/721

The refactor to the rolloutContext introduced a problem with canary rollouts where the revision annotated on the ReplicaSet and Rollout was under by one. This is because it calculated the revision number by the count of older RSs, which was not calculated correctly during the refactor.

This change also makes it possible for E2E tests to run without an instanceID, so that they may run against a normally installed cluster controller.

NOTE: this PR is layered ontop of 9d1424e so only look at last commit


khhirani (approved):
LGTM

Issue #720, opened 2020-09-19
Rollout can pause twice for same canary pause step


jessesuen:
## Summary 

The TestCanarySetCanaryScale e2e test sometimes fails, which reproduces the following problem where

1. the Rollout controller pauses a Rollout when reaching a canary pause step
2. the e2e test unpauses the rollout (as part of a promote)
3. the Rollout controller pauses it again because it is operating on a stale rollout in the informer cache

The e2e test then times out because it never reaches the next step (since it's indefinitely paused).

The underlying cause is that it's possible the rollout is operating on stale data in the informer and decides it is time to pause the rollout even though it had previously paused it in the last reconciliation (but doesn't know it has).

NOTE: this may not be a real-world scenario because in order to reproduce this bug, the rollout must *immediately* be unpaused after it becoming paused (fractions of seconds), faster than the informer cache can become updated. In fact the e2e test can only reproduce this some of the time.

## Diagnostics

v0.9.0

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #726, opened 2020-09-20
feat: writeback rollout updates to informer to prevent stale data

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/720

Currently the rollout controller may operate on stale information in the informer cache. This can easily happen when a rollout is requeued into the workqueue while already in the middle of reconciliation. When this happens, the controller immediately re-reconciles the rollout, but this time operating on a stale version of the rollout in the cache (it is stale because the changes to the rollout saved in first reconciliation, have not yet made the round trip into the cache).

At best, this simply duplicates work. At worst, it causes incorrect behavior (such as re-pausing after an unpause as in #720)

There is a technique which argo workflows uses that writes back the updated resource back into the informer cache. Using this technique, even if a rollout is immediately re-reconciled, it at least has the updates saved in the previous reconciliation.

Before this change, TestCanarySetCanaryScale could reproduce the double-pause problem in #720 quite easily. After this fix, I am unable to reproduce the problem.

### DIFF
--- a/rollout/context.go
+++ b/rollout/context.go
@@ -15,6 +15,8 @@ type rolloutContext struct {
 	log *log.Entry
 	// rollout is the rollout being reconciled
 	rollout *v1alpha1.Rollout
+	// newRollout is the rollout after reconciliation. used to write back to informer
+	newRollout *v1alpha1.Rollout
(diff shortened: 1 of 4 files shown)


khhirani (approved):
LGTM

Issue #596, opened 2020-07-27
[Feature Request] rollout status


markrity:
There would be an awesome if cli had a functionality to get a current status of a rollout.

Command:
`kubectl argo rollouts status rollout guestbook` 
Returns:
`Healthy`

or as a flag.
`kubectl argo rollouts get rollout guestbook --status`

To know if it Healthy,Processing or Degraded.

markrity:
In addition I would add that knowing also the status of each `ReplicaSet` (blue/green) , would be great.
Lets say that overall `Rollout` got `Degraded` , but I still would like to know if `blue-ReplicaSet` aka `preview` , is `Healthy` or not .

something like
`kubectl argo rollouts get rollout guestbook --statusPreview`
or
`kubectl argo rollouts get rollout guestbook --status --preview`

returns
`Healthy` 

it is really useful for building fail-less CD Pipelines.

PR #730, opened 2020-09-23
feat(cli): get rollout/experiment print options

cronik:
Add common k8s print flags to get rollout/experiments commands.

```
-o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
      --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
```

Example usages:
```
# Output json
kubectl argo rollouts get rollout guestbook -o json

# Print status
kubectl argo rollouts get rollout guestbook -o jsonpath='{.spec.status}'
Healthy

# Print Blue Green preview replica set status
kubectl argo rollouts get rollout guestbook -o jsonpath='{.spec.replicaSets[?(@.spec.preview==true)].spec.status}'
Progressing
```

Refs #596

Checklist:

* [x] I've signed the CLA.

### DIFF
--- a/pkg/kubectl-argo-rollouts/cmd/get/get.go
+++ b/pkg/kubectl-argo-rollouts/cmd/get/get.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/info"
@@ -98,6 +99,7 @@ type GetOptions struct {
 	Watch   bool
 	NoColor bool
 
+	PrintFlags *genericclioptions.PrintFlags
 	options.ArgoRolloutsOptions
 }
 
--- a/pkg/kubectl-argo-rollouts/cmd/get/get_experiment.go
+++ b/pkg/kubectl-argo-rollouts/cmd/get/get_experiment.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/juju/ansiterm"
 	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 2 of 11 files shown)


jessesuen:
Hi @cronik , I don't agree we should be exposing the RolloutInfo / ExperimentInfo objects as consumable entities to users/CLI:
* those objects are considered an implementation detail to power the CLI, not intended to be consumed
* it implies we would need to commit to stability/backwards compatibility for this spec since CLI automation would rely on it
* we would have created two specs the Rollout spec vs. RolloutInfo spec, which will add to confusion
* the new spec would need to be documented
* Commands like `kubectl argo rollouts get rollout guestbook -o json` would return a completely different payload than `kubectl get rollout guestbook -o json`, which will be confusing

If there is a need/desire to boil down the overall status of a Rollout or Experiment into a single status code such as `Healthy`, `Progressing`, etc... I think we should introduce a new CLI command to do this instead of exposing a new object. An analogous example would be the `kubectl rollout status` command (the difference being that `kubectl rollout status` reduces it to a message vs. a status code).

cronik:
@jessesuen  that all makes sense and I agree to all those points. This was a bit of a learning exercise with regard to the k8s generic printer options and figured I submit it to see if any of it stuck. 

There definitely seems to be a desire within my own org and looking at the feature requests for either some command to get the status via the cli and/or a usable `kubectl wait` condition for rollout states. 

All the logic to determine the rollout status (i.e. Health, Degraded, Processing, ...) is based on the rollout status itself, would it make sense to add a property to the RolloutStatus to capture the current status. Then instead of introducing a new cli command one could just use the built in print options on the Rollout resource.

```
kubectl get ro foo -o jsonpath='{.status.rollout}'
```

This can already by done for experiments using
```
kubectl get experiment foo -o jsonpath='{.status.phase}'
```

jessesuen:
> would it make sense to add a property to the RolloutStatus to capture the current status.

@cronik  i think an informational-only field to capture this under  .status would be reasonable so that clients do not have to re-implement the algorithm. Please feel free to file an issue.

Issue #728, opened 2020-09-22
web metric jsonPath always returns values as strings


jessesuen:
## Summary 

According to https://github.com/argoproj/argo-rollouts/pull/705#discussion_r488051520, the json path library we use to parse arbitrary payloads in the web metric, returns result values as strings. It should return the result in the original format, so that functions like `asFloat` are not necessary if the data is already in the desired format.

## Diagnostics

v0.9

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
We're not using jsonParser to its fullest potential because we call the Execute convenience method which loses the information about the type. If we switch to jsonParser.FindResults(), we will get reflect.Values which indicate the original type.

```golang
	err = p.jsonParser.Execute(buf, data)   // this loses information
	if err != nil {
		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Could not find JSONPath in body: %s", err)
	}
	out := buf.String()

```

Issue #718, opened 2020-09-19
insecureSkipHostVerify option for web metric


jessesuen:
# Summary

Web metric should allow communicating to a host with a self-signed certificate, and thus an insecureSkipHostVerify option

# Use Cases

Dev and test purposes

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Fixed

Issue #717, opened 2020-09-19
jsonPath should not be required for web metric


jessesuen:
Got this error when using the web metric:

```
The AnalysisRun \"web-background-success\" is invalid: spec.metrics.provider.web.jsonPath: Required value\n" rollout=rollout-bg-analysis-success
```

If jsonPath is omitted, the `result` placeholder object should simply be the whole return payload.

jessesuen:
Fixed in v0.9.1

PR #731, opened 2020-09-23
feat: web metrics preserve data types, allow insecure tls, and make jsonPath optional

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/728, https://github.com/argoproj/argo-rollouts/issues/718, https://github.com/argoproj/argo-rollouts/issues/717

1. When using jsonPath, the web metric always converted the result into a string. This is why the `asFloat` and `asInt` helpers were introduced, in to convert these strings into numbers so that mathematical comparison could be performed. This change preserves the original types so that `asFloat` and `asInt` are no longer necessary (as long as the original data types are numeric).

2. Previously, `jsonPath` was a required field. This change allows `jsonPath` to be omitted, in which case the entire body is the payload.

3. Added an `insecure: true` field that allows the web metric to disable TLS host verification when making HTTP requests.

### DIFF
--- a/docs/features/analysis.md
+++ b/docs/features/analysis.md
@@ -837,7 +837,7 @@ data:
 
 ## Web Metrics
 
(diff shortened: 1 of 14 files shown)


khhirani (approved):
LGTM

PR #735, opened 2020-09-24
fix:scale down not  available stableRS

wangzhipeng:
After the program version of crashloopbackoff is deployed for the first time, the number of copies of the old version is not deleted after rolling out the version that can be run

### DIFF
--- a/utils/replicaset/canary.go
+++ b/utils/replicaset/canary.go
@@ -163,7 +163,14 @@ func CalculateReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS *appsv1.Re
 	minAvailableReplicaCount := rolloutSpecReplica - MaxUnavailable(rollout)
 
 	totalAvailableOlderReplicaCount := GetAvailableReplicaCountForReplicaSets(oldRSs)
-	scaleDownCount := GetReplicasForScaleDown(newRS) + GetReplicasForScaleDown(stableRS) + totalAvailableOlderReplicaCount - minAvailableReplicaCount
+	stableRSReplicasForScaleDown := GetReplicasForScaleDown(stableRS)
+	scaleDownCount := GetReplicasForScaleDown(newRS) + stableRSReplicasForScaleDown + totalAvailableOlderReplicaCount - minAvailableReplicaCount
+
+	if scaleDownCount <= 0 && stableRSReplicasForScaleDown == 0 {
+		// if stableRS AvailableReplicas is 0 , container  not running  rollout replicas to 0
+		stableRSReplicaCount = stableRSReplicasForScaleDown
+		return newRSReplicaCount, stableRSReplicaCount
+	}
 
 	if scaleDownCount <= 0 {
 		// Cannot scale down stableRS or newRS without going below min available replica count
--- a/utils/replicaset/canary_test.go
+++ b/utils/replicaset/canary_test.go
@@ -450,6 +450,22 @@ func TestCalculateReplicaCountsForCanary(t *testing.T) {
 			expectedStableReplicaCount: 10,
 			expectedCanaryReplicaCount: 2,
 		},
+		{
+			name:                "scale down not  available stableRS",
+			rolloutSpecReplicas: 10,
+			setWeight:           30,
+			maxSurge:            intstr.FromInt(1),
+			maxUnavailable:      intstr.FromInt(0),
+
+			stableSpecReplica:      4,
+			stableAvailableReplica: 0,
+
+			canarySpecReplica:      0,
+			canaryAvailableReplica: 0,
+
+			expectedStableReplicaCount: 0,
+			expectedCanaryReplicaCount: 3,
+		},
 	}
 	for i := range tests {
 		test := tests[i]


jessesuen:
@wangzhipeng thanks for your contribution. Just so I can understand the bug a bit better, can you provide the steps to reproduce the issue? I think I may want to write an e2e test for this later.

wangzhipeng:
create   revision1.yaml  ,this  `command` is fail
```
kind: Rollout
metadata:
  name: ping
spec:
  replicas: 2
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      app: ping
  template:
    metadata:
      labels:
        app: ping
    spec:
      containers:
      - name: ping
        image: alpine
        command:
        - pin
        args:
        - baidu.com
  strategy:
    canary:
      steps:
      - setWeight: 100
```
```
kubectl apply -f revision1.yaml
```

fix command, edit  revision1.yaml  to revision2.yaml
```
kind: Rollout
metadata:
  name: ping
spec:
  replicas: 2
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      app: ping
  template:
    metadata:
      labels:
        app: ping
    spec:
      containers:
      - name: ping
        image: alpine
        command:
        - ping
        args:
        - baidu.com
  strategy:
    canary:
      steps:
      - setWeight: 100
```
```
kubectl apply -f revision2.yaml
```

this ping  revision1  rs Replica is 2  ,not scale  down

jessesuen:
Thanks. This is a good find. I was able to reproduce the problem and was able to write an e2e test to catch the issue. However, I think I have some questions I need to ask @dthomson25 about regarding the implementation to see if this is the right way to fix the problems.

utils/replicaset/canary.go
```diff
@@ -163,7 +163,14 @@ func CalculateReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS *appsv1.Re
 	minAvailableReplicaCount := rolloutSpecReplica - MaxUnavailable(rollout)
 
 	totalAvailableOlderReplicaCount := GetAvailableReplicaCountForReplicaSets(oldRSs)
-	scaleDownCount := GetReplicasForScaleDown(newRS) + GetReplicasForScaleDown(stableRS) + totalAvailableOlderReplicaCount - minAvailableReplicaCount
+	stableRSReplicasForScaleDown := GetReplicasForScaleDown(stableRS)
+	scaleDownCount := GetReplicasForScaleDown(newRS) + stableRSReplicasForScaleDown + totalAvailableOlderReplicaCount - minAvailableReplicaCount
+
+	if scaleDownCount <= 0 && stableRSReplicasForScaleDown == 0 {
```
jessesuen, line 169: Although this works for the scenario you provided, I don't think this is a complete fix, because  `stableRSReplicasForScaleDown == 0` is just one case and I think it would be possible for `stableRSReplicasForScaleDown > 0` and we would still face the problem. E.g. for the initial version, 1 out of 2 pods were available, but the other was CrashLoopBackoff-ing

jessesuen:
@dthomson25 - the current implementation of GetReplicasForScaleDown returns either Spec.Replicas, or Status.AvailableReplicas:

```golang
// GetReplicasForScaleDown returns the number of replicas to consider for scaling down.
func GetReplicasForScaleDown(rs *appsv1.ReplicaSet, new bool) int32 {
	if rs == nil {
		return int32(0)
	}
	if *rs.Spec.Replicas < rs.Status.AvailableReplicas {
		// The ReplicaSet is already going to scale down replicas since the availableReplica count is bigger
		// than the spec count. The controller uses the .Spec.Replicas to prevent the controller from
		// assuming the extra replicas (availableReplica - .Spec.Replicas) are going to remain available.
		// Otherwise, the controller use those extra replicas to scale down more replicas and potentially
		// violate the min available.
		return *rs.Spec.Replicas
	}
	return rs.Status.AvailableReplicas
}
```

However, it's possible for a situation where Spec.Replicas >= Status.Replicas > Status.AvailableReplicas, in which case we would choose `Status.AvailableReplicas`

The real world example is a `stableRS` in this state:

```yaml
spec:
  replicas: 2
status:
  replicas: 2
  availableReplicas: 0
```

Here we would say there are zero replicas available in stable to scale down, even though `status.replicas: 2`

My question is: why isn't `GetReplicasForScaleDown` just always returning spec.replicas? I tried making the change but then unit tests started failing. 

One change I made which *did* pass unit tests was:

```golang
// GetReplicasForScaleDown returns the number of replicas to consider for scaling down.
func GetReplicasForScaleDown(rs *appsv1.ReplicaSet, new bool) int32 {
	if rs == nil {
		return int32(0)
	}
	if *rs.Spec.Replicas < rs.Status.AvailableReplicas {
		// The ReplicaSet is already going to scale down replicas since the availableReplica count is bigger
		// than the spec count. The controller uses the .Spec.Replicas to prevent the controller from
		// assuming the extra replicas (availableReplica - .Spec.Replicas) are going to remain available.
		// Otherwise, the controller use those extra replicas to scale down more replicas and potentially
		// violate the min available.
		return *rs.Spec.Replicas
	}
	if rs.Status.AvailableReplicas < rs.Status.Replicas {
		return rs.Status.Replicas
	}
	return rs.Status.AvailableReplicas
}
```

But I'm not sure it was the correct solution.

dthomson25:
Based on https://github.com/argoproj/argo-rollouts/pull/141/files, I believe that GetReplicasForScaleDown returns the RS's Status.AvailableReplicas sometimes to make sure it honor's the Rollout's maxUnavailable value, and I think we need to make sure that the fix still upholds that maxUnavailable. I think part of the issue is that the replicaset 1 is being set to stable without it ever becoming healthy

jessesuen:
> One change I made which did pass unit tests was:

The "fix" I attempted below (while it passes unit tests), is entirely unsafe, and is an incorrect solution:

```golang
	if rs.Status.AvailableReplicas < rs.Status.Replicas {
		return rs.Status.Replicas
	}
	return rs.Status.AvailableReplicas
}
```

It fails to honor maxSurge and maxUnavailable and undoes the fix in https://github.com/argoproj/argo-rollouts/pull/141

jessesuen:
OK looking into this very deeply for several days, and after extensive testing, I can say this PR doesn't cover all scenarios:
* it only works when the stableRS is 100% completely unavailable (`availableReplicas == 0`)
* when the new logic kicks in, it scales the canary/stable counts immediately to the desired setWeight without honoring maxSurge/Unavailable, which it should still do even if `availableReplicas == 0`

I've filed an alternative fix here https://github.com/argoproj/argo-rollouts/pull/739, which addresses the above

wangzhipeng:
@jessesuen good  , close this PR

PR #739, opened 2020-09-27
fix: unavailable stable RS was not scaled down to make room for canary

jessesuen:
This is an alternative fix for the issue described in https://github.com/argoproj/argo-rollouts/pull/735

Scenario:
1. Apply a "bad" rollout (with or without steps) with an image which will crashloopbackoff, and the following surge/unavailable:

```yaml
spec:
  replicas: 4
  strategy:
    canary:
      maxSurge: 1
      maxUnavailable: 0
```

2.  Update rollout with a new spec which will not crash
3.  Rollout will surge up, such that there are 5 total replicas (1 new, 4 old)
4. Currently, rollout will now be stuck forever never being able to reach the second revision, because the controller decides it cannot scale down the stableRS (even when it should).

The correct behavior should be that the Rollout scales down one stableRS pod, in order to make room for another canary

The fix is that when calculating the replica counts when scaling down the stableRS, it should consider the Spec.Replicas when it is higher than Status.AvailableReplicas.

This also adds many e2e tests to verify scaling behavior


utils/replicaset/canary.go
```diff
@@ -160,23 +160,23 @@ func CalculateReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS *appsv1.Re
 		}
 	}
 
-	minAvailableReplicaCount := rolloutSpecReplica - MaxUnavailable(rollout)
+	if GetReplicaCountForReplicaSets(oldRSs) > 0 {
```
jessesuen, line 163: Note this was changed from `GetAvailableReplicaCountForReplicaSets` to `GetReplicaCountForReplicaSets` to avoid the same problem with older RSs.

jessesuen: Also to simplify logic, i move the check to return early if there are older RSs up here, so all the logic below only has to deal with new and stable RS, and not have to factor in number of older RSs.

utils/replicaset/canary.go
```diff
@@ -160,23 +160,23 @@ func CalculateReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS *appsv1.Re
...
 
-	if scaleDownCount <= totalAvailableOlderReplicaCount {
-		//Need to scale down older replicas before scaling down the newRS or stableRS.
-		return newRSReplicaCount, stableRSReplicaCount
-	}
-	scaleDownCount = scaleDownCount - totalAvailableOlderReplicaCount
+	scaleDownCount := replicasToScaleDown - minAvailableReplicaCount
 
-	if newRS != nil && *newRS.Spec.Replicas > desiredNewRSReplicaCount && scaleDownCount > 0 {
+	if newRS != nil && *newRS.Spec.Replicas > desiredNewRSReplicaCount {
```
jessesuen, line 179: `&& scaleDownCount > 0` check is/was unnecessary because `replicasToScaleDown <= minAvailableReplicaCount` covers that.

danny-stytch (approved):
I can't really comment on the e2e tests, but the actual code change LGMT.

khhirani (approved):
LGTM

PR #751, opened 2020-10-01
fix: bluegreen postPromotion abort did not work properly

jessesuen:
Writing e2e tests for blueGreen pre/postpromotion analysis revealed a bunch of issues with blueGreen analysis including:
* creating jobs over and over in a loop
* not aborting properly and switching active selector back to stable

This PR also adds a bunch of analysis tests surrounding blue-green strategy and pre/post promotion analysis.


khhirani (approved):
LGTM

PR #752, opened 2020-10-01
chore: update k8s dependencies to v1.18.2. revert preserveUnknownFields=false

jessesuen:
* Updates K8s libraries to v1.18.2, along with all method signature changes
* reverts CRD change to `preserveUnknownFields: false` for reasons described in https://github.com/argoproj/argo-rollouts/pull/708#issuecomment-701614697


khhirani (approved):
LGTM

Issue #750, opened 2020-09-30
Rollout.Status.AvailableReplicas is null during rollout


dudadornelles:
## Summary 

During a rollout with bluegreen strategy, `Rollout.Status.AvailableReplicas` changes to null for a little bit, which can trip monitoring alerts, when in fact replicas have been available the whole time (thats the whole point of progressive delivery).

## Diagnostics

What version of Argo Rollouts are you running? 0.9.0

## logs

controller doesn't have useful logs for this, but watching `status.availableReplicas` in the rollout resource during a release shows 

```
$ while true; do kc get rollout/<app> -o json | jq .status.availableReplicas; sleep 1; done
2
2
null
null
null
null
null
null
null
null
null
null
1
2
2
2
````

I was expecting this number to stay at 2 the whole time.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

**Comment by jessesuen** [comment]
It's null (omitted) when the value is zero:

```
	// Total number of available pods (ready for at least minReadySeconds) targeted by this rollout.
	// +optional
	AvailableReplicas int32 `json:"availableReplicas,omitempty"`
```

Is availableReplicas truly zero? or is this an accounting problem?

**Comment by dudadornelles** [comment]
I looked at  `readyReplicas` during the rollout as well and the logs are:

```
$ while true; do kc get rollout/pipeline-tools -o json | jq .status.readyReplicas; sleep 1; done
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
3
3
3
3
4
4
4
4
4
4
4
3
3
3
2
2
2
```

So Im quite positive that at all times during the deployment there are `availableReplicas`, I think it is how we are updating that value during the rollout that is not quite right

dudadornelles:
Potentially here? https://github.com/argoproj/argo-rollouts/blob/master/rollout/bluegreen.go#L244

Maybe we should be passing the old replica set there too?

jessesuen:
I think I may have broken this recently in https://github.com/argoproj/argo-rollouts/pull/704.

PR #757, opened 2020-10-03
fix(controller): calculate bluegreen available replicas from active ReplicaSet

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/750

Changes behavior of blue-green replica calculation to derive information from the active replicaset (if available). Otherwise, they are derived from all replicasets.


khhirani (approved):
LGTM

PR #767, opened 2020-10-06
test: add e2e test for previewReplicaCount and bluegreen-to-canary

jessesuen:
Adds two e2e test to:
* verify the bluegreen previewReplicaCount feature
* verify behavior when switching strategies from blue-green to canary


khhirani (approved):
LGTM

Issue #756, opened 2020-10-02
BlueGreen trapped in Progressing until ScaleDownDelay is exceeded


jessesuen:
## Summary 

At the end of a successful blue-green update, we have a rollout with a status like:

```yaml
  status:
    HPAReplicas: 1
    availableReplicas: 1
    blueGreen:
      activeSelector: 6874845b5d
      previewSelector: 6874845b5d
    canary: {}
    conditions:
    - lastTransitionTime: "2020-10-02T07:52:21Z"
      lastUpdateTime: "2020-10-02T07:52:21Z"
      message: Rollout has minimum availability
      reason: AvailableReason
      status: "True"
      type: Available
    - lastTransitionTime: "2020-10-02T07:52:26Z"
      lastUpdateTime: "2020-10-02T07:52:28Z"
      message: ReplicaSet "bluegreen-abort-and-update-6874845b5d" has successfully
        progressed.
      reason: NewReplicaSetAvailable
      status: "True"
      type: Progressing 
    currentPodHash: 6874845b5d
    observedGeneration: 5d4d9cd75b
    readyReplicas: 2
    replicas: 2
    selector: app=bluegreen-abort-and-update,rollouts-pod-template-hash=6874845b5d
    stableRS: 6874845b5d
    updatedReplicas: 1
```

Currently the kubectl plugin (and I believe also Argo CD), considers the above Rollout in a Progressing state because we have less updated replicas than status.replicas:

The check which decides this is in the rollout_info:

```golang
	if ro.Status.UpdatedReplicas < ro.Status.Replicas {
		// more replicas need to be updated
		return "Progressing"
	}
```

This calculation is misguided because with blue-green, we may have many old ReplicaSets lying around which are scaled up, because of the `scaleDownRevisionLimit` feature

The controller in the case of blueGreen, should update Status.Replicas to reflect the number of replicas which the active service is pointing to, rather than the sum of all replicas of all involved replicasets. Currently it is calculated as:

```golang
	newStatus.Replicas = replicasetutil.GetActualReplicaCountForReplicaSets(c.allRSs)
```

The above is appropriate for replica-weighted canary, but not appropriate for blue-green.

## Diagnostics

What version of Argo Rollouts are you running? v0.9.1

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
The kubectl rollout plugin does not match the health check in Argo CD. I may be the problem

[...]

PR #768, opened 2020-10-07
fix: blue-green scaleDownDelay was delaying Healthy status

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/756

Unlike canary, in the case of blue-green, the plugin should **_not_** consider a Rollout as still Progressing if it has more Status.Replicas than Status.UpdatedReplicas. This is because the blue-green has a scaleDownDelay, where it can leave old replicasets scaled up for a long time. These scaled up replicasets do not impact the service (unlike canary) because they are not reachable from the active service. This logic is already in Argo CD Rollout health check, but not in the plugin.

Also, this PR changes the output of the plugin such that messages and errors show up in a new "messages" row, in a consistent manner like Argo CD. See below examples:

```
Name:            canary-demo
Namespace:       jesse-test
Status:          ◌ Progressing
Message:         more replicas need to be updated
```

```
Name:            rollout-invalid
Namespace:       default
Status:          ✖ Degraded
Message:         InvalidSpec: The Rollout "rollout-invalid" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"app":"doesnt-match"}: `+"`selector`"+` does not match template `+"`labels`"+`
Strategy:        Canary
```


khhirani (approved):
LGTM

PR #772, opened 2020-10-08
feat: Add New Relic metricprovider

### LINKED ISSUE #736: NewRelic metric provider
# Summary

Expanding the [New Relic](https://newrelic.com/) support to include standard analysis instead of only Mann-Whitney experimentation with Kayenta will cover more use cases. 

# Use Cases

This provides a way to run background analysis and regular canary analysis without dealing with all of the setup and configuration required to run Kayenta. Additionally, this will provide options for analysis for services that are sensitive to replica count changes like Kafka consumers. 

The [newrelic-client-go](https://github.com/newrelic/newrelic-client-go) package provides an easy way to run NRQL queries against NRDB. The existing web metricprovider is not sufficient for working with New Relic's GraphQL API. 

---

jwelch92:
Closes #736 

Due to the different result format for NRQL queries I had to deviate a bit from some of the common result representations in the other providers. I hope the documentation makes it clear how to use the results correctly. 

Checklist:

* [x] I've signed the CLA.


jwelch92:
cc/ @ctrombley @sanderblue @zlesnr

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -130,6 +130,8 @@ type MetricProvider struct {
 	Web *WebMetric `json:"web,omitempty"`
 	// Wavefront specifies the wavefront metric to query
 	Wavefront *WavefrontMetric `json:"wavefront,omitempty"`
+	// NewRelic specifies the newrelic metric to query
+	NewRelic *NewRelicMetric `json:"newrelic,omitempty"`
```
jessesuen, line 134: The API violation in `violation_exceptions.list` is complaining that `NewRelic` has the `R` in the field name uppercased, but the JSON tag is all lowercase.  The json tag should be `newRelic` to avoid the violation.

docs/features/analysis.md
```diff
@@ -881,3 +881,40 @@ was greater than `0.90`
...
+spec:
+  args:
+  - name: application-name
+  metrics:
+  - name: success-rate
+    successCondition: result.successRate >= 0.95
+    provider:
+      newrelic:
+        region: "us"
+        accountID: 1234
```
jessesuen, line 902: I have some questions about the appropriate level of where region and accountID should be configured. i.e. could/should this be configuration at the controller level vs. AnalysisTemplate level? Maybe these clarifying questions would help me better understand:

* can the same account ID be used in different regions? If not, then I think we should eliminate `region` from the analysis spec, and configure the region centrally (same place where the accountID api-key is specified).

* is it common to have multiple accountIDs in the same organization? If most organizations use a single accountID, I would propose to make their experience better / less configuration by not require accountID to be specified in the AnalysisTemplate, and provide a way where a single accountID could be configured centrally. Then AnalysisTemplates would only need to specify `query` and nothing else.

If there is a need / use case to support multiple accountIDs within the same cluster/organization, I think accountID could still be made an **_optional_** field in the analysis spec, but when omitted, the controller could either use designated a "default" accountID / region, or infer the accountID/region/key if there is only a single accountID configured.

Note that a very similar discussion just happened with the Datadog provider:
https://github.com/argoproj/argo-rollouts/pull/705#discussion_r492486540

With DataDog, it was decided that it would be uncommon for a single organization to configure multiple DataDog API/APP keys, so currently argo-rollouts will look to a single Datadog secret for its configuration:

```
apiVersion: v1
kind: Secret
metadata:
   name: datadog
data:
  address: https://api.datadoghq.com
  api-key: igq7imbxlm97vnlw61bo
  app-key: 35bu8ofvc0osyjmqxtgi
```

This doesn't preclude the possibility in the future to support multiple Datadog secrets/api-keys, but it makes the end-user/developer experience better since they did not need to concern themselves with API/APP keys in the AnalysisTemplate.

Another example is with wavefront. It is common to use different wavefront hosts. So, it's configuration looks like this:

Centrally configured Host to API tokens:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: wavefront-api-tokens
type: Opaque
data:
  example1.wavefront.com: <token1>
  example2.wavefront.com: <token2>
```

In the AnalysisTemplate, a reference to **_which_** host is configured:

```yaml
kind: AnalysisTemplate
metadata:
  name: success-rate
spec:
  args:
  - name: service-name
  metrics:
  - name: success-rate
    interval: 5m
    successCondition: result >= 0.95
    provider:
      wavefront:
        address: example.wavefront.com
```

Given the DataDog and Wavefront examples, which approach would NewRelic more closely follow?

pkg/apis/rollouts/v1alpha1/analysis_types.go
jwelch92: Thanks I'll get that fixed up!

docs/features/analysis.md
jwelch92: Thanks for the detailed feedback @jessesuen!

Based on your feedback I am working on updating this PR with changes that will more closely follow the pattern used by DataDog. New Relic API keys are specific to an account ID and accounts are specific to a region of New Relic. Given that, the model of putting all of the configuration into the secret to be a global config makes a lot more sense.

I am adding support for multiple secrets out of the gate because we have the need to support multiple accounts within a given cluster. It's fairly common for New Relic users to have primary accounts with sub-accounts that each have their own monitored applications and separate API keys.

jwelch92:
Updated the New Relic configuration to more closely match other consumers of the [newrelic-client-go](https://github.com/newrelic/newrelic-client-go) library like the [newrelic-cli](https://github.com/newrelic/newrelic-cli) which uses profiles for credentials. 

```
newrelic profile add --help
Add a new profile

The add command creates a new profile for use with the New Relic CLI.
API key and region are required. An Insights insert key is optional, but required
for posting custom events with the `newrelic events`command.

Usage:
  newrelic profile add [flags]

Examples:
newrelic profile add --name <profileName> --region <region> --apiKey <apiKey> --insightsInsertKey <insightsInsertKey>

Flags:
      --apiKey string              your personal API key
  -h, --help                       help for add
      --insightsInsertKey string   your Insights insert key
  -n, --name string                unique profile name to add
  -r, --region string              the US or EU region

Global Flags:
      --format string   output text format [Text, YAML, JSON] (default "JSON")
      --plain           output compact text
```

jwelch92:
Hey @jessesuen can you please give this another look when you get a chance? I'll try to clear the new merge conflicts as soon as possible. Thanks!!

docs/features/analysis.md
```diff
@@ -920,3 +920,41 @@ data:
...
+  name: success-rate
+spec:
+  args:
+  - name: application-name
+  metrics:
+  - name: success-rate
+    successCondition: result.successRate >= 0.95
+    provider:
+      newRelic:
+        profileSecretName: my-newrelic-secret  # optional, defaults to 'newrelic'
```
jessesuen, line 941: I like the idea of profiles, which match the new relic CLI. However, I'd prefer to keep the fact that profiles are stored as secrets, an implementation detail from the spec. So can we rename this to just `profile` ?

jessesuen:
@jwelch92 this looks just about ready to go. The only change I think needs to be done is renaming `profileSecretName` to `profile`. 

In the future, this would free us to change the implementation of how profiles are stored, and the spec would still make sense. For example, newrelic profiles could be changed to be configured in a central configmap, and reference arbitrarily named secrets.

manifests/install.yaml
```diff
@@ -5647,8 +5661,6 @@ spec:
                     properties:
                       datadog:
                         properties:
-                          address:
```
jwelch92, line 5650: Hmm not sure why this diff is showing for datadog. I rebased upstream master and ran codegen and rebuilt manifests.

jwelch92:
Thanks again @jessesuen! Updated with your requested changes.

jwelch92:
Build failed on tests unrelated to my changes..is master broken right now? I rebased yesterday when I was fixing up my changes.

jessesuen:
Let me check. Heads up, I might commit to your fork/branch if it's something I can easily fix.

jessesuen (approved):

jessesuen:
Seems it just needed to resubmit and it passed.

Great work!

PR #780, opened 2020-10-14
fix(controller): blue-green with analysis was broken

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/774

The major change here is that we were using status.activeRS == currentPodHash to determine if rollout was fully promoted. However this is no longer true, because we now use status.stableRS == currentPodHash. 

fix: postPromotionAnalysis was broken
fix: blue-green fast-tracked rollbacks would still start analysis templates
fix: prePromotionAnalysis with previewReplicaCount did not start prePromotionAnalysis
feat(plugin): indicate the stable ReplicaSet for blue-green rollouts in plugin

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -98,26 +98,43 @@ KUBECONFIG=~/.kube/minikube make test-e2e
 (or test regex):
 
 ```shell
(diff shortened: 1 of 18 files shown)


khhirani (approved):
LGTM

Issue #777, opened 2020-10-13
Canary rollouts with a failed AnalysisRun immediately terminate all canary pods and ignores maxUnavailable


abatilo:
## Summary 

When doing a canary rollout with `maxUnavailable: 0`, if I have an `AnalysisRun` fail, all of the canary pods are immediately terminated even though the previous revision has no available pods.

## Diagnostics

Versions of all tools:
```
argo-rollouts 0.9.1
helm 3.2.1
kind 0.8.1
kubectl 1.18.2
```

Repro yaml can be found [here](https://gist.github.com/abatilo/e98186261db7e555f962c9657e9abd2c).

Steps to repro:
```
#!/bin/bash
kind create cluster
helm repo add argo https://argoproj.github.io/argo-helm
helm upgrade --install argo-rollouts argo/argo-rollouts --version 0.3.7
sleep 10
kubectl apply -f https://gist.githubusercontent.com/abatilo/e98186261db7e555f962c9657e9abd2c/raw/b98d5e779e428901af9a708eba3b63012e869069/argo-rollouts-issue-repro.yml
sleep 60
kubectl argo rollouts set image nginx nginx=nginx:1.19
```

The moment the `AnalysisRun` fails you'll see:
```
⇒  kubectl argo rollouts get rollout nginx

Name:            nginx
Namespace:       default
Status:          ✖ Degraded
Strategy:        Canary
  Step:          0/2
  SetWeight:     0
  ActualWeight:  0
Images:          nginx:1.18 (stable)
Replicas:
  Desired:       10
  Current:       10
  Updated:       0
  Ready:         1
  Available:     1

NAME                                                                KIND         STATUS               AGE  INFO
⟳ nginx                                                             Rollout      ✖ Degraded           97s
├──# revision:2
│  ├──⧉ nginx-56696d5c7b                                            ReplicaSet   • ScaledDown         37s  canary
│  │  ├──□ nginx-56696d5c7b-88k94                                   Pod          ◌ Terminating        37s  ready:1/1
│  │  ├──□ nginx-56696d5c7b-k757l                                   Pod          ◌ Terminating        37s  ready:1/1
│  │  ├──□ nginx-56696d5c7b-xkpbt                                   Pod          ◌ Terminating        37s  ready:1/1
│  │  ├──□ nginx-56696d5c7b-tjnt5                                   Pod          ◌ Terminating        29s  ready:1/1
[...]

PR #786, opened 2020-10-16
fix(controller): controller did not honor maxUnavailable during abort

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/777

The calculation change made in https://github.com/argoproj/argo-rollouts/pull/739, only accounted for the case when we were scaling up the canary, but not when we were scaling down the canary (e.g. in the case of abort). This caused a regression where we did not honor maxUnavailable during an abort.


khhirani (approved):
LGTM

Issue #771, opened 2020-10-07
Argo CD can fire PostSync hooks prematurely (before Rollout is Healthy)


ferwasy:
Hello. I'm using argo-cd version 1.6.2 with argo-rollouts version 0.7.2 to perform BlueGreen deployments. I've configured a bunch of PostSync hooks. My expectation is that these hooks should be triggered after the BlueGreen deploy finishes (the new version is already deployed, all the new pods have been created and the "active" service is pointing to the new version). However, I noticed that the hooks are triggered while the pods of the new application version are still being created.
Is this a bug or my expectation was incorrect? If my expectation is incorrect, how could I implement something like a PostSync job but for argo-rollouts?
Thanks in advance!

jessesuen:
> However, I noticed that the hooks are triggered while the pods of the new application version are still being created.
Is this a bug or my expectation was incorrect

This sounds like a bug, but given the versions, possibly addressed in later releases. It's affected by the health.lua script that is baked into Argo CD. The hooks should only fire once the rollout becomes healthy.

hoantran3108:
Hello. I'm using B/G deployment and struggling the same issue which needs to perform a cache purge after the new version successfully rollouts. Do we have any workaround for this? I'm using ArgoCD 1.7.5 and Argo Rollout 0.9.1.

Thanks

ferwasy:
@jessesuen and @hoantran3108 thank you very much for your quick responses. I was thinking in upgrading both ArgoCD and Argo Rollouts versions supposing that the new releases would contain the fix for this issue, but after @hoantran3108 comment, I'm starting to doubt about it. Any recommendations?

jessesuen:
Thanks for the data points. I will need some time to reproduce this.

ferwasy:
I've being doing some tests and I noticed that the PostSync job gets triggered BEFORE `status.blueGreen.activeSelector `is equals to `status.blueGreen.previewSelector `which is the condition for "healthy" state in `health.lua`

ferwasy:
We can confirmed that these also happens with ArgoCD version 1.7.7

ferwasy:
We found a workaround. We run the following bash script as the PostSync job with sync-wave: 1. That script waits until the preview service and the active service point to the same replica set (we noticed that when the first PostSync job is triggered, they point to different replica sets already). Once this script finishes, it means that the new rollout is healthy, so the following PostSync jobs (sync-wave: 2 and so forth) are triggered.
```

[...]

Issue #440, opened 2020-03-12
Remove .Status.Canary.StableRS


dthomson25:
We need to remove this field in v0.9.0 as we deprecated it with the changes in https://github.com/argoproj/argo-rollouts/issues/348

PR #789, opened 2020-10-19
feat(controller): use CRD status subresource

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/771
Resolves https://github.com/argoproj/argo-rollouts/issues/440

This change updates Argo Rollouts to enable the use of the [CRD status subresource](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource). The reason for doing so, is to allow the rollout controller to record the numeric `metadata.generation` into `status.observedGeneration`. This allows us to have a reliable indicator of a Rollout who's spec has (or has not yet) been observed by the controller (for example if the argo-rollouts controller was down). This is needed in situations when tooling needs to wait/block on a Rollout to become healthy, after making a spec change (e.g. Argo CD PostSync hook).

A huge consequence of this change, is that Rollouts can no longer update `spec/metadata` at the same time as `status`, since the latter requires a separate call to the rollout's `/status` endpoint. Most changes are in test expectations that currently expect updates on the resource and not resource/status. 

The main code change is in `rollout/sync.go`, which splits the update of a rollout into two API calls, and `promote.go` which also needs to make two API calls during promotion.

This change also:
* removes the deprecated `status.canary.stableRS` field.
* reorganizes installation manifests so that there is only one role file to update

TODO: 
* [x] make sure this works with rollouts with an invalid spec. (e.g. can we still persist rollouts that have invalid specs)


pkg/kubectl-argo-rollouts/info/rollout_info.go
```diff
@@ -122,10 +122,24 @@ func RolloutErrorConditions(ro *v1alpha1.Rollout) []string {
...
+	return int64(observedGen) == ro.Generation
+}
+
 // RolloutStatusString returns a status and message for a rollout
 // This logic is more or less the same as the Argo CD rollouts health.lua check
 // Any changes to this function should also be changed there
 func RolloutStatusString(ro *v1alpha1.Rollout) (string, string) {
+	if !isGenerationObserved(ro) {
+		return "Progressing", "waiting for rollout spec update to be observed"
+	}
```
jessesuen, line 142: This is the check that allows us to detect if rollout controller observed the new spec.

khhirani (approved):
LGTM

PR #797, opened 2020-10-20
feat: adding support for valueFrom in analysis arguments.

### LINKED ISSUE #778: Analysis Arguments support for valueFrom
# Summary
Currently Analysis step under Canary can only read arguments value from value , and it would be more dynamic to use valueFrom, to refer to metadata labels.

When would you use this?

In case of keeping majority of things in kustomize base , so that overlays can just be overriding or populating only things that are not standard, it would be much better to have feature to refer metadata labels etc using valueFrom and then adding a FieldRef to it. In this way , the kustomize base can be dynamic and this information will be propagated to overlays as well during kustomize build , and hence , leads to cleaner and non-repetitive overlays templates.

An example would be like below, in kustomize-base.
```
          analyses:
          - name: some-analysis
            templateName: some-analysis
            args:
            - name: app
              valueFrom:
                fieldRef:
                  fieldPath: metadata.labels['app']
            - name: env
              valueFrom:
                fieldRef:
                  fieldPath: metadata.labels['env']
            - name: region
              valueFrom:
                fieldRef:
                  fieldPath: metadata.labels['region']

```

---

inspiremirnal:
Checklist:

* [x ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x ] The title of the PR is "fix(valueFrom support): updated arguments to now resolve metadata. Fixes #778"`.  
* [x] I've signed the CLA.
* [x ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ x] My builds are green. Try syncing with master if they are not. 
* [ x] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).


utils/analysis/factory.go
```diff
@@ -4,24 +4,31 @@ import (
...
+			if arg.ValueFrom.PodTemplateHashValue != nil {
+				switch *arg.ValueFrom.PodTemplateHashValue {
+				case v1alpha1.Latest:
+					value = newRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
+				case v1alpha1.Stable:
+					value = stableRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
+				}
+			}else {
+				if arg.ValueFrom.FieldRef != nil {
+					value, _ = fieldpath.ExtractFieldPathAsString(r, arg.ValueFrom.FieldRef.FieldPath)
```
jessesuen, line 28: In order to safely ignore this error, we need to improve the validation library to make sure user is making valid references.

Can you improve pkg/apisrollouts/validation.ValidateRolloutSpec() to do this? If so, we can ignore this error.

jessesuen (changes requested):
Please fix linting issues by running `make lint`  and add documentation for the feature

rollout/controller_test.go
```diff
@@ -113,6 +113,10 @@ func newRollout(name string, replicas int, revisionHistoryLimit *int32, selector
 			Annotations: map[string]string{
 				annotations.RevisionAnnotation: "1",
 			},
+			Labels: map[string]string{
+				"app": "test",
+				"env": "test",
+			},
```
jessesuen, line 119: Why was this added? Doesn't appear to be used

docs/features/analysis.md
```diff
@@ -425,6 +425,48 @@ spec:
...
 ```
+Analysis arguments also support valueFrom for reading metadata fields and passing them as arguments to AnalysisTemplate.
+An example would be to reference metadata labels like env and region and passing them along to AnalysisTemplate.
+```yaml
+kind: Rollout
+metadata:
+  name: guestbook
+  env: test
+  region: us-west-2
```
jessesuen, line 436: These need to go under `metadata.labels` not directly under metadata

docs/features/analysis.md
```diff
@@ -425,6 +425,48 @@ spec:
...
+        - name: api-url
+          value: http://other-api
+        # pod template hash from the stable ReplicaSet
+        - name: stable-hash
+          valueFrom:
+            podTemplateHashValue: Stable
+        # pod template hash from the latest ReplicaSet
+        - name: latest-hash
+          valueFrom:
+            podTemplateHashValue: Latest
```
jessesuen, line 458: To simplify the documentation, I think we can omit these entirely since they do not highlight the feature.

docs/features/analysis.md
inspiremirnal: @jessesuen - you are right , this was a typo , i have resolved it.

docs/features/analysis.md
inspiremirnal: ok , good idea , i have removed this part as per your suggestion.

jessesuen:
I tried this out end-to-end, but it does not seem to be working.

```yaml
kind: AnalysisTemplate
metadata:
  name: sleep-job
spec:
  args:
  - name: duration
    value: 0s
  - name: exit-code
    value: "0"
  metrics:
  - name: sleep-job
    provider:
      job:
        spec:
          template:
            spec:
              containers:
              - name: sleep-job
                image: nginx:1.19-alpine
                command: [sh, -c, -x]
                args: ["sleep {{args.duration}} && exit {{args.exit-code}}"]
              restartPolicy: Never
          backoffLimit: 0
```

```yaml
kind: Rollout
metadata:
  name: rollout-inline-analysis
  metadata:
    labels:
      duration: 600s
spec:
  strategy:
    canary:
      steps:
      - setWeight: 10
      - analysis:
          templates:
          - templateName: sleep-job
          args:
          - name: duration
            valueFrom:
              fieldRef:
                fieldPath: metadata.labels['duration']
  selector:
    matchLabels:
      app: rollout-inline-analysis
  template:
    metadata:
      labels:
        app: rollout-inline-analysis
    spec:
      containers:
      - name: rollouts-demo
        image: nginx:1.19-alpine
        resources:
          requests:
            memory: 16Mi
            cpu: 5m
```

The resulting job did not have the value substituted:

```yaml
    spec:
      containers:
      - args:
        - sleep  && exit 0
        command:
        - sh
        - -c
        - -x
```

jessesuen:
Great work!

Issue #796, opened 2020-10-20
SecurityPrivileged Forbidden


shiransky:
## Summary 

When upgrading to argo rollouts v0.9.2 we encountered the following issue:

Invalid: spec.template.spec.initContainers[0].securityPrivileged: Forbidden: disallowed by cluster policy

in the version 0.7.2 we didn't have this issue

From the log:
time="2020-10-20T12:01:31Z" level=info msg="Started syncing rollout at (2020-10-20 12:01:31.117536654 +0000 UTC m=+1.199094785)" namespace=default
rollout=knginx
time="2020-10-20T12:01:31Z" level=error msg="The Rollout \"knginx\" is invalid: spec.template.spec.initContainers[0].securityContext.privileged: Fo
rbidden: disallowed by cluster policy" namespace=default rollout=knginx
time="2020-10-20T12:01:31Z" level=info msg="Reconciliation completed" namespace=default rollout=knginx time_ms=1.8316409999999999

jessesuen:
Can you share your rollout?

jessesuen:
This is almost certainly a bug with your Rollout spec. Rollouts v0.9. added validation of specs, which is the error you are seeing now. We re-use the validation from Deployments, so if you were to convert your rollout to deployment, you should get the same error.

jessesuen:
> This is almost certainly a bug with your Rollout spec 

I've looked into this more, and I see a place where the validation library that we use in Kubernetes disallow privileged by default. Since we are only using this for the context of validation, rollouts should allow this. I think this is a bug.

jessesuen:
@khhirani can you take a look at this? The issue is here:

https://github.com/kubernetes/kubernetes/blob/b46612a74224b0871a97dae819f5fb3a1763d0b9/pkg/apis/core/validation/validation.go#L5785-L5789

The validation library which we reuse calls `capabilities.Get()` but Get() will always return false unless initialized with a library call.

To reproduce this, you need to deploy a Rollout which sets `spec.template.spec.containers[].securityContext.privileged`.

jessesuen:
@khhirani nevermind i have a fix for this

jessesuen:
@shiransky the fix is now in an official release v0.9.3

PR #802, opened 2020-10-24
fix(controller): validation should not consider securityContext.privileged

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/796


khhirani (approved):
LGTM

Issue #424, opened 2020-03-02
Dependency License Issue


mattnaides:
The project depends on a project with a rough license for testing.  Would it be possible to remove this dependency?

"Copyright Bouke van der Bijl

I do not give anyone permissions to use this tool for any purpose. Don't use it."

https://github.com/bouk/monkey/blob/v1.0.0/LICENSE.md

PR #803, opened 2020-10-24
chore: migrate off bouk/monkey to go-mpatch

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/424


khhirani (approved):
I see some warnings for `Incorrect conversion between integer types`, but besides that, LGTM

PR #809, opened 2020-10-29
fix: rollout kustomize transform analysis ref should use templateName instead of name

jessesuen:
The rollout-transform was incorrect and didn't match our type definitions. AnalysisTemplates reference use the field `templateName:` and not `name:`


khhirani (approved):
LGTM

Issue #808, opened 2020-10-28
kubectl plugin create should create objects using dynamic client


jessesuen:
## Summary 

Currently the plugin is tightly coupled with a specific rollout version because it unmarshals files into the typed objects (which might be out-of-date from the version running in the cluster.

For example, if the controller is running a newer version of rollouts, and the plugin is outdated, during a create the plugin will drop all unknown fields which are not known by the plugin.

We should instead use the dynamic client and deal with unstructured objects. This way, an older plugin will still be forwards compatible with newer rollout versions

## Diagnostics

What version of Argo Rollouts are you running?

v0.9.x

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

khhirani:
@jessesuen Should the plugin use the dynamic client only for object creation, or should the dynamic client replace the RO client entirely?

jessesuen:
> @jessesuen Should the plugin use the dynamic client only for object creation, or should the dynamic client replace the RO client entirely?

It should do it for wherever it could cause problem of dropping fields. For example, the patches that happen in promote, abort, restart do not need to switch, since they have no risk of dropping fields (patches will not drop fields) 

However, creating an analysis run from a template, or local file has risk of dropping unknown fields, as does create.

khhirani:
> However, creating an analysis run from a template, or local file has risk of dropping unknown fields, as does create.

Just to clarify, only creation has the risk of dropping unknown fields?

PR #812, opened 2020-11-02
feat: Add undo command in kubectl plugin. Fixes #575

### LINKED ISSUE #575: kubectl argo rollouts undo rollout ROLLOUTNAME
Hi, I've been progressing a PoC of Argo-rollouts.

If I want to roll back to the specific revision, how can I do it?
I cannot find a solution to resolve that.
Is it the only way to just apply the manifest of the specific revision?

In case of using `Deployment`, it is so easy: `kubectl rollout undo deploy $deploy_name --to-revision $specific_version`
I've tried this for `Rollout`, but it wasn't supported.

hi1280:
I created `kubectl argo rollouts undo` command in the same way as `kubectl rollout undo` command.
I referred to the following program.
https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollback.go

Refs #575 

Checklist:

* [x] I've signed the CLA.

### DIFF
--- a/pkg/kubectl-argo-rollouts/cmd/cmd.go
+++ b/pkg/kubectl-argo-rollouts/cmd/cmd.go
@@ -13,6 +13,7 @@ import (
 	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/cmd/retry"
 	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/cmd/set"
 	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/cmd/terminate"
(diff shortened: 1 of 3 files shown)


jessesuen (commented):
This is great!

Do you think it would be possible to switch to using the dynamic client instead of the rollout client? The reason is we want to make the plugin compatible with future versions of rollouts. See:

https://github.com/argoproj/argo-rollouts/issues/808

If not, we can address it later as part of https://github.com/argoproj/argo-rollouts/issues/808

hi1280:
@jessesuen I changed to use dynamic client.

jessesuen (approved):

jessesuen:
Appreciate the unit tests!

Issue #754, opened 2020-10-02
Faster InvalidSpec recovery


jessesuen:
## Summary 

1. Apply a Rollout with InvalidSpec because it references a service that doesnt exist
2. Apply the missing service

Rollout will not reconcile until much later (resync period) before clearing the InvalidSpec condition. Ideally when the service is applied, the Rollout will recover immediately.

## Diagnostics

What version of Argo Rollouts are you running? v0.9.1

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

Issue #713, opened 2020-09-17
No InvalidSpec condition when referencing a missing analysis template


jessesuen:
## Summary 

There needs to be an validation check failure when a rollout references an analysis template which doesn't exist

## Diagnostics

What version of Argo Rollouts are you running? v0.9.0

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Looks like we have checks for everything but background analysis:

```
const (
	PrePromotionAnalysis  AnalysisTemplateType = "PrePromotionAnalysis"
	PostPromotionAnalysis AnalysisTemplateType = "PostPromotionAnalysis"
	CanaryStep            AnalysisTemplateType = "CanaryStep"
)

```

jessesuen:
Fixed

PR #814, opened 2020-11-03
fix: background analysis refs were not verified. requeue InvalidSpec rollouts

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/754, https://github.com/argoproj/argo-rollouts/issues/713

1. When a Rollout has an InvalidSpec, we will requeue the Rollout every ~20 seconds to detect if the situation has fixed itself (e.g. the referenced Service was created).

2. BackgroundAnalysis was not being verified

3. Rename CanaryStep to InlineAnalysis


khhirani (approved):
LGTM

Issue #573, opened 2020-07-08
promote --skip-all-steps doesn't work on blue/green


jessesuen:
The promote flag `--skip-all-steps` which is intended to do a full promotion to skip analysis and steps, does not support blue-green. 

Also, perhaps we should rename the flag to `--full`.

PR #828, opened 2020-11-09
feat(controller): add full rollout promotion (skip analysis, pause, steps)

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/573

Introduces the `--full` flag to the `promote` command, for fully promoting a rollout which is in the middle of update, including blue-green rollouts. Adds controller changes to support it. Fully promoting a rollout will cause the rollout to ignore Abort and Pause conditions, steps, and skip analysis. 

```shell
kubectl argo rollouts promote --full ROLLOUT
```

Implementation details: full promotion is accomplished by introducing:

```yaml
status:
  promoteFull: true
```

which the plugin (and eventually Argo CD) will set to true when full promotion is desired. The controller understands when this flag is set, and skips abort/pause/analysis to get to the new version ASAP.

### DIFF
--- a/analysis/controller_test.go
+++ b/analysis/controller_test.go
@@ -174,7 +174,7 @@ func (f *fixture) runController(analysisRunName string, startInformers bool, exp
 	}
 
 	if len(f.actions) > len(actions) {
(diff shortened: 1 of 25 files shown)


khhirani (approved):
LGTM

PR #829, opened 2020-11-09
fix: (controller boot) Do not cache all cluster Secrets - fetch them on demand. Fixes #798

### LINKED ISSUE #798: do not attempt to cache secrets for analysis templates
## Summary 

My colleague (@mrak) and I expect the Argo Rollouts controller to successfully boot. In one our Kubernetes clusters, we have over 600MB of secrets and the Controller is unable to succeed at booting.

Issue #420 is where this functionality was introduced.

Given that we can just query for specific secrets at resolveArgs time, we want to grab these lazily, not cache everything.

We are planning to go do this now, so we can use Argo Rollouts in this cluster. We'll open a PR and reference this bug.

## Diagnostics

0.9.2.

```

The logs aren't useful -- they just show you never get past "Waiting for controller's informer caches to sync"
```

---

mrak:
Using a list cache of secrets is not needed as we do not need to react
to any changes and we can always fetch the individual secrets we need
for analysis arguments when the time comes. For large kubernetes
clusters with thousands of secrets this causes the argo-rollouts
controller to be unbootable.

Checklist:

* [x] Either ~(a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community,~ (b) this is a bug fix, ~or (c) this is a chore.~
* [ ] I've signed the CLA.
* [x] This file does not exist ~My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).~

### DIFF
--- a/analysis/analysis.go
+++ b/analysis/analysis.go
@@ -1,6 +1,7 @@
 package analysis
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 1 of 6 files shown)


jessesuen (approved):
Thank you!

PR #834, opened 2020-11-11
fix: kubectl plugin should use dynamic client

### LINKED ISSUE #808: kubectl plugin create should create objects using dynamic client

khhirani:
Closes #808 

Signed-off-by: Kareena Hirani kareena_hirani@intuit.com


pkg/kubectl-argo-rollouts/cmd/create/create.go
```diff
@@ -290,11 +290,17 @@ func NewCmdCreateAnalysisRun(o *options.ArgoRolloutsOptions) *cobra.Command {
...
 				return err
 			}
-			fmt.Fprintf(createOptions.Out, "analysisrun.argoproj.io/%s created\n", created.Name)
+			var un unstructured.Unstructured
+			unmarshal(runBytes, un)
+			_, err = createOptions.DynamicClient.Resource(v1alpha1.AnalysisRunGVR).Namespace(ns).Create(ctx, &un, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(createOptions.Out, "analysisrun.argoproj.io/%s created\n", run.Name)
```
khhirani, line 303: @jessesuen Did I convert the AnalysisRun to an unstructured object correctly?

jessesuen (changes requested):
`analysisutil.NewAnalysisRunFromClusterTemplate` and `analysisutil.NewAnalysisRunFromTemplate` are two more instances where we will drop fields because we are getting a version of AnalysisRun which is tied to the client. We'll need a version of that which deals with, and returns an unstructured object, that the CLI will use.

khhirani:
@jessesuen Would I convert the AnalysisTemplate/ClusterAnalysisTemplate to an Unstructured object, make changes to it, and then create it?

utils/analysis/helpers.go
```diff
@@ -328,6 +329,41 @@ func flattenMetrics(templates []*v1alpha1.AnalysisTemplate, clusterTemplates []*
...
+
+	// Change kind to AnalysisRun
+	unstructured.SetNestedField(obj.Object, "AnalysisRun", "kind")
+
+	unstructured.SetNestedField(obj.Object, name, "metadata", "name")
+	unstructured.SetNestedField(obj.Object, generateName, "metadata", "generateName")
+	unstructured.SetNestedField(obj.Object, namespace, "metadata", "namespace")
+
+	// Set args
+	unstructured.SetNestedField(obj.Object, newArgs, "spec", "args")
```
khhirani, line 362: @jessesuen This line is causing an error: ```panic: cannot deep copy *[]v1alpha1.Argument```

jessesuen: yea you have to convert []v1alpha1.Argument to map[string]interface

jessesuen (approved):

PR #844, opened 2020-11-14
fix: controller was not given privileges to update status subresource

jessesuen:


khhirani (approved):
LGTM

Issue #845, opened 2020-11-14
Restarting a single replica Rollout in middle of canary should only restart 1 pod


jessesuen:
## Summary 

I had a rollout (replicas: 1) which was paused during an update (1 x canary + 1 x stable), and performed a restart. The controller restarted both pods at the same time but it should only restart one:

## Diagnostics

Version v0.10.0

# Paste the logs from the rollout controller

```
time="2020-11-14T11:43:31Z" level=info msg="Reconcile pod restart (replicas:1, available:2, maxUnavailable:0, effectiveMinAvailable:0, concurrentRestart:1, canRestart:2)" Reconciler=PodRestarter namespace=istio-rollout rollout=istio-rollout
time="2020-11-14T11:43:31Z" level=info msg="restarting Pod that's older than restartAt Time" CreatedAt="2020-11-14T11:35:11Z" Pod=istio-rollout-67c9d9c96-w7zm4 Reconciler=PodRestarter RestartAt="2020-11-14T11:43:31Z" namespace=istio-rollout rollout=istio-rollout
time="2020-11-14T11:43:31Z" level=info msg="restarting Pod that's older than restartAt Time" CreatedAt="2020-11-14T11:36:03Z" Pod=istio-rollout-bbbc78968-lfrrl Reconciler=PodRestarter RestartAt="2020-11-14T11:43:31Z" namespace=istio-rollout rollout=istio-rollout
time="2020-11-14T11:43:31Z" level=info msg="all 2 pods are current. setting restartedAt" Reconciler=PodRestarter namespace=istio-rollout rollout=istio-rollout
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #856, opened 2020-11-18
fix: restart was restaring too much when available > spec.replicas

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/845


jessesuen:
Found a bug with implementation. Still working on it

khhirani (approved):
LGTM

PR #875, opened 2020-12-03
fix: plugin incorrectly treated v0.9 rollout as v0.10 when it had numeric observedGeneration

jessesuen:
Checking for numeric observedGeneration is not a reliable indicator of a v0.9 rollout. We also have to check if that observedGeneration is > metadata.generation, since that would indicate that `metadata.generation` isn't being used as observedGeneration.


khhirani (approved):
LGTM

PR #818, opened 2020-11-05
fix: namespaced scoped controller support

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/712
Resolves https://github.com/argoproj/argo-rollouts/issues/334

Fixes support for the namespaced install of argo-rollouts by introducing `--namespaced` flag to controller and conditionally starting the clusteranalysistemplate informer.

Signed-off-by: Jesse Suen <jesse_suen@intuit.com>

PR #877, opened 2020-12-03
fix(manifests): Add missing RoleBinding file.

tdongsi:
Signed-off-by: tdongsi <dongsi.tuecuong@gmail.com>

I believe the RoleBinding is missing and want to create a quick fix.

Checklist:

* [x] this is a bug fix.
* [ ] I've signed the CLA.

### DIFF
--- a/manifests/namespace-install/kustomization.yaml
+++ b/manifests/namespace-install/kustomization.yaml
@@ -5,6 +5,7 @@ bases:
 - ../crds
 - ../base
 - ../role
+- argo-rollouts-rolebinding.yaml
 
 patchesStrategicMerge:
 - add-namespaced-flag.yaml


tdongsi:
I do have the regenerated `namespace-install.yaml` but I didn't check in. 
The `make manifests` does not work with the latest script `hack/update-manifests.sh`. Note that PR #818 uses

```
kustomize build --load_restrictor none "${SRCROOT}/manifests/namespace-install" >> "${SRCROOT}/manifests/namespace-install.yaml"
```

which would work for my change. I believe `kubectl kustomize` is a couple versions behind the standalone "kustomize build". 

@jessesuen FYI.

manifests/namespace-install/kustomization.yaml
```diff
@@ -5,6 +5,7 @@ bases:
 - ../crds
 - ../base
 - ../role
+- argo-rollouts-rolebinding.yaml
```
jessesuen, line 8: @tdongsi this should actually be under resources.

Also, you will need to run `make manifests` after the kustomization.yaml change.

jessesuen:
I rebased your branch and fixed the kustomization.yaml in https://github.com/argoproj/argo-rollouts/pull/899 so we can close this.

PR #899, opened 2020-12-14
fix(manifests): Add missing RoleBinding file

jessesuen:
This supersedes https://github.com/argoproj/argo-rollouts/pull/877 and fixes update-manifests.sh to use `kubectl kustomize` to generate output for consistency.


khhirani (approved):
LGTM

PR #901, opened 2020-12-15
feat: metric fields can be parameterized by analysis arguments

### LINKED ISSUE #825: Make AnalysisTemplate more configurable
# Summary

There are a number of fields in AnalysisTemplate/ClusterAnalysisTemplate that are static (`consecutiveErrorLimit`, `interval`, `count`, etc). It would mean a lot of practical benefits to make them dynamically configurable.

# Use Cases
Having a single AnalysisTemplate you could define different threasholds for different applications (since they may have different tolerance to failures). The preferred experience IMO would be something like
```yaml
postPromotionAnalysis:
    templates:
    - templateName: error-rate
      failureLimit: 3
      interval: 5m
      count: 10
```

---

khhirani:
Closes #825 


jessesuen (changes requested):
I see you made changes to all of the types, but I don't see any tests which actually test the feature. Where is a test which actually does substitution for a metric like:

```yaml
  metrics:
  - name: error-rate
    provider:
      prometheus:
        interval: "{{args.interval}}"
        count: "{{args.count}}"
```

jessesuen:
Also, an e2e would be nice, even if it is just modifying an existing one.

jessesuen:
@khhirani you'll need to rebase on master to fix the codegen problem. I found a toolchain discrepancy between macOS kubectl vs. github's ubuntu-latest kubectl that causes the difference in manifests. I switched to using `kustomize build` instead of `kubectl kustomize`

khhirani:
@jessesuen Fixed

docs/features/bluegreen.md
```diff
@@ -57,14 +57,14 @@ spec:
...
       antiAffinity: object
       previewService: string
       prePromotionAnalysis: object
       postPromotionAnalysis: object
-      previewReplicaCount: *int32
-      scaleDownDelaySeconds: *int32
-      scaleDownDelayRevisionLimit: *int32
+      previewReplicaCount: intOrString
+      scaleDownDelaySeconds: intOrString
+      scaleDownDelayRevisionLimit: intOrString
```
jessesuen, line 67: This is wrong -- you didn't change these types

utils/defaults/defaults.go
```diff
@@ -111,8 +111,8 @@ func GetAutoPromotionEnabledOrDefault(rollout *v1alpha1.Rollout) bool {
 }
 
 func GetConsecutiveErrorLimitOrDefault(metric *v1alpha1.Metric) int32 {
-	if metric.ConsecutiveErrorLimit != nil {
-		return *metric.ConsecutiveErrorLimit
+	if &metric.ConsecutiveErrorLimit != nil && metric.ConsecutiveErrorLimit.IntValue() > 0 {
+		return int32(metric.ConsecutiveErrorLimit.IntValue())
```
jessesuen, line 115: I think this is a bug. Doesn't this logic mean you are not allowing for `consecutiveErrorLimit: 0` ? Why was the check added?

rollout/analysis_test.go
```diff
@@ -21,14 +21,16 @@ import (
...
 	return &v1alpha1.AnalysisTemplate{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: metav1.NamespaceDefault,
 		},
 		Spec: v1alpha1.AnalysisTemplateSpec{
 			Metrics: []v1alpha1.Metric{{
-				Name: "example",
+				Name:  "example",
+				Count: &count,
```
jessesuen, line 33: This seems to be changing the base template used by the tests. I would expect tests to behave differently because of this change because `count: nil` is different than `count: 0`.  Is this change necessary for some reason?

analysis/analysis_test.go
```diff
@@ -1325,39 +1332,19 @@ func TestSecretNotFound(t *testing.T) {
...
-	tasks := []metricTask{{
-		metric: v1alpha1.Metric{
-			Name:             "metric-name",
-			SuccessCondition: "{{args.secret-empty}}",
-		},
-		incompleteMeasurement: nil,
-	}}
-	_, _, err := c.resolveArgs(tasks, args, metav1.NamespaceDefault)
-	assert.Equal(t, "arg 'secret-empty' does not contain a secret reference", err.Error())
-}
```
jessesuen, line 1359: What makes this test not valid anymore?

analysis/analysis_test.go
```diff
@@ -1302,6 +1308,7 @@ func TestSecretContentReferenceAndMultipleArgResolutionSuccess(t *testing.T) {
 						}},
 					},
 				},
+				Count: &count,
```
jessesuen, line 1311: Why can't count be nil?

jessesuen (changes requested):
I see that everywhere where count used to be nil, is now doing `count := intstr.FromInt(0)` and I'd like to understand why, since nil and zero mean different things.

utils/defaults/defaults.go
khhirani: That was my mistake, I had a pointer error in `analysis_types.go`. Fixed now.

analysis/analysis_test.go
khhirani: I modified the logic `analysis.go` so this error no longer occurs.

docs/features/bluegreen.md
```diff
@@ -57,7 +57,7 @@ spec:
   strategy:
     blueGreen:
       autoPromotionEnabled: boolean
-      autoPromotionSeconds: *int32
+      autoPromotionSeconds: intOrString
```
jessesuen, line 60: I don't think you changed this to intOrString

jessesuen (approved):
LGTM, please revert unintentional change to bluegreen.md before merging

Issue #883, opened 2020-12-08
Missing labels and annotations after last step of canary rollout


Yorkerrr:
## Summary 

I want to implement a canary release with istio as a traffic manager. Everything works as expected but in a result replica-set istio-related annotations and labels are missing. Here is a diff between the initial pod manifest and a final one:

<img width="960" alt="Screen Shot 2020-12-07 at 5 42 20 PM" src="https://user-images.githubusercontent.com/1927829/101427359-90870000-38b3-11eb-84d3-6b8eed8a81a6.png">

## Diagnostics

argo-rollouts:v0.10.1
argocd:v1.7.9
istio:v1.7.3

```
spec:
  strategy:
    canary:
      canaryMetadata:
        labels:
          role: canary
      stableMetadata:
        labels:
          role: stable
      analysis:
        templates:
         - templateName: success-rate
        startingStep: 2
        args:
        - name: service-name
          value: demo-canary.canary.svc.cluster.local
      steps:
      - setWeight: 20
      - pause:
          duration: "15s"
      - setWeight: 50
      - pause:
          duration: "1m"
      canaryService: demo-canary
      stableService: demo
      trafficRouting:
        istio:
           virtualService: 
            name: istio-vs
            routes:
            - primary 
```

# Paste the logs from the rollout controller
```
2020-12-08T01:35:46.541924016Z time="2020-12-08T01:35:46Z" level=info msg="Started syncing rollout at (2020-12-08 01:35:46.54067444 +0000 UTC m=+9900.077589128)" namespace=canary rollout=demo
2020-12-08T01:35:46.543161892Z time="2020-12-08T01:35:46Z" level=info msg="Reconciling TrafficRouting with type 'Istio'" namespace=canary rollout=demo
2020-12-08T01:35:46.543174592Z time="2020-12-08T01:35:46Z" level=info msg="Reconciling StableRS" namespace=canary rollout=demo
2020-12-08T01:35:46.543180793Z time="2020-12-08T01:35:46Z" level=info msg="No StableRS exists to reconcile or matches newRS" namespace=canary rollout=demo
2020-12-08T01:35:46.543186893Z time="2020-12-08T01:35:46Z" level=info msg="Reconciling new ReplicaSet 'demo-799cb5ffc6'" namespace=canary rollout=demo
2020-12-08T01:35:46.543192593Z time="2020-12-08T01:35:46Z" level=info msg="Reconciling old replica sets" namespace=canary rollout=demo
[...]

PR #906, opened 2020-12-15
fix: ephemeral metadata injection was dropping metadata injected by mutating webhooks

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/883

The ephemeral metadata injection feature did not account for mutating webhooks when it modified pods. The incorrect implementation was blindly copying the ReplicaSet's `spec.template.metadata` to the child Pods, not considering that the Pod could have had metadata injected by a mutating webhook (e.g. istio sidecar injector).

This change makes it such that when removing metadata, the rollout controller will only remove metadata that it had previously injected. 

### DIFF
--- a/docs/getting-started/setup/index.md
+++ b/docs/getting-started/setup/index.md
@@ -8,7 +8,9 @@ purposes.
 Some dependencies are installable via the Helm stable repository:
 
 ```shell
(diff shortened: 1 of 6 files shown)


khhirani (approved):
LGTM

Issue #878, opened 2020-12-03
requiredForCompletion does not work


jessesuen:
## Summary 

Tried using `requiredForCompletion` feature for rollout/experiment/analysis but my experiment kept running forever.

```yaml
spec:
  strategy:
    canary:
      steps:
      - experiment:
          duration: 10s
          templates:
          - name: baseline
            specRef: stable
          - name: canary
            specRef: canary
          analyses:
          - name : echo-pod-hash
            templateName: echo-pod-hash
            requiredForCompletion: true
```

## Diagnostics

What version of Argo Rollouts are you running? v0.10.0

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #907, opened 2020-12-15
fix: requiredForCompletion did not work for an experiment started by a rollout

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/878

The type definitions for a canary experiment step was missing the `requiredForCompletion` field, and so it was impossible to use this feature for an Experiment which was launched by a Rollout. The fix is to add the `requiredForCompletion` field to RolloutExperimentStepAnalysisTemplateRef, and carry over the value when creating the Experiment.

E2E test has been updated to exercise this feature. 


khhirani (approved):
LGTM

Issue #908, opened 2020-12-15
Pod restart feature is not honoring PDBs


jessesuen:
## Summary 

The restart pod feature uses normal Pod delete API, and not the evict API. This means that any pod disruption budget is not being honored.

## Diagnostics

What version of Argo Rollouts are you running? v0.10.1

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Fixed in v0.10.2 (now released)

jessesuen:
> @jessesuen Why should the pods be evicted instead of being deleted by honoring the maxUnavailable provided as part of .spec.strategy.(canary|blueGreen).maxUnavailable ?

Eviction is equivalent to a delete but honoring PDBs. My counter question is why would we want to bypass PDB restrictions and use delete instead of evict?

We do honor maxUnavailable as long as it is greater than 0. We cannot support 0 because restaring rollout is different than restarting Deployments (which is a rolling update). It needs to be different because we support restarting a Rollout at any point of an update, including in the middle of a paused canary/blue-green.

PR #910, opened 2020-12-16
fix: switch pod restart to use evict API to honor PDBs

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/908

This change switches the controller to use the evict API instead of the delete API for the pod restart feature. By using evict instead of delete, any PodDisruptionBudget targeting rollout pods will be honored.


khhirani (approved):
LGTM

PR #962, opened 2021-01-25
bugfix:argo create command not support namespace in yaml file

zhaoyun006:
modify before:
[root@k8s-master-1 argo-rollouts]# grep namespace examples/rollout-canary.yaml
  namespace: base
[root@k8s-master-1 argo-rollouts]# kubectl argo create -f  examples/rollout-canary.yaml
Error: the namespace of the provided object does not match the namespace sent on the request

modify after:
[root@k8s-master-1 argo-rollouts]# grep namespace examples/rollout-canary.yaml
  namespace: base
[root@k8s-master-1 argo-rollouts]# kubectl argo create -f  examples/rollout-canary.yaml
rollout.argoproj.io/rollout-canary created
[root@k8s-master-1 argo-rollouts]#

Signed-off-by: zhaozhiqiang <zhaozhiqiang@lixiang.com>

Checklist:

* [ ] I've signed the CLA.

### DIFF
--- a/pkg/kubectl-argo-rollouts/cmd/create/create.go
+++ b/pkg/kubectl-argo-rollouts/cmd/create/create.go
@@ -137,6 +137,20 @@ func unmarshal(fileBytes []byte, obj interface{}) error {
 	}
 }
 
+func (c *CreateOptions) getNamespace(un unstructured.Unstructured) string {
+	ns := c.ArgoRolloutsOptions.Namespace()
+	if md, ok := un.Object["metadata"]; ok {
+		if md == nil {
+			return ns
+		}
+		metadata := md.(map[string]interface{})
+		if internalns, ok := metadata["namespace"]; ok {
+			ns = internalns.(string)
+		}
+	}
+	return ns
+}
+
 func (c *CreateOptions) createResource(path string) (runtime.Object, error) {
 	ctx := context.TODO()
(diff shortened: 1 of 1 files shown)


zhaoyun006:
@zcc35357949

pkg/kubectl-argo-rollouts/cmd/create/create.go
```diff
@@ -137,6 +137,17 @@ func unmarshal(fileBytes []byte, obj interface{}) error {
 	}
 }
 
+func (c *CreateOptions) getNamespace(un unstructured.Unstructured) string {
+	ns := c.ArgoRolloutsOptions.Namespace()
+	if md, ok := un.Object["metadata"]; ok {
+		metadata := md.(map[string]interface{})
```
zcc35357949, line 143: may panic here if metadata is not a map in the resource file.

zcc35357949 (commented):
when we use "kubectl create -f xxx -n myns", if the namespace in file is different from flag，kubectl will return not match error.
I think "kubectl argo rollouts create -f" should return like so in that situation.

pkg/kubectl-argo-rollouts/cmd/create/create.go
zhaoyun006: not write matadata name is invalid resource file,
it's error

jessesuen (approved):

PR #974, opened 2021-02-05
feat(controller): Add support for  ephemeral metadata on BlueGreen rollouts. Fixes #973

### LINKED ISSUE #973: Add Ephemeral Metadata for Blue Green strategy
# Summary

Functionality for setting ephemeral metadata for Canary rollouts was proposed in #455 and implemented in #770. I would like to extend this functionality to the BlueGreen strategy as well. This would be done by adding `previewMetadata` and `activeMetadata` to the BlueGreen spec and using similar reconciliation logic to apply them to the replicasets and pods.

# Use Cases

Same use cases as ephemeral metadata for canary rollouts (i.e. tagging metrics with consistent labels). My specific use case is to tag Prometheus metrics with a label if the pod is active and use that for selecting custom HPA metrics.

---

cooperbenson-qz:
Checklist:

* [x] I've signed the CLA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,20 @@ For these reasons, in large scale high-volume production environments, a rolling
 To learn more about Argo Rollouts go to the [complete documentation](https://argoproj.github.io/argo-rollouts/).
 
 ## Who uses Argo Rollouts?
+Organizations below are **officially** using Argo Rollouts. Please send a PR with your organization name if you are using Argo Rollouts.
+
+1. [ADP](https://www.adp.com)
+1. [Intuit](https://www.intuit.com/)
+1. [Nozzle](https://nozzle.io)
+1. [PayPay](https://paypay.ne.jp/)
+1. [Twilio SendGrid](https://sendgrid.com)
+1. [Ubie](https://ubie.life/)
+1. [VISITS Technologies](https://visits.world/en)
+1. [Spotify](https://www.spotify.com/)
+1. [New Relic](https://newrelic.com/)
+1. [Quipper](https://www.quipper.com/)
+1. [Devtron Labs](https://github.com/devtron-labs/devtron)
+1. [Quizlet](https://quizlet.com)
 
 [Official Argo Rollouts User List](https://github.com/argoproj/argo-rollouts/blob/master/README.md)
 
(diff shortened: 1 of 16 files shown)


cooperbenson-qz:
Current status:
- Working on getting permission to sign the CLA from my org, should have that in an hour or so
- I'm a little worried that I've made the e2e tests flaky. I had to add sleeps to get them to work which seems suspect to me
- A documentation update is required, but I'm not sure how to go about it since the canary and blue-green terms are different and I don't want to add a bunch of redundant listing of terms

jessesuen:
This looks great!

> Working on getting permission to sign the CLA from my org, should have that in an hour or so

DCO is sufficient. We've already moved off the CLA and will be disabling the bot.

> I'm a little worried that I've made the e2e tests flaky. I had to add sleeps to get them to work which seems suspect to me

There is a ongoing issue with e2e tests flaking because of `waiting for rollout spec update to be observed` so I will ignore those failures.

> A documentation update is required, but I'm not sure how to go about it since the canary and blue-green terms are different and I don't want to add a bunch of redundant listing of terms

We often refer to it as: desired (aka active/preview) vs. stable. For documentation, I'm thinking we can replace the use of canary with desired, along with an explanation that desired means canary or preview depending on the strategy.

jessesuen:
The change looks straight forward and pretty much good to go. The e2e tests are not your fault so I think the only thing left to do is to update documentation.

Please add in the documentation markdown, a note saying that this feature is a v1.0 feature so that < 0.10 users won't be confused when it doesn't work. e.g.:

```
!!! important

    Blue-green ephemeral metadata available in v1.0
```

cooperbenson-qz:
Thanks for the suggestion for the documentation. Just pushed the update to the relevant docs page and took the opportunity to fix the link to USERS.md

jessesuen (approved):
Great job and appreciate the e2e test for this!

bobbyvacco:
> The change looks straight forward and pretty much good to go. The e2e tests are not your fault so I think the only thing left to do is to update documentation.
> 
> Please add in the documentation markdown, a note saying that this feature is a v1.0 feature so that < 0.10 users won't be confused when it doesn't work. e.g.:
> 
> ```
> !!! important
> 
>     Blue-green ephemeral metadata available in v1.0
> ```

is this meant to say available in v.0.10?

PR #977, opened 2021-02-05
feat: Allow user to handle NaN result in Analysis

### LINKED ISSUE #459: User controlled handling of NaN metric
I know NaN as treat as 'Inconclusive' after #180. 
But, I think the treat of NaN is very dependent on use cases.

- Some of the users may want NaN as treat as 'Inconclusive'.
- On the other hand, some other users may want NaN as treat as 0 or false...something else.

To allow users to handle NaN himself, 

- A default treatment of NaN should be override per 'AnalyzeTemplate'.
- 'utils/evaluate/evaluate.go' should export more math-related functions.

khhirani:
Closes #459


docs/features/analysis.md
```diff
@@ -643,3 +643,112 @@ spec:
           value: "Bearer {{ args.api-token }}" 
 ```
 
+## Handling Metric Results - NaN and Infinity
+Metric providers can sometimes return values of NaN (not a number) and infinity. Users can edit the `successCondition` and `failureCondition` fields
```
jessesuen, line 647: I think we should mention the functions in the docs. Can you explain the fact that we provide two math functions `isNaN` and `isInf` as a convenience to allow users to customize the handling of it and what they return (it may not be obvious to everyone). Also note that if not handled, a `NaN` result could result in Failed measurements.

```
!!! important

    available since v1.0
```

jessesuen (approved):
Conditional approval with improved docs

Bunny15738:
image

Function isNaN(result) giving error.

PR #981, opened 2021-02-05
fix: Set Canary Strategy default maxUnavailable to 25%

### LINKED ISSUE #938: Canary strategy is not defaulting maxUnavailable to 25%
## Summary 

The canary strategy is supposed to behave like a Deployment rollingUpdate strategy by defaulting both maxUnavailable and maxSurge to 25%.

While maxSurge does seem to be defaulting to 25%, maxUnavailable appears to be defaulting to 0.

## Diagnostics

v0.10.2 but I tested with v0.8.2 and v0.6.3 those versions also exhibit the same behavior.

---

khhirani:
Closes #938 

### DIFF
--- a/utils/defaults/defaults.go
+++ b/utils/defaults/defaults.go
@@ -14,7 +14,7 @@ const (
 	// DefaultMaxSurge default number for the max number of additional pods that can be brought up during a rollout
 	DefaultMaxSurge = "25"
 	// DefaultMaxUnavailable default number for the max number of unavailable pods during a rollout
-	DefaultMaxUnavailable = 0
+	DefaultMaxUnavailable = "25"
 	// DefaultProgressDeadlineSeconds default number of seconds for the rollout to be making progress
 	DefaultProgressDeadlineSeconds = int32(600)
 	// DefaultScaleDownDelaySeconds default seconds before scaling down old replicaset after switching services
@@ -57,7 +57,7 @@ func GetMaxUnavailableOrDefault(rollout *v1alpha1.Rollout) *intstr.IntOrString {
 	if rollout.Spec.Strategy.Canary != nil && rollout.Spec.Strategy.Canary.MaxUnavailable != nil {
 		return rollout.Spec.Strategy.Canary.MaxUnavailable
 	}
-	defaultValue := intstr.FromInt(DefaultMaxUnavailable)
+	defaultValue := intstr.FromString(DefaultMaxUnavailable)
 	return &defaultValue
 }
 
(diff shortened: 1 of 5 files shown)


jessesuen (commented):
Looks good. Can you update both the documentation and the types.go to reflect this?

utils/defaults/defaults.go
```diff
@@ -14,7 +14,7 @@ const (
 	// DefaultMaxSurge default number for the max number of additional pods that can be brought up during a rollout
 	DefaultMaxSurge = "25"
 	// DefaultMaxUnavailable default number for the max number of unavailable pods during a rollout
-	DefaultMaxUnavailable = 0
+	DefaultMaxUnavailable = "25"
```
jessesuen, line 17: Do you know why this is `25` and not `25%`. Can you verify that we aren't defaulting max unavailable to 25 pods instead of 25% of pods?

khhirani: I copied the format for MaxSurge, which also defaults to 25%. 
surge, err := intstrutil.GetValueFromIntOrPercent(intstrutil.ValueOrDefault(maxSurge, intstrutil.FromInt(0)), int(desired), true)
	if err != nil {
		return 0, 0, err
	}
	unavailable, err := intstrutil.GetValueFromIntOrPercent(intstrutil.ValueOrDefault(maxUnavailable, intstrutil.FromInt(0)), int(desired), false)
	if err != nil {
		return 0, 0, err
	}

Both of them are resolved

khhirani: The [code which resolves both MaxSurge and MaxUnavailable](https://github.com/argoproj/argo-rollouts/blob/master/utils/replicaset/replicaset.go#L366-L374) uses the method `intstrutil.GetValueFromIntOrPercent`. I confirmed that it works with some manual tests.

jessesuen: OK, thanks for confirming. I see that `GetValueFromIntOrPercent` interprets all string values as percentages and not integer values, even if the value is missing a `%` character:

```golang
func GetValueFromIntOrPercent(intOrPercent *IntOrString, total int, roundUp bool) (int, error) {
	if intOrPercent == nil {
		return 0, errors.New("nil value for IntOrString")
	}
	value, isPercent, err := getIntOrPercentValue(intOrPercent)
	if err != nil {
		return 0, fmt.Errorf("invalid value for IntOrString: %v", err)
	}
	if isPercent {
		if roundUp {
			value = int(math.Ceil(float64(value) * (float64(total)) / 100))
		} else {
			value = int(math.Floor(float64(value) * (float64(total)) / 100))
		}
	}
	return value, nil
}

func getIntOrPercentValue(intOrStr *IntOrString) (int, bool, error) {
	switch intOrStr.Type {
	case Int:
		return intOrStr.IntValue(), false, nil
	case String:
		s := strings.Replace(intOrStr.StrVal, "%", "", -1)
		v, err := strconv.Atoi(s)
		if err != nil {
			return 0, false, fmt.Errorf("invalid value %q: %v", intOrStr.StrVal, err)
		}
		return int(v), true, nil
	}
	return 0, false, fmt.Errorf("invalid type: neither int nor percentage")
}
```

jessesuen (approved):

Issue #617, opened 2020-08-05
Support Istio canary via multiple DestinationRule subsets (versus canary/stable services)


hexsel:
Currently, according to https://argoproj.github.io/argo-rollouts/features/traffic-management/istio/, Istio canary deployments require separate services for stable and canary.  This is undesirable, as I understand in-mesh communications won't hit the virtualservice directly (there's no DNS entry added by default for those in the Kubernetes cluster).  This forces microservice-to-microservice communication to choose whether to hit the stable service or the canary (hardcoded), or go to the gateway, or DNS entries to be added for the virtualservices, which is cumbersome and poorly documented.

All Istio canary examples I could find use a single service, but have a destinationrule that splits them into 2 groups based on label selectors (for canary vs stable, or per version), then the virtualservice links this single service and splits the traffic accordingly.

Maybe a canary rollout should follow this principle? Or I may be misunderstanding this problem and this can be fixed with some more documentation?

jessesuen:
> All Istio canary examples I could find use a single service, but have a destinationrule that splits them into 2 groups based on label selectors (for canary vs stable, or per version), then the virtualservice links this single service and splits the traffic accordingly.

@hexsel I'd like to understand your proposal a little better. Could you provide the alternative resulting DestinationRule / VirtualService that achieves the canary using that means?

jessesuen:
OK, my memory is a bit hazy (and I also might be wrong about this), but I think one of the motivations for current design was that two separate services were able to produce istio prometheus metrics that were not available with a single DestinationRule and VirtualService. In other words, it would be possible to graph and compare the metrics of a canary vs. stable service, but not two DestinationRule subsets.

jessesuen:
That said, I think it is possible to enhance the rollout istio canary implementation to reference a DestinationRule which would be updated by the controller with the canary/stable rollouts-pod-template-hash values.

hexsel:
Regarding the metrics issue you mentioned, I admit I do not know if it is possible, but I found this:
https://istio.io/latest/docs/tasks/observability/metrics/querying-metrics/

![image](https://user-images.githubusercontent.com/299385/92748168-df73d680-f352-11ea-957b-8e4f135c35e7.png)

Which _seems_ to indicate you could query metrics for different labels on the same virtual service.

jessesuen:
Thanks for providing the use case. I believe what you are saying checks out.

>Which seems to indicate you could query metrics for different labels on the same virtual service.

[...]

Issue #587, opened 2020-07-15
503  Service Unavailable when  canary release with istio traffic manage


TaylorHere:
``` yaml
# Source: demo/templates/rollout.yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: demo-rollout-deploy
spec:
  replicas: 1
  strategy:
    rollingUpdate:
      maxSurge: 100%
      maxUnavailable: 0
    type: RollingUpdate
  selector:
    matchLabels:
      app: demo
      release: demo
      app.kubernetes.io/name: demo
      app.kubernetes.io/instance: demo
  template:
    metadata:
      annotations:
        checksum/config: 819666c3c37d0b9a43595609f9bf286d3810e9e129472fdf6cb60cbacae6e1d3
      labels:
        app: demo
        release: demo
        app.kubernetes.io/name: demo
        app.kubernetes.io/instance: demo
    spec:
      imagePullSecrets:
        - name: lilith-image
      serviceAccountName: demo
      containers:
        - name: demo
          env:
            - name: LC_ALL
              value: en_US.utf-8
            - name: LANG
              value: en_US.utf-8
          securityContext:
            {}
          image: "demo:v1"
          imagePullPolicy: IfNotPresent
          ports:
            - name: http
              containerPort: 5000
              protocol: TCP
          livenessProbe:
            httpGet:
              path: /api/live
              port: 5000
          readinessProbe:
            httpGet:
              path: /api/ready
              port: 5000
          resources:
            limits:
              cpu: "1"
              memory: 256Mi
            requests:
              cpu: "1"
              memory: 256Mi
          volumeMounts:
            - mountPath: /app/anti_fraud_server/configs/
              name: demo
      volumes:
        - name: demo
          configMap:
              name: demo
  minReadySeconds: 10
  revisionHistoryLimit: 3
  progressDeadlineSeconds: 600
  strategy:
    canary:
      canaryService: demo-canary-svc
      stableService: demo-prod-svc
      maxUnavailable: 0
      maxSurge: 25%
      steps:
      - pause: {}
      - setWeight: 0
      - pause: {}
      - setWeight: 50
      - pause: {}
      - setWeight: 100
      trafficRouting:
        istio:
          virtualService: 
            name: demo
            routes:
            - demo-vpc
            - demo-public-debug

```

when I apply my rollout to k8s, a New ReplicaSet created, with istio vsvc traffic 100% to demo-prod-svc

[...]

PR #985, opened 2021-02-10
feat: support canarying using Istio DestinationRule subsets

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/617
Resolves https://github.com/argoproj/argo-rollouts/issues/587

This change makes it possible to canary on Istio using two DestinationRule subsets as an alternative to two Kubernetes Services, which opens the door to other, more flexible ways of canarying.

It also resolves https://github.com/argoproj/argo-rollouts/issues/587 by making no longer required to have two Kubernetes services point to the same set of pods, which caused 503s in Istio 1.7 (fixed in Istio 1.8).


khhirani (approved):
LGTM

PR #1001, opened 2021-02-22
feat: Implement rollout status command. Fixes #596

### LINKED ISSUE #596: [Feature Request] rollout status

PedroArvela:
Closes #596

This replicates the `kubectl rollout status deployment [--watch] <name>` command, but for Rollouts.

This is implemented as:

```
# Returns Rollout status: Healthy, Progressing, Degraded, etc
kubectl argo rollouts status --watch=false <name>

# Waits until status is Healthy or Degraded (default behaviour without flags)
kubectl argo rollouts status --watch <name>
```

The flags and defaults for them replicate the `kubectl rollout status deployment <name>` command. As in it, `--watch` is default and the default timeout is 5 minutes.

---

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, ~(b) this is a bug fix, or (c) this is a chore~.
  * (a) See #596

### DIFF
--- a/pkg/kubectl-argo-rollouts/cmd/status/status.go
+++ b/pkg/kubectl-argo-rollouts/cmd/status/status.go
@@ -12,9 +12,8 @@ import (
 )
 
 const (
-	statusLong = `Watch the progress of a rollout until it is done. The return will be success if
-the Rollout ends as Healthy and error if the Rollout ends degraded or if the
-timeout is hit before the progress finishes.`
+	statusLong = `Watch rollout until it finishes or the specified timeout is exceeded. Returns
+success if the rollout is healthy upon completion and an error otherwise.`
 	statusExample = `
 	# Watch the rollout until it succeeds
 	%[1]s status gestbook
--- a/pkg/kubectl-argo-rollouts/cmd/status/status.go
+++ b/pkg/kubectl-argo-rollouts/cmd/status/status.go
@@ -3,6 +3,7 @@ package status
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/info"
 	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/options"
@@ -40,7 +41,8 @@ func NewCmdStatus(o *options.ArgoRolloutsOptions) *cobra.Command {
 			}
 			name := args[0]
 			controller := viewcontroller.NewRolloutViewController(o.Namespace(), name, statusOptions.KubeClientset(), statusOptions.RolloutsClientset())
-			ctx := context.Background()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
 			controller.Start(ctx)
 
 			ri, err := controller.GetRolloutInfo()
@@ -48,7 +50,17 @@ func NewCmdStatus(o *options.ArgoRolloutsOptions) *cobra.Command {
 				return err
 			}
 
-			statusOptions.PrintStatus(ri)
+			if !statusOptions.Watch {
+				fmt.Println(ri.Status)
+			} else {
+				rolloutUpdates := make(chan *info.RolloutInfo)
+				defer close(rolloutUpdates)
+				controller.RegisterCallback(func(roInfo *info.RolloutInfo) {
+					rolloutUpdates <- roInfo
+				})
+				go statusOptions.WatchStatus(ctx.Done(), cancel, rolloutUpdates)
+				controller.Run(ctx)
+			}
 
 			return nil
 		},
@@ -57,6 +69,27 @@ func NewCmdStatus(o *options.ArgoRolloutsOptions) *cobra.Command {
 	return cmd
 }
 
-func (o *StatusOptions) PrintStatus(roInfo *info.RolloutInfo) {
-	fmt.Println(roInfo.Status)
+func (o *StatusOptions) WatchStatus(stopCh <-chan struct{}, cancelFunc context.CancelFunc, rolloutUpdates chan *info.RolloutInfo) {
+	ticker := time.NewTicker(time.Second)
+	var roInfo *info.RolloutInfo
+	var preventFlicker time.Time
+
+	for {
+		select {
+		case roInfo = <-rolloutUpdates:
+		case <-ticker.C:
+		case <-stopCh:
+			return
+		}
+		if roInfo != nil && time.Now().After(preventFlicker.Add(200*time.Millisecond)) {
+			fmt.Printf("%s - %s\n", roInfo.Status, roInfo.Message)
+
+			if roInfo.Status == "Healthy" || roInfo.Status == "Degraded" {
+				cancelFunc()
+				return
+			}
+
+			preventFlicker = time.Now()
+		}
+	}
 }
--- a/pkg/kubectl-argo-rollouts/cmd/status/status.go
+++ b/pkg/kubectl-argo-rollouts/cmd/status/status.go
@@ -12,13 +12,21 @@ import (
 )
 
 const (
-	statusExample     = ``
-	statusUsage       = ``
-	statusUsageCommon = ``
+	statusLong = `Watch the progress of a rollout until it is done. The return will be success if
+the Rollout ends as Healthy and error if the Rollout ends degraded or if the
+timeout is hit before the progress finishes.`
+	statusExample = `
+	# Watch the rollout until it succeeds
+	%[1]s status gestbook
+
+	# Watch the rollout until it succeeds, fail if it takes more than 60 seconds
+	%[1]s status --timeout 60 guestbook
+	`
 )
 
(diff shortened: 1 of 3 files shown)


pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/info"
+	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/options"
+	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/viewcontroller"
+	"github.com/spf13/cobra"
+)
+
+const (
+	statusLong = `Watch the progress of a rollout until it is done. The return will be success if
+the Rollout ends as Healthy and error if the Rollout ends degraded or if the
+timeout is hit before the progress finishes.`
```
jsoref, line 17: I'm not a fan of `the progress finishes`

Can you explain what you're trying to say?

Note that this uses `hit` as opposed to `within` (`hit` is vaguely ok, `within` as noted below isn't, although `hit` isn't really great)

pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+type StatusOptions struct {
+	Watch   bool
+	Timeout int64
+
+	options.ArgoRolloutsOptions
+}
+
+// NewCmdStatus returns a new instance of an `rollouts status` command
```
jsoref, line 34:
```suggestion
// NewCmdStatus returns a new instance of a `rollouts status` command
```

pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+				go statusOptions.WatchStatus(ctx.Done(), cancel, statusOptions.Timeout, rolloutUpdates)
+				controller.Run(ctx)
+
+				finalRi, err := controller.GetRolloutInfo()
+				if err != nil {
+					return err
+				}
+
+				if finalRi.Status == "Degraded" {
+					return fmt.Errorf("The rollout is in degraded state with message: %s", finalRi.Message)
```
jsoref, line 78:
```suggestion
					return fmt.Errorf("The rollout is in a degraded state with message: %s", finalRi.Message)
```

pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+				} else if finalRi.Status != "Healthy" {
+					return fmt.Errorf("The rollout progress did not finish within the specified timeout")
```
jsoref, line 80: I don't understand `progress` in this context (see my previous request about the same)

Also, you can't use `timeout` like this. Timeout is a fixed quantity, whereas `within` operates on an interval.

Possibly something like:
```suggestion
					return fmt.Errorf("Timeout reached before rollout completed")
```

pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,123 @@
...
+				}
+			}
+
+			return nil
+		},
+	}
+	cmd.Flags().BoolVarP(&statusOptions.Watch, "watch", "w", true, "Watch the status of the rollout until it's done")
+	cmd.Flags().Int64VarP(&statusOptions.Timeout, "timeout", "t", 300, "The length of time in seconds to wait before ending watch, zero means never.")
```
jsoref, line 88: Something like this?
```suggestion
	cmd.Flags().Int64VarP(&statusOptions.Timeout, "timeout", "t", 300, "The length of time in seconds to watch before giving up, zero means wait forever.")
```

pkg/kubectl-argo-rollouts/cmd/status/status.go
PedroArvela: That is indeed not ideally worded.

If the Rollout is still either in Progressing or Paused by the time the timeout is reached the commands returns with an error.

This is the same behaviour that happens in `kubectl rollout status` when a Deployment doesn't finish its rollout before the specified timeout.

jsoref: `Watch rollout until it finishes or if a timeout is specified the timeout is exceeded`?
`Returns success if the rollout is healthy upon completion and an error otherwise`?

pkg/kubectl-argo-rollouts/cmd/status/status.go
PedroArvela: Maybe using the same analogy as in `kubectl rollout status` might make sense here: https://github.com/kubernetes/kubectl/blob/e909d3918e23a879d24ca1fd624827037c9920a7/pkg/polymorphichelpers/rollout_status.go#L78

> deployment %q exceeded its progress deadline

pkg/kubectl-argo-rollouts/cmd/status/status.go
PedroArvela: There is a default timeout specified, but I think that could work:

> Watch rollout until it finishes or the specified timeout is exceeded. Returns success if the rollout is healthy upon completion and an error otherwise.

jsoref: This sounds pretty good, but, if there's a default timeout, then drop `specified`.

pkg/kubectl-argo-rollouts/cmd/status/status.go
jsoref: `exceeded` sounds good. Based on what you've written elsewhere, I don't think you can use `deadline`.

PedroArvela: The only thing I can think of is

> Rollout progress exceeded timeout

But I'm not sure if that is the best phrasing.

pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/argoproj/argo-rollouts/pkg/kubectl-argo-rollouts/viewcontroller"
+	"github.com/spf13/cobra"
+)
+
+const (
+	statusLong = `Watch rollout until it finishes or the timeout is exceeded. Returns success if
+the rollout is healthy upon completion and an error otherwise.`
+	statusExample = `
+	# Watch the rollout until it succeeds
+	%[1]s status gestbook
```
jsoref, line 19:
```suggestion
	%[1]s status guestbook
```
(I haven't contributed my spell checker to Argo yet, but it'd scream if this merged as is)

pkg/kubectl-argo-rollouts/cmd/status/status.go
```diff
@@ -0,0 +1,122 @@
...
+
+// NewCmdStatus returns a new instance of a `rollouts status` command
+func NewCmdStatus(o *options.ArgoRolloutsOptions) *cobra.Command {
+	statusOptions := StatusOptions{
+		ArgoRolloutsOptions: *o,
+	}
+
+	var cmd = &cobra.Command{
+		Use:          "status ROLLOUT_NAME",
+		Short:        "Show the status of a rollout.",
```
jsoref, line 41: Argo isn't entirely consistent here, but generally doesn't include a `.`:
```suggestion
		Short:        "Show the status of a rollout",
```

jsoref (changes requested):
With this, I think this is the best I can offer as feedback. Someone more core should be responsible for a code review & approval

keithmattix:
Is there anything you preventing this from being merged? This would be huge for my organization's CD pipeline

jessesuen:
> The flags and defaults for them replicate the kubectl rollout status deployment <name> command. As in it, --watch is default and the default timeout is 5 minutes.

It seems that kubectl defaults to at timeout of 0s (never). Instead of defaulting to 5m, can we replicate kubectl behavior?

```
$ kubectl rollout status --help
...
      --timeout=0s: The length of time to wait before ending watch, zero means never. Any other values should contain a
corresponding time unit (e.g. 1s, 2m, 3h).
```

PedroArvela:
> It seems that kubectl defaults to at timeout of 0s (never). Instead of defaulting to 5m, can we replicate kubectl behavior?

I must have misread when I saw the timeout default. I'll update it to replicate that default.

PedroArvela:
Fixed the timeout to match the official one in kubectl itself.

jessesuen (approved):

jessesuen:
Great work!

PR #1002, opened 2021-02-23
fix: Clear ProgressDeadlineExceeded Condition in paused BlueGreen Ro…

### LINKED ISSUE #939: ProgressDeadlineExceeded condition not cleared in a paused blueGreen rollout
## Summary 

When investigating a blueGreen rollout which was considered degraded because it had a ProgressDeadlineExceeded timeout condition, I noticed the pods were actually ready/available even though the condition was still set.

I think there may be a bug where the controller doesn't ever clear ProgressDeadlineExceeded once it reaches that condition.

## Diagnostics

What version of Argo Rollouts are you running? v0.10.2

---

khhirani:
Closes #939 


test/e2e/bluegreen_test.go
```diff
@@ -158,3 +158,121 @@ spec:
...
+		When().
+		ApplyManifests().
+		WaitForRolloutReplicas(2).
+		WaitForRolloutStatus("Healthy").
+		UpdateSpec().
+		WaitForRolloutStatus("Paused").
+		PromoteRollout().
+		WaitForRolloutStatus("Healthy").
+		Then().
+		ExpectActiveRevision("2")
```
jessesuen, line 219: Can the test be improved so that after UpdateSpec(), it waits for Degraded status and then Paused? We want to confirm it recovered from Degraded automatically.

test/e2e/bluegreen_test.go
```diff
@@ -158,3 +158,121 @@ spec:
...
 		ExpectRevisionPods("revision 2 has active metadata2", "2", podsHaveActiveMetadata2)
 }
+
+func (s *BlueGreenSuite) TestBlueGreenProgressDeadlineExceededWithPause() {
+	s.Given().
+		RolloutObjects(`
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
+  name: rollout-bluegreen
```
jessesuen, line 168: Can we use a unique name for the rollout which is descriptive for the test? It helps with grepping controller logs during debugging

test/e2e/bluegreen_test.go
```diff
@@ -158,3 +158,121 @@ spec:
...
+    matchLabels:
+      app: rollout-bluegreen
+  template:
+    metadata:
+      labels:
+        app: rollout-bluegreen
+    spec:
+      containers:
+      - name: rollouts-demo
+        image: argoproj/rollouts-demo:blue
```
jessesuen, line 183: Could we use the nginx image instead? All the other e2e tests use it and it helps with keeping the e2e tests running quicker since it does not need to pull images.

test/e2e/bluegreen_test.go
```diff
@@ -158,3 +158,121 @@ spec:
...
+        imagePullPolicy: Always
```
jessesuen, line 184: remove `imagePullPolicy: Always` since it slows down tests.

test/e2e/bluegreen_test.go
```diff
@@ -158,3 +158,121 @@ spec:
...
+        ports:
+        - containerPort: 8080
+        readinessProbe:
+          initialDelaySeconds: 20
```
jessesuen, line 188: Can we try: `progressDeadlineSeconds: 5` and `initialDelaySeconds: 10` to speed up the test?

test/e2e/bluegreen_test.go
```diff
@@ -158,3 +158,121 @@ spec:
...
+
+func (s *BlueGreenSuite) TestBlueGreenProgressDeadlineExceededWithPause() {
+	s.Given().
+		RolloutObjects(`
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
+  name: rollout-bluegreen
+spec:
+  replicas: 2
```
jessesuen, line 170: Remove replicas so we only need to start 1 container, so we can speed up the test.

test/e2e/bluegreen_test.go
khhirani: Is it possible for me to use a readiness probe when using the nginx image? I need to configure a readiness probe to set `initialDelaySeconds`.

jessesuen: Yes it just need to be modified to the port which nginx listens on instead of rollouts-demo

jessesuen (approved):

PR #1015, opened 2021-02-27
feat: Argo Rollouts api-server and UI

rbreeze:
Signed-off-by: Remington Breeze <remington@breeze.software>

Checklist:


pkg/apiclient/rollout/rollout.proto
```diff
@@ -0,0 +1,92 @@
...
+}
+
+message UndoQuery {
+    string rollout = 1;
+    int64 revision = 2;
+}
+
+message RolloutWatchEvent {
+    string type = 1;
+    github.com.argoproj.argo_rollouts.pkg.apis.rollouts.v1alpha1.RolloutInfo rolloutInfo = 2;
```
jessesuen, line 28: Instead of adding RolloutInfo can we add it to rollout.proto?

pkg/kubectl-argo-rollouts/cmd/get/get_rollout.go
```diff
@@ -66,12 +75,13 @@ func NewCmdGetRollout(o *options.ArgoRolloutsOptions) *cobra.Command {
 	}
 	cmd.Flags().BoolVarP(&getOptions.Watch, "watch", "w", false, "Watch live updates to the rollout")
 	cmd.Flags().BoolVar(&getOptions.NoColor, "no-color", false, "Do not colorize output")
+	cmd.Flags().IntVar(&getOptions.TimeoutSeconds, "timeout-seconds", -1, "Timeout after specified seconds")
```
jessesuen, line 78: Nit: usually a timeout of 0 (not -1) indicates no timeout.

server/server.go
```diff
@@ -0,0 +1,416 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	kubeinformers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	appslisters "k8s.io/client-go/listers/apps/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
```
jessesuen, line 44: Nit: can we follow golang standard and group argo-rollouts imports under 3rd party imports

pkg/apiclient/rollout/rollout.proto
```diff
@@ -0,0 +1,179 @@
...
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "google/protobuf/empty.proto";
+
+package rollout;
+
+message RolloutQuery {
+    string name = 1;
+}
+
+message SetImageQuery {
```
jessesuen, line 15: Since this is an action and not a get, I think this should be named `SetImageRequest` instead of `SetImageQuery`

pkg/apiclient/rollout/rollout.proto
```diff
@@ -0,0 +1,179 @@
...
+    string rollout = 1;
+    string container = 2;
+    string image = 3;
+    string tag = 4;
+}
+
+message UndoQuery {
```
jessesuen, line 22: Same for UndoQuery: `s/UndoQuery/UndoRequest/`

pkg/apiclient/rollout/rollout.proto
```diff
@@ -0,0 +1,179 @@
...
+        option (google.api.http).get = "/api/v1/rollout/set/{rollout}/{container}/{image=**}/{tag}";
+    }
+
+    rpc UndoRollout(UndoQuery) returns (google.protobuf.Empty) {
+        option (google.api.http).get = "/api/v1/rollout/undo/{rollout}/{revision}";
+    }
+
+    rpc RetryRollout(RolloutQuery) returns (google.protobuf.Empty) {
+        option (google.api.http).get = "/api/v1/rollout/retry/{name}";
+    }
```
jessesuen, line 174: 1. All of these mutation API's which currently use the same `RolloutQuery` object, need to have individual Request objects. e.g.

* PromoteRolloutRequest
* AbortRolloutRequest
* UndoRolloutRequest
* RetryRolloutRequest

This is because each of these APIs potentially will have their own options, and we cannot reuse the RolloutQuery object to have all the options for each endpoint.

2. To follow REST semantics, these should be PUT requests instead of GET, with the subfield coming after the rollout name. e.g. 

```
option (google.api.http).put = "/api/v1/rollout/{name}/abort";
```

3. Instead of returning a `google.protobuf.Empty`, it would be better to return the updated Rollout object with the modifications.

jessesuen (changes requested):

jessesuen (approved):
Great work!

Issue #1012, opened 2021-02-26
Wavefront needs longer lookback period


jessesuen:
## Summary 

Currently, our wavefront implementation makes a query using the start/end times using the time when the measurement is being made. However, this results in the problem where there may be no results in the time series, resulting in the error: `No TimeSeries found in response from Wavefront`.

We need to lower the start time in order to capture more data points in the series. Instead of start/end times using the measurement time, we could use now() - interval as a better default.

## Diagnostics

What version of Argo Rollouts are you running? v0.10.2

jessesuen:
After further testing & investigation, I discovered that the wavefront query parameter `SeriesOutsideTimeWindow` affects the results the most.

Here is the wavefront description of this option from their query API:
```
whether series with only points that are outside of the query window will be returned (defaults to true)
```
https://code.vmware.com/apis/714/wavefront-rest#/Query/queryApi

I played around the options and found the following:
* maxPoints: 1, SeriesOutsideTimeWindow: false =  no datapoints returned
* maxPoints: "", SeriesOutsideTimeWindow: false = no datapoints returned
* maxPoints: 1, SeriesOutsideTimeWindow: true = 1 datapoints returned, drift: 490
* maxPoints: "", SeriesOutsideTimeWindow: true = 2 datapoint returned, drift: 413

All of the below use startTime and endTime to be the same value.

I think the fix for this is that we should set SeriesOutsideTimeWindow to be true by default. However, this might have the adverse effect of returning very stale data in the query. However, I believe this is preferable versus erroring the query entirely.

Issue #1008, opened 2021-02-24
AnalysisRun query error was not retried in 10 seconds


jessesuen:
## Summary 

When an AnalysisRun hits an `Error`, it is supposed to retry after 10 seconds. Instead, it retried the user specified interval, which was 5 minutes.

## Diagnostics

What version of Argo Rollouts are you running? v0.10.2

```
INFO[2021-02-24T11:56:06-08:00] Started syncing Analysis at (2021-02-24 11:56:06.001213 -0800 PST m=+299.912326539)  analysisrun=success-rate-hlzhz namespace=test
INFO[2021-02-24T11:56:06-08:00] running overdue measurement                   analysisrun=success-rate-hlzhz metric=success-rate namespace=test
INFO[2021-02-24T11:56:06-08:00] taking 1 measurements                         analysisrun=success-rate-hlzhz namespace=test
INFO[2021-02-24T11:56:06-08:00] measurement completed Error                   analysisrun=success-rate-hlzhz metric=success-rate namespace=test
WARN[2021-02-24T11:56:06-08:00] measurement had error: server returned 403 Forbidden
You are not allowed to perform this operation.  analysisrun=success-rate-hlzhz metric=success-rate namespace=test
INFO[2021-02-24T11:56:06-08:00] enqueueing analysis after 4m59.999891106s     analysisrun=success-rate-hlzhz namespace=test
INFO[2021-02-24T11:56:06-08:00] Patch status successfully                     analysisrun=success-rate-hlzhz namespace=test
INFO[2021-02-24T11:56:06-08:00] Reconciliation completed                      analysisrun=success-rate-hlzhz namespace=test time_ms=927.487469
```

Issue #1006, opened 2021-02-24
Analysis evaluation errors are not surfaced


jessesuen:
## Summary 

When trying a metric query, I made a user error comparing mismatched types (array to float). However, the rollout controller did not surface this.

## Diagnostics

What version of Argo Rollouts are you running? v0.10.2

```
time="2021-02-24T09:42:15Z" level=warning msg="invalid operation: >= (mismatched types []float64 and float64)" analysisrun=success-rate-9prjw metric=success-rate namespace=argo-rollouts
time="2021-02-24T09:42:15Z" level=info msg="measurement completed Error" analysisrun=success-rate-9prjw metric=success-rate namespace=argo-rollouts```
```

```yaml
status:
  metricResults:
  - consecutiveError: 4
    error: 4
    measurements:
    - finishedAt: "2021-02-24T09:26:59Z"
      metadata:
        timestamps: '[1614158812,1614158812,1614158782,1614158812,1614158812,1614158782,1614158812,1614158812,1614158782,1614158812,1614158782,1614158782,1614158812,1614158782]'
      phase: Error
      startedAt: "2021-02-24T09:26:47Z"
      value: '[1.60,12.00,1.57,1.07,12.00,1.10,3.50,0.57,0.50,1.23,1.40,1.80,1.33,2.37]'
    - finishedAt: "2021-02-24T09:32:05Z"
      metadata:
        timestamps: '[1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112,1614159112]'
      phase: Error
      startedAt: "2021-02-24T09:31:59Z"
      value: '[1.60,12.00,1.67,0.97,12.00,1.40,3.50,0.40,1.43,1.30,0.90,4.50,1.17,0.40]'
    - finishedAt: "2021-02-24T09:37:11Z"
      metadata:
        timestamps: '[1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412,1614159412]'
      phase: Error
      startedAt: "2021-02-24T09:37:05Z"
      value: '[1.60,12.00,1.63,0.93,12.00,1.60,3.50,0.63,1.63,1.07,0.83,4.50,1.37,0.50]'
    - finishedAt: "2021-02-24T09:42:15Z"
      metadata:
[...]

PR #1021, opened 2021-03-03
fix: wavefront queries would return no datapoints. surface evaluate errors. retry errors faster

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1012
Resolves https://github.com/argoproj/argo-rollouts/issues/1006
Resolves https://github.com/argoproj/argo-rollouts/issues/1008

fix: Wavefront query will now allow data points to return outside the time window
fix: expr evaluation errors were not surfaced in the message
fix: metrics which errored, were not retried at 10s error interval

Also, datapoint drift is calculated as a convenience for debugging.


khhirani (approved):
LGTM

Issue #972, opened 2021-02-03
Support Ambassador AES as a traffic router for canary rollouts


leoluz:
# Summary

Add support for Ambassador AES to be used as a traffic router during canary rollouts.

What change needs making?

Provide a new `TrafficRoutingReconciler` implementation leveraging Ambassador AES canary deployment capabilities.

# Use Cases

When would you use this?

As a user of ArgoRollouts I want to be able to choose my existing Ambassador API gateway to control traffic during canary deployments.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

leoluz:
# Overview

The general idea of this implementation is leveraging [Ambassador canary releases feature](https://www.getambassador.io/docs/latest/topics/using/canary/) integrating it and making it compatible with Argo Rollouts canary deployments. Ambassador canary releases have a very similar weight based concept used to route traffic. Ambassador will be monitoring its `Mapping` resource in order to define how to route traffic to a specific service. It expects to have at least 2 Mappings available during the canary release:
- The base `Mapping` used to route all traffic to services
- The canary `Mapping` which is going to define the canary service as well as the desired weight

The example bellow shows how those `Mapping`s would look like:

```yaml
---
apiVersion: getambassador.io/v2
kind:  Mapping
metadata:
  name:  quote-backend
spec:
  prefix: /backend/
  service: quote
---
apiVersion: getambassador.io/v2
kind:  Mapping
metadata:
  name:  quote-backend2
spec:
  prefix: /backend/
  service: quote-canary
  weight: 10
```

# Proposal

In order to have Argo-Rollouts integrated the current idea is dynamically create the canary `Mapping` based on the provided base `Mapping`. In order to have that it is required to add the following change in Argo-Rollouts CRD allowing users to provide something like:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
spec:
  strategy:
    canary:
      stableService: quote
      canaryService: quote-canary
      trafficRouting:
        ambassador:
          mapping: quote-backend
```

With this configuration a canary `Mapping` can be dynamically generated and have the `weight` attribute updated regarding the steps configuration.

# Implementation

There is branch created implementing this proposal available in the following fork: 

https://github.com/datawire/argo-rollouts/commits/ambassador/canary

We are currently testing it and addressing some issues.

PR #1025, opened 2021-03-04
feat: Implement Ambassador to be used as traffic router for canary deployments

leoluz:
This PR is related to the following proposal:
[Support Ambassador AES as a traffic router for canary rollouts](https://github.com/argoproj/argo-rollouts/issues/972)

Checklist:


jessesuen:
Thanks for the contribution! I'll take time to look at this in the coming days

docs/features/traffic-management/ambassador.md
```diff
@@ -0,0 +1,95 @@
...
+kind: Rollout
+...
+spec:
+  strategy:
+    canary:
+      stableService: someapp-stable
+      canaryService: someapp-canary
+      trafficRouting:
+        ambassador:
+          apiVersion: getambassador.io/v2
```
jessesuen, line 48: For TrafficSplit, Istio VirtualService/DestinationRule CRs, we made apiVersion configuration a controller-level decision with the `--istio-api-version` and `--traffic-split-api-version` flags to the controller. In that way, the rollout controller would choose to work with a single apiVersion of a CRD, and leverage the natural conversion process (e.g. conversion webhooks) to handle any object.

Given TrafficSplit and Istio implementations are deciding apiVersion at the controller level, would it also make sense to do the same for Mapping objects and not expose it here?

leoluz: I saw those flags in the controller when I implemented this attribute. What have driven me to make this as part of the CRD were mainly 2 reasons:
1. It is more flexible to users who might have different mapping versions applied in the cluster to be able to specify which one to use at deployment time. Setting this as argo-rollout flag forces users to use only one mapping version.
2. I found that keeping it at the CRD level allows the Ambassador related code to be mainly centralized in one package making the code cleaner and easier to reason about.

I also understand if you want to keep some level of consistency with other traffic router implementations by adding this as an additional flag.

Please let me know your thoughts.

jessesuen: > It is more flexible to users who might have different mapping versions applied in the cluster to be able to specify which one to use at deployment time. Setting this as argo-rollout flag forces users to use only one mapping version.

So My assumption here is that the user's preference of which mapping version to use, doesn't really matter in the end, because they can always retrieve their preferred version.

HPA is a good analogy for this. In kubernetes, you can deploy an `autoscaling/v1` object but still retrieve it from the API as `autoscaling/v2beta1` because kubernetes will convert it on the fly. For example you can run the following two commands to get two versions of the same HPA object

```yaml
kubectl get hpa.v1.autoscaling my-hpa

kubectl hpa.v2beta1.autoscaling my-hpa
```

Is my understanding of CRD conversion incorrect?

leoluz: > Is my understanding of CRD conversion incorrect?

No. You are right that it doesn't necessarily forces users to use only one version as long as they define the latest one to avoid missing fields during the canary mapping generation.

I guess another flexibility problem related to specifying it as flags is that it forces users to reconfigure and redeploy argo-rollouts on the eventual release of a new Ambassador API version. It seems to me that having the ability to specify it at the CRD level provides a better user experience. What do you think?

jessesuen: > I guess another flexibility problem related to specifying it as flags is that it forces users to reconfigure and redeploy argo-rollouts on the eventual release of a new Ambassador API version.

This is true, but it's better to put the onus of this on the operator, and not the developers who are actually managing the Rollout objects. In other words, developers will GitOps their Rollout rollout with say a v1 mapping. But when v2 is introduced into the cluster, all of the Rollout objects will then have to be updated to use a v2 mapping. By making it a controller level decision, the end users (developers) do not have to do anything. For this reason I think it's better to remove the apiVersion and have this be a controller level flag.

leoluz: Makes sense. I'll implement the changes and let you know.
Thank you

eilonmonday: any updates or estimated time for merge?

leoluz: The PR is updated following @jessesuen's suggestion above.
We are now trying to make all checks green again after master rebase.

jessesuen (approved):
LGTM! Awesome work!

docs/features/traffic-management/ambassador.md
eilonmonday: @leoluz amazing thank you!
there is an issue with the sticky session.
the splitting to mapping canary or active is happening before the session affinity action.

PR #1044, opened 2021-03-23
fix: Fixes the regression of dropping resources from argo-rollouts crds. Fixes #1043

### LINKED ISSUE #1043: Resource requests/limits are ignored by Kubernetes
## Summary 

When trying to apply resource requests/limits to a rollout the apply succeeds, but the resulting object in the API service has resources set as an empty object.

Anderson Silva in the Slack channel found this was related to the CRD for Rollouts defining the `resources` just as an object without any elements. By patching the additional elements the CRD would work.

This seems like a regression caused by https://github.com/argoproj/argo-rollouts/pull/480 and I think it is caused by field pruning: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning

### Steps to reproduce

1. Apply https://gist.github.com/PedroArvela/66dec9c51ce0872caf272f90603e28e8
2. Run `kubectl get rollout -o yaml borker`

### Expected Results

Returned yaml should have:

```yaml
          resources:
            requests:
              cpu: 300m
              memory: 512Mi
            limits:
              cpu: 600m
              memory: 1Gi
```

### Actual Results

Resources are:

```yaml
resources: {}
```

## Diagnostics

* Example resource file that shows the problem: https://gist.github.com/PedroArvela/66dec9c51ce0872caf272f90603e28e8
* Argo Rollouts 0.10.2
* Logs for the entire controller: https://gist.github.com/PedroArvela/a18a4362f57c1063db32a329101c4b6c
* Logs for a specific rollout: https://gist.github.com/PedroArvela/7cf5256fb5485de729a3960ff0aea301

---

anderson4u2:
Closes #1043

As Pedro Arvela explained in #1043, #480 removed the validation for 'resources' field due to the inability to support float types for the validation. However we believe this is causing [k8s to prune the undefined fields in the schema. ](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning)

In this MR, I am re-adding the validations removed by #480 (and a couple that were probably missed) as fields that won't be pruned, will not break with float values, and so "resources" won't be pruned and will work as expected.

Checklist:


jessesuen:
I'm unable to reproduce this issue using the sample gist in https://gist.github.com/PedroArvela/66dec9c51ce0872caf272f90603e28e8. Is this related to kubernetes version?

jessesuen:
Note that when we run controller-gen, we do *not* run it with preserveUnknownFields=false, meaning we are telling Kubernetes *not* to prune unknown fields: 

```golang
	crdYamlBytes, err := exec.Command(
		"controller-gen",
		"paths=./pkg/apis/rollouts/...",
		"crd:trivialVersions=true",
		// cannot use preserveUnknownFields=false until controller-gen generates proper support for
		// resource.Quantity, which we remove validation for
		//"crd:preserveUnknownFields=false",
		"crd:crdVersions=v1beta1",
		"output:crd:stdout",
	).Output()
```

So I do not know understand how this could happen.

jessesuen:
OK I understand what's going on. From kubernetes documentation:

https://v1-18.docs.kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#preserving-unknown-fields

> For CustomResourceDefinitions created in apiextensions.k8s.io/v1, structural OpenAPI v3 validation schemas are required and pruning is enabled and cannot be disabled (note that CRDs converted from apiextensions.k8s.io/v1beta1 to apiextensions.k8s.io/v1 might lack structural schemas, and spec.preserveUnknownFields might be true).

In other words, kubernetes no longer supports the preservation of unknown fields as a top-level option, as of `apiextensions.k8s.io/v1`. But because Argo Rollouts still vends `apiextensions.k8s.io/v1beta1` CRD definitions, we were able to get away with disabling this. So instead of relying on the deprecated `spec.preserveUnknownFields` ability, we must preserve unknown fields explicitly through the custom open API annotations which is being added in this PR. 

I notice that when I get the v1 CRD object, i see the following condition:

```yaml
status:
  conditions:
  - lastTransitionTime: "2021-03-19T11:33:45Z"
    message: 'spec.preserveUnknownFields: Invalid value: true: must be false'
    reason: Violations
    status: "True"
    type: NonStructuralSchema
 ```

So technically speaking, Argo Rollouts is vending a CRD that violates current CRD requirements and this PR is fixing that violation.

**Comment by jessesuen** [comment]
@anderson4u2 - this PR is valid, but I'm still curious how you wound up in this situation, given that Argo Rollouts vends the v1beta1 CRDs.

**Comment by jsoref** [comment]
The argo-helm repo is vending 1.0: https://github.com/argoproj/argo-helm/blob/master/charts/argo-rollouts/templates/crds/rollout-crd.yaml

And it's impossible to merge *anything* into that repository without vending 1.0.

File: manifests/crds/experiment-crd.yaml [hunk 1]
```diff
@@ -2334,21 +2349,9 @@ spec:
...
                                               properties:
                                                 limits:
-                                                  additionalProperties:
-                                                    anyOf:
-                                                    - type: integer
-                                                    - type: string
-                                                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                                                    x-kubernetes-int-or-string: true
-                                                  type: object
+                                                  x-kubernetes-preserve-unknown-fields: true
```
jessesuen, line 2352: I'm still understanding the implications of dropping the resource quantity validation here, and why this only applied to ephemeral containers. Once I do that, the PR looks good to go.

jsoref: I think in theory one would want to provide a full spec. But if you can't, then providing a "we don't know what belongs here, we're labeling it dragons -- `here be dragons`". This allows the dragons to live here.

Ideally someone will later spend the time to properly spec it out and replace the dragon gate.

But, also, when they do that, (probably in a new version,) it'll force people to revalidate the data that's hiding there instead naively trusting it (which is the reason k8s to force default omit -- to prevent pre-poisoning)

anderson4u2:
Hello, I've pushed another commit to upgrade the API to v1 CRD. Let me know your thoughts. I've removed the function `removeNestedItems` as it seems it hasn't been run in the argo-helm templates, and we're running that without any problems. Any feedback greatly appreciated, thanks!

manifests/crds/experiment-crd.yaml
anderson4u2: FYI I'm only setting the not-prune on values of resources/requests and resources/limits. So it limits the scope for failure to only those fields, not its subfields nor its parent fields, which limits the scope for failure significantly.

jessesuen:
Great! will take another look

jessesuen:
> I've removed the function removeNestedItems as it seems it hasn't been run in the argo-helm templates, and we're running that without any problems.

In order for you to have notice, you would have to have had a volume like:

```yaml
  volumes:
    # You set volumes at the Pod level, then mount them into containers inside that Pod
    - name: config
      configMap:
        # Provide the name of the ConfigMap you want to mount.
        name: game-demo
        # An array of keys from the ConfigMap to create as files
        items:
        - key: "game.properties"
          path: "game.properties"
        - key: "user-interface.properties"
          path: "user-interface.properties"
```

Before removing this, can you test that this works with the new CRD?

anderson4u2:
> > I've removed the function removeNestedItems as it seems it hasn't been run in the argo-helm templates, and we're running that without any problems.
> 
> In order for you to notice any problems, you would have to have had a volume like:
> 
> ```yaml
>   volumes:
>     # You set volumes at the Pod level, then mount them into containers inside that Pod
>     - name: config
>       configMap:
>         # Provide the name of the ConfigMap you want to mount.
>         name: game-demo
>         # An array of keys from the ConfigMap to create as files
>         items:
>         - key: "game.properties"
>           path: "game.properties"
>         - key: "user-interface.properties"
>           path: "user-interface.properties"
> ```
> 
> Before removing this, can you test that this works with the new CRD?

Hi, I've checked that it works with k8s 1.17+. Tested it with k8s 1.16 and got [this error](https://github.com/kubernetes/kubernetes/issues/84880), the merge that fixed it was [this](https://github.com/kubernetes/kubernetes/pull/85223), which was released in 1.17. I've also added a test in the test-e2e that fails in k8s 1.16 but succeeds in 1.17+.

jessesuen:
> Hi, I've checked that it works with k8s 1.17+. Tested it with k8s 1.16 and got this error, the merge that fixed it was this, which was released in 1.17. I've also added a test in the test-e2e that fails in k8s 1.16 but succeeds in 1.17+.

I just polled the community in slack and looks like we still have a significant portion of our users on v1.16 or below (4/10 responses at time of writing).

So I think it's better to preserve the existing behavior in order to continue to support our v1.16/v1.15 users. Could you add back removeNestedItems to support this?

anderson4u2:
> > Hi, I've checked that it works with k8s 1.17+. Tested it with k8s 1.16 and got this error, the merge that fixed it was this, which was released in 1.17. I've also added a test in the test-e2e that fails in k8s 1.16 but succeeds in 1.17+.
> 
> I just polled the community in slack and looks like we still have a significant portion of our users on v1.16 or below (4/10 responses at time of writing).
> 
> So I think it's better to preserve the existing behavior in order to continue to support our v1.16/v1.15 users. Could you add back removeNestedItems to support this?

Hi, I've checked with the refactored function but I still got errors when trying to deploy a configmap volume in k8s 1.16. I've unvalidated the whole `volumes` block and it now works. I've added this unvalidation to all CRDs. Let me know your thoughts :)

jessesuen (approved):
Great improvement!

hack/gen-crd-spec/main.go
```diff
@@ -44,23 +52,36 @@ func removeValidation(un *unstructured.Unstructured, path string) {
...
-		// cannot use preserveUnknownFields=false until controller-gen generates proper support for
-		// resource.Quantity, which we remove validation for
-		//"crd:preserveUnknownFields=false",
-		"crd:crdVersions=v1beta1",
+		// The only possible value is 'false' since 'apiextensions.k8s.io/v1'
+		// https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning
+		// It is possible though to opt-out of pruning for specifc sub-trees of fields by adding x-kubernetes-preserve-unknown-fields: true
+		// by using the 'setValidationOverride' function in this file.
+		"crd:preserveUnknownFields=false",
+		"crd:crdVersions=v1",
```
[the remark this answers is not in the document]
anderson4u2: np, glad to help!

PR #1053, opened 2021-03-29
feat: support a custom base URL for the new relic provider

jemisonf:
Enables use of the provider by people accessing new relic via a proxy. Users should set either a `region` key or a `base-url` key in their new relic secret, and if neither is provided the provider will continue to default to the US region.

Checklist:

### DIFF
--- a/metricproviders/newrelic/newrelic.go
+++ b/metricproviders/newrelic/newrelic.go
@@ -155,15 +155,20 @@ func NewNewRelicAPIClient(metric v1alpha1.Metric, kubeclientset kubernetes.Inter
 
 	newrelicOptions := []newrelic.ConfigOption{newrelic.ConfigPersonalAPIKey(apiKey), newrelic.ConfigUserAgent(userAgent)}
 
-	// support either base-url or region; default to us region if neither is used
-	if _, ok := secret.Data["base-url"]; ok {
-		newrelicOptions = append(newrelicOptions, newrelic.ConfigBaseURL(string(secret.Data["base-url"])))
-	} else {
-		region := "us"
-		if _, ok := secret.Data["region"]; ok {
-			region = string(secret.Data["region"])
-		}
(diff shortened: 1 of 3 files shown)


jemisonf:
Based on discussion in slack: https://argoproj.slack.com/archives/CKTN88XFH/p1616688533036100

I don't think there's a corresponding issue but I can open one if that would be helpful

russellwhitaker:
@jemisonf if you could associate an issue with this, yes, that would be great, thanks.

jessesuen (commented):
This looks good to go. Do you mind also documenting this as well?

metricproviders/newrelic/newrelic.go
```diff
@@ -155,15 +155,20 @@ func NewNewRelicAPIClient(metric v1alpha1.Metric, kubeclientset kubernetes.Inter
...
+	newrelicOptions = append(newrelicOptions, newrelic.ConfigRegion(region))
+
+	// base URL for the new relic REST API
+	if _, ok := secret.Data["base-url-rest"]; ok {
+		newrelicOptions = append(newrelicOptions, newrelic.ConfigBaseURL(string(secret.Data["base-url-rest"])))
+	}
+
+	// base URL for the nerdgraph (graphQL) API
+	if _, ok := secret.Data["base-url-nerdgraph"]; ok {
+		newrelicOptions = append(newrelicOptions, newrelic.ConfigNerdGraphBaseURL(string(secret.Data["base-url-nerdgraph"])))
```
jemisonf: this is a change from the original approach, based on reading the docs. Looks like we need to define separate REST API and NerdGraph (aka GraphQL) API base URLs in addition to the region.

jemisonf: there are some other base URL config keys available but these are the only two that should be relevant for this specific application

jessesuen (approved):
LGTM!

Issue #1045, opened 2021-03-24
controller can panic on startup when starting up with invalid object


jessesuen:
## Summary 

The recent use of tolerant informers helped deal with the problem when the controller was **_introduced_** to bad rollout objects. However, I am still able to crash the controller if I change the order of steps. I can induce a crash if I introduce the bad rollout *before* the controller starts e.g.:

1. stop the controller
2. introduce a bad rollout (e.g. like below)

```yaml
spec:
  replicas: 1
  selector:
    matchLabels:
      app: borker
      env: development
  strategy:
    canary:
      maxSurge: 50%
      maxUnavailable: 0
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: borker
        env: development
    spec:
      containers:
      - image: argoproj/rollouts-demo:blue
        name: main
        resources:
          requests:
            cpu: 0.1
            memory: 256Mi
      dnsPolicy: ClusterFirst
      ephemeralContainers:
      - image: argoproj/rollouts-demo:blue
        name: main
        resources:
          requests:
            cpu: asdfadsf       # <<<<<< bad
            memory: 256Mi
```

3. start the controller

Then I am able to crash the controller:

```
E0324 16:13:36.084998   47797 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 152 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x27593a0, 0x3847730)
        /Users/jsuen/go/src/github.com/argoproj/argo-rollouts/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /Users/jsuen/go/src/github.com/argoproj/argo-rollouts/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x89
panic(0x27593a0, 0x3847730)
        /usr/local/Cellar/go/1.15.5/libexec/src/runtime/panic.go:969 +0x1b9
github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1.(*Rollout).GetNamespace(0x0, 0x0, 0x2c8f280)
        <autogenerated>:1 +0x5
[...]

PR #1055, opened 2021-04-01
fix: controller panic with ephemeral containers with bad resource qty

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1045

Fixes a controller panic recently introduced in pre-v1.0


khhirani (approved):
LGTM

Issue #478, opened 2020-04-17
Introduce scaleDownDelay in canary (with traffic routing)


jessesuen:
Just like we have an implicit scaleDownDelay (default 30s) in the blue green strategy, I think we may also need an implicit scaleDownDelay with canary (with traffic routing). Here is the scenario:

```yaml
  strategy:
    canary:
      steps:
        - setWeight: 20
        - pause:
            duration: 30
        - setWeight: 50
        - pause:
            duration: 30
      trafficRouting:
        istio:
...
```

When the final step is reached we do several things **_simultaneously_**:
1. setWeight to 100
2. switch stable selector to new ReplicaSet
2. scale down the old replicaset to 0

However, as with blue-green, I do not believe it is possible for the effects of the `setWeight: 100` and changing service selectors to take effect immediately, especially across all mesh/ingress providers implementations.

Therefore, I propose that an implicit scaleDownDelay is introduced in the canary strategy, which delays the scale down of the old replicaset for 30 seconds (default), to give a chance for the 100% weight and service selectors to propagate properly.

NOTE that scaleDownDelay is only applicable for canary **_with_** trafficRouting since we cannot delay the scaleDown when using replica-count based canary when it need to reach 100%. (by definition 100% means the old replicaset is scaled down to zero).

A workaround in the meantime is to add two more steps to the canary steps

```yaml
        - setWeight: 100
        - pause:
            duration: 30
```

jessesuen:
Relevant discussion in slack:
https://argoproj.slack.com/archives/CKTN88XFH/p1587117236048300

PR #1056, opened 2021-04-01
feat: support scaleDownDelaySeconds in canary w/ traffic routing

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/478

This change introduces scaleDownDelaySeconds and scaleDownDelayRevisionLimit to the canary strategy when used with traffic routing:

```yaml
  strategy:
    canary:
      scaleDownDelaySeconds: 30        # default 30s
      scaleDownDelayRevisionLimit: 1
      trafficRouting:
...
```

The purpose of a scaleDownDelaySeconds for canary with traffic routing, is the same reason for blue-green scaleDownDelaySeconds. We want to avoid scaling down the stable stack prematurely in order to give some time for service selectors to propagate to mesh providers and ingress controllers, and avoid 500 errors. This change will introduce a default 30s delay before scaling down stable stack just like in blue-green.


khhirani (approved):
LGTM

Issue #1067, opened 2021-04-06
spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema


jessesuen:
## Summary 

This is a pre-1.0 issue and must/will be fixed by v1.0. When applying the `apiextensions.k8s.io/v1` CRD against a cluster which had previously installed  `apiextensions.k8s.io/v1beta1` version of the rollout CRDs, the kubectl apply will fail with the following error:

```
...
Resource: "apiextensions.k8s.io/v1, Resource=customresourcedefinitions", GroupVersionKind: "apiextensions.k8s.io/v1, Kind=CustomResourceDefinition"
Name: "rollouts.argoproj.io", Namespace: ""
for: "./manifests/crds": CustomResourceDefinition.apiextensions.k8s.io "rollouts.argoproj.io" is invalid: spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema
```

This means anyone upgrading from v0.10 to v1 will face this issue (so we must fix this before releasing v1.0).

The problem is, spec.preserveUnknownFields was defaulted to false in `apiextensions.k8s.io/v1beta1`, but is now defaulted to true in `apiextensions.k8s.io/v1`. And because neither version of our v0.10 or v1.0 CRDs currently specifies this value, when kubectl apply happens, it preserves the existing version of true, which is disallowed in `apiextensions.k8s.io/v1`.

To fix this, we should explicitly set `spec.preserveUnknownFields: false` in the v1 CRD, in order to allow a smooth upgrade.

## Diagnostics

What version of Argo Rollouts are you running? v1.0 pre-release

```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME>
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #1069, opened 2021-04-06
fix: spec.preserveUnknownFields must be explicitly false to allow upgrades from apiextensions.k8s.io/v1beta1

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1067

After switching from v1beta1 to v1 CRD, the default value of `spec.preserveUnknownFields` in Kubernetes went from true to false. Because this field is unmanaged in our previous and current set of manifests, when upgrading from the v1beta1 CRD version to v1, it would encounter the following error:

> spec.preserveUnknownFields: Invalid value: true: must be false in order to use defaults in the schema

This is because when the v1 CRD was attempted to be applied (over the existing v1beta1 version), K8s rejected the update stating the value must be false (since a v1 manifest was being applied). This change explicitly sets it to false so that upgrades can happen from v0.10 to v1.0

Also removes our removeDescriptions workaround now that `controller-gen` has the ability to remove descriptions.

### DIFF
--- a/hack/gen-crd-spec/main.go
+++ b/hack/gen-crd-spec/main.go
@@ -82,6 +82,7 @@ func NewCustomResourceDefinition() []*extensionsobj.CustomResourceDefinition {
 		// by using the 'setValidationOverride' function in this file.
 		"crd:preserveUnknownFields=false",
 		"crd:crdVersions=v1",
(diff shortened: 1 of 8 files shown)


alexmt (approved):
LGTM

PR #1071, opened 2021-04-08
fix: verify analysis arguments name with those in the rollout

### LINKED ISSUE #1052: Unresolved argument errors to templates does not surface as validation error
## Summary 

When debugging a rollout that had an analysis template and the rollout did not supply a required argument to the template, the rollout simply stops with an unhelpful message:

```
Name:            cloud-ops-rollout-vite-fat-jar2-e2e-usw2
Namespace:       cloud-ops-vitefatjar2-usw2-e2e
Status:          ◌ Progressing
Message:         more replicas need to be updated
```

In reality, it's not a problem when more replicas needing to be updated. The real problem is that the rollout did not supply necessary arguments to analysis template, as seen in controller logs:

```
time="2021-03-26T20:59:57Z" level=info msg="Reconciliation completed" namespace=cloud-ops-vitefatjar2-usw2-e2e rollout=cloud-ops-rollout-vite-fat-jar2-e2e-usw2 time_ms=6.004360999999999
time="2021-03-26T20:59:57Z" level=error msg="rollout syncHandler error: args.service-name was not resolved" namespace=cloud-ops-vitefatjar2-usw2-e2e rollout=cloud-ops-rollout-vite-fat-jar2-e2e-usw2
```

The bug is that we should Degrade the rollout with a validation error when arguments are not supplied.

## Diagnostics

What version of Argo Rollouts are you running? v0.10.2

---

huikang:
- return error if arguments do not match between rollout spec and analysis template

close #1052

### DIFF
--- a/pkg/apis/rollouts/validation/validation_references.go
+++ b/pkg/apis/rollouts/validation/validation_references.go
@@ -65,7 +65,7 @@ func ValidateRolloutReferencedResources(rollout *v1alpha1.Rollout, referencedRes
 		allErrs = append(allErrs, ValidateService(service, rollout)...)
 	}
 	for _, template := range referencedResources.AnalysisTemplateWithType {
-		allErrs = append(allErrs, ValidateAnalysisTemplateWithType(template)...)
+		allErrs = append(allErrs, ValidateAnalysisTemplateWithType(rollout, template)...)
 	}
 	for _, ingress := range referencedResources.Ingresses {
 		allErrs = append(allErrs, ValidateIngress(rollout, ingress)...)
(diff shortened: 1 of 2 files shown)


jessesuen:
@khhirani can you review this?

huikang:
@jessesuen , are we expecting the following case to be passed?

```yaml
kind: AnalysisTemplate
metadata:
  name: sleep-job
spec:
  args:
  - name: duration
    value: 0s
  - name: exit-code
    value: "0"
  - name: count
    value: "1"
```

```yaml
kind: Rollout
metadata:
  name: canary-promote-full
spec:
  replicas: 3
  strategy:
    canary:
      maxUnavailable: 0
      analysis:
        templates:
        - templateName: sleep-job
```

In this case, the Rollout does not provide any args, but the AnalysisTemplate seems having some default value. Is this a valid usage?

jessesuen:
> In this case, the Rollout does not provide any args, but the AnalysisTemplate seems having some default value. Is this a valid usage?

Yes, defaulting is a feature of the AnalysisTemplate.

huikang:
> 
> Yes, defaulting is a feature of the AnalysisTemplate.

Thanks. I have updated PR to consider the default values.

khhirani (approved):
Hi @huikang , your PR looks good to merge

PR #1072, opened 2021-04-08
feat: support reference model for workloads (#676)

### LINKED ISSUE #676: Support reference model for workloads
# Summary

What change needs making?

Consider change the embedded mode to reference mode, so we could have more flexibility on workload side.

```yaml
kind: Rollout
metadata:
  name: rollout-bluegreen
spec:
  workloadRef: 
    apiVersion: apps/v1
    kind: Deployment
    name: demo1
  strategy:
    blueGreen: 
      activeService: rollout-bluegreen-active
      previewService: rollout-bluegreen-preview
      autoPromotionEnabled: false
```

```yaml
kind: Rollout
metadata:
  name: rollout-bluegreen
spec:
  workloadRef: 
    apiVersion: apps/v1
    kind: ReplicaSet
    name: demo2
  strategy:
    blueGreen: 
      activeService: rollout-bluegreen-active
      previewService: rollout-bluegreen-preview
      autoPromotionEnabled: false
```

# Use Cases

To support we use Deployment or our own stateless workload (like CloneSet).

---

alexmt:
Signed-off-by: Alexander Matyushentsev <AMatyushentsev@gmail.com>

Closes #676 

* [x] First draft
* [x] Unit tests & comments
* [x] E2E tests
* [x] Documentation


examples/rollout-template-ref.yaml
```diff
@@ -0,0 +1,42 @@
...
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
+  name: rollout-ref-deployment
+spec:
+  replicas: 5
+  revisionHistoryLimit: 2
+  selector:
+    matchLabels:
+      app: rollout-ref-deployment
```
jessesuen, line 33: Can we make this not required for Deployment? Because we can already get selector from the deployment spec.

examples/rollout-template-ref.yaml
```diff
@@ -0,0 +1,42 @@
...
+  templateRef:
```
jessesuen, line 34: I think `workloadRef` may be better than `templateRef`

examples/rollout-template-ref.yaml
alexmt: Ah, good point. Agree.

examples/rollout-template-ref.yaml
alexmt: I've changed `workloadRef` to `templateRef` to make it more explicit. Howerver, after looking at comment https://github.com/argoproj/argo-rollouts/pull/1072#discussion_r610208462, I agree. Changing back to `workloadRef`

alexmt: done

examples/rollout-template-ref.yaml
alexmt: done

docs/migrating.md
```diff
@@ -36,8 +41,78 @@ spec:
       - pause: {}
 ```
 
-## Other Considerations
+!!! warning
+    When migrating a Deployment which is already serving live production traffic, a Rollout should
+    run next to the Deployment before deleting the Deployment. **Not following this approach might result in
```
jessesuen, line 46: > before deleting the Deployment

Change to "before deleting the Deployment or scaling down the Deployment"

docs/migrating.md
```diff
@@ -36,8 +41,79 @@ spec:
...
+          image: argoproj/rollouts-demo:blue
+          imagePullPolicy: Always
+          ports:
+            - containerPort: 8080
+```
+
+Consider following if your Deployment runs in production:
+
+**Running Rollout and Deployment side-by-side**
+
```
Sonam11, line 105: So, just to completely understand the workflow, the migration will happen like below:
1. Create rollout object to refer to the deployment object.
2. Apply the rollout object, it will create different set of pods with the same spec.
3. Now when all is ready we will scale down the deployment replicas to 0.
4. Then going forward the deployment object can keep to have the pod specs and rollout can keep referring to it and any change in the deployment object will trigger the rollout. And the pods will be managed by rollout as deployment is scaled to 0 (So as in any point of time we can go back to deployment object in control by just changing the replica number).

Please let me know if my understanding is correct.

alexmt: hello @Sonam11, yes. That is correct.

jessesuen (commented):
We also need to update the rbac so that the controller can list/get/watch Deployments, otherwise, things won't work.

jessesuen (approved):
Please fix rbac necessary for this to work.

Impressed you were able to do this by unmarshalling tricks and barely touching controller code!

rollout/temlateref.go
```diff
@@ -0,0 +1,266 @@
...
+	defer r.informersLock.Unlock()
+	if r.cancelContext != nil {
+		r.cancelContext()
+	}
+	ctx, cancelContext := context.WithCancel(context.TODO())
+	r.ctx = ctx
+	r.cancelContext = cancelContext
+}
+
+func remashalMap(objMap map[string]interface{}, res interface{}) error {
```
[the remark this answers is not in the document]
alexmt: thank you! fixed

PR #1073, opened 2021-04-08
feat: [datadog metricprovider] Allow Datadog API and APP keys to be c…

jzyeezy:
## Why
My organization has a pattern where keys are stored in a secret cloud provider which get auto-synced as generic k8s secrets across regions. Rather than duplicate these keys in their own `datadog` k8s-secret, it would be much more convenient for us to mount the original secret values as environment variables for argo-rollouts. This also eliminates the need for granting the argo-rollouts pod the API permission to access the `datadog` k8s-secret.

Checklist:

### DIFF
--- a/metricproviders/datadog/datadog.go
+++ b/metricproviders/datadog/datadog.go
@@ -28,6 +28,9 @@ const (
 	//ProviderType indicates the provider is datadog
 	ProviderType            = "Datadog"
 	DatadogTokensSecretName = "datadog"
+	DatadogApiKey           = "api-key"
+	DatadogAppKey           = "app-key"
+	DatadogAddress          = "address"
 )
 
 // Provider contains all the required components to run a Datadog query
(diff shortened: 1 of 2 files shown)


metricproviders/datadog/datadog.go
```diff
@@ -165,18 +167,36 @@ func (p *Provider) GarbageCollect(run *v1alpha1.AnalysisRun, metric v1alpha1.Met
...
-	address := ""
-	if _, hasAddress := secret.Data["address"]; hasAddress {
-		address = string(secret.Data["address"])
+	appKey, err := getSecretValue("app-key", ns, kubeclientset, false)
+	if err != nil {
+		return nil, err
+	}
+	address, err := getSecretValue("address", ns, kubeclientset, true)
+	if err != nil {
+		return nil, err
```
jessesuen, line 199: The change will result in three API calls to k8s instead of the previous one. Could you refactor such that it still only makes a single API call for when we need to get the secret from the API server?

jessesuen (changes requested):
@jzyeezy the feature is a good improvement, but we'll need to reduce the API calls to k8s API server when getting each secret value.

Also, do you mind signing off for the DCO on your commit for us to accept this PR?

```
git commit --amend -s
git push -f
```

jzyeezy:
Okie @jessesuen I've addressed your comments. Thanks for the review!

metricproviders/datadog/datadog.go
```diff
@@ -165,18 +168,36 @@ func (p *Provider) GarbageCollect(run *v1alpha1.AnalysisRun, metric v1alpha1.Met
...
+
+	return string(k8sSecret.Data[key]), nil
+}
+
 func NewDatadogProvider(logCtx log.Entry, kubeclientset kubernetes.Interface) (*Provider, error) {
 	ns := defaults.Namespace()
 	secret, err := kubeclientset.CoreV1().Secrets(ns).Get(context.TODO(), DatadogTokensSecretName, metav1.GetOptions{})
 	if err != nil {
 		return nil, err
 	}
```
jessesuen, line 189: Note that by leaving the logic to fetch the secret here, it does not satisfy your previous benefit:

> This also eliminates the need for granting the argo-rollouts pod the API permission to access the datadog k8s-secret.

Not sure if you care about this as a feature, but I just wanted to call out the fact that it does not really eliminate the RBAC necessary to get the datadog secret.

jzyeezy: Wow you are right about that. 🤦🏻‍♀️ Thanks for catching that!

Let me address this one more time, properly.

jzyeezy: Alrighty pushed up 1308004, 3rd time's a charm?

jessesuen (approved):

PR #1095, opened 2021-04-20
chore: make codegen more consistent across environments

jessesuen:
All toolchain binaries (with the exception of `protoc`, `swagger-codegen`, and `yarn`) are now built on the fly using `go get <package>@<version>` and installed into the `dist` directory instead of using the system GOBIN path. They also have mostly been removed from `go.mod` as a dependency, except for the ones needed to be vendored by kubernetes.


rbreeze (approved):
Looks great!

Issue #942, opened 2021-01-13
Rollout stuck in "Progressing" indefinitely after failed experiment


artem-nefedov:
## Summary 

What happened/what you expected to happen?

Rollout would be forever stuck in "Progressing" after experiment with analysis failed.
Running `kubectl argo rollouts abort` does not help, but `kubectl argo rollouts restart` does.

It seems that it is only reproduced if the same experiment was already run successfully in the same rollout once prior to this.

Rollout strategy:

```
  strategy:
    canary:
      steps:
      - experiment:
          templates:
          - name: canary-prefiew
            specRef: canary
            selector:
              matchLabels:
                app: httpbin-test-ingress-private-preview
            metadata:
              labels:
                app: httpbin-test-ingress-private-preview
          analyses:
          - name: json-ok
            templateName: json-ok
            requiredForCompletion: true
            args:
            - name: service
              value: httpbin-test-ingress-private-preview.default.svc.cluster.local:8000
```

## Diagnostics

What version of Argo Rollouts are you running?

v0.10.2

# Paste the logs from the rollout controller

attached because of size

[logs.txt](https://github.com/argoproj/argo-rollouts/files/5809573/logs.txt)

# Screenshot of kubectl plugin output

![2](https://user-images.githubusercontent.com/7116010/104475290-fa3fab80-55cf-11eb-9786-e0083f467138.jpg)

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

huikang:
@artem-nefedov , why there is only `canary-prefiew` in the experiment template? Should there be a stable replicaset?

artem-nefedov:
@huikang There is a stable replicaset, but experiment runs only on canary

artem-nefedov:
More info about reproduction:
It seems the problem occur if experiment fails when upgrading to the "older" version that was stable before.
E.g.:

1. Deploy Rollout with image 1.0.0 (this image WOULD fail the experiment, but because it's the first deployed version, steps are ignored)
2. Update to image 1.0.1 (this image passes the experiment)
3. Downgrade to image 1.0.0 (this fails the experiment and cause Rollout to be forever stuck in progressing)

huikang:
> @huikang There is a stable replicaset, but experiment runs only on canary
> (as a way to test new version before actually passing any real traffic to it without using "traffic routing/service mesh integration" feature)

[...]

PR #1099, opened 2021-04-21
fix: rollout status always in progressing if analysis fails

huikang:
- when analysis fails but rollout completes, the status should
be degraded

address #942

### DIFF
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -585,7 +585,7 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	// In such a case, we should simply not estimate any progress for this rollout.
 	currentCond := conditions.GetRolloutCondition(c.rollout.Status, v1alpha1.RolloutProgressing)
 
-	isCompleteRollout := newStatus.Replicas == newStatus.AvailableReplicas && currentCond != nil && currentCond.Reason == conditions.NewRSAvailableReason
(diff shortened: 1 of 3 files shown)


khhirani (approved):
LGTM

rollout/sync.go
```diff
@@ -585,7 +585,7 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	// In such a case, we should simply not estimate any progress for this rollout.
 	currentCond := conditions.GetRolloutCondition(c.rollout.Status, v1alpha1.RolloutProgressing)
 
-	isCompleteRollout := newStatus.Replicas == newStatus.AvailableReplicas && currentCond != nil && currentCond.Reason == conditions.NewRSAvailableReason
+	isCompleteRollout := newStatus.Replicas == newStatus.AvailableReplicas && currentCond != nil && currentCond.Reason == conditions.NewRSAvailableReason && currentCond.Type != v1alpha1.RolloutProgressing
```
perenesenko, line 588: I just reviewed the `calculateRolloutConditions` to learn what's going on inside and found one interesting place where I'm a little bit confused:

The next expression `currentCond.Type != v1alpha1.RolloutProgressing` will always be `false` as the `currentCond` is a condition filtered by the v1alpha1.RolloutProgressing type.

That means `isCompleteRollout` is always `false`. Maybe we will remove it?

@jessesuen @huikang Please correct me here if I'm wrong.

Issue #1096, opened 2021-04-20
Support integration with istio virtual-service in another namespace for canary traffic routing


gaurav517:
# Summary

In current canary traffic routing with istio, we can refer to a istio virtual service but that virtual service needs to be in same namespace as rollout. We should be able to refer to a virtual service that is not in rollout namespace.

Current:
```
  strategy:
    canary:
      canaryService: canary-svc  
      stableService: stable-svc 
      trafficRouting:
        istio:
          virtualService: 
            name: rollout-vsvc   # expects virtual service to be in same namespace
            routes:
            - primary
```

Solution should enable following:
```
  strategy:
    canary:
      canaryService: canary-svc
      stableService: stable-svc
      trafficRouting:
        istio:
          virtualService: 
            name: rollout-vsvc.vsvc-ns  # rollout-vsvc in in vscv-ns namespace. 
            routes:
            - primary
```

# Use Cases

In some cases istio virtual service can live in different namespace.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
It would be great to support this, and there was an attempt to do this last year but it was incomplete:
https://github.com/argoproj/argo-rollouts/pull/650

It's probably safe to say that @sakajunquality may not be working on it at the moment and someone else can pick this up.

huikang:
@jessesuen , I can help with this one. Please assign it to me. Thanks.

huikang:
Hi, @gaurav517 , could you test and see if we can close this issue?

gaurav517:
Thanks a lot @huikang. Will test it over weekend assuming `argoproj/argo-rollouts:latest` docker tag has the changes.

gaurav517:
looks good on my side. Thanks a lot.

PR #1114, opened 2021-04-26
feat: istio virtualservice and rollout in different namespaces

huikang:
- A new field namespace to istio trafficrouting
- canarySvc and stableSvc could be FQDN in virtualService,
  so extract the host name and compare with the hosts defined
  in rollout
- Updated doc to reflect the change:
    docs/features/traffic-management/istio.md

- Address #1096 

### DIFF
--- a/docs/features/traffic-management/istio.md
+++ b/docs/features/traffic-management/istio.md
@@ -58,8 +58,9 @@ spec:
 
 The VirtualService must contain an HTTP route with a name referenced in the Rollout, containing
 two route destinations with `host` values that match the `canaryService` and `stableService` 
-referenced in the Rollout.  If the VirtualService is defined in a different namespace then the rollout, a `namespace` value can be added in addition to the virtuService `name`.
-Note that Istio require that all weights add to 100, so the initial weights can be be 100% to stable, and 0% to canary.
+referenced in the Rollout.  If the VirtualService is defined in a different namespace then the rollout,
+its name should be `rollout-vsvc.<vsvc namespace name>`. Note that Istio require that all weights add to
+100, so the initial weights can be be 100% to stable, and 0% to canary.
 
 ```yaml
 apiVersion: networking.istio.io/v1alpha3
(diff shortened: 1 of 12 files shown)


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -342,6 +342,8 @@ type IstioTrafficRouting struct {
 type IstioVirtualService struct {
 	// Name holds the name of the VirtualService
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+	// Namespace holds the namespace of the VirtualService
+	Namespace string `json:"namespace,omitempty"`
```
jessesuen, line 346: Since the VirtualService name needs to be specified like `stable-svc.<namespace>` when referencing across namespaces, is there any value in introducing a `Namespace` field to the VirtualService reference?

Can we just infer namespace instead? This eliminates the possibility of the two values disagreeing with each other due to user error.

huikang: I added the `namespace` field just because it is easy to process in the code.
After second thought, I agreed that inferring namespace in `stable-svc.<namespace>` is better liking inferring service endpoint. I will make the change. Thanks.

huikang: @jessesuen , PR is updated. Please take a another look at it. Thanks.

utils/istio/istio.go
```diff
@@ -48,13 +49,33 @@ func GetIstioDestinationRuleGVR() schema.GroupVersionResource {
 	}
 }
 
+func GetVirtualServiceNamespaceName(vsv string) (string, string) {
+	namespace := ""
+	name := ""
+
+	if idx := strings.Index(vsv, "."); idx > 0 {
+		name = vsv[:idx]
+		namespace = vsv[idx+1:]
```
jessesuen, line 58: I dont think this will work for fully qualified virtualservice names which have multiple `.`. e.g:

```
my-vsvc.my-namespace.svc.cluster.local
```

Namespace will be parsed as `my-namespace.svc.cluster.local` which won't work to subsequent K8s API call

huikang: Thanks. You are right. I switched to using `strings.Split` in the updated code.

jessesuen (approved):

docs/features/traffic-management/istio.md
```diff
@@ -84,6 +85,7 @@ spec:
 
 Finally, a canary and stable Service should be deployed. The selector of these Services will be
 modified by the Rollout during an update to target the canary and stable ReplicaSet pods.
+Note that if virtualservice and destionation host reside in different namespaces (e.g., virtualservice and rollout are not in the same namespace), we should use FQDN as the destination host like `stable-svc.<namespace>`.
```
ngms06, line 88: 1. Nit: Typo in destination
2. Is it always true that if that virtualservice and rollout are not in the same namespace, the virtualservice and destination host reside in different namespaces? I think it is more likely that the virtualservice and service are in the same namespace

PR #1115, opened 2021-04-26
feat: richer prometheus stats and Kubernetes events

jessesuen:
* Introduces a new rollout EventRecorder library to replace kubernetes EventRecorder. The new event recorder additionally logs events with structured logging, and increments a new Prometheus counter for rollout events. This will be the basis for rollout notification support.
* Refactor entire codebase to use new EventRecorder
* Introduce new EventReasons:
    * e.g. RolloutUpdated, RolloutAborted
* Refactor/rename existing EventReasons for consistency. e.g.:
    * AnalysisRunChanged -> AnalysisRunPending
    * Successful -> ExperimentSuccessful
    * SetStable -> RolloutCompleted
    * SetStepIndex -> RolloutStepCompleted
    * ... and more ...
* Consolidates prometheus metric definitions into single file `prommetrics.go`


controller/metrics/experiments.go
```diff
@@ -56,21 +38,18 @@ func (c *experimentCollector) Collect(ch chan<- prometheus.Metric) {
...
-	addGauge(descExperimentInfo, 1)
+	addGauge(MetricExperimentInfo, 1, string(calculatedPhase))
 
-	calculatedPhase := ex.Status.Phase
-	addGauge(descExperimentPhase, boolFloat64(calculatedPhase == v1alpha1.AnalysisPhasePending || calculatedPhase == ""), string("Pending"))
-	addGauge(descExperimentPhase, boolFloat64(calculatedPhase == v1alpha1.AnalysisPhaseError), string("Error"))
-	addGauge(descExperimentPhase, boolFloat64(calculatedPhase == v1alpha1.AnalysisPhaseSuccessful), string("Successful"))
-	addGauge(descExperimentPhase, boolFloat64(calculatedPhase == v1alpha1.AnalysisPhaseRunning), string("Running"))
-	addGauge(descExperimentPhase, boolFloat64(calculatedPhase == v1alpha1.AnalysisPhaseInconclusive), string("Inconclusive"))
+	// DEPRECATED
```
harikrongali, line 49: looks good overall..one comment i have is i see deprecated like  these for metrics..is there a reason why we say deprecated and if so what is the alternative and why do we need to still maintain these if deprecated.

jessesuen: Dashboards (including our own) are still built on these deprecated metrics. So there will be a period of time when not all clusters are migrated to the new metrics. Until then, you need a release which has both metrics so that when 100% of your rollout installations are producing the new metrics, then you can update the dashboard.

rbreeze (approved):

khhirani (approved):
LGTM

Issue #1107, opened 2021-04-23
The rollout controller keeps creating AnalysisRun runs.


saharMlowell:
## Summary 
What happened/what you expected to happen? 
Happened: The rollout controller keeps creating AnalysisRun runs even though they are completed. 
Expected: Canary deploys successfully after AnalysisRun gets completed.

## Diagnostics
What version of Argo Rollouts are you running? 0.10.2 was also able to reproduce with version 0.10.0 or above. The sync works properly using argo-rollouts version `0.8.3`, `0.9.0`, `0.9.1`, `0.9.2` or `0.9.3`. Didn't test it with the versions prior to `0.8.3`.

**To Reproduce**
1. Install any argo-rollouts helm chart version that includes argoproj/argo-rollouts version 0.10.0 or above (latest version at the time of reporting the issue is:  0.10.3).
2. Have `canary` enabled in the deploy.
3. Include some `AnalysisTemplate` as part of the canary deploy.
4. Trigger the rollout by syncing the application via ArgoCD UI.
5. Sync status gets stuck in `Progressing` and the controller keeps creating AnalysisRun.

**Screenshots**

![Screen Shot 2021-04-23 at 11 02 29 AM](https://user-images.githubusercontent.com/26151167/115891371-01beee00-a424-11eb-83d2-b2b680d2c12a.png)

**Additional context**
Simplified AnalysisTemplate:
```
apiVersion: argoproj.io/v1alpha1
kind: AnalysisTemplate
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"argoproj.io/v1alpha1","kind":"AnalysisTemplate", ….}
  labels:
    argocd.argoproj.io/instance: percolator
  managedFields:
  - apiVersion: argoproj.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1: …
    manager: argocd-application-controller
    operation: Update
  name: percolator-router
  namespace: percolator
spec:
  metrics:
  - count: 1
    failureCondition: result[0] == 1
    interval: 1m
    name: malformed_inbound_requests_rate_increased
    provider:
      prometheus:
        address: https://thanos-dt-percolator-sandbox-2-dev.eks.usw2.fanops.net
        query: |-
          (
            sum(rate(requests_inbound_malformed_total{service="percolator-router-canary"}[5m]))
            / sum(rate(requests_inbound_total{service="percolator-router-canary"}[5m]))
          ) > bool (
[...]

PR #1118, opened 2021-04-27
fix: AnalysisRuns could be created continuously due to non-deterministic specs

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1107

When we create an AnalysisRun from the AnalysisTemplate, we were inadvertently randomizing the list of metrics and arguments from the merged AnalysisTemplate. This caused a problem where we would create many AnalysisTemplates, since when we compared the existing run to the run we want to create, it was considered different and we would recreate the run.

This change also partially reverts https://github.com/argoproj/argo-rollouts/commit/ee06d5cc0969049427e1044110ed3a6b2a4a2320, which caused a regression where multiple analysis templates could not be merged properly.


khhirani (approved):
LGTM

PR #1133, opened 2021-04-30
docs: fix missing apiVersion and selector lables in workloadRef

huikang:

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -40,17 +40,6 @@ Istio [DestinationRule Subsets](https://istio.io/latest/docs/reference/config/ne
 instead.
 
 
-# v1.1+
-
-## Webhook Notifications
-
-[Issue #369](https://github.com/argoproj/argo-rollouts/issues/369)
-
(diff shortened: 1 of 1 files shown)


docs/roadmap.md
```diff
@@ -79,8 +79,12 @@ metadata:
 spec:
   replicas: 5
   workloadRef:
+    apiVersion: apps/v1
     kind: Deployment
     name: guestbook
+  selector:
+    matchLabels:
+      app: guestbook
```
jessesuen, line 87: I think we should remove this section since workloadRef is in v1.0

Also, selector is not necessary for Deployments -- if it is a Deployment, we get the selector from the deployment.

huikang: > I think we should remove this section from roadmap since workloadRef is in v1.0
> 
> Also, selector is not necessary for Deployments -- if it is a Deployment, we get the selector from the deployment.

You are right. I was testing using v0.10.2 and saw the error of missing `selector` labels. The latest version does not requires them.

How about moving the section to v1.0?

jessesuen (approved):

Issue #1123, opened 2021-04-27
workloadRef: when referenced Deployment goes missing, we delete pods


jessesuen:
## Summary 

What happened/what you expected to happen?

Using the new workloadRef feature, a user deleted the referenced deployment. When this happened, he noticed that the rollout also deleted/scaled down the RelicaSet which was created from the Deployment definition.

Instead of scaling down the ReplicaSet, we should start emitting Warning events and put rollout in degraded state, and leave the ReplicaSet running, so that it does not cause any downtime.

## Diagnostics

What version of Argo Rollouts are you running? pre v1.0 release (as of today)

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

huikang:
Hi, @jessesuen , I can look into this bug. Please assign it to me. Thanks.

huikang:
> ## Summary
> 
> What happened/what you expected to happen?
> 
> Using the new workloadRef feature, a user deleted the referenced deployment. When this happened, he noticed that the rollout also deleted/scaled down the RelicaSet which was created from the Deployment definition.
> 

Hi, @jessesuen , I tested workloadRef and noted that when the reference depoyment is deleted, the RelicaSet of rollout is still running (which is different than what you described) with the controller emitting the following message

```
ERRO[2021-04-30T00:01:22-04:00] rollout syncHandler error: deployments.apps "default/rollout-demo-deploy" not found  namespace=default rollout=rollouts-demo-canary
E0430 00:01:22.320382    4388 controller.go:172] deployments.apps "default/rollout-demo-deploy" not found
ERRO[2021-04-30T00:01:22-04:00] rollout syncHandler error: deployments.apps "default/rollout-demo-deploy" not found  namespace=default rollout=rollouts-demo-canary
E0430 00:01:22.342081    4388 controller.go:172] deployments.apps "default/rollout-demo-deploy" not found
ERRO[2021-04-30T00:01:22-04:00] rollout syncHandler error: deployments.apps "default/rollout-demo-deploy" not found  namespace=default rollout=rollouts-demo-canary
E0430 00:01:22.387388    4388 controller.go:172] deployments.apps "default/rollout-demo-deploy" not found
```

Did I miss something in reproducing the behavior?

jessesuen:
> @jessesuen , I don't think the controller will remove the replicasets if the referenced deployment is deleted because

[...]

PR #1136, opened 2021-05-04
fix: set rollout status to degraded for non-existent deploy 

huikang:
- the rollout status is set to degraded if the referenced
  deployment doesn't exist, while the replicaset of the
  rollout remains

Address #1123

### DIFF
--- a/rollout/controller.go
+++ b/rollout/controller.go
@@ -353,6 +353,12 @@ func (c *Controller) syncHandler(key string) error {
 	r := remarshalRollout(rollout)
 
 	if err := c.refResolver.Resolve(r); err != nil {
+		invalidSpecCond := conditions.NewRolloutCondition(v1alpha1.InvalidSpec, corev1.ConditionTrue, conditions.InvalidSpecReason, err.Error())
+		cr := r.DeepCopy()
+		cr.Status.Conditions = append(cr.Status.Conditions, *invalidSpecCond)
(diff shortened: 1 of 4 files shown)


rollout/controller.go
```diff
@@ -353,6 +353,12 @@ func (c *Controller) syncHandler(key string) error {
 	r := remarshalRollout(rollout)
 
 	if err := c.refResolver.Resolve(r); err != nil {
+		invalidSpecCond := conditions.NewRolloutCondition(v1alpha1.InvalidSpec, corev1.ConditionTrue, conditions.InvalidSpecReason, err.Error())
+		cr := r.DeepCopy()
+		cr.Status.Conditions = append(cr.Status.Conditions, *invalidSpecCond)
+		_, err = c.argoprojclientset.ArgoprojV1alpha1().Rollouts(r.Namespace).UpdateStatus(
+			context.TODO(), cr, metav1.UpdateOptions{},
+		)
```
jessesuen, line 361: This seems like it would continue to add the InvalidSpec condition on every reconciliation. Can you make it so it doesn't continuously add to the condition list?

Also, a unit test would be nice.

huikang: Thanks for reviewing! I will update the PR.

huikang: @jessesuen , add some unit tests, but still failed to meet the 80% threshold

jessesuen (approved):
Thanks for also including tests!

Issue #1142, opened 2021-05-05
Simplify plugin promote options (deprecate `--skip-current-step`)


jessesuen:
# Summary

We currently have three options to "promote" a rollout in the plugin:

```
        # Promote a paused rollout
        kubectl argo rollouts promote guestbook

        # Fully promote a rollout to desired version, skipping analysis, pauses, and steps
        kubectl argo rollouts promote guestbook --full

        # Skip currently running canary step of a rollout
        kubectl argo rollouts promote guestbook --skip-current-step
```

The different options can be hard to explain/understand, and users may not know when to use which option. To simplify things, I think the behavior of `kubectl argo rollouts promote guestbook --skip-current-step` (which increments the step index) should be merged into the existing `kubectl argo rollouts promote` command so that there are only two options for promote:

* `kubectl argo rollouts promote` - Unpause or promote the rollout to the next step
* `kubectl argo rollouts promote --full` - Fully promote a rollout to desired version, skipping analysis, pauses, and steps 

With this change, users no longer have to think about when to run `kubectl argo rollouts promote` vs. `kubectl argo rollouts promote guestbook --skip-current-step` when they just want the rollout to move past the current step.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Another motivation for this is that the UI does not show `kubectl argo rollouts promote guestbook --skip-current-step` as an option, so the UI would benefit from having only the two options: promote and promote-full

harikrongali:
will pick it up, please assign

jessesuen:
Resolved in https://github.com/argoproj/argo-rollouts/pull/1156

PR #1156, opened 2021-05-10
fix: deprecate skip-current-step command

harikrongali:
Checklist:

* [X] Resolves https://github.com/argoproj/argo-rollouts/issues/1142.
* [X ] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [X ] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)


pkg/kubectl-argo-rollouts/cmd/promote/promote.go
```diff
@@ -150,13 +149,24 @@ func getPatches(rollout *v1alpha1.Rollout, skipCurrentStep, skipAllStep, full bo
...
+			unifiedPatch = []byte(unpauseAndClearPauseConditionsPatch)
+		} else {
+			_, index := replicasetutil.GetCurrentCanaryStep(rollout)
+			// At this point, the controller knows that the rollout is a canary with steps and GetCurrentCanaryStep returns 0 if
+			// the index is not set in the rollout
+			if *index < int32(len(rollout.Spec.Strategy.Canary.Steps)) {
+				*index++
+			}
+			statusPatch = []byte(fmt.Sprintf(setCurrentStepIndex, *index))
+			unifiedPatch = statusPatch
```
jessesuen, line 168: The specPatch and statusPatch to nill out pause and pause conditions should happen regardless of strategy, so I don't think the check for: `if rollout.Spec.Strategy.BlueGreen != nil` should be added. Instead, we should check if we need to increment steps after the existing code

```
...
if len(rollout.Status.PauseConditions) > 0 {
...
}

if rollout.Spec.Strategy.Canary != nil {
    replicasetutil.GetCurrentCanaryStep(rollout)
...
}
```


With the current code, it does not handle the case where 
`spec.paused: true`
and it is at a canary step because code will only do one of the patches.

rollout/pause.go
```diff
@@ -185,7 +185,10 @@ func (pCtx *pauseContext) CompletedCanaryPauseStep(pause v1alpha1.RolloutPause)
 	rollout := pCtx.rollout
 	pauseCondition := getPauseCondition(rollout, v1alpha1.PauseReasonCanaryPauseStep)
 
-	if pause.Duration != nil {
+	if rollout.Status.ControllerPause && pauseCondition == nil {
+		pCtx.log.Info("Rollout has been unpaused")
+		return true
+	} else if pause.Duration != nil {
```
jessesuen, line 191: Can you describe the bug which made this fix necessary? If it's reproducible can we modify an e2e test to catch the bug?

pkg/kubectl-argo-rollouts/cmd/promote/promote.go
harikrongali: that is an OR condition , the first condition will be true for both canary and Bluegreen and else will be for canary only..to avoid confusion, I will make those changes...
one issue is if we increment step, then pause status will be redundant..

rollout/pause.go
harikrongali: if the step is pause with duration, then if we do promote, it wont get promoted as expected as the first condition is checking for duration and if not duration met, it just returns without checking for pause condition.
checking for pause condition first will fix the issue.. i have a test that validated the use case.

and in promote command, if we do step increment always for canary, then this code is redundant..

jessesuen (approved):

Issue #1059, opened 2021-04-01
Traffic Routing ability with Experiment steps


jessesuen:
# Summary

Currently, if you are using canary along with traffic routing, and have an experiment step, you cannot split traffic to the experiment stacks in a fine-grained manner. We should improve the Rollout Experiment step to also allow traffic to be shifted to experiment pods. Something like:

```yaml
    canary:
      steps:
      - experiment:
          templates:
          - name: baseline
            replicas: 1
            specRef: stable
            weight: 5
          - name: canary
            replicas: 1
            specRef: canary
            weight: 5
      - setWeight: 25
      - pause: {duration: 10m}
      - setWeight: 50
      - pause: {duration: 10m}
      - setWeight: 75
      - pause: {duration: 10m}
```

In the above example,  during an update, the first step would start a baseline vs. canary experiment. When pods are ready (Experiment enters Running phase), the rollout would direct 5% of traffic to canary and 5% to baseline, leaving the remaining 90% of traffic to the old stack.

# Use Cases

Today, Experiment steps are only useful with basic canary, but not very useful when used with traffic routing. 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

huikang:
Hi, @jessesuen , I'd like to do some investigation on this enhancement. If no one has been looking into this issue, please assign it to me. Thanks.

jessesuen:
Some design notes:

1. When the experiment step (with weights) is used in conjunction with setWeight, the combined amount will be subtracted from 100 until the experiment is completed. For example:

```yaml
    canary:
      steps:
      - setWeight: 25       # 25% to canary, 75% to stable
      - experiment:           # 25% to canary, 5% to baseline template, 5% to test template, 65% to stable
          templates:
          - name: baseline
            replicas: 1
            specRef: stable
            weight: 5
          - name: test
            replicas: 1
            specRef: canary
            weight: 5
      - pause: {}            # back to 25% canary, 75% stable
```

2. SetWeight will be updated to take an *optional* list of additional destinations (excluding canary), in which to split traffic:

```golang
type TrafficRoutingReconciler interface {
	// SetWeight sets the canary weight to the desired weight.
	// Optionally splits the weight to N additional destinations. Combined weights must not exceed 100.
	// Remaining weight will direct to the stable.
	SetWeight(desiredWeight int32, additionalDestinations ...WeightDestination) error
...
}

type WeightDestination struct {
[...]

PR #1158, opened 2021-05-11
feat: ability to auto-create Services for each template in an Experiment

khhirani:
As described in the comments of #1059 


experiments/service.go
```diff
@@ -0,0 +1,99 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var experimentKind = v1alpha1.SchemeGroupVersion.WithKind("Experiment")
+
+func (c *Controller) getServicesForExperiment(experiment *v1alpha1.Experiment) (map[string]*corev1.Service, error) {
+	svcList, err := c.serviceLister.Services(experiment.Namespace).List(labels.Everything())
```
harikrongali, line 21: will it throw error if there are no services..as in current flow, if create service is false, then we. don't create services...so do we need to check for the flag and return nil?

experiments/service_test.go
```diff
@@ -0,0 +1,13 @@
+package experiments
+
+import (
+	"testing"
+)
+
+func TestCreatedServiceHasExpectedAnnotations(t *testing.T) {
```
harikrongali, line 7: can you add negative tests for capturing errors/failures?

experiments/replicaset.go
```diff
@@ -199,6 +204,56 @@ func (ec *experimentContext) isReplicaSetSemanticallyEqual(newRS, existingRS *ap
...
+		}
+		return rsIsUpdated, nil
+	} else {
+		// If RS already has non-zero scaleDownDelayDeadline set, then we don't do anything
+		if replicasetutil.HasScaleDownDeadline(rs) {
+			return rsIsUpdated, nil
+		}
+	}
+
+	deadline := metav1.Now().Add(scaleDownDelaySeconds * time.Second).UTC().Format(time.RFC3339)
```
harikrongali, line 230: scaleDownDelaySeconds * time.Second ..what this mean? is it not addition instead of multiplying?

harikrongali:
can you add/update e2e tests with adding setWeight and scaleDownDelay seconds?

experiments/replicaset.go
khhirani: `deadline := metav1.Now().Add(scaleDownDelaySeconds * time.Second).UTC().Format(time.RFC3339)`

This line creating a deadline is adding `scaleDownDelaySeconds` to the current time.

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -413,6 +413,8 @@ type RolloutExperimentTemplate struct {
 	// use the same selector as the Rollout
 	// +optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty" protobuf:"bytes,5,opt,name=selector"`
+	// Weight sets the percentage of traffic the template's replicas should receive
+	Weight *int32 `json:"weight,omitempty" protobuf:"varint,6,opt,name=weight"`
```
jessesuen, line 417: To keep things cleaner, can we move the introduction of this field with the separate PR which integrates experiments with traffic routing weights?

experiments/experiment.go
```diff
@@ -132,9 +144,56 @@ func (ec *experimentContext) reconcileTemplate(template v1alpha1.TemplateSpec) {
...
+					if err != nil {
+						ec.log.Warnf("Unable to get rs '%s' with added scaleDownDelay", rs.Name)
+					} else {
+						rs = modifiedRS
+						ec.templateRSs[template.Name] = modifiedRS
+					}
+				}
+			}
+
+		}
```
jessesuen, line 165: The addition of service creation to `reconcileTemplate` has nearly doubled the length of this method, making it quite complex.

Can we refactor this to better organize service creation into its own set of method(s)? This should make it easier to unit test as well.

experiments/experiment.go
```diff
@@ -132,9 +144,56 @@ func (ec *experimentContext) reconcileTemplate(template v1alpha1.TemplateSpec) {
...
+					scaleDownAtTime, err := time.Parse(time.RFC3339, scaleDownAtStr)
+					if err != nil {
+						ec.log.Warnf("Unable to read scaleDownAt label on rs '%s'", rs.Name)
+					}
+					now := metav1.Now()
+					scaleDownAt := metav1.NewTime(scaleDownAtTime)
+					if scaleDownAt.After(now.Time) {
+						ec.log.Infof("RS '%s' has not reached the scaleDownTime", rs.Name)
+						remainingTime := scaleDownAt.Sub(now.Time)
+						if remainingTime < ec.resyncPeriod {
```
jessesuen, line 180: Instead of copying and pasting this code which is now duplicated in rollout controller, the scaledown helper logic should be moved to its own set of helper libraries that can be reused between the  rollout and experiment controller.

pkg/apis/rollouts/v1alpha1/experiment_types.go
```diff
@@ -111,6 +120,10 @@ type TemplateStatus struct {
 	// LastTransitionTime is the last time the replicaset transitioned, which resets the countdown
 	// on the ProgressDeadlineSeconds check.
 	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty" protobuf:"bytes,9,opt,name=lastTransitionTime"`
+	// ServiceName is the name of the service which corresponds to this experiment
+	ServiceName string `json:"serviceName,omitempty" protobuf:"bytes,10,opt,name=serviceName"`
+	// PodTemplateHash is the
```
jessesuen, line 125: incomplete sentence.

pkg/apis/rollouts/v1alpha1/experiment_types.go
```diff
@@ -67,6 +74,8 @@ type TemplateSpec struct {
 	Selector *metav1.LabelSelector `json:"selector" protobuf:"bytes,4,opt,name=selector"`
 	// Template describes the pods that will be created.
 	Template corev1.PodTemplateSpec `json:"template" protobuf:"bytes,5,opt,name=template"`
+	// CreateService determines if a service should be created for the template
+	CreateService bool `json:"createService,omitempty" protobuf:"varint,6,opt,name=createService"`
```
jessesuen, line 78: I think a boolean might be short-sighted in case we need knobs to control **_how_** services should be generated (e.g. with a specific name) or possibly referenced and modified (like blue-green activeService).

Instead of a simple boolean, I think we can use an empty struct to indicate it should be created:

```yaml
spec:
  templates:
  - name: foo
    service: {}
```

Then in the future (not as a part of this PR), we could introduce:

```yaml
spec:
  templates:
  - name: foo
    service:
      name: foo
```

or:

```yaml
spec:
  templates:
  - name: foo
    service:
      generateName: foo-
```

experiments/service.go
```diff
@@ -0,0 +1,99 @@
...
+			}},
+		},
+	}
+
+	service, err := ec.kubeclientset.CoreV1().Services(ec.ex.Namespace).Create(ctx, newService, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return service, nil
+}
```
jessesuen, line 82: This logic is not idempotent. Meaning if we create the service and it already exists (and it's ours), then we should not fail. Can you make sure we handle that case.

experiments/experiment.go
```diff
@@ -151,6 +210,31 @@ func (ec *experimentContext) reconcileTemplate(template v1alpha1.TemplateSpec) {
...
 
+	// Create Service for template
+	// Use same Name and rollout-pod-template-hash as ReplicaSet
+	if template.CreateService && rs != nil {
+		podTemplateHash := rs.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
+		svc := ec.templateServices[template.Name]
+		if desiredReplicaCount == 0 {
+			ec.deleteTemplateService(svc, templateStatus, template.Name)
+		} else if svc == nil || svc.Name != rs.Name {
+			newService, err := ec.CreateService(rs.Name, template, rs.Labels)
```
jessesuen, line 221: Due to the way ingress controllers work such as AWS LoadBalancer controller, Service creation needs to actually happen **_before_** pods are created, otherwise they will not get readiness gates and won't be targetted properly by the load balancer. 

So service creation needs to be moved to happen *before* we create any pods from the template.

I think the best way to handle this is to:

1. Create the ReplicaSet with `spec.replicas: 0`
2. Get PodTemplateHash of ReplicaSet (to use in Service creation)
3. Scale up ReplicaSet to desired count.

This will require some refactoring.

jessesuen (changes requested):

jessesuen (approved):
Great job!

Issue #941, opened 2021-01-13
Option to scale down current stack upon abort


artem-nefedov:
## Summary 

What happened/what you expected to happen?

If you run analysis as a step after `setCanaryScale` and it fails, canary replicas created by `setCanaryScale` will run indefinitely (until new version is deployed) after rollback is performed, even though they are never going to be used for anything. It should either be cleaned up by default (similar to replicas created by `setWeight`), or there should be an option to do it.

In the described scenario, there are no `setWeight` steps prior to `setCanaryScale`.

## Diagnostics

What version of Argo Rollouts are you running?

v0.10.2

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
When you say "after rollback is performed" do you mean when the rollout is aborted? If so, this is the current design. The rationale was that retrying analysis would not have to redo all the spin-up time of scheduling pods, and warm up. Note that BlueGreen has a similar behavior in that the new stack is not terminated when rollback is aborted caused by analysis failure. 

> or there should be an option to do it.

This is not the first time this behavior has been requested. If we were to introduce a flag like `scaleDownOnAbort: true` would that satisfy your use case?

artem-nefedov:
> When you say "after rollback is performed" do you mean when the rollout is aborted?

Yes, I mean when it is aborted automatically due to new version failing on analysis.

> If we were to introduce a flag like scaleDownOnAbort: true would that satisfy your use case?

Yes, I believe that would do it.

huikang:
@jessesuen , do you think the proposed change is also applicable to an aborted b/g rollout?

huikang:
This seems related to #793.

@jessesuen , I will write up a change proposal to handle these user stories. Please assign this to me. Thanks.

huikang:
@jessesuen , I think we can close this issue.

jessesuen:
Fixed in v1.1

Issue #793, opened 2020-10-19
Scale down preview replica set when the analysis run fails in bg


jgquach3:
# Summary
- Currently within bg, the preview replica set does not scale down when an analysis run fails. 

What change needs making?
- The scale down annotation can be added on the preview replica set in the case of an analysis run failure.
PR - https://github.com/argoproj/argo-rollouts/compare/master...jgquach3:preview-scale-down?expand=1
(currently by default it will scale down, but another key can be added onto the rollout CR to add the annotation or not)

# Use Cases

When would you use this?
If a preview replica set is left up, it could be taking up k8s resources. It also makes sense to remove the preview in prod settings to prevent potential problems. 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Closing as duplicate of https://github.com/argoproj/argo-rollouts/issues/941

PR #1160, opened 2021-05-11
feat: introduce abortScaleDownDelaySeconds to control scale down of preview/canary upon abort

huikang:
Address #941 #793

### DIFF
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -115,9 +115,9 @@ spec:
       # down. Defaults to nil
       scaleDownDelayRevisionLimit: 2
 
-      # Adds a delay before scaling down the p preview replicaset if update is
-      # aborted. Default is 0, meaning preview replicaset  are not scaled down.
-      AbortScaleDownDelaySeconds: 30
+      # Adds a delay before scaling down the preview replicaset if update is
+      # aborted. Default is 0, meaning preview replicaset won't be scaled down.
+      abortScaleDownDelaySeconds: 30
 
       # Anti Affinity configuration between desired and previous ReplicaSet.
       # Only one must be specified
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -58,6 +58,11 @@ spec:
   # than or equal to this value.
   restartAt: "2020-03-30T21:19:35Z"
 
+  # Scale down the preview replicasets or canary replicasets (if
+  # canary's setCanaryScale.replicas is set) on aborted update
+  # Defaults to false
+  scaleDownOnAbort: false
+
   strategy:
 
     # Blue-green update strategy
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -58,11 +58,6 @@ spec:
   # than or equal to this value.
   restartAt: "2020-03-30T21:19:35Z"
 
-  # Scale down the preview replicasets or canary replicasets (if
-  # canary's setCanaryScale.replicas is set) on aborted update
-  # Defaults to false
-  scaleDownOnAbort: false
-
   strategy:
 
     # Blue-green update strategy
@@ -119,6 +114,10 @@ spec:
       # Limits the number of old RS that can run at once before getting scaled
       # down. Defaults to nil
       scaleDownDelayRevisionLimit: 2
+
+      # Adds a delay before scaling down the p preview replicaset if update is
+      # aborted. Default is 0, meaning preview replicaset  are not scaled down.
+      AbortScaleDownDelaySeconds: 30
 
       # Anti Affinity configuration between desired and previous ReplicaSet.
       # Only one must be specified
@@ -299,6 +298,11 @@ spec:
         smi:
           rootService: root-svc # optional
           trafficSplitName: rollout-example-traffic-split # optional
+
+      # Add a delay before scaling down the canary pods when update
+      # is aborted for canary strategy using replicas of setCanaryScale.
+      # Default is 0, meaning canary pods are not scaled down.
+      AbortScaleDownDelaySeconds: 30
 
 status:
   pauseConditions:
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -115,8 +115,8 @@ spec:
       # down. Defaults to nil
       scaleDownDelayRevisionLimit: 2
 
-      # Adds a delay before scaling down the preview replicaset if update is
-      # aborted. Default is 0, meaning preview replicaset won't be scaled down.
+      # Add a delay in second before scaling down the preview replicaset
+      # if update is aborted. 0 means not to scale down. Default is 30 second
       abortScaleDownDelaySeconds: 30
 
       # Anti Affinity configuration between desired and previous ReplicaSet.
@@ -299,10 +299,10 @@ spec:
           rootService: root-svc # optional
           trafficSplitName: rollout-example-traffic-split # optional
 
-      # Add a delay before scaling down the canary pods when update
-      # is aborted for canary strategy using replicas of setCanaryScale.
-      # Default is 0, meaning canary pods are not scaled down.
-      AbortScaleDownDelaySeconds: 30
+      # Add a delay in second before scaling down the canary pods when update
+      # is aborted for canary strategy with traffic routing (not applicable for basic canary).
+      # 0 means canary pods are not scaled down. Default is 30 seconds.
+      abortScaleDownDelaySeconds: 30
 
 status:
   pauseConditions:
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -287,7 +287,7 @@ type CanaryStrategy struct {
 	// +optional
 	ScaleDownDelayRevisionLimit *int32 `json:"scaleDownDelayRevisionLimit,omitempty" protobuf:"varint,12,opt,name=scaleDownDelayRevisionLimit"`
 	// AbortScaleDownDelaySeconds adds a delay before scaling down the canary pods when update
-	// is aborted for canary strategy using replicas of setCanaryScale.
+	// is aborted for canary strategy with traffic routing (not applicable for basic canary).
 	// Default is 0, meaning canary pods are not scaled down.
 	// +optional
 	AbortScaleDownDelaySeconds *int32 `json:"abortScaleDownDelaySeconds,omitempty" protobuf:"varint,13,opt,name=abortScaleDownDelaySeconds"`
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -202,8 +202,9 @@ type BlueGreenStrategy struct {
 	// ActiveMetadata specify labels and annotations which will be attached to the active pods for
 	// the duration which they act as a active pod, and will be removed after
 	ActiveMetadata *PodTemplateMetadata `json:"activeMetadata,omitempty" protobuf:"bytes,13,opt,name=activeMetadata"`
-	// AbortScaleDownDelaySeconds adds a delay before scaling down the preview replicaset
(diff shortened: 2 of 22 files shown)


huikang:
Hi, @jessesuen , could you please review this PR at your convenience? Thanks.

jessesuen:
Sure I'll take a look after v1.0 release this week. This will need to wait for v1.1 since v1.0 is about to release.

huikang:
Hi, @jessesuen , I added two test cases in e2e. Please take a look when you can. Thanks.

jessesuen:
Jad a chance to look at this. The use case is valid, but I'd like to tweak the spec a bit to make this fit with the current design. Some considerations:

1. If we scale down the canary/desired ReplicaSet immediately upon the abort, there is a possibility of 500 routing errors due to the fact that mesh/ingress controllers can take time to shift traffic. Usually, this happens within seconds, but in general, you want to give it 30s minimum.

2. I think the time frame in which to scale down the canary after abort should be configurable. A simple boolean would make retrying an aborted rollout go through the whole scale up process and pod provisioning phase.

Therefore, I am suggesting the following change:

Instead of a `scaleDownOnAbort: true` boolean. I think this should be a duration field, which acts similarly to `spec.strategy.canary.scaleDownDelaySeconds`.

```yaml
spec:
  strategy:
    canary:
      abortScaleDownDelay: 1h
``` 

When a rollout aborts, it would immediately attach a deadline timestamp in the future to the ReplicaSet. At that timestamp, the rollout controller will scaledown the canary stack. During that delay, traffic would have already shifted back to the stable stack.

Having a delay also allows Rollouts to be retried much easier/quicker. If the Rollout is retried within the abortScaleDownDelay, then there is no time wasted provisioning pods because the ReplicaSet of the canary is already running.

huikang:
@jessesuen , thanks for detailing the proposed design. I agree that having a scaledown delay duration would improve the traffic shitting cases. Following your example of canary, I think the spec for b/g would be

```
spec:
  strategy:
    blueGreen:
      abortScaleDownDelay: 30s
```

Is my understanding correct?

huikang:
Hi, @jessesuen , please take a look at the PR at your convenience. Thanks.

rollout/replicaset.go
```diff
@@ -120,10 +119,53 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 	if err != nil {
 		return false, err
 	}
+
+	abortScaleDownDelaySeconds := time.Duration(defaults.GetAbortScaleDownDelaySecondsOrDefault(c.rollout))
+	if c.isScaleDownOnabort() {
+		c.log.Info("scale down on abort for the newRS")
```
ngms06, line 125: Would it be helpful to add the newRS being scaled down to the log line?

rollout/replicaset.go
```diff
@@ -120,10 +119,53 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
...
+
+		// if the newRS has scale down annotation and to be scaled down now, set
```
ngms06, line 127: Is the comma supposed to be after set and not now at the end of the comment?

rollout/replicaset.go
```diff
@@ -120,10 +119,53 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
...
+		// Update newReplicasCount = int32(0)
+		if scaleDownAtStr, ok := c.newRS.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey]; ok {
+			c.log.Infof("New rs '%s' has scaledown delay annotation: %s", c.newRS.Name, scaleDownAtStr)
```
ngms06, line 130: The comment says "rs has **delay** annotation, but the previous line checks for **deadline** annotation (DefaultReplicaSetScaleDownDeadlineAnnotationKey). Is that correct?

rollout/replicaset.go
```diff
@@ -120,10 +119,53 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
...
+			scaleDownAtTime, err := time.Parse(time.RFC3339, scaleDownAtStr)
+			if err != nil {
+				c.log.Warnf("Unable to read scaleDownAt label on rs '%s'", c.newRS.Name)
+			} else {
+				now := metav1.Now()
+				scaleDownAt := metav1.NewTime(scaleDownAtTime)
+				if scaleDownAt.After(now.Time) {
+					c.log.Infof("RS '%s' has not reached the scaleDownTime", c.newRS.Name)
```
ngms06, line 138: Should we add scaleDownAt to log line?

docs/features/specification.md
```diff
@@ -115,6 +115,10 @@ spec:
       # down. Defaults to nil
       scaleDownDelayRevisionLimit: 2
 
+      # Adds a delay before scaling down the p preview replicaset if update is
```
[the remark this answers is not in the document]
huikang: Fixed

rollout/replicaset.go
huikang: Since line 129 logs the scale down time `scaleDownAtStr`, I think we don't replicate it here.

huikang (commented):
@ngms06, thanks for reviewing. The PR is updated.

rollout/replicaset.go
huikang: The `DefaultReplicaSetScaleDownDeadlineAnnotationKey` is correct. Updated the log message to make it clear.

rollout/replicaset.go
huikang: Updated the log message. Thanks.

rollout/replicaset.go
huikang: log message is updated. Thanks.

jessesuen:
@huikang I'm finally reviewing this now but during my review, I realized some of my assumptions of v1.0 behavior were not correct. It appears we **_already_** scale down canary upon abort, but a bug is preventing this from happening when used with `setCanaryScale`.

So this PR will be our opportunity to get some consistency of our behavior between all the strategies (blue-green, canary, canary w/ traffic routing, etc...).

Let me test some behavior with the various options and document what the desired behavior should/will be with all the options:
* abortScaleDownDelaySeconds = nil
* abortScaleDownDelaySeconds = 0
* abortScaleDownDelaySeconds > 0

docs/features/specification.md
```diff
@@ -295,6 +299,11 @@ spec:
           rootService: root-svc # optional
           trafficSplitName: rollout-example-traffic-split # optional
 
+      # Add a delay before scaling down the canary pods when update
+      # is aborted for canary strategy using replicas of setCanaryScale.
+      # Default is 0, meaning canary pods are not scaled down.
+      AbortScaleDownDelaySeconds: 30
```
jessesuen, line 305: abortScaleDownDelaySeconds  should be lowercased

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -282,6 +286,11 @@ type CanaryStrategy struct {
 	// ScaleDownDelayRevisionLimit limits the number of old RS that can run at one time before getting scaled down
 	// +optional
 	ScaleDownDelayRevisionLimit *int32 `json:"scaleDownDelayRevisionLimit,omitempty" protobuf:"varint,12,opt,name=scaleDownDelayRevisionLimit"`
+	// AbortScaleDownDelaySeconds adds a delay before scaling down the canary pods when update
+	// is aborted for canary strategy using replicas of setCanaryScale.
+	// Default is 0, meaning canary pods are not scaled down.
```
jessesuen, line 291: Let's add a comment stating that this field is only used with traffic routing and not applicable for basic canary.

rollout/replicaset.go
```diff
@@ -120,10 +119,52 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 	if err != nil {
 		return false, err
 	}
+
+	abortScaleDownDelaySeconds := time.Duration(defaults.GetAbortScaleDownDelaySecondsOrDefault(c.rollout))
```
jessesuen, line 123: Let's move this inside the else clause since it's only used there.

jessesuen (changes requested):
Sorry for the delay in reviewing this. So here is the matrix of behavior for when abort happens, and what I feel the v1.1 behavior should be:

| strategy | v1.0 behavior | abortScaleDownDelaySeconds | desired v1.1 behavior |
|---------|---------|---------|---------|
| blue-green | does not scale down | nil | scales down after 30 seconds |
| blue-green | does not scale down | 0 | does not scale down |
| blue-green | does not scale down | N | scales down after N seconds |
| basic canary | scales down |  N/A | scales down |
| canary w/ traffic routing | scales down immediately |  nil | scales down after 30 seconds |
| canary w/ traffic routing | scales down immediately |  0 | does not scale down |
| canary w/ traffic routing | scales down immediately |  N | scales down after N seconds |
| canary w/ traffic routing + setCanaryScale | does not scale down (bug) | * | should behave like above |

@huikang  - I think v1.1 should make the behavior consistent across all strategies. Basically, upon an abort, we should scale down the new replicaset for all strategies. Users can then choose to leave the new replicaset scaled up indefinitely by setting  `abortScaleDownDelaySeconds` to 0, or adjust the value to something larger (or smaller).

huikang:
> Sorry for the delay in reviewing this. So here is the matrix of behavior for when abort happens, and what I feel the v1.1 behavior should be:
> strategy 	v1.0 behavior 	abortScaleDownDelaySeconds 	desired v1.1 behavior
> blue-green 	does not scale down 	nil 	scales down after 30 seconds
> blue-green 	does not scale down 	0 	does not scale down
> blue-green 	does not scale down 	N 	scales down after N seconds
> basic canary 	rolling update back to stable 	N/A 	rolling update back to stable
> canary w/ traffic routing 	scales down immediately 	nil 	scales down after 30 seconds
> canary w/ traffic routing 	scales down immediately 	0 	does not scale down
> canary w/ traffic routing 	scales down immediately 	N 	scales down after N seconds
> canary w/ traffic routing + setCanaryScale 	does not scale down (bug) 	* 	should behave like above
> 
> @huikang - I think v1.1 should make the behavior consistent across all strategies. Basically, upon an abort, we should scale down the new replicaset for all strategies. Users can then choose to leave the new replicaset scaled up indefinitely by setting `abortScaleDownDelaySeconds` to 0, or adjust the value to something larger (or smaller).


@jessesuen , thanks for summarizing. I will update the PR to reflect the descried behavior as listed in the table. In addition, I will update the doc to include the table.

huikang:
Hi, @jessesuen , the PR has been updated based upon the desired behavior of v1.1. Please help review at your convenience. Thanks.

jessesuen (approved):
Great work!

dudadornelles:
Was this released? Really looking forward to use this feature, thanks for the hard work folks!

alexandresavicki:
> Was this released? Really looking forward to use this feature, thanks for the hard work folks!

I'm try to use this parameter using version 1.0.4 but i'm getting error:
```
# rollouts.argoproj.io "dummy-app" was not valid:
# * : Invalid value: "The edited file failed validation": ValidationError(Rollout.spec.strategy.canary): unknown field "abortScaleDownDelaySeconds" in io.argoproj.v1alpha1.Rollout.spec.strategy.canary
```

jessesuen:
Not yet. It's a v1.1 feature. You can try the feature out by installing the manifests from master, which use the `:latest` images

alexandresavicki:
@jessesuen Thanks a lot for your reply. Just one more doubt i'm looking at this parameter trying bypass some downtime when canary fails for any reason (eg.: Analysis step not match specified conditions) so canary will be aborted and rollback will be started. I understand the rollback process it's very aggressive such as traffic shift it's slow and can cause downtime in my application.
There is an another way to  bypass this scenario?

PR #1176, opened 2021-05-14
feat: lint supporting rollout in multiple doc

huikang:
- to support use cases where the yaml file may contain other objects like
  deploy, service, etc

The current lint implementation does not report error if an invalid rollout yaml contains another valid resource like service - added example in unit test.


jessesuen (approved):
Great improvement!

PR #1178, opened 2021-05-14
feat: record events for RolloutPaused and RolloutResumed

jessesuen:
This emits kubernetes events for:
* when a rollout becomes paused by the controller (RolloutPaused)
* when a rollout automatically resumes after a pause (RolloutResumed)

Also adds more details about events for NewReplicaSetCreated and ScalingReplicaSet

### DIFF
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -250,7 +250,8 @@ func (c *rolloutContext) createDesiredReplicaSet() (*appsv1.ReplicaSet, error) {
 	}
 
 	if !alreadyExists && newReplicasCount > 0 {
-		c.recorder.Eventf(c.rollout, record.EventOptions{EventReason: conditions.ScalingReplicaSetReason}, "Scaled up replica set %s to %d", createdRS.Name, newReplicasCount)
+		revision, _ := replicasetutil.Revision(createdRS)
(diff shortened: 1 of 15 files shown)


rollout/sync.go
```diff
@@ -698,16 +701,35 @@ func (c *rolloutContext) persistRolloutStatus(newStatus *v1alpha1.RolloutStatus)
...
+
+	c.sendStateChangeEvents(&prevStatus, newStatus)
 	logCtx.Infof("Patched: %s", patch)
 	c.newRollout = newRollout
 	return nil
 }
 
+// sendStateChangeEvents emit rollout events on significant state changes
+func (c *rolloutContext) sendStateChangeEvents(prevStatus, newStatus *v1alpha1.RolloutStatus) {
+	prevPaused := len(prevStatus.PauseConditions) > 0
```
harikrongali, line 719: i see in lot of places we check for len of pausedconditions..is there a case where there can be more than one pause conditions applies?

harikrongali:
LGTM apart from one clarification

khhirani (approved):
LGTM

rollout/sync.go
jessesuen: Good question. A background analysis run could pause a canary rollout because of inconclusive analysis, but a rollout could also be automatically paused from a pause step. If the pause step had a duration, we want to automatically clear the pause step (by removing that condition), but leave the pause condition for inconclusive analysis.

Issue #1180, opened 2021-05-14
rollout status doesn't wait until the end after restart


DanTulovsky:
## Summary 

I triggered a `restart` and then ran `rollout status`, but it finished with `healthy` before the rollout was complete. It looked like it was returning `healthy` after each individual Pod was restarted... Is this just a bad interaction with `restart`?

## Diagnostics

What version of Argo Rollouts are you running?

1.00-rc1

Here's the log of my terminal. The rollout was running throughout this whole time.

```
% k argo rollouts --namespace "myns" restart myns-service                                                                                                                                    
rollout 'myns-service' restarts in 0s
dant@imacLS :) [12:02:56] 
% k argo rollouts --namespace "myns" status myns-service --timeout=600                                                                                                                       
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - waiting for rollout spec update to be observed
Healthy
dant@imacLS :) [12:03:39] 
% k argo rollouts --namespace "myns" status myns-service --timeout=600                                                                                                                       
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Healthy
dant@imacLS :) [12:04:38]
% k argo rollouts --namespace "myns" status myns-service --timeout=600                                                                                                                       
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Healthy
dant@imacLS :) [12:05:34] 
% k argo rollouts --namespace "myns" status myns-service --timeout=600                                                                                                                       
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Progressing - updated replicas are still becoming available
Healthy
```

```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME>
```

---

**Message from the maintainers**:

[...]

PR #1186, opened 2021-05-14
feat: rollout phase/message calculated server-side

jessesuen:
Currently, all clients which need to assess the health of a rollout (such as the plugin, as well as the Argo CD health lua script), need to perform a complex calculation by evaluating `status.availableReplicas`, etc...

This logic is hard to repeat outside of these clients. Also, the logic could become out-of-date over time which makes it necessary for clients to be updated with controller. This PR introduces two status fields: phase and message, which makes client side calculation simpler. e.g.:

```yaml
status:
  phase: Progressing
  message: waiting for all steps to complete
```

Phase is one of four states: `Healthy`, `Paused`, `Progressing`, `Degraded` and message.

The clients would then be reduced to only performing two checks:
1. Ensure `status.observedGeneration` == `metadata.generation`, if not the controller has not seen the update to spec
2. If observedGeneration is current, then `status.phase` and `status.message` could be surfaced

This PR also resolves https://github.com/argoproj/argo-rollouts/issues/1180


harikrongali:
LGTM..

khhirani (approved):
LGTM

PR #1190, opened 2021-05-17
fix: Add edge case handling to traffic routing

zezaeoh:
Checklist:

### DIFF
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -100,8 +100,8 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 		// when we are fully promoted. desired canary weight should be 0
 	} else if c.pauseContext.IsAborted() {
 		// when promote aborted. desired canary weight should be 0
-	} else if c.newRS.Status.AvailableReplicas == 0 {
-		// when newRS available replicas num is 0. never weight to canary
+	} else if c.newRS == nil || c.newRS.Status.AvailableReplicas == 0 {
+		// when newRS is not available or replicas num is 0. never weight to canary
 	} else if index != nil {
 		atDesiredReplicaCount := replicasetutil.AtDesiredReplicaCountsForCanary(c.rollout, c.newRS, c.stableRS, c.otherRSs)
 		if !atDesiredReplicaCount {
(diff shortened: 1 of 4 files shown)


rollout/trafficrouting.go
```diff
@@ -98,6 +98,10 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 	desiredWeight := int32(0)
 	if c.rollout.Status.StableRS == c.rollout.Status.CurrentPodHash {
 		// when we are fully promoted. desired canary weight should be 0
+	} else if c.pauseContext.IsAborted() {
+		// when promote aborted. desired canary weight should be 0
+	} else if c.newRS.Status.AvailableReplicas == 0 {
+		// when newRS available replicas num is 0. never weight to canary
```
jessesuen, line 104: This change is significant, and a surprising oversight if we are not handling this correctly. I will take some time to understand the implications of the change.

jessesuen:
Looks like we have a few tests (both unit and e2e) that are asserting the opposite of this change. So at the very least the existing tests would need to change.

But we would also want new tests to verify this change. This is assuming the change is correct -- I still need to play around and what the behavior should be vs. what is actually happening.

zezaeoh:
Hi, @jessesuen 
Thanks a lot for your review.

I would like to cover next cases with this changes

1. abort case
```go
	} else if c.pauseContext.IsAborted() {		
		// when promote aborted. desired canary weight should be 0	
	} ...
``` 
---
when promotion is aborted, controller scale down replicas before change canary weight to 0
because of this logic
```go
                atDesiredReplicaCount := replicasetutil.AtDesiredReplicaCountsForCanary(c.rollout, c.newRS, c.stableRS, c.otherRSs)
		if !atDesiredReplicaCount {
			// Use the previous weight since the new RS is not ready for a new weight
			for i := *index - 1; i >= 0; i-- {
				step := c.rollout.Spec.Strategy.Canary.Steps[i]
				if step.SetWeight != nil {
					desiredWeight = *step.SetWeight
					break
				}
			}
		} ...
```
this behavior leads some request drops 

2. when promotion aborted and then start `promotefull` action
```go
	} else if c.newRS == nil || c.newRS.Status.AvailableReplicas == 0 {
		// when newRS available replicas num is 0. never weight to canary
	} ...
```
---
Suppose we have a rollout resource with a canary strategy like this:
```yaml
    canary:
      steps:
        - setWeight: 20
        - pause: {}
        - setWeight: 40
        - pause: {}
        - setWeight: 60
        - pause: {}
        - setWeight: 100
        - pause: {}
```

when promotion is aborted, then canary service's endpoint will be empty.
If `promotefull` action is executed in this situation,
despite the fact that newRS has just been created, trafficrouting sets the canary weight to 100

this behavior leads request drops

jessesuen:
Thanks for the explanation. The change makes sense now and looks like all tests are passing.

jessesuen (approved):

PR #1197, opened 2021-05-18
chore: use a base image that has a shell for the kubectl plugin image

keithmattix:
Checklist:

* [x ] My builds are green. Try syncing with master if they are not. 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,7 +69,7 @@ RUN make ${MAKE_TARGET}
 ####################################################################################################
 # Kubectl plugin image
 ####################################################################################################
-FROM ubuntu:20.10 as kubectl-argo-rollouts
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -69,7 +69,7 @@ RUN make ${MAKE_TARGET}
 ####################################################################################################
 # Kubectl plugin image
 ####################################################################################################
-FROM scratch as kubectl-argo-rollouts
+FROM ubuntu:20.10 as kubectl-argo-rollouts
```
jessesuen, line 72: Could you use the fully qualified name `docker.io/library/ubuntu:20.10` instead? This makes it more compatible in environments which do not assume dockerhub

jessesuen (approved):

PR #1198, opened 2021-05-19
feat: add workload-ref/generation to rollout

### LINKED ISSUE #1189: workloadRef rollouts should record `status.workloadObservedGeneration`
# Summary

Rollouts that use the new workloadRef feature, currently do not benefit from client-side checks where `status.observedGeneration == metadata.generation` because changes to the referenced workload object, do not increment the rollout's `metadata.generation`. This can cause a rollout's phase (Healthy/Degraded/Progressing/Paused) to be interpreted incorrectly because a client has no way to understand if both the rollout's & workload's spec has been observed by the rollout controller.

To support more reliable detection of changes to workloadRef resources, to help detect when the workload's spec has been observed by the rollout controller, I think we can  make the following improvements:

1. Currently the informer which is watching workload resources simply enqueues the relevant rollout for reconciliation. Instead of simply enqueuing the rollout to the workqueue, it should instead *patch* the rollout with the `metadata.generation` of the workload under a new rollout annotation. e.g.:

```yaml
annotations:
  rollout.argoproj.io/workload-generation: "5"
```

2. Then when reconciling a rollout using workloadRefs, the rollout controller would then record that generation under the rollout status:

```yaml
status:
  workloadObservedGeneration: 5
```

3. Clients such as the plugin and Argo CD should then be improved to check both:
* `metadata.generation == status.observedGeneration`
* `metadata.annotations["rollout.argoproj.io/workload-generation"] == `status.workloadObservedGeneration`

in order to decide if both the rollout spec and the workload spec has been observed by the controller

It should be noted that while this technique will **_increase_** the probability that workload spec changes are observed by the rollout controller, it does not **_guarantee_** it. This is because it still relies on a running controller to patch the rollout upon workload spec changes. That said, the addition of a controller dedicated to patching the rollout upon spec changes, should help make it generally more reliable because it would not be affected by say saturation of the rollout workers.

---

huikang:
- Add a "rollout.argoproj.io/workload-generation" annotation to
  the rollout metadata, which equals to the generation of reference
  workload
- workload-generation is updated when the referenced workload is
  updated.
- status.workloadObservedGeneration records the observed generation
  of the rollout
- e2e test is updated

close #1189 

Checklist:

### DIFF
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -691,6 +691,14 @@ func (c *rolloutContext) persistRolloutStatus(newStatus *v1alpha1.RolloutStatus)
 
 	prevStatus := c.rollout.Status
 	c.pauseContext.CalculatePauseStatus(newStatus)
+	if c.rollout.Spec.TemplateResolvedFromRef {
+		workloadRefObservation, _ := annotations.GetWorkloadGenerationAnnotation(c.rollout)
+		currentWorkloadObservedGeneration, _ := strconv.ParseInt(newStatus.WorkloadObservedGeneration, 10, 32)
+		if workloadRefObservation != int32(currentWorkloadObservedGeneration) {
+			newStatus.WorkloadObservedGeneration = strconv.Itoa(int(workloadRefObservation))
+		}
+	}
+
 	newStatus.ObservedGeneration = strconv.Itoa(int(c.rollout.Generation))
 	newStatus.Phase, newStatus.Message = rolloututil.CalculateRolloutPhase(c.rollout.Spec, *newStatus)
 
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -697,6 +697,8 @@ func (c *rolloutContext) persistRolloutStatus(newStatus *v1alpha1.RolloutStatus)
 		if workloadRefObservation != int32(currentWorkloadObservedGeneration) {
 			newStatus.WorkloadObservedGeneration = strconv.Itoa(int(workloadRefObservation))
 		}
+	} else {
+		newStatus.WorkloadObservedGeneration = ""
 	}
 
 	newStatus.ObservedGeneration = strconv.Itoa(int(c.rollout.Generation))
--- a/rollout/temlateref.go
+++ b/rollout/temlateref.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/workqueue"
 )
 
 const (
@@ -61,7 +60,6 @@ type informerBasedTemplateResolver struct {
 	discoClient            discovery.DiscoveryInterface
 	ctx                    context.Context
 	cancelContext          context.CancelFunc
-	rolloutWorkQueue       workqueue.Interface
 	rolloutsInformer       cache.SharedIndexInformer
 	argoprojclientset      clientset.Interface
 }
@@ -72,7 +70,6 @@ func NewInformerBasedWorkloadRefResolver(
 	dynamicClient dynamic.Interface,
 	discoClient discovery.DiscoveryInterface,
 	agrgoProjClientset clientset.Interface,
-	rolloutWorkQueue workqueue.Interface,
 	rolloutsInformer cache.SharedIndexInformer,
 ) *informerBasedTemplateResolver {
 	ctx, cancelContext := context.WithCancel(context.TODO())
@@ -97,7 +94,6 @@ func NewInformerBasedWorkloadRefResolver(
 		argoprojclientset:      agrgoProjClientset,
 		dynamicClient:          dynamicClient,
 		discoClient:            discoClient,
-		rolloutWorkQueue:       rolloutWorkQueue,
 		rolloutsInformer:       rolloutsInformer,
 	}
 }
@@ -178,11 +174,11 @@ func (r *informerBasedTemplateResolver) Resolve(rollout *v1alpha1.Rollout) error
 	}
 
 	// initialize rollout workload-generation annotation
-	roMeta, err := meta.Accessor(obj)
+	workloadMeta, err := meta.Accessor(obj)
 	if err != nil {
 		return err
 	}
-	generation := strconv.FormatInt(roMeta.GetGeneration(), 10)
+	generation := strconv.FormatInt(workloadMeta.GetGeneration(), 10)
 	annotations.SetRolloutWorkloadRefGeneration(rollout, generation)
 
 	return nil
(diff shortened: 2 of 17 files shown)


huikang:
Hi, @jessesuen , this PR is ready for review. (I was trying to increase the test coverage, but still fall short of the required target. Please provide any advice if you have. Thanks.)

jessesuen:
Thanks for working on this. Will be taking a look at this in the coming days.

jessesuen:
@alexmt thinks there might be a better way to address the issue. I will ask him to put an alternative proposal in the issue.

huikang:
> @alexmt thinks there might be a better way to address the issue. I will ask him to put an alternative proposal in the issue.

Sounds good. Will follow up the thread.

jessesuen:
> @alexmt thinks there might be a better way to address the issue. I will ask him to put an alternative proposal in the issue.
> Sounds good. Will follow up the thread.

I checked in with @alexmt about this again. He gave it some more thought but didn't have a better proposal. So we will go with the original improvement as discussed in the issue.

rollout/temlateref.go
```diff
@@ -219,7 +234,23 @@ func (r *informerBasedTemplateResolver) requeueReferencedRollouts(obj interface{
...
+				if err != nil {
+					log.Errorf("Cannot update the workload-ref/annotation for %s/%s", rollout.GetName(), rollout.GetNamespace())
+				}
+			}
+
+		}
+
 		if key, err := cache.MetaNamespaceKeyFunc(ro); err == nil {
 			r.rolloutWorkQueue.Add(key)
 		}
```
jessesuen, line 256: There's actually no need to enqueue the rollout to `r.rolloutWorkQueue` anymore with this new approach because the act of annotating the rollout automatically gets the rollout into the workqueue. I think we should remove these three  lines because we may be operating on the rollout prematurely (where the annotation is not reflected in the rollout), which could cause ConflictErrors during reconciliation.

Can we remove these lines?
```
		if key, err := cache.MetaNamespaceKeyFunc(ro); err == nil {
			r.rolloutWorkQueue.Add(key)
		}
```

test/e2e/canary_test.go
```diff
@@ -407,7 +407,7 @@ spec:
 		When().
 		PromoteRollout().
 		WaitForRolloutStatus("Degraded").
-		WaitForRolloutStatus("Healthy")
+		WaitForRolloutStatus("Healthy", time.Second*90)
```
jessesuen, line 410: Is there a reason for the 90 seconds?

jessesuen (commented):
Change looks very good. Some minor points

rollout/temlateref.go
huikang: You are correct. Thanks for the explanation.

test/e2e/canary_test.go
huikang: This is just to avert the flaky e2e test.

rollout/temlateref.go
huikang: HI, @jessesuen , I tested remove the above 3 lines. However, the unit test failed at

https://github.com/argoproj/argo-rollouts/blob/460262ee353abe60acbae431265a78c5fb86e58b/rollout/temlateref_test.go#L366-L367

The `Get()` does not return any object. Any advice?

jessesuen: I think test needs to be modified since we no longer update workqueue directly. So I think this part of test can be removed as well.

Maybe in place of this, we verify the rollout was annotated.

huikang: True. Will change the test to verify annotation is added.

rollout/sync.go
```diff
@@ -691,6 +691,14 @@ func (c *rolloutContext) persistRolloutStatus(newStatus *v1alpha1.RolloutStatus)
 
 	prevStatus := c.rollout.Status
 	c.pauseContext.CalculatePauseStatus(newStatus)
+	if c.rollout.Spec.TemplateResolvedFromRef {
+		workloadRefObservation, _ := annotations.GetWorkloadGenerationAnnotation(c.rollout)
+		currentWorkloadObservedGeneration, _ := strconv.ParseInt(newStatus.WorkloadObservedGeneration, 10, 32)
+		if workloadRefObservation != int32(currentWorkloadObservedGeneration) {
+			newStatus.WorkloadObservedGeneration = strconv.Itoa(int(workloadRefObservation))
+		}
+	}
```
jessesuen, line 700: To handle the case someone switches from workloadRef to inlined template, can we have an else case that does:

if c.rollout.Spec.TemplateResolvedFromRef {
 ...
} else {
  newStatus.WorkloadObservedGeneration = ""
}

rollout/temlateref.go
jessesuen: In fact, can we remove the `rolloutWorkQueue` field entirely from NewInformerBasedWorkloadRefResolver?

rollout/sync.go
huikang: Added the else block. Thanks.

rollout/temlateref.go
huikang: Hi, @jessesuen , I updated the test to verifying the annotations are added for the rollout. However, it looks like the test clientset can't turn `interface{}` to rollout. Could you help to test a look? 

https://github.com/argoproj/argo-rollouts/pull/1198/files#diff-52c4f59609dee33fdeab90f1c17716000fca7922b7524c4e718e114cc1e41afdR243

rollout/temlateref.go
```diff
@@ -198,36 +209,51 @@ func (r *informerBasedTemplateResolver) newInformerForGVK(gvk schema.GroupVersio
...
+	}, workloadMeta.GetNamespace()))
 	if err != nil {
 		return
 	}
+
+	generation := strconv.FormatInt(workloadMeta.GetGeneration(), 10)
 	for _, ro := range rollouts {
-		if key, err := cache.MetaNamespaceKeyFunc(ro); err == nil {
-			r.rolloutWorkQueue.Add(key)
+		un, ok := ro.(*unstructured.Unstructured)
```
huikang, line 243: @jessesuen , in the e2e test, the ro can be mappted to `*unstructured.Unstructured` without any issue. However, in the test code, the rollout can't be resolved. Do you have any idea on this? Thanks.

huikang: FYI, to reproduce: `go test -timeout 999s -run ^TestRequeueReferencedRollouts$ ./rollout -v`

utils/rollout/rolloututil.go
```diff
@@ -39,6 +45,19 @@ func isGenerationObserved(ro *v1alpha1.Rollout) bool {
...
+	if _, ok := annotations.GetWorkloadGenerationAnnotation(ro); !ok {
+		return true
+	}
+	workloadGeneration, _ := annotations.GetWorkloadGenerationAnnotation(ro)
+	observedWorkloadGen, err := strconv.Atoi(ro.Status.WorkloadObservedGeneration)
+	if err != nil {
+		return true
+	}
+
+	return int32(observedWorkloadGen) == workloadGeneration
```
jessesuen, line 58: 1. The code scanner is catching a valid concern with the use of strconv.Atoi. We should use `strconv.ParseInt(ro.Status.WorkloadObservedGeneration, 10, 32)` instead

2. I just noticed that we may not handle the case properly where we still may have a workloadGenerationAnnotation on the rollout (leftover from when it switched from workload to non-workload ref) even if rollout switched to using inline template.

I think this function needs to check if `ro.Spec.WorkloadRef != nil` before proceeding. At the same time, I think we should remove the workload generation annotation from the rollout, since I don't think that gets cleaned up at the moment.

huikang: Thanks for the feedback. The PR is updated and an e2e test case is added to verify the workload generation annotation  is removed after it switches to inline template.

jessesuen (approved):

PR #1224, opened 2021-05-27
fix: workload rollout spec is invalid template is not empty

### LINKED ISSUE #1222: InvalidSpec not properly set in some circumstances
## Summary 

A user applied the following invalid rollout (which specifies both workloadRef and template/selector):

```yaml
spec:
  selector:
    matchLabels:
      app: credential-adapter
  strategy:
    blueGreen:
      activeService: credential-adapter-appd-stable-service
  template:
    metadata:
      labels:
        app: credential-adapter
  workloadRef:
    apiVersion: apps/v1
    kind: Deployment
    name: credential-adapter-appd-deployment
```

In this scenario, the controller did not update the rollout status with a condition/phase about the invalid spec.

## Diagnostics

What version of Argo Rollouts are you running? v1.0.0

```

kubectl logs -n argo-rollouts deployment/argo-rollouts

```

---

huikang:

close #1222 

Checklist:

### DIFF
--- a/pkg/apis/rollouts/validation/validation.go
+++ b/pkg/apis/rollouts/validation/validation.go
@@ -83,6 +83,12 @@ func ValidateRolloutSpec(rollout *v1alpha1.Rollout, fldPath *field.Path) field.E
 		}
 	}
 
+	// WorkloadRef and template can not be set at the same time
+	if spec.WorkloadRef != nil && !spec.EmptyTemplate() {
+		allErrs = append(allErrs, field.InternalError(fldPath.Child("template"),
+			fmt.Errorf("template must be empty for workload reference rollout")))
+	}
+
 	selector, err := metav1.LabelSelectorAsSelector(spec.Selector)
 	if err != nil {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("selector"), spec.Selector, "invalid label selector"))
--- a/pkg/apis/rollouts/validation/validation.go
+++ b/pkg/apis/rollouts/validation/validation.go
@@ -83,8 +83,9 @@ func ValidateRolloutSpec(rollout *v1alpha1.Rollout, fldPath *field.Path) field.E
 		}
 	}
 
-	// WorkloadRef and template can not be set at the same time
-	if spec.WorkloadRef != nil && !spec.EmptyTemplate() {
+	if !rollout.Spec.TemplateResolvedFromRef && (spec.WorkloadRef != nil && !spec.EmptyTemplate()) {
+		// WorkloadRef and template can not be set at the same time for lint plugin
+		// During reconciliation, TemplateResolvedFromRef is true and will not reach here
 		allErrs = append(allErrs, field.InternalError(fldPath.Child("template"),
 			fmt.Errorf("template must be empty for workload reference rollout")))
 	}
--- a/rollout/temlateref.go
+++ b/rollout/temlateref.go
@@ -125,6 +125,12 @@ func (r *informerBasedTemplateResolver) Resolve(rollout *v1alpha1.Rollout) error
 		return nil
 	}
 
+	// When workloadRef is resolved for the first time, TemplateResolvedFromRef = false.
+	// In this case, template must not be set
+	if !rollout.Spec.TemplateResolvedFromRef && !rollout.Spec.EmptyTemplate() {
+		return fmt.Errorf("template must be empty for workload reference rollout")
+	}
+
 	gvk := schema.FromAPIVersionAndKind(rollout.Spec.WorkloadRef.APIVersion, rollout.Spec.WorkloadRef.Kind)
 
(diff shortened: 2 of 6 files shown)


huikang:
@jessesuen , after some investigation, I found that https://github.com/argoproj/argo-rollouts/blob/ff65b39f21a0be79744d45d7169de6ee84e991a6/rollout/controller.go#L462 does not return any error in this scenario.

The check could be placed during resolving the workload ref for the **first** time. Is there any corner case the fix does not cover?

jessesuen:
@huikang - the original problem was that it was possible for a user to make a mistake in the spec (using my example) without that error surfacing into a condition (there was no update to status). And so the user did not know why the rollout was stuck. Will this fix solve that?

huikang:
> @huikang - the original problem was that it was possible for a user to make a mistake in the spec (using my example) without that error surfacing into a condition (there was no update to status). And so the user did not know why the rollout was stuck. Will this fix solve that?

I am not sure if I understand your question.

With the PR, if the uses submit a rollout like your example, the get rollout will show the following error

```
Name:            rollouts-demo-canary
Namespace:       default
Status:          ✖ Degraded
Message:         InvalidSpec: The Rollout "rollouts-demo-canary" is invalid: template must be empty for workload reference rollout
```

Since the change occurs to `Resolve()`, it surfaces into a condition.

To prevent the user makes a mistake, do you mean change is suppose to throw an error when user submit the manifest (or even using `kubectl-argo-rollouts lint`)?

rollout/temlateref.go
```diff
@@ -125,6 +125,12 @@ func (r *informerBasedTemplateResolver) Resolve(rollout *v1alpha1.Rollout) error
 		return nil
 	}
 
+	// When workloadRef is resolved for the first time, TemplateResolvedFromRef = false.
+	// In this case, template must not be set
+	if !rollout.Spec.TemplateResolvedFromRef && !rollout.Spec.EmptyTemplate() {
+		return fmt.Errorf("template must be empty for workload reference rollout")
+	}
+
```
jessesuen, line 133: Instead of in here, can we put this in validation library so that the linter can catch the problem?

jessesuen:
> I am not sure if I understand your question.
> With the PR, if the uses submits a rollout like your example, the get rollout will show the following error

Yes, this answers my question! With the original problem, the rollout  `status` field was never populated by the controller.

pkg/apis/rollouts/validation/validation.go
```diff
@@ -130,6 +130,12 @@ func ValidateRolloutSpec(rollout *v1alpha1.Rollout, fldPath *field.Path) field.E
 
 	allErrs = append(allErrs, ValidateRolloutStrategy(rollout, fldPath.Child("strategy"))...)
 
+	// WorkloadRef and template can not be set at the same time
+	if spec.WorkloadRef != nil && !spec.EmptyTemplate() {
+		allErrs = append(allErrs, field.InternalError(fldPath.Child("template"),
+			fmt.Errorf("template must be empty for workload reference rollout")))
+	}
+
```
huikang, line 138: @jessesuen , I moved the check to validation. Could you take another look?

BTW, when I run the lint command on your example, I got

```
Error: spec.template.spec.containers: Required value
```

So the lint sub command works without the PR.

huikang: After further investigation, I found that https://github.com/argoproj/argo-rollouts/blob/04201fb068dab917218e0533231f8007d2cba71c/pkg/apis/rollouts/validation/validation.go#L117

returns an error for empty template.spec.containers if using the lint sub-command. However, if we apply the invalid spec, the spec.containers are filled as default value at some where, which causes the issue

huikang: so I think the fix could be move the validation to somewhere before controller fills out the template.spec.containers field so that the error can be caught earlier. what do you think?

huikang: My above thoughts kind of complicate the problem. Please see if the updated commit looks good to you. Thanks.

rollout/temlateref.go
huikang: Updated. The method is moved to validation package. However, lint still can not catch the problem as it reports the following error. 

```
Error: spec.template.spec.containers: Required value
```

Please see the test cases I added to `validation_test.go`. During reconciliation time, since the `template.spec.containers` is filled by clientset.Get(), validation method will only report this error -`template must be empty for workload reference rollout`.

huikang: Hi, @jessesuen , the validation are placed in both Resolve and ValidateRolloutSpec method. In `ValidateRolloutSpec`, an additional condition (`!rollout.Spec.TemplateResolvedFromRef`) is added because during reconciliation template is filled prior to validation; otherwise it results in false alarm.

An e2e test is added.

jessesuen (approved):

Issue #1193, opened 2021-05-17
Argo rollouts often hangs with long running deploys


MarkSRobinson:
## Summary 

When deploying roll outs with a long step time (>10m), Argo rollouts will often hang on the last step.

AR will get to 100% deployed, with all traffic on the new pods but will get stuck tearing down the old pods.  Using kubectl plugin, the message `Message:         old replicas are pending termination` will be displayed for a considerable amount of time (10-30 minutes).  Eventually Argo will be able to terminate the pods.

Ideally Argo should be able to kill these pods in less time.

## Diagnostics

Argo: 0.10.2
K8s: 1.17

There's a lot of log data
https://gist.github.com/MarkSRobinson/e44ce06689aa02dd2d7886482c152fe2

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Are you using scaleDownDelaySeconds?
Which version of the plugin? Is that also `0.10.2` as well?

MarkSRobinson:
I'm not using `scaleDownDelaySeconds` and the plugin version is `v0.10.0+3837eab`

MarkSRobinson:
My setup is using canary.

jessesuen:
Are you using canary + traffic routing? There is a fix in v1.0 plugin which may address this:

https://github.com/argoproj/argo-rollouts/commit/31afa28d59f5e98ebb2fc4c3121ead29bf8b3256#diff-e6177f3a3015ec10cd7a2cfcb6293603125bbfb53bfdbd51f342507d084e6d2e

You can download v1.0-rc1 plugin here:
https://github.com/argoproj/argo-rollouts/releases/tag/v1.0.0-rc1

MarkSRobinson:
Yes, I'm using canary with traffic routing (linkerD).

The problem really isn't with the plugin, it's that Argo can't scale down old pods for a considerable time.

jessesuen:
> Argo can't scale down old pods for a considerable time.

Argo should set the `spec.replicas` of the old ReplicaSet stack to `0` shortly after the promotion of the canary. After that step, it's really up to the Pods to shut themselves down since they will get SIGTERMed by kubelet while honoring the pod's grace period seconds before getting SIGKILLed by kubelet.

[...]

PR #1243, opened 2021-06-02
fix: Mitigate the bug where items are re-added constantly to the workqueue. #1193

MarkSRobinson:
There is a deep bug where items are added to the rollouts workqueue constantly.  This is a problem because there is an exponential back-off for items so each add extends the back-off by a factor of two.  The backoff maxes out at 16.6 minutes.

This fix will prevent Argo Rollouts from hanging for up to 16 minutes at a time if this case happens.  This change reduces the maximum back-off to 10 seconds.

Checklist:


jessesuen (commented):
This is a great find! Could you describe the scenario where this happens? I'm just surprised we have not come across this.

Also, can you fix linting errors? After that it looks good to merge.

MarkSRobinson:
I'm not entirely sure since it can be hard to reproduce, but the big thing is high pod count (>20) and analysis runs that are frequent and don't terminate.  So every 10s for analysis checks.  It also correlates with long deployment times (>10m)

jessesuen (approved):

PR #1268, opened 2021-06-10
fix: rollout paused longer than progressDeadlineSeconds would briefly degrade

### LINKED ISSUE #1225: status command exits before timout
## Summary 

I have a rollout that was being watched via `status` with a `--timeout=3h` flag.  It exited with a `degraded` after 20 minutes.

## Diagnostics

What version of Argo Rollouts are you running?  v1.0.1 (kubectl-argo-rollouts = v1.0.0)

```
2021-05-27 14:13:33 EDT Status: Downloaded newer image for gcr.io/helpful-cat-109717/kubectl-argo-rollouts:v1.0.0
2021-05-27 14:13:34 EDT Executing command: kubectl get deployment -l "app.kubernetes.io/name=managed-satellite-envoy" -o name | cut -d '/' -f 2 | xargs -n 1 kubectl-argo-rollouts status --timeout=3h
2021-05-27 14:33:51 EDT Error: The rollout is in a degraded state with message: ProgressDeadlineExceeded: ReplicaSet "public-satellite-pools-managed-satellite-envoy-v2-6f9df7c4bf" has timed out progressing.
```

14:13 is when helm pushed the change.
14:33 is when the command failed.

The rollout itself completed successfully shortly thereafter. 

(there is only one rollout returned by the first kubectl, and I've omitted the extra flags to kubectl for brevity)

```

Logs attached. The server doesn't show any errors.  Possibly this is a client bug. I've seen this happen twice now.  Is there anything else I can provide to help here?

kubectl logs -n argo-rollouts depl
[filtered2.csv](https://github.com/argoproj/argo-rollouts/files/6556579/filtered2.csv)
oyment/argo-rollouts

```

---

khhirani:
Closes #1225

### DIFF
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -965,6 +965,25 @@ func (f *fixture) getPatchedRollout(index int) string {
 	return string(patchAction.GetPatch())
 }
 
+func (f *fixture) getPatchedRolloutWithoutConditions(index int) (string, error) {
+	action := filterInformerActions(f.client.Actions())[index]
+	patchAction, ok := action.(core.PatchAction)
+	if !ok {
+		f.t.Fatalf("Expected Patch action, not %s", action.GetVerb())
+	}
+	ro := make(map[string]interface{})
+	err := json.Unmarshal(patchAction.GetPatch(), &ro)
+	if err != nil {
+		return "", err
+	}
+	unstructured.RemoveNestedField(ro, "status", "conditions")
+	roBytes, err := json.Marshal(ro)
+	if err != nil {
+		return "", err
+	}
+	return string(roBytes), nil
+}
+
 func (f *fixture) getPatchedRolloutAsObject(index int) *v1alpha1.Rollout {
 	action := filterInformerActions(f.client.Actions())[index]
 	patchAction, ok := action.(core.PatchAction)
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -965,7 +965,7 @@ func (f *fixture) getPatchedRollout(index int) string {
 	return string(patchAction.GetPatch())
 }
 
-func (f *fixture) getPatchedRolloutWithoutConditions(index int) (string, error) {
+func (f *fixture) getPatchedRolloutWithoutConditions(index int) string {
 	action := filterInformerActions(f.client.Actions())[index]
 	patchAction, ok := action.(core.PatchAction)
 	if !ok {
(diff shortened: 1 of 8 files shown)


rollout/sync.go
```diff
@@ -657,7 +660,7 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 			if c.newRS != nil {
 				msg = fmt.Sprintf(conditions.ReplicaSetTimeOutMessage, c.newRS.Name)
 			}
-			condition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionFalse, conditions.TimedOutReason, msg)
+			condition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.TimedOutReason, msg)
```
jessesuen, line 663: If we are timed out, why do we set Progressing condition to be True?

jessesuen:
The e2e tests are failing because the github environment doesn't have the kubectl argo rollouts plugin installed.

```
Error: unknown command "argo" for "kubectl"
Run 'kubectl --help' for usage.
```

So you'll need to invoke the plugin logic another way. You can follow how I use the Promote() logic of the plugin from the e2e tests.

rollout/controller_test.go
```diff
@@ -965,6 +965,25 @@ func (f *fixture) getPatchedRollout(index int) string {
...
+	action := filterInformerActions(f.client.Actions())[index]
+	patchAction, ok := action.(core.PatchAction)
+	if !ok {
+		f.t.Fatalf("Expected Patch action, not %s", action.GetVerb())
+	}
+	ro := make(map[string]interface{})
+	err := json.Unmarshal(patchAction.GetPatch(), &ro)
+	if err != nil {
+		return "", err
+	}
```
jessesuen, line 978: Let's simplify this method signature such that it doesn't ever return error. You can call `f.t.Fatalf` like you do above, rather than returning error.

test/e2e/functional_test.go
```diff
@@ -1237,3 +1240,51 @@ func (s *FunctionalSuite) TestControllerMetrics() {
...
+    metadata:
+      labels:
+        app: rollout-canary
+    spec:
+      containers:
+      - name: rollouts-demo
+        image: argoproj/rollouts-demo:blue
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 8080
```
jessesuen, line 1267: Can we use the nginx image like we do for the rest of the e2e tests? It will speed up tests since it should already be loaded into the kubernetes environment.

Also, `imagePullPolicy: Always` will slow down tests, and the container port for nginx is 80

jessesuen (approved):
LGTM. Please wait for e2e / unit tests to pass before merging (resubmit e2e as necessary due to flakiness)

PR #1273, opened 2021-06-12
feat: allow VirtualService HTTPRoute to be inferred if there is single route

jessesuen:
Currently, the istio configuration in the rollout requires a VirtualService HTTPRoute name to be explicitly specified.
This change makes it possible to infer the HTTP route in the event that there is only a single route specified, simplifying the Rollout config. e.g.:

```yaml
  strategy:
    canary:
      trafficRouting:
        istio:
          virtualService: 
            name: istio-subset-split-single-route-vsvc
          destinationRule:
            name: istio-subset-split-single-route-destrule
            canarySubsetName: canary
            stableSubsetName: stable
```

### DIFF
--- a/rollout/trafficrouting/istio/istio.go
+++ b/rollout/trafficrouting/istio/istio.go
@@ -83,10 +83,6 @@ func (patches virtualServicePatches) patchVirtualService(httpRoutes []interface{
 func (r *Reconciler) generateVirtualServicePatches(httpRoutes []VirtualServiceHTTPRoute, desiredWeight int64) virtualServicePatches {
 	canarySvc := r.rollout.Spec.Strategy.Canary.CanaryService
 	stableSvc := r.rollout.Spec.Strategy.Canary.StableService
-	routes := map[string]bool{}
-	for _, r := range r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualService.Routes {
-		routes[r] = true
-	}
 	canarySubset := ""
 	stableSubset := ""
 	if r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.DestinationRule != nil {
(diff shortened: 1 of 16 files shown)


rollout/trafficrouting/istio/istio_test.go
```diff
@@ -171,6 +193,41 @@ func TestReconcileVirtualServiceNotFound(t *testing.T) {
...
+	assert.NoError(t, err)
+	actions := client.Actions()
+	assert.Len(t, actions, 1)
+	assert.Equal(t, "update", actions[0].GetVerb())
+
+	// Verify we actually made the correct change
+	vsvcUn, err := client.Resource(istioutil.GetIstioVirtualServiceGVR()).Namespace(ro.Namespace).Get(context.TODO(), "vsvc", metav1.GetOptions{})
+	assert.NoError(t, err)
+	routes, _, _ := unstructured.NestedSlice(vsvcUn.Object, "spec", "http")
+	route := routes[0].(map[string]interface{})
```
huikang, line 226: Shall we add an `assert.Equal` to verify the route name is correctly inferred?

rollout/trafficrouting/istio/istio.go
```diff
@@ -466,6 +462,7 @@ func (r *Reconciler) SetWeight(desiredWeight int32) error {
 	}
 	_, err = client.Update(ctx, modifiedVsvc, metav1.UpdateOptions{})
 	if err == nil {
+		r.log.Infof("UpdatedVirtualService: %s", modifiedVsvc)
```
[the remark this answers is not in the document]
jessesuen: I actually did not mean to keep this as info level. But I did mean to keep the search word same as the event name. Will fix log level to debug.

rollout/trafficrouting/istio/istio_test.go
jessesuen: Actually the test VirtualService intentionally does not have a route name so there's no name to verify:

```yaml
kind: VirtualService
metadata:
  name: vsvc
  namespace: default
spec:
  gateways:
  - istio-rollout-gateway
  hosts:
  - istio-rollout.dev.argoproj.io
  http:
  - route:
    - destination:
        host: 'stable'
      weight: 100
    - destination:
        host: canary
      weight: 0
```

rollout/trafficrouting/istio/istio.go
```diff
@@ -83,25 +83,21 @@ func (patches virtualServicePatches) patchVirtualService(httpRoutes []interface{
...
-	}
 	canarySubset := ""
 	stableSubset := ""
 	if r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.DestinationRule != nil {
 		canarySubset = r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.DestinationRule.CanarySubsetName
 		stableSubset = r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.DestinationRule.StableSubsetName
 	}
 
+	// err can be ignored because we already called ValidateHTTPRoutes earlier
+	routeIndexesToPatch, _ := getRouteIndexesToPatch(r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualService.Routes, httpRoutes)
```
alexmt, line 94: nitpick: I would panic just in case. If for some reason `getRouteIndexesToPatch` returns error then `routeIndexesToPatch` is nil/empty slice. The crash is better than silently do nothing.

alexmt (approved):
LGTM. Added one comment

rollout/trafficrouting/istio/istio.go
jessesuen: I did consider that but determined it was safe to proceed, even on error because we would only iterate a nil slice, which is safe. Having a panic would be equivalent to allowing a nil pointer dereference.

huikang:
LGTM

PR #1285, opened 2021-06-16
feat: add an up panel to the example dashboard

huikang:
- reflect if the controller is running

Checklist:

### DIFF
--- a/examples/dashboard.json
+++ b/examples/dashboard.json
@@ -19,6 +19,65 @@
   "links": [],
   "panels": [
     {
+      "type": "stat",
+      "title": "up(controller)",
+      "gridPos": {
+        "x": 0,
+        "y": 0,
+        "w": 3,
+        "h": 4
+      },
+      "id": 32,
+      "targets": [
+        {
+          "expr": "up{kubernetes_pod_name=~\"argo-rollouts-.*\"}",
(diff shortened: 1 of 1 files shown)


huikang:
Screen Shot 2021-06-16 at 2 17 10 PM

examples/dashboard.json
```diff
@@ -18,6 +18,65 @@
...
+      "gridPos": {
+        "x": 0,
+        "y": 0,
+        "w": 3,
+        "h": 4
+      },
+      "id": 32,
+      "targets": [
+        {
+          "expr": "up{kubernetes_pod_name=~\"argo-rollouts-.*\"}",
```
jessesuen, line 33: I believe  `kubernetes_pod_name` presumes user is running `kube-state-metrics` which probably isn't a safe assumption to make in all clusters.

In Argo CD, we rely on an "uptime" gauge, which is based on `process_start_time_seconds `, which is a standard prometheus metric I think already exported by rollout controller:
https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics

image

WDY think about a singlestat panel that shows uptime, instead of this?

huikang:
@jessesuen , agreed. Thanks.

PR #1304, opened 2021-06-24
fix: avoid invalid replica in workloadRef

### LINKED ISSUE #1302: Validation error when using workloadRef
## Summary 

This rollout works without any issues:

```yaml
kind: Rollout
metadata:
  name: test-service
spec:
  workloadRef:
    apiVersion: apps/v1
    kind: Deployment
    name: test-service
  strategy:
    canary:
      maxSurge: 10%
      maxUnavailable: 10%
      steps:
        - setWeight: 3
        - pause: {duration: 1m}
        - setWeight: 20
        - pause: {duration: 1m}
```

However, when I add analysis, like below, I get an incorrect error message. I would also expect this configuration to work. It just loops forever complaining that the rollout spec is not valid.

```yaml
kind: Rollout
metadata:
  name: test-service
spec:
  workloadRef:
    apiVersion: apps/v1
    kind: Deployment
    name: test-service
  strategy:
    canary:
      maxSurge: 10%
      maxUnavailable: 10%
      analysis:
        templates:
        - templateName: error-rate
        args:
        - name: service-name
          value: test-service
        - name: environment
          value: dev
      steps:
        - setWeight: 3
        - pause: {duration: 1m}
        - setWeight: 20
        - pause: {duration: 1m}
```

## Diagnostics

What version of Argo Rollouts are you running? **1.0.2**

These messages repeat forever and the rollout never completes.
```
time="2021-06-23T21:45:49Z" level=info msg="Started syncing rollout" generation=9 namespace=default resourceVersion=251502897 rollout=test-service
time="2021-06-23T21:45:49Z" level=info msg="Defaulting .spec.replica to 1" generation=9 namespace=default resourceVersion=251502897 rollout=test-service
time="2021-06-23T21:45:49Z" level=error msg="rollout syncHandler error: Rollout.argoproj.io \"test-service\" is invalid: [spec.selector: Invalid value: \"null\": spec.selector in body must be of type object: \"null\", spec.template.spec.containers: Invalid value: \"null\": spec.template.spec.containers in body must be of type array: \"null\"]" namespace=default rollout=test-service
time="2021-06-23T21:45:49Z" level=info msg="rollout syncHandler queue retries: 287 : key \"default/test-service\"" namespace=default rollout=test-service
```

---

huikang:
- defaulting replica after workload ref is resolved

close #1302 

Checklist:

### DIFF
--- a/rollout/controller.go
+++ b/rollout/controller.go
@@ -360,14 +360,6 @@ func (c *Controller) syncHandler(key string) error {
 		return nil
 	}
 
-	// In order to work with HPA, the rollout.Spec.Replica field cannot be nil. As a result, the controller will update
-	// the rollout to have the replicas field set to the default value. see https://github.com/argoproj/argo-rollouts/issues/119
-	if rollout.Spec.Replicas == nil {
-		logCtx.Info("Defaulting .spec.replica to 1")
-		r.Spec.Replicas = pointer.Int32Ptr(defaults.DefaultReplicas)
-		_, err := c.argoprojclientset.ArgoprojV1alpha1().Rollouts(r.Namespace).Update(ctx, r, metav1.UpdateOptions{})
-		return err
-	}
 	defer func() {
 		duration := time.Since(startTime)
 		c.metricsServer.IncRolloutReconcile(r, duration)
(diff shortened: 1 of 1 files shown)


jessesuen:
@huikang I'm having trouble understanding how moving the defaulting of spec.replicas fixes the problem and why the problem only happens if there is an analysis in the steps. Could you explain how this fixes it?

huikang:
> @huikang I'm having trouble understanding how moving the defaulting of spec.replicas fixes the problem and why the problem only happens if there is an analysis in the steps. Could you explain how this fixes it?

Sure thing. In the original implementation, if the rollout refers to a deployment and does not have replica set, it will call the update method before resolving the template. Since the template is empty (no container in the strcut), update will return error.

https://github.com/argoproj/argo-rollouts/blob/e1840b58a79fb8dc6d84af1e0b6b8b76e71ddaab/rollout/controller.go#L367-L368

This PR changes the order of update() and resolve() such that update can return successfully with defaulted replica.
Hope it makes sense to you.

jessesuen:
> Since the template is empty, i.e., no container in the struct), update will return error.

This is what was surprising to me. I didn't think the Update() would fail in that manner. In any case I see this fixes the issue. Thanks!

jessesuen (approved):

Issue #1278, opened 2021-06-15
Docs: inconsistent urls and broken links


kostis-codefresh:
(Reported by Bryan Taylor in the CNCF slack channel)

The documentation pages have mixed rendering results where sometimes the extension .html works and sometimes it doesn't

https://argoproj.github.io/argo-rollouts/features/traffic-management/istio/ -> renders a broken page
https://argoproj.github.io/argo-rollouts/features/traffic-management/istio.html -> works ok

But on the other hand

https://argoproj.github.io/argo-rollouts/getting-started/istio/ -> works ok
https://argoproj.github.io/argo-rollouts/getting-started/istio.html -> renders a broken page.

The issue here (apart from the inconsistency) is that the broken links are indexed by Google confusing users.

We should either fix all links to have a consistent rendering or make the wrong ones return 404 instead of broken renderings

![problematic-urls](https://user-images.githubusercontent.com/39800303/122034515-476fb580-cdda-11eb-9f26-e4c3c9c405db.png)

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Fixed in https://github.com/argoproj/argo-rollouts/pull/1321

PR #1316, opened 2021-06-29
docs: Fix links on architecture.md

shuheiktgw:
It seems the links on the architecture page are broken and point to the wrong places so I fixed it. 

Checklist:


kostis-codefresh:
I think this is related to https://github.com/argoproj/argo-rollouts/issues/1278 

Something seems to have changed the way docs are rendered (with or without .html at the end)

shuheiktgw:
Yes, I think something seems to be wrong with the links on the documentation yet the links on the architecture page points to the wrong places so that's one problem.

kostis-codefresh (approved):

jessesuen:
I fixed the source of the problem so this PR is unnecessary. See: https://argoproj.github.io/argo-rollouts/architecture/

Our site generation was recently affected by a change config requirements mkdocs.yaml and we use `:latest` mkdocs image to generate the docs. I fixed our config to adhere to new mkdocs.yaml config requirements

PR #1322, opened 2021-07-02
fix: github action to publish docs using incorrect make target

jessesuen:


khhirani (approved):
LGTM

PR #1331, opened 2021-07-09
fix: abort scaledown stable RS for canary with traffic routing

### LINKED ISSUE #1292: Rollout abort scales down stable ReplicaSet if there is no rollout in progress
## Summary 

For _canary_ strategy, calling `rollouts abort` when there's **no** current rollout would scale the **stable** _ReplicaSet_ to 0.

This was surprising, I would have thought this action to be a no-op or at most revert to the _ReplicaSet_ prior to stable.

This is what [the doc](https://argoproj.github.io/argo-rollouts/generated/kubectl-argo-rollouts/kubectl-argo-rollouts_abort.html ) for this command says

> This command stops progressing the current rollout and reverts all steps. The previous ReplicaSet will be active.

## Diagnostics

`v0.10.2`

```
Started syncing rollout at (2021-06-15 23:29:12.699651128 +0000 UTC m=+754407.503616433)
Reconciling TrafficRouting with type 'Istio'
Reconciling StableRS
No StableRS exists to reconcile or matches newRS
Reconciling new ReplicaSet 'foo-859bb5fdfb'
Not finished reconciling new ReplicaSet 'foo-859bb5fdfb'
Not finished reconciling ReplicaSets
Event(v1.ObjectReference{Kind:\"Rollout\", UID:\"...\", APIVersion:\"argoproj.io/v1alpha1\", ResourceVersion:\"xyz\", FieldPath:\"\"}): type: 'Normal' reason: 'ScalingReplicaSet' Scaled down replica set foo-859bb5fdfb from 7 to 0
Reconciliation completed

kubectl logs -n argo-rollouts deployment/argo-rollouts

```

---

huikang:

close #1292 

Checklist:

### DIFF
--- a/test/e2e/istio_test.go
+++ b/test/e2e/istio_test.go
@@ -197,3 +197,30 @@ func (s *IstioSuite) TestIstioSubsetSplitSingleRoute() {
 		}).
 		ExpectRevisionPodCount("1", 1) // don't scale down old replicaset since it will be within scaleDownDelay
 }
+
+func (s *IstioSuite) TestIstioAbortUpdate() {
+	s.Given().
+		RolloutObjects("@istio/istio-host-split.yaml").
+		When().
+		ApplyManifests().
+		WaitForRolloutStatus("Healthy").
+		Then().
+		When().
+		AbortRollout().
+		WaitForRolloutStatus("Degraded").
+		Then().
+		ExpectRevisionPodCount("1", 1).
+		When().
+		UpdateSpec().
+		WaitForRolloutStatus("Paused").
+		Then().
+		When().
+		PromoteRollout().
+		WaitForRolloutStatus("Healthy").
+		Then().
+		When().
+		AbortRollout().
+		WaitForRolloutStatus("Degraded").
+		Then().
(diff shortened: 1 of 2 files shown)


jessesuen (changes requested):
@huikang I think we should be fixing this in a different way. 

Instead of modifying `CalculateReplicaCountsForCanary` to consider if we see `ro.Status.Abort`,  I think it will be cleaner to modify `GetCanaryReplicasOrWeight` so that it always returns 100 if we are in a fully promoted state.

Something like:

```golang
// GetCanaryReplicasOrWeight either returns a static set of replicas or a weight percentage
func GetCanaryReplicasOrWeight(rollout *v1alpha1.Rollout) (*int32, int32) {
	if rollout.Status.PromoteFull || rollout.Status.CurrentPodHash == rollout.Status.StableRS {
		return nil, 100
	}
```

WDYT?

huikang:
Hi, @jessesuen , your approach is more generic than my last one. PR is updated. Thanks for your feedback.

jessesuen (approved):

PR #1338, opened 2021-07-12
feat: Support CloudWatch as a metric provider

### LINKED ISSUE #1014: Support CloudWatch as a metric provider
# Summary

It would be useful to gather metrics from AWS CloudWatch to drive analysis

# Use Cases

For users who are using integrations such as ALB ingress, error rate and latency would be metrics that could be used to drive rollout promotion.

---

ryota-sakamoto:
Signed-off-by: Ryota Sakamoto <sakamo.ryota+github@gmail.com>

close #1014 

Checklist:


ryota-sakamoto:
I have no idea why the codecov report is weird.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -193,6 +195,12 @@ type JobMetric struct {
 	Spec     batchv1.JobSpec   `json:"spec" protobuf:"bytes,2,opt,name=spec"`
 }
 
+// CloudWatchMetric defines the cloudwatch query to perform canary analysis
+type CloudWatchMetric struct {
+	Interval          DurationString `json:"interval,omitempty" protobuf:"bytes,1,opt,name=interval,casttype=DurationString"`
+	MetricDataQueries string         `json:"metricDataQueries" protobuf:"bytes,2,opt,name=metricDataQueries"`
```
jessesuen, line 201: I think we can make the following improvement to the spec:

1. Instead of a string for `metricDataQueries`, I feel it would be better to make this field a list, where each element is a single `cloudwatch.MetricDataQuery`. This matches the AWS client API better.

In other words, I am proposing

```golang
type CloudWatchMetricDataQuery string

type CloudWatchMetric struct {
	Interval          DurationString `json:"interval,omitempty" protobuf:"bytes,1,opt,name=interval,casttype=DurationString"`
	MetricDataQueries []CloudWatchMetricDataQuery         `json:"metricDataQueries" protobuf:"bytes,2,opt,name=metricDataQueries"`
```

Then the spec would look like:

```yaml
    provider:
      cloudWatch:
        interval: 15m
        metricDataQueries:
        - |
            {
              "Id": "rate",
              "Expression": "errors / requests"
            }
        - |
            {
              "Id": "errors",
              "MetricStat": {
                "Metric": {
                  "Namespace": "app",
                  "MetricName": "errors"
                },
                "Period": 300,
                "Stat": "Sum",
                "Unit": "Count"
              },
              "ReturnData": false
            }
        - |
            {
              "Id": "requests",
              "MetricStat": {
                "Metric": {
                  "Namespace": "app",
                  "MetricName": "requests"
                },
                "Period": 300,
                "Stat": "Sum",
                "Unit": "Count"
              },
              "ReturnData": false
            }
```

2. I think we should allow `MetricDataQueries` to accept both JSON and YAML. So this would also be valid:

```yaml
    provider:
      cloudWatch:
        interval: 15m
        metricDataQueries:
        - |
            Id: errors
            MetricStat:
              Metric:
                Namespace: app
                MetricName: errors
              Period: 300
              Stat: Sum
              Unit: Count
            ReturnData: false
        - |
            Id: requests
            MetricStat:
              Metric:
                Namespace: app
                MetricName: requests
              Period: 300
              Stat: Sum
              Unit: Count
            ReturnData: false

```

metricproviders/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	result, err := p.api.Query(interval, query)
+	if err != nil {
+		return metricutil.MarkMeasurementError(measurement, err)
+	}
+
+	for _, message := range result.Messages {
+		p.logCtx.Warnf("CloudWatch returned the following messages: %+v", message)
+	}
```
jessesuen, line 83: Instead of or in addition to logging a warning, the `Measurement` object has a `Metadata` field, which can be used to store any information you want to surface in the AnalysisRun. The end-user will likely not have access to the controller logs, so having messages stored in the measurement metadata will help these users debug problems.

metricproviders/cloudwatch/cloudwatch.go
```diff
@@ -0,0 +1,131 @@
...
+	status, err := evaluate.EvaluateResult(result.MetricDataResults, metric, p.logCtx)
+	if err != nil {
+		return metricutil.MarkMeasurementError(measurement, err)
+	}
+
+	value := [][]float64{}
+	for _, result := range result.MetricDataResults {
+		value = append(value, result.Values)
+	}
+	measurement.Value = fmt.Sprintf("%+v", value)
```
jessesuen, line 94: We assign Value after `EvaluateResult`. This makes it impossible for the end user to see what the return value was and why it was evaluated to be Failed/Error. Can we store the value first and then perform evaluation?

jessesuen (changes requested):

jessesuen:
@ryota-sakamoto we're seeing if this feature could make the v1.1 release in next few weeks. Did you have thoughts about my proposed changes to the spec?

ryota-sakamoto:
@jessesuen 
Sorry for being late.
Your idea is great and suitable, and then I'm working and testing.
There is complete when this week.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -193,6 +195,40 @@ type JobMetric struct {
...
+	Dimensions []CloudWatchMetricStatMetricDimension `json:"Dimensions,omitempty" protobuf:"bytes,1,opt,name=Dimensions"`
+	MetricName *string                               `json:"MetricName,omitempty" protobuf:"bytes,2,opt,name=MetricName"`
+	Namespace  *string                               `json:"Namespace,omitempty" protobuf:"bytes,3,opt,name=Namespace"`
+}
+
+type CloudWatchMetricStatMetricDimension struct {
+	Name  *string `json:"Name,omitempty" protobuf:"bytes,1,opt,name=Name"`
+	Value *string `json:"Value,omitempty" protobuf:"bytes,2,opt,name=Value"`
+}
+
```
jessesuen, line 231: I like this new approach you use to use structured objects, instead of my suggestion of making `CloudWatchMetricDataQuery` being a string field. However, I think we'll need to modify this some more. I believe issues 1 and 2 may be because you copied exact datastructures from AWS go client and used as our own types:

1. There are now a lot of new "API rule violations" because of the casing of the json field name. Kubernetes APIs require names to be camel-cased always starting with lowercase letter, but here we use uppercase in the field names. e.g.:

```golang
Name  *string `json:"Name,omitempty" protobuf:"bytes,1,opt,name=Name"`
```

2. Some fields were made as pointers, when they should always have a value required. For example, I don't think these should be allowed to be pointers. Can we make everything non-pointers unless we need to explicitly distinguish between not supplied vs. empty?

```golang
type CloudWatchMetricStatMetricDimension struct {
	Name  *string `json:"Name,omitempty" protobuf:"bytes,1,opt,name=Name"`
	Value *string `json:"Value,omitempty" protobuf:"bytes,2,opt,name=Value"`
}
```

3. With this approach for using structured data (instead of a giant string field), we will not allow some things to be parameterized with Analysis arguments. For example, the `Period` field cannot be parameterized because it is an `int32`. 
```
	Period     *int32                `json:"Period,omitempty" protobuf:"varint,5,opt,name=Period"`
```

My question is: is `Period` something we should allow to be parameterized? If so, i think this should be made into an `intstrutil.IntOrString` instead of an `int32`

ryota-sakamoto: 1. I've fixed
2. I've fixed the pointer which isn't necessary I refer https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDataQuery.html
3. I've fixed for that I use `intstrutil.IntOrString` instead of `int32`

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -193,6 +195,40 @@ type JobMetric struct {
 	Spec     batchv1.JobSpec   `json:"spec" protobuf:"bytes,2,opt,name=spec"`
 }
 
+// CloudWatchMetric defines the cloudwatch query to perform canary analysis
+type CloudWatchMetric struct {
+	Interval          DurationString              `json:"interval,omitempty" protobuf:"bytes,1,opt,name=interval,casttype=DurationString"`
```
harikrongali, line 200: DurationString to follow camelcase

harikrongali:
@jessesuen can we merge this PR? seems @ryota-sakamoto addressed review comments

jessesuen (approved):
Great work!

PR #1351, opened 2021-07-14
feat: TrafficRouting SMI with Experiment Step in Canary

khhirani:


rollout/trafficrouting.go
```diff
@@ -110,7 +101,6 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 				step := c.rollout.Spec.Strategy.Canary.Steps[i]
 				if step.SetWeight != nil {
 					desiredWeight = *step.SetWeight
-					break
 				}
```
jessesuen, line 104: Can you explain the removal of the break? I don't think it is correct.

test/e2e/smi_test.go
```diff
@@ -0,0 +1,95 @@
...
+			assert.Equal(s.T(), "rollout-smi-experiment-canary", ts.Spec.Backends[0].Service)
+			assert.Equal(s.T(), int64(5), ts.Spec.Backends[0].Weight.Value())
+
+			ex := t.GetRolloutExperiments().Items[0]
+			exServiceName := ex.Status.TemplateStatuses[0].ServiceName
+			assert.Equal(s.T(), int64(5), ts.Spec.Backends[1].Weight.Value())
+			assert.Equal(s.T(), exServiceName, ts.Spec.Backends[1].Service)
+
+			assert.Equal(s.T(), "rollout-smi-experiment-stable", ts.Spec.Backends[2].Service)
+			assert.Equal(s.T(), int64(90), ts.Spec.Backends[2].Weight.Value())
```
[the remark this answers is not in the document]
jessesuen: Likewise for the other places in this test (assert we went back to exactly two backends).

rollout/trafficrouting/alb/alb.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
```
jessesuen, line 8: This import is not in proper location

rollout/trafficrouting/smi/smi_test.go
```diff
@@ -432,3 +433,116 @@ func TestReconcileRolloutDoesNotOwnTrafficSplitError(t *testing.T) {
...
+		// check experiment service backends
+		assert.Equal(t, weightDestinations[0].ServiceName, ts2.Spec.Backends[1].Service)
+		assert.Equal(t, int(weightDestinations[0].Weight), ts2.Spec.Backends[1].Weight)
+
+		assert.Equal(t, weightDestinations[1].ServiceName, ts2.Spec.Backends[2].Service)
+		assert.Equal(t, int(weightDestinations[1].Weight), ts2.Spec.Backends[2].Weight)
+
+		// check stable backend
+		assert.Equal(t, "stable-service", ts2.Spec.Backends[3].Service)
+		assert.Equal(t, 80, ts2.Spec.Backends[3].Weight)
```
jessesuen, line 513: This test is calling `r.SetWeight(10)` with no additional weight destinations. So why is the test expecting to be four total backends? There should be only two, right?

rollout/trafficrouting/istio/istio.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
```
jessesuen, line 9: import is not in correct location

rollout/trafficrouting/nginx/nginx.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
```
jessesuen, line 8: import is not in correct location

rollout/trafficrouting/smi/smi.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
```
jessesuen, line 7: import is not in correct location

rollout/trafficrouting/smi/smi_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
```
jessesuen, line 9: import is not in correct location

rollout/trafficrouting.go
```diff
@@ -3,6 +3,8 @@ package rollout
 import (
 	"time"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
+
```
jessesuen, line 7: extra newline

rollout/trafficrouting/ambassador/ambassador.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/argoproj/argo-rollouts/rollout/trafficrouting"
+
```
jessesuen, line 13: unsorted import

jessesuen (commented):
Do you mind sorting all imports? Not just the ones I'm calling out.

rollout/trafficrouting.go
```diff
@@ -122,9 +112,33 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
...
+				templateWeight := getTemplateWeight(templateStatus.Name)
+				if templateStatus.ServiceName != "" || templateStatus.PodTemplateHash != "" {
+					weightDestinations = append(weightDestinations, trafficrouting.WeightDestination{
+						ServiceName:     templateStatus.ServiceName,
+						PodTemplateHash: templateStatus.PodTemplateHash,
+						Weight:          *templateWeight,
+					})
+				}
+			}
+		}
```
jessesuen, line 138: What prevents us from prematurely setting the weight to the templates before the ReplicaSet is available?

khhirani: I added L130 because the trafficRouting providers were setting backends/annotations for the weightDestinations with an empty service.

jessesuen: Are you implying that `.ServiceName` and `.PodTemplateHash` fields are only set when the Pods of the template are fully available?

Shouldn't we be checking the Running status of the experiment somewhere?

jessesuen: Also, if experiment fails and the replicasets scale down. Do we still split traffic to the scaled down experiments?

khhirani: `.ServiceName` and `.PodTemplateHash` are set when the ReplicaSet is initialized with zero replicas, not when the Pods are fully available. When the ReplicaSet scales down, the service is deleted and `.ServiceName` is set to an empty string, so L130 prevents traffic from being sent to scaled down replicas.

khhirani: Should we be checking the Running status of the experiment or each experiment template?

jessesuen: I would say the entire Experiment should be `Running` in order to consider splitting weights to its templates. Conversely, when the Experiment exits out of `Running`, then we should **_not_** split traffic to it. The Running status of the Experiment should be sufficient to know it is safe to target. I think checking against ServiceName and .PodTemplateHash are a round-about way of checking if it is safe, and it also doesn't cover the case when Experiment is still Pending.

jessesuen (approved):
LGTM. Please squash all 91 commits, and clean the commit message when merging.

Issue #1356, opened 2021-07-16
Incorrect pod template change detection after upgrade from v0.10 to v1.0


jessesuen:
## Summary 

We updated a controller from v0.10.x to v1.0.2. After the update, the rollout controller incorrectly decided that a steady-state rollout was updated, as indicated by this log:

```log
time="2021-07-01T19:24:49Z" level=info msg="Pod template change detected (new: 5f74ffd77f, old: 849cf644f5)" 
```

This caused the rollout to go through an update, even though the pod specs were identical.

I captured the ReplicaSet yaml of each version and verified there was no difference. So this must be a change in the way pod templates are compared in v1.0.

## Diagnostics

What version of Argo Rollouts are you running? v1.0.2

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
I can reproduce this with the following rollout first deployed in v0.10, then upgrading the controller to v1.0:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  labels:
  name: foo
spec:
  selector:
    matchLabels:
      app: foo
  strategy:
    canary: {}
  template:
    metadata:
      labels:
        app: foo
    spec:
      containers:
      - image: argoproj/rollouts-demo:blue
        name: app
      serviceAccountName: default
```

The detected difference is that the `serviceAccount` field (which has been deprecated in k8s) is no longer defaulted using 

live object:
```yaml
metadata:
  creationTimestamp: null
  labels:
    app: foo
spec:
  containers:
  - image: argoproj/rollouts-demo:blue
    imagePullPolicy: IfNotPresent
    name: app
    resources: {}
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
  dnsPolicy: ClusterFirst
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 30
```

defaulted object:
```yaml
metadata:
  creationTimestamp: null
  labels:
    app: foo
spec:
  containers:
  - image: argoproj/rollouts-demo:blue
    imagePullPolicy: IfNotPresent
    name: app
    resources: {}
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
  dnsPolicy: ClusterFirst
  restartPolicy: Always
  schedulerName: default-scheduler
[...]

PR #1367, opened 2021-07-22
fix: unsolicited rollout after upgrade from v0.10->v1.0 when pod was using service account

jessesuen:
Partially fixes https://github.com/argoproj/argo-rollouts/issues/1356

### DIFF
--- a/utils/replicaset/replicaset.go
+++ b/utils/replicaset/replicaset.go
@@ -494,6 +494,15 @@ func PodTemplateEqualIgnoreHash(live, desired *corev1.PodTemplateSpec) bool {
 	}
 	corev1defaults.SetObjectDefaults_PodTemplate(&podTemplate)
 	desired = &podTemplate.Template
+
+	// Do not allow the deprecated spec.serviceAccount to factor into the equality check. In live
+	// ReplicaSet pod template, this field will be populated, but in the desired pod template
+	// it will be missing (even after defaulting), causing us to believe there is a diff
+	// (when there really wasn't), and hence causing an unsolicited update to be triggered.
+	// See: https://github.com/argoproj/argo-rollouts/issues/1356
+	desired.Spec.DeprecatedServiceAccount = ""
+	live.Spec.DeprecatedServiceAccount = ""
+
 	return apiequality.Semantic.DeepEqual(live, desired)
 }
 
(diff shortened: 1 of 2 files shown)


huikang:
Hi, @jessesuen , I am just wondering what if the existing pod has both serviceAccountName and serviceAccount in the template. Will this PR handle the case properly? Thanks.

khhirani (approved):
LGTM

jessesuen:
> Hi, @jessesuen , I am just wondering what if the existing pod has both serviceAccountName and serviceAccount in the template. Will this PR handle the case properly? Thanks.

First, serviceAccountName and serviceAccount should always be the same value. I just verified that if you try to create a Deployment with different values, serviceAccount will get overridden with value from serviceAccountName.

Given that, `serviceAccountName` is the only thing we should care about. Users should be able to modify this field (add/remove/update), and the rollout should properly detect this as a change and perform an update.

However, the `serviceAccount` field in the pod is safe to ignore completely. It is a deprecated kubernetes field which has special treatment in k8s codebase. AFIK, there is not any other field in the pod spec which has this special treatment. The serviceAccount field is auto-populated in a special way (not by pod spec defaulting) because it derives its value from serviceAccount. We incorrectly detect the absence of this in the desired spec as a change in the pod template.

NOTE for this to happen, ComputeHash would have had to have changed meaning from underneath us. Something which I will address in a v1.1 specific PR.

huikang:
> > Hi, @jessesuen , I am just wondering what if the existing pod has both serviceAccountName and serviceAccount in the template. Will this PR handle the case properly? Thanks.
> 
> First, serviceAccountName and serviceAccount should always be the same value. I just verified that if you try to create a Deployment with different values, serviceAccount will get overridden with value from serviceAccountName.
> 
> Given that, `serviceAccountName` is the only thing we should care about. Users should be able to modify this field (add/remove/update), and the rollout should properly detect this as a change and perform an update.
> 
> However, the `serviceAccount` field in the pod is safe to ignore completely. It is a deprecated kubernetes field which has special treatment in k8s codebase. AFIK, there is not any other field in the pod spec which has this special treatment. The serviceAccount field is auto-populated in a special way (not by pod spec defaulting) because it derives its value from serviceAccountName. We incorrectly detect the absence of serviceAccount in the desired spec as a change in the pod template, which causes the unsolicited update.
> 
> NOTE for this to happen, ComputeHash would have had to have changed meaning from underneath us. Something which I will address in a v1.1 specific PR.

Thanks for the detailed explanation

Issue #1151, opened 2021-05-07
Investigate ordering of service patching vs. replicaset creation


jessesuen:
## Summary 

In our environment, it was observed that some pods of a Rollout's ReplicaSet never had readiness gate injected by the aws-loadbalancer-controller, but other Pods of the ReplicaSet were.

If we create a new replicaset before we patch the preview/canary service, its possible the pods of the replicaset will be created without a readiness gate because with AWS load balancer controller, it chooses to only inject readiness gates on pods that are targeted by a Service that is associated with a Load Balancer.

Need to investigate ordering on how we create RSs vs. patching services and see if we handle this correctly.

## Diagnostics

What version of Argo Rollouts are you running? v1.0.0-rc1

```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME>
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Looks like we reconcile the new replicaset before preview service. So we can try swapping the ordering.

```golang
	err = c.reconcileBlueGreenReplicaSets(activeSvc)
	if err != nil {
		return err
	}

	err = c.reconcilePreviewService(previewSvc)
	if err != nil {
		return err
	}
```

jessesuen:
> Looks like we reconcile the new replicaset before preview service. So we can try swapping the ordering.

Actually, after investigating, `reconcileBlueGreenReplicaSets` is not where we create the new ReplicaSet, so this is not as simple as switching the ordering. We actually create the ReplicaSets much earlier:

```golang
func (c *rolloutContext) rolloutBlueGreen() error {
	previewSvc, activeSvc, err := c.getPreviewAndActiveServices()
	if err != nil {
		return err
	}
	c.newRS, err = c.getAllReplicaSetsAndSyncRevision(true)
	if err != nil {
		return err
	}
```

`c.newRS` is assigned at that point, which is then needed by:

```
func (c *rolloutContext) reconcilePreviewService(previewSvc *corev1.Service) error {
	if previewSvc == nil {
		return nil
	}
	newPodHash := c.newRS.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
[...]

PR #1368, opened 2021-07-22
fix: retarget blue-green preview service before scaling up preview replicaset

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1151

This change implements the following behavior when creating the preview service and changing selectors.

1. create preview replicaset with size 0
2. retarget preview service to replicaset hash
3. scale up the preview replicaset

Previously we would create the preview replicaset with the desired size (without the separate create and scale steps). This will allow *all* pods of the preview to be considered for readiness gate injection by the AWS LoadBalancer controller

### DIFF
--- a/docs/features/bluegreen.md
+++ b/docs/features/bluegreen.md
@@ -68,6 +68,25 @@ spec:
       scaleDownDelayRevisionLimit: *int32
 ```
 
+## Sequence of Events
+
+The following describes the sequence of events that happen during a blue-green update.
+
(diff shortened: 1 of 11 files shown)


huikang:
@jessesuen , could you update the timeline in the b/g rollout doc: https://github.com/argoproj/argo-rollouts/blob/master/docs/features/bluegreen.md#previewservice? I think this change obsoletes some of the description. Thanks.

jessesuen:
Sure I'll update that as well. Thank you for the feedback.

khhirani (approved):
LGTM

PR #1375, opened 2021-07-27
fix: zero-value abortScaleDownDelay was not honored with setCanaryScale

jessesuen:
Recent PR did not rebase ontop of new scaledowndelay on abort behavior and the two caused a conflict in behavior

### DIFF
--- a/rollout/replicaset.go
+++ b/rollout/replicaset.go
@@ -53,7 +53,7 @@ func (c *rolloutContext) addScaleDownDelay(rs *appsv1.ReplicaSet, scaleDownDelay
 		}
 		return nil
 	}
-	deadline := metav1.Now().Add(scaleDownDelaySeconds * time.Second).UTC().Format(time.RFC3339)
+	deadline := metav1.Now().Add(scaleDownDelaySeconds).UTC().Format(time.RFC3339)
 	patch := fmt.Sprintf(addScaleDownAtAnnotationsPatch, v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey, deadline)
 	_, err := c.kubeclientset.AppsV1().ReplicaSets(rs.Namespace).Patch(ctx, rs.Name, patchtypes.JSONPatchType, []byte(patch), metav1.PatchOptions{})
 	if err == nil {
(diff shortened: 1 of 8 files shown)


huikang (approved):

jessesuen:
Still trying to fix tests.

rollout/replicaset.go
```diff
@@ -53,7 +53,7 @@ func (c *rolloutContext) addScaleDownDelay(rs *appsv1.ReplicaSet, scaleDownDelay
 		}
 		return nil
 	}
-	deadline := metav1.Now().Add(scaleDownDelaySeconds * time.Second).UTC().Format(time.RFC3339)
+	deadline := metav1.Now().Add(scaleDownDelaySeconds).UTC().Format(time.RFC3339)
```
jessesuen, line 56: Fixed confusing method behavior where duration was expected to be supplied in nanoseconds.

harikrongali: just for clarification, earlier scaleDownDelaySeconds * time.Second ==> wrong calculation? 
as scaleDownDelaySeconds is already in seconds

rollout/replicaset.go
```diff
@@ -144,9 +145,8 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 					newReplicasCount = int32(0)
 				}
 			}
-		} else {
-			abortScaleDownDelaySeconds := time.Duration(defaults.GetAbortScaleDownDelaySecondsOrDefault(c.rollout))
-			err = c.addScaleDownDelay(c.newRS, abortScaleDownDelaySeconds)
+		} else if abortScaleDownDelaySeconds != nil {
```
harikrongali, line 148: is this condition check required as it is verified in shouldDelayScaleDownOnAbort

utils/defaults/defaults_test.go
```diff
@@ -233,8 +263,20 @@ func TestGetAbortScaleDownDelaySecondsOrDefault(t *testing.T) {
...
+	{
+		// basic canary should not have scaledown delay seconds
+		canaryDefaultValue := &v1alpha1.Rollout{
+			Spec: v1alpha1.RolloutSpec{
+				Strategy: v1alpha1.RolloutStrategy{
+					Canary: &v1alpha1.CanaryStrategy{},
+				},
+			},
+		}
+		assert.Equal(t, time.Duration(0), *GetAbortScaleDownDelaySecondsOrDefault(canaryDefaultValue))
```
harikrongali, line 277: is GetAbortScaleDownDelaySecondsOrDefault nil here?

rollout/replicaset.go
jessesuen: All callers of this method were supplying scaleDownDelaySeconds in nanosecond duration. I also changed the callers to simply supply proper durations.

rollout/replicaset.go
jessesuen: You're right that it's not required. But I was trying to be defensive here in case there was a bug or change in behavior to shouldDelayScaleDownOnAbort().

khhirani (approved):
LGTM

PR #1380, opened 2021-07-29
feat: Adding support for Istio VirtualService spec.tls[]

agrawroh:

This PR adds support for TLS routes `spec.tls[]` in the Istio VirtualService. After this change Argo will start modifying the destination weights of any TLS routes specified using `tlsRoutes` mapped to the port numbers of the TLS routes in Istio VS.

This PR also includes the necessary documentation changes for Istio Traffic Routing sections.
[Preview Link](https://github.com/argoproj/argo-rollouts/blob/464fc86bddb1d230db8c6fd1028eaef1c4d5849e/docs/getting-started/istio/index.md)

## Checklist

## Testing Notes

We manually tested these changes by building a docker image including these changes and spinning `argo-rollouts` container with the new custom image.

This is the `trafficRouting` section which had both HTTP/TLS routes:

```
trafficRouting: {
  istio: {
    virtualService: {
      name: "dummyservice",
      routes: [
        "http-dummyservice"
      ],
      tlsRoutes: [
        {
          port: 38686,
          sniHosts: [
            "*.databricks.com",
            "*.svc.cluster.local",
            "localhost"
          ],
        },
      ],
    },
    destinationRule: {
      name: "dummyservice",
      canarySubsetName: CANARY_VERSION_NAME,
      stableSubsetName: STABLE_VERSION_NAME,
    },
  },
},
...
```

This was the initial state of the VirtualService before applying the `rollout` and as you can see the initial weights for TLS routes were 67/33.

initial-status

As you can see in the below screenshots, once the `rollout` got applied argo started modifying the destination weights of both the HTTP as well as TLS routes as per the rollout steps.

upd-status-1

upd-status-2


rollout/trafficrouting/istio/istio.go
```diff
@@ -26,8 +26,8 @@ const Type = "Istio"
 // NewReconciler returns a reconciler struct that brings the Virtual Service into the desired state
 func NewReconciler(r *v1alpha1.Rollout, client dynamic.Interface, recorder record.EventRecorder, virtualServiceLister, destinationRuleLister dynamiclister.Lister) *Reconciler {
 	return &Reconciler{
-		rollout: r,
-		log:     logutil.WithRollout(r),
+		rollout:               r,
+		log:                   logutil.WithRollout(r),
```
huikang, line 30: missing go fmt?

rollout/trafficrouting/istio/istio.go
```diff
@@ -57,30 +58,50 @@ const (
...
+			}
+			destinations, ok := route["route"].([]interface{})
+			if !ok {
+				return fmt.Errorf(invalidCasting, "tls[].route", "[]interface")
+			}
+			destination, ok := destinations[patch.destinationIndex].(map[string]interface{})
+			if !ok {
+				return fmt.Errorf(invalidCasting, "tls[].route[].destination", "map[string]interface")
+			}
+			destination["weight"] = float64(patch.weight)
```
huikang, line 94: Line 94-97 are duplicate to line 76-79. Please consider refactoring it to out of if...else

rollout/trafficrouting/istio/istio_types.go
```diff
@@ -30,23 +48,23 @@ type VirtualServiceHTTPRouteDestination struct {
 
 // Destination fields within the VirtualServiceDestination struct of the Virtual Service that the controller modifies
 type VirtualServiceDestination struct {
-	Host   string `json:"host,omitempty"`
+	Host string   `json:"host,omitempty"`
```
huikang, line 51: go fmt?

rollout/trafficrouting/istio/istio_types.go
```diff
@@ -30,23 +48,23 @@ type VirtualServiceHTTPRouteDestination struct {
...
 }
 
 type DestinationRuleSpec struct {
 	Subsets []Subset `json:"subsets,omitempty"`
 }
 
 type Subset struct {
-	Name   string            `json:"name,omitempty"`
-	Labels map[string]string `json:"labels,omitempty"`
+	Name string                  `json:"name,omitempty"`
```
huikang, line 66: go fmt?

huikang (commented):
Took a quick review. But I think it would be ideal if the doc change could be within this PR. Thanks.

rollout/trafficrouting/istio/istio_types.go
agrawroh: Sorry I forgot to run `go fmt` before pushing the changes. Will run and update the commit. Thanks!

rollout/trafficrouting/istio/istio_types.go
agrawroh: Sorry I forgot to run `go fmt` before pushing the changes. Will run and update the commit. Thanks!

rollout/trafficrouting/istio/istio.go
agrawroh: Sorry I forgot to run `go fmt` before pushing the changes. Will run and update the commit. Thanks!

rollout/trafficrouting/istio/istio.go
agrawroh: Done!

agrawroh:
> Took a quick review. But I think it would be ideal if the doc change could be within this PR. Thanks.

Thanks, @huikang! I have added the doc changes.

agrawroh:
> Took a quick review. But I think it would be ideal if the doc change could be within this PR. Thanks.

Thanks, @huikang! I have added the doc changes.

docs/getting-started/istio/index.md
```diff
@@ -33,15 +33,19 @@ spec:
...
+            - https-3000
 ...
 ```
 
 The VirtualService and route referenced in `trafficRouting.istio.virtualService` is required
-to have a HTTP route which splits between the stable and canary Services, referenced in the rollout.
-In this guide, those Services are named: `rollouts-demo-stable` and `rollouts-demo-canary` 
-respectively. The weight values for these services used should be initially set to 100% stable, 
+to have either an HTTP or a TLS route spec that splits between the stable and canary Services,
+referenced in the rollout. If the route is HTTP/TLS route it of type then it needs to be
```
jessesuen, line 45: Shouldn't this sentence read: "If the route is HTTPS/TLS route"

(missing S in HTTPS)

docs/getting-started/istio/index.md
```diff
@@ -33,15 +33,19 @@ spec:
           virtualService:
             # Reference to a VirtualService which the controller updates with canary weights
             name: rollouts-demo-vsvc
+            # optional if there is a single route in VirtualService, required otherwise
             routes:
-            - primary # optional if there is a single route in VirtualService, required otherwise
+            - http-primary
+            - https-3000
```
jessesuen, line 39: It seems the approach you are taking relies on convention and prefixing to determine if we need to modify HTTP routes vs TLS routes. Instead of convention, can we take a more explicit approach and introduce a new `tlsRoutes` field? Istio VirtualService also separates the two and I feel this would match their spec better.

The existing `routes` would continue to modify only HTTP routes, matching by name, and the new tlsRoutes would match by port numbers. The spec would look like this:

```yaml
          virtualService:
            name: rollouts-demo-vsvc
            routes:
            - primary
            tlsRoutes:
            - 3000
```

agrawroh: Sure, if you think that's cleaner then we can do that. Does it make sense to rename `routes` -> `httpRoutes` or it'll be a huge breaking change?

docs/getting-started/istio/index.md
agrawroh: That's correct! Thanks for pointing it out.

docs/getting-started/istio/index.md
jessesuen: > Sure, if you think that's cleaner then we can do that

Yep, I think it will be cleaner, and we will get type safety for free.

> or it'll be a huge breaking change?

Right. We can't break this, so we'll have to live with `routes` meaning HTTP routes.

agrawroh: Done the above changes as per suggestion. @jessesuen Could you please take another look?

jessesuen: @agrawroh - I took a closer look at the Istio VirtualService spec and I think we are oversimplifying by making tlsRoutes only match on numeric ports. It turns out port is an optional field for TLSMatchAttributes, and the tls match could be against something else (e.g. SNI hosts). For example the user may have a VirtualService like:

```yaml
  tls:
  - match:
    - port: 443
      sniHosts:
      - login.bookinfo.com
    route:
    - destination:
        host: login.prod.svc.cluster.local
  - match:
    - port: 443
      sniHosts:
      - reviews.bookinfo.com
    route:
    - destination:
        host: reviews.prod.svc.cluster.local
```

In the above example, port 443 is used in both routes. The user may want to split traffic on `reviews.bookinfo.com` but not `login.bookinfo.com`. I think the spec has to change to accommodate this. Instead of a port list, we should do something like:

```yaml
          virtualService:
            name: rollouts-demo-vsvc
            routes:
            - primary
            tlsRoutes:
            - port: 443
              sniHosts:
              - reviews.bookinfo.com
```

Both port and sniHosts would be optional but the user has to pick at least one. NOTE that there are also other fields like gateways, sourceLabels, sourceNamespace which can match. We don't necessarily need to support all of them off the bat, but we need to leave room to support them in the future.

agrawroh: Yeah, that make sense. I can add support for SNI in this PR. Do you think we need to support more things as of now or is it okay to do just port & SNI as a start.

jessesuen: I think port + SNI will be sufficient for this PR.

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -363,8 +363,10 @@ type IstioTrafficRouting struct {
 type IstioVirtualService struct {
 	// Name holds the name of the VirtualService
 	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
-	// Routes are list of routes within VirtualService to edit. If omitted, VirtualService must have a single route
+	// A list of HTTP routes within VirtualService to edit. If omitted, VirtualService must have a single route of any type.
 	Routes []string `json:"routes,omitempty" protobuf:"bytes,2,rep,name=routes"`
+	// A list of TLS/HTTPS routes (port numbers) within VirtualService to edit. If omitted, VirtualService must have a single route of any type.
+	TlsRoutes []int64 `json:"tlsRoutes,omitempty" protobuf:"bytes,3,rep,name=tlsRoutes"`
```
jessesuen, line 369: I think the golang struct field name should be TLSRoutes

agrawroh: Will change it.

docs/getting-started/istio/index.md
agrawroh: @jessesuen I pushed the change if you wanna take another look at it.

docs/getting-started/istio/index.md
```diff
@@ -33,15 +33,28 @@ spec:
...
+              - reviews.bookinfo.com
+              - localhost
 ...
 ```
 
 The VirtualService and route referenced in `trafficRouting.istio.virtualService` is required
-to have a HTTP route which splits between the stable and canary Services, referenced in the rollout.
-In this guide, those Services are named: `rollouts-demo-stable` and `rollouts-demo-canary` 
-respectively. The weight values for these services used should be initially set to 100% stable, 
+to have either an HTTP or a TLS route spec that splits between the stable and the canary services,
```
huikang, line 51: can we have both http and tls route in a rollout spec? Could not infer the answer from the description.

docs/getting-started/istio/index.md
```diff
@@ -55,15 +68,27 @@ spec:
   hosts:
   - rollouts-demo.local
   http:
-  - name: primary  # Should match spec.strategy.canary.trafficRouting.istio.virtualService.routes
+  - name: http-primary  # Should match spec.strategy.canary.trafficRouting.istio.virtualService.routes
```
huikang, line 71: spec.strategy.canary.trafficRouting.istio.virtualService.routes -> `rollout.spec.strategy.canary.trafficRouting.istio.virtualService.routes`

docs/getting-started/istio/index.md
```diff
@@ -55,15 +68,27 @@ spec:
...
+    route:
+    - destination:
+        host: rollouts-demo-stable  # Should match spec.strategy.canary.stableService
```
huikang, line 74: ditto

docs/getting-started/istio/index.md
```diff
@@ -55,15 +68,27 @@ spec:
...
+      weight: 100
+    - destination:
+        host: rollouts-demo-canary  # Should match spec.strategy.canary.canaryService
```
huikang, line 77: ditto

docs/getting-started/istio/index.md
```diff
@@ -55,15 +68,27 @@ spec:
...
+      weight: 0
+  tls:
+  - match:
+    - port: 3000  # Should match the port number of the route defined in spec.strategy.canary.trafficRouting.istio.virtualService.tlsRoutes
```
huikang, line 81: ditto

docs/getting-started/istio/index.md
```diff
@@ -55,15 +68,27 @@ spec:
...
+      sniHosts: # Should match all the SNI hosts of the route defined in spec.strategy.canary.trafficRouting.istio.virtualService.tlsRoutes
```
huikang, line 82: ditto

utils/evaluate/evaluate.go
```diff
@@ -168,3 +168,20 @@ func asFloat(in interface{}) float64 {
 	}
 	panic(fmt.Sprintf("asFloat() not supported on %v %v", reflect.TypeOf(in), in))
 }
+
+// Check whether two slices of type string are equal or not.
+func Equal(a, b []string) bool {
```
huikang, line 173: This method seems not used anywhere in the PR. Shall we remove it?

agrawroh: It's actually getting used [here](https://github.com/argoproj/argo-rollouts/pull/1380/files#diff-bc2f3a4d5d2fd799a169ceccb8bf97ae8b8a45f6a69009525c24e41fd846ef5eR644) while comparing the SNI Hosts list from the Rollout Def and the actual Istio VirtualService.

docs/getting-started/istio/index.md
agrawroh: Will change! Thanks.

docs/getting-started/istio/index.md
agrawroh: Will do! Thanks

utils/evaluate/evaluate.go
huikang: The link doesn't point to the right place and search in the browser does not show me anything.

agrawroh: Sorry about that. It's **L#644** in file `rollout/trafficrouting/istio/istio.go`.

Screen Shot 2021-08-10 at 4 28 29 PM

jessesuen (approved):
Great work!

docs/getting-started/istio/index.md
agrawroh: @huikang I've addressed all the README related feedback in #1414

PR #1381, opened 2021-07-29
feat: support management of multiple Istio VirtualService objects

### LINKED ISSUE #1100: Support having multiple VirtualService objects
# Summary

Rollout does not seem to support having (specifically, updating weights in) multiple VirtualService objects.

Per the Rollout specification currently at
https://argoproj.github.io/argo-rollouts/features/specification/, in particular, this part:
```
        # Istio traffic routing configuration
        istio:
          virtualService: 
            name: rollout-vsvc  # required
            routes:
            - primary # At least one route is required
```
it _does_ support having multiple Istio HTTPRoute sub-objects in a VirtualService (via having that `routes` member be a list of HTTPRoute names).

However, Rollout does _not_ support having multiple VirtualService objects (since the value of member `virtualService` is single object)

If member `virtualService`, which currently gives one sub-object specifying a VirtualService, were replaced with a member (e.g., `virtualServices`) listing _multiple_ such specification sub-objects, then Rollout would support having multiple VirtualServices (each with its own name and own list of HTTPRoute names).

For backward compatibility, Rollout could keep `virtualService` when adding `virtualServices` (making it a semantic error to specify both).

# Use Cases

This enhancement is needed to support Istio-based setups in which a service's routes are given in multiple VirtualService objects rather than just a single one. 

(Recall that Istio allows splitting up a service's routing into multiple VirtualService objects. Also, note that mapping from ingress host names to services might involve multiple host names, per–URI-path routing, and multiple target services. For multiple routes going to the same target service, sometimes splitting into multiple VirtualService objects makes things (e.g., matching rules) much simpler.)

---

darshanhs09:
Fixes #1100 

Currently, Argo Rollout spec only supports a single VirtualService (VS) in its Istio integrated canary.
Multiple VirtualService support is needed in Istio-based setups in which a service's routes are given in multiple VirtualService objects rather than just a single one.

Use cases:
This feature is required to enable canary/blue-green traffic shifting for services with multiple port delegate VirtualServices. 
It's very common to have multiple VirtualService's pointing to a single backend service.

Signed-off-by: Darshan Hassan Shashikumar   hasssh@usc.com

Note: Users are recommended to use "trafficRouting.istio.virtualServices" instead of "trafficRouting.istio.virtualService". Because it provides more flexibility in adding one or more virtualServices. I am still keeping virtualService is to provide the backward compatibility. Going forward, we can deprecate "trafficRouting.istio.virtualService".

Checklist:
 Either (a) I've created an enhancement proposal and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
 The title of the PR is (a) conventional, (b) states what changed, and (c) suffixes the related issues number. E.g. "fix(controller): Updates such and such. Fixes #1234".
 I've signed my commits with DCO
 I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
 My builds are green. Try syncing with master if they are not.
 My organization is added to USERS.md.


docs/getting-started/istio/index.md
```diff
@@ -30,32 +30,63 @@ spec:
...
+  http:
+  - name: primary  # Should match spec.strategy.canary.trafficRouting.istio.virtualServices.routes
+    route:
+    - destination:
+        host: rollouts-demo-stable  # Should match spec.strategy.canary.stableService
+      weight: 100
+    - destination:
+        host: rollouts-demo-canary  # Should match spec.strategy.canary.canaryService
+      weight: 0
+
```
huikang, line 75: Changes made to the example override the existing single virtual service cases, so it is a bit hard for beginners to follow.

Instead of mixing them, can we separate these two cases by adding a new example for `trafficRouting.istio.virtualServices`

docs/getting-started/istio/index.md
```diff
@@ -30,32 +30,63 @@ spec:
...
+          virtualServices:
+          # One or more virtualServices can be configured
+          # Reference to a VirtualService which the controller updates with canary weights.
+          - name: rollouts-demo-vsvc1
             routes:
             - primary # optional if there is a single route in VirtualService, required otherwise
+          # Reference to a VirtualService which the controller updates with canary weights
+          - name: rollouts-demo-vsvc2
+            routes:
+            - secondary # optional if there is a single route in VirtualService, required otherwise
```
huikang, line 42: Can we change `secondary` to `primary` like the example in `ocs/getting-started/istio/multipleVirtualsvc.yaml` and put some note saying the route name should match in virtualservice? If not, could you add some explanation the difference between primary and secondary routes?

docs/getting-started/istio/rollout.yaml
```diff
@@ -10,10 +10,13 @@ spec:
...
         istio:
-          virtualService:
-            name: rollouts-demo-vsvc
+          virtualServices:
+          - name: rollouts-demo-vsvc1 # At least one virtualService is required
             routes:
             - primary # At least one route is required
+          - name: rollouts-demo-vsvc2
+            routes:
+            - secondary # At least one route is required
```
huikang, line 19: Same here, can we add a new file named `rollout-multiple-vsvs.yaml`, instead of changing the existing one.

BTW, should this be primary, because there is no secondary route in `rollouts-demo-vsvc2`

pkg/apis/rollouts/validation/validation_references.go
```diff
@@ -214,25 +214,45 @@ func ValidateIngress(rollout *v1alpha1.Rollout, ingress v1beta1.Ingress) field.E
...
+		virtualServices = rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualServices
+	} else {
+		fldPath = field.NewPath("spec", "strategy", "canary", "trafficRouting", "istio", "virtualService", "name")
+		virtualServices = []v1alpha1.IstioVirtualService{rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualService}
 	}
-	err = istio.ValidateHTTPRoutes(rollout, httpRoutes)
-	if err != nil {
-		msg := fmt.Sprintf("Istio VirtualService has invalid HTTP routes. Error: %s", err.Error())
-		allErrs = append(allErrs, field.Invalid(fldPath, vsvcName, msg))
+
```
huikang, line 229: Since the PR makes both VirtualService and VirtualServices as optional, we need to check if virtualservice is empty the validation. Therefore, I think the new function `ValidateRolloutReferencedResources` needs to be called here to make it consistent with the validation logic. WDYT?

rollout/trafficrouting/istio/controller_test.go
```diff
@@ -78,6 +79,118 @@ func TestGetReferencedVirtualServices(t *testing.T) {
...
+		c := NewFakeIstioController(vService)
+		_, err := c.GetReferencedVirtualServices(&ro)
+		expectedErr := field.Invalid(field.NewPath("spec", "strategy", "canary", "trafficRouting", "istio", "virtualServices", "name"), "istio-vsvc2-name", "virtualservices.networking.istio.io \"istio-vsvc2-name\" not found")
+		assert.Equal(t, expectedErr.Error(), err.Error())
+	})
+}
+
+func TestGetReferencedMultipleVirtualServicesInvalidConfig(t *testing.T) {
+
+	multipleVirtualService := []v1alpha1.IstioVirtualService{{Name: "istio-vsvc1-name", Routes: nil}}
```
huikang, line 140: Please add another invalid case where both are configured.

rollout/trafficrouting/istio/controller.go
```diff
@@ -174,32 +174,67 @@ func (c *IstioController) EnqueueRolloutFromIstioVirtualService(vsvc interface{}
...
 		canary := ro.Spec.Strategy.Canary
 		if canary.TrafficRouting != nil && canary.TrafficRouting.Istio != nil {
 			var vsvc *unstructured.Unstructured
 			var err error
-			vsvcNamespace, vsvcName := istioutil.GetVirtualServiceNamespaceName(canary.TrafficRouting.Istio.VirtualService.Name)
-			if vsvcNamespace == "" {
-				vsvcNamespace = ro.Namespace
+			var vsvcs []v1alpha1.IstioVirtualService
+
+			err = c.validateRolloutVirtualServicesConfig(ro)
```
huikang, line 205: As mentioned earlier, this validation should be moved to the earlier validation path.

rollout/trafficrouting/istio/controller_test.go
```diff
@@ -78,6 +79,118 @@ func TestGetReferencedVirtualServices(t *testing.T) {
...
+	t.Run("get referenced virtualService - fail", func(t *testing.T) {
+		c := NewFakeIstioController(vService)
+		_, err := c.GetReferencedVirtualServices(&ro)
+		fldPath := field.NewPath("spec", "strategy", "canary", "trafficRouting", "istio")
+		expected := fmt.Sprintf("%s: Internal error: either VirtualService or VirtualServices must be configured", fldPath)
+		assert.Equal(t, expected, err.Error())
+	})
+}
+
+func TestGetReferencedNoVirtualServicesConfig(t *testing.T) {
```
huikang, line 170: This case can be added to `TestGetReferencedMultipleVirtualServicesInvalidConfig` to reduce some duplication.

rollout/trafficrouting/istio/istio.go
```diff
@@ -436,34 +436,45 @@ func (r *Reconciler) SetWeight(desiredWeight int32) error {
...
+		modifiedVirtualService, modified, err := r.reconcileVirtualService(vsvc, virtualService.Routes, desiredWeight)
+		if err != nil {
+			return err
+		}
+		if !modified {
+			continue
+		}
+		_, err = client.Update(ctx, modifiedVirtualService, metav1.UpdateOptions{})
+		if err == nil {
+			r.log.Debugf("UpdatedVirtualService: %s", modifiedVirtualService)
```
huikang, line 475: nit: `Updated VirtualService`

utils/istio/istio.go
```diff
@@ -64,19 +64,42 @@ func GetVirtualServiceNamespaceName(vsv string) (string, string) {
...
+	if rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualServices != nil {
+		return true
+	}
+	return false
+}
+
 // GetRolloutVirtualServiceKeys gets the referenced VirtualService and its namespace from a Rollout
 func GetRolloutVirtualServiceKeys(ro *v1alpha1.Rollout) []string {
+	var virtualServices []v1alpha1.IstioVirtualService
+	var virtualServiceKeys []string
```
huikang, line 77: Are we getting keys or just `namespace+name`? The variable name is kind of confusing.

utils/istio/istio.go
```diff
@@ -64,19 +64,42 @@ func GetVirtualServiceNamespaceName(vsv string) (string, string) {
 	return namespace, name
 }
 
+func MultipleVirtualServiceConfigured(rollout *v1alpha1.Rollout) bool {
+	if rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualServices != nil {
```
huikang, line 68: how about changing to `return rollout.Spec.Strategy.Canary.TrafficRouting.Istio.VirtualServices != nil`

utils/istio/istio.go
```diff
@@ -64,19 +64,42 @@ func GetVirtualServiceNamespaceName(vsv string) (string, string) {
...
+	if canary == nil || canary.TrafficRouting == nil || canary.TrafficRouting.Istio == nil ||
+		(canary.TrafficRouting.Istio.VirtualServices == nil && canary.TrafficRouting.Istio.VirtualService.Name == "") {
 		return []string{}
 	}
 
-	namespace, name := GetVirtualServiceNamespaceName(canary.TrafficRouting.Istio.VirtualService.Name)
-	if namespace == "" {
-		namespace = ro.Namespace
+	if MultipleVirtualServiceConfigured(ro) {
+		if canary.TrafficRouting.Istio.VirtualService.Name != "" {
```
huikang, line 86: this should not happen, right? Because the validation logic will not allow to reach this place. If so, this is a panic.
Or we can remove line 86 - 88

huikang (commented):
Thanks for the contributions. My review comments are added.

docs/getting-started/istio/index.md
darshanhs09: The reason of overriding this existing VS doc is Users are recommended to use "trafficRouting.istio.virtualServices" instead of "trafficRouting.istio.virtualService". Because it provides more flexibility in adding one or more virtualServices. I am still keeping virtualService is to provide the backward compatibility.
Going forward, we can deprecate "trafficRouting.istio.virtualService".

huikang: Got it. Then I think you need to 1) update the specification doc https://github.com/argoproj/argo-rollouts/blob/master/docs/features/specification.md and 2) put the above explanation in the PR. Thanks.

pkg/apis/rollouts/validation/validation_references.go
darshanhs09: This is already taken care here 
https://github.com/argoproj/argo-rollouts/pull/1381/files#diff-7fecb947ba065e9c2f022d05ca775a0655d2c94e1776b89b95291fed79924df9R177
this is called by GetReferencedVirtualServices method which is called before the validation.

rollout/trafficrouting/istio/controller_test.go
darshanhs09: TestGetReferencedMultipleVirtualServicesInvalidConfig -> This test case is for both virtualService and virtualServices configured.

rollout/trafficrouting/istio/controller.go
darshanhs09: GetReferencedVirtualServices method iterates to rollout virtualServices and stores the k8s virtualService Record.
It makes sense to validate here before proceeding to the rollout virtualServices iteration.

utils/istio/istio.go
darshanhs09: Basically, we get the list of keys. each key will be in the format of <namespace>/<virtualService Name>

docs/getting-started/istio/index.md
darshanhs09: I have added the difference between primary and secondary routes. please check.

utils/istio/istio.go
darshanhs09: Updated the PR

docs/getting-started/istio/rollout.yaml
darshanhs09: Updated the PR. Can you please check?

shakti-das:
Hi @jessesuen, could you please review this PR when you find the time. We want to get Argo Rollouts to production soon and this issue is a blocker right now. Thank you.

pkg/apis/rollouts/validation/validation_references.go
huikang: @darshanhs09, what I meant is that the new validation function validateRolloutVirtualServicesConfig should be called before GetReferencedVirtualServices. validateRolloutVirtualServicesConfig should be condolidated with existing virtual service check at

https://github.com/argoproj/argo-rollouts/blob/2620490c9ab96e9da43b9c014fdf802c5e87dfa3/pkg/apis/rollouts/validation/validation_references.go#L216

which is done before reconcile.

rollout/trafficrouting/istio/controller.go
huikang: Please see my comment about where to put the validation: https://github.com/argoproj/argo-rollouts/pull/1381#discussion_r682662879

utils/istio/istio.go
huikang: so how about vsvsNamespaceName? Key seems have different meanings in k8s.

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
+kind: VirtualService
+metadata:
+  name: rollouts-demo-vsvc1
+spec:
+  gateways:
+  - rollouts-demo-gateway
+  hosts:
+  - rollouts-demo-vsvc1.local
+  http:
+  - name: primary  # Should match spec.strategy.canary.trafficRouting.istio.virtualServices.routes
```
huikang, line 69: "Should match spec.strategy...." -> "Should match rollout.spec.strategy...."

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
+    route:
+    - destination:
+        host: rollouts-demo-stable  # Should match spec.strategy.canary.stableService
```
huikang, line 72: ditto

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
+        port:
+          number: 15372
+      weight: 100
+    - destination:
+        host: rollouts-demo-canary  # Should match spec.strategy.canary.canaryService
```
huikang, line 77: ditto

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
+  name: rollouts-demo-vsvc2
 spec:
   gateways:
   - rollouts-demo-gateway
   hosts:
-  - rollouts-demo.local
+  - rollouts-demo-vsvc2.local
   http:
-  - name: primary  # Should match spec.strategy.canary.trafficRouting.istio.virtualService.routes
+  - name: secondary  # Should match spec.strategy.canary.trafficRouting.istio.virtualServices.routes
```
huikang, line 95: ditto

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
     route:
     - destination:
         host: rollouts-demo-stable  # Should match spec.strategy.canary.stableService
```
huikang, line 98: ditto

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
+        port:
+          number: 15373
       weight: 100
     - destination:
         host: rollouts-demo-canary  # Should match spec.strategy.canary.canaryService
```
huikang, line 103: ditto

rollout/trafficrouting/istio/controller.go
```diff
@@ -174,32 +174,67 @@ func (c *IstioController) EnqueueRolloutFromIstioVirtualService(vsvc interface{}
 	}
 }
 
+func (c *IstioController) validateRolloutVirtualServicesConfig(r *v1alpha1.Rollout) error {
```
huikang, line 177: This one should be called at https://github.com/argoproj/argo-rollouts/pull/1381/files#diff-ee634f180c1529c412a4207985224a3d0670f7332d00e967fa72f090af273f28L216 as well

docs/getting-started/istio/index.md
```diff
@@ -30,38 +30,79 @@ spec:
...
+`trafficRouting.istio.virtualServices`. `trafficRouting.istio.virtualServices` helps in adding
+one or more virtualServices unlike `trafficRouting.istio.virtualService` where only single virtualService can be added.
+This is required to have a HTTP route which splits between the stable and canary Services, referenced in the rollout.
+In this example, primary route splits the http route with port number 15372 between stable and canary services;
+secondary route splits the http route with port number 15373 between stable and canary services.
 In this guide, those Services are named: `rollouts-demo-stable` and `rollouts-demo-canary` 
 respectively. The weight values for these services used should be initially set to 100% stable, 
 and 0% on the canary. During an update, these values will be modified by the controller.
+If there are multiple VirtualService then weight values for stable and canary service of each VirtualService
+will be modified by the controller.
```
huikang, line 56: how about "... each VirtualService will be modified by the controller." -> "each VirtualService will be modified by the controller simultaneously" ?

huikang (commented):
Thanks for the update. Looks much better.

pkg/apis/rollouts/validation/validation_references.go
darshanhs09: @huikang 
Please check the below code. getRolloutReferencedResources  is called before ValidateRolloutReferencedResources.
As i mentioned, the reason for introducing this validation in GetReferencedVirtualServices is because this method iterates the virtualservices to fill the k8s virtualservice record references. So, it doesn't really make sense to introduce this validation in ValidateVirtualService. But i can do one thing if required, before calling the method GetReferencedVirtualServices i can validate the multi VSes invalid config.
[https://github.com/argoproj/argo-rollouts/blob/2620490c9ab96e9da43b9c014fdf802c5e87dfa3/rollout/controller.go#L468](url)

Please let  me know your thoughts

rollout/trafficrouting/istio/controller.go
darshanhs09: [copy pasting the same response from previous review comment.]

@huikang 
Please check the below code. getRolloutReferencedResources  is called before ValidateRolloutReferencedResources.
As i mentioned, the reason for introducing this validation in GetReferencedVirtualServices is because this method iterates the virtualservices to fill the k8s virtualservice record references. So, it doesn't really make sense to introduce this validation in ValidateVirtualService. But i can do one thing if required, before calling the method GetReferencedVirtualServices i can validate the multi VSes invalid config.
[https://github.com/argoproj/argo-rollouts/blob/2620490c9ab96e9da43b9c014fdf802c5e87dfa3/rollout/controller.go#L468](url)

Please let  me know your thoughts

pkg/apis/rollouts/validation/validation_references.go
huikang: Hi, @darshanhs09, 

> But i can do one thing if required, before calling the method GetReferencedVirtualServices i can validate the multi VSes invalid config.

This is exactly what I suggested in my last comment (https://github.com/argoproj/argo-rollouts/pull/1381#discussion_r682662879), right? Please do so. Thanks.

utils/istio/istio.go
huikang: Such scenario should be detected earlier in the validationVirtualServiceConfig, I believe this if and return empty string array may introduce a subtle bug in the future.

rollout/trafficrouting/istio/istio_test.go
```diff
@@ -616,3 +617,255 @@ func TestUpdateHashDestinationRuleNotFound(t *testing.T) {
...
+	route1 := routes1[0].(map[string]interface{})
+	checkDestination(t, route1, "stable", 90)
+	checkDestination(t, route1, "canary", 10)
+
+	vsvcUn2, err := client.Resource(istioutil.GetIstioVirtualServiceGVR()).Namespace(ro.Namespace).Get(context.TODO(), "vsvc2", metav1.GetOptions{})
+	assert.NoError(t, err)
+	routes2, _, _ := unstructured.NestedSlice(vsvcUn2.Object, "spec", "http")
+	route2 := routes2[0].(map[string]interface{})
+	checkDestination(t, route2, "stable", 90)
+	checkDestination(t, route2, "canary", 10)
```
huikang, line 870: The above check can be done in a for  look to reduce code duplication

```go
for _, vsvName := range []string{"vsvc1", "vsvc2"} {
        vsvcUn, err := client.Resource(istioutil.GetIstioVirtualServiceGVR()).Namespace(ro.Namespace).Get(context.TODO(), vsvName, metav1.GetOptions{})
	assert.NoError(t, err)
	routes1, _, _ := unstructured.NestedSlice(vsvcUn.Object, "spec", "http")
	route1 := routes1[0].(map[string]interface{})
	checkDestination(t, route1, "stable", 90)
	checkDestination(t, route1, "canary", 10)
}
```

huikang:
Hi, @jessesuen , I almost finished reviewing this PR. Do you mind kicking off the CI flow? Thanks.

rollout/trafficrouting/istio/istio_test.go
darshanhs09: Thanks!

pkg/apis/rollouts/validation/validation_references.go
darshanhs09: @huikang not really! It doesn't really make sense to validate (validateRolloutVirtualServicesConfig) in ValidateVirtualService because it is called after getRolloutReferencedResources. As i mentioned, i am validating before GetReferencedVirtualServices. Please check the latest update.

darshanhs09:
Hi @jessesuen , I have update the PR and the review comments are being addressed. 
Can you please kickoff the CI flow again?

rollout/trafficrouting/istio/istio_test.go
huikang: Hi,  @darshanhs09, please leave it unresolved as you have not made the change as request.

utils/istio/istio.go
huikang: not resolved yet

utils/istio/istio.go
huikang: please resolve this comment

huikang (changes requested):
Please do not mark the comment resolved as you have not made any change.

shakti-das:
@darshanhs09 could you please address the open review comments. Thanks.

utils/istio/istio.go
darshanhs09: The only way to identify the rollout is with namespace and name. So, having the key as the variable name would make more sense. 
I didn't change this in my latest update. Let me know your thoughts.

utils/istio/istio.go
darshanhs09: Yes, this check is already taken care. I have removed it.
Please check, let me know if you are okay with this change then I will resolve it.

rollout/trafficrouting/istio/istio_test.go
darshanhs09: Hi @huikang, I missed to update this change. Sorry for that.
Please check the latest.

darshanhs09:
> @darshanhs09 could you please address the open review comments. Thanks.

Hi @shakti-das, I was out of town so couldn't check. 
I have updated the changes now.
Thanks

huikang:
Could you resolve the conflict? Else LGTM.

sboobna:
@darshanhs09

darshanhs09:
TlsRoutes is a new field introduced in virtualService.
Now, I cannot commit my changes directly not just because it is conflicting with some of my changes.
I have to also handle this new field in virtualServices.
So, I made the changes to handle this new field in multiple virtual services. But I am yet to test it.
Once I am done with my testing I will update the PR.

Thanks

huikang:
LGTM

darshanhs09:
@huikang Thanks for reviewing.
@jessesuen Can you please run the CI flow ?

jessesuen (approved):

jessesuen:
Great work!

Issue #1029, opened 2021-03-10
Ability to control stable scale when using traffic routing


jessesuen:
# Summary

When using traffic routing with canary, the stable stack is never scaled down, until the rollout is fully promoted. This means that we are double the number of total replicas in the environment eventually (when the traffic weight reaches 100). The reasons for doing so are explained here: https://github.com/argoproj/argo-rollouts/issues/430#issuecomment-595995434

> The reason for this is that when a mesh or ingress like Istio is being used, users will be now be able to shift traffic much more rapidly/sporadically (e.g. go from 1% to 99% in an extreme example).

> During design, it was an important requirement to be able to abort a rollout and immediately go from XX% traffic percentage back down to 0% instantly, without it becoming delayed/prevented by external factors such as replicaset and pod orchestration. So the decision was made to keep the canary stack size the same as the stable stack (e.g. simply keep replica counts in-sync), for the duration of the update.

However, some users have requested that the stable stack be scaled inversely to the canary weight, in order to save on resources (e.g. a baremetal clusters where resource provisioning cannot be doubled). For these users, it is acceptable to rollback slowly during an abort, and they would be willing to sacrifice instantaneous rollbacks.

This proposal is to offer a `setStableScale` option, which would set the number of stable replicas to be the inverse of the canary weight.

```yaml
steps:
- setStableScale:
    matchTrafficWeight: true  # it would be inverse of canary weight (default when using basic canary)
- setStableScale:
    weight: 100   # default behavior of canary using trafficRouting
- setStableScale:
    replicas: 4     # not sure anyone would need this option, but can have it to be consistent with setCanaryScale
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

nestorsokil:
Hey, came in to create a ticket for this and then saw it's already here :) 
I think that's a very important feature to have. For my use case the bloated pods during canary is probably the only problem with using rollouts.

huikang:
Hi, @jessesuen , I can take this one if no significant effort has been done on it.

jessesuen:
Thinking about this some more, we probably don't need the flexibility for this to be implemented as a new canary step type and can simply have a boolean or constant values to indicate the behavior of the stable scale. e.g.:

```yaml
spec:
  strategy:
    canary:
      proportionalStableScale: true  # need better name
```

```yaml
spec:
  strategy:
    canary:
[...]

PR #1382, opened 2021-07-30
Ability to control stable scale when using traffic routing

perenesenko:
Checklist:


perenesenko:
Fix for https://github.com/argoproj/argo-rollouts/issues/1029

bpoland:
@perenesenko thank you for working on this!  I have one question/edge case that I'm wondering about.  I will preface this by saying that I'm not at all familiar with the code and so this may be a silly question :)

With this change, will we end up scaling down the stable ReplicaSet a little bit too early particularly if someone uses `setCanaryScale` to manually inflate the number of canary pods?

For example if we had
```
- setWeight: 0
- setCanaryScale:
  weight: 100
- pause: {}
```
would that result in the `stable` ReplicaSet being scaled down to 0?

I also would want to make sure that when increasing the canary percentage by a large amount, that we wait for the "new" canary pods to be up and ready before scaling down the stable pods (or at least that the `maxUnavailable` is respected throughout the process).  I only see mention of the `AvailableReplicas` in the rollback scenario but maybe I'm missing something.

Hopefully this makes sense, and again thanks for working on this change.  I am excited to use it once it's released!

perenesenko:
Hi @bpoland Interesting. Let me check that scenario

perenesenko:
Hi @bpoland 

1. If we set the `setCanaryScale.weight:100` then yes we will scale down the `stable` to 0. That's the expected by new dynamic scaling behavior.
2. When we _increasing the canary percentage by a large amount_ yes, we're wait for the new pods are ready. So here we're using the same logic for the pods calculation as for the basic canary (based on `maxUnavailable` and `maxSurge`)

bpoland:
> Hi @bpoland
> 
> 1. If we set the `setCanaryScale.weight:100` then yes we will scale down the `stable` to 0. That's the expected by new dynamic scaling behavior.

Thanks for confirming.  It seems like it might be more intuitive to use `(100% - <percentage of traffic actually going to the canary pods>)` as the stable percentage, instead of using the canary scale. I think that would also help with the nginx 503 issue you mentioned.  But looking at the code it seems like that would be more complex.

If maintaining the current logic in this PR, I think it might be good to clearly document this since it might lead to unexpected behavior when using both `setCanaryScale` and `dynamicStableScale` together. 

> 2. When we _increasing the canary percentage by a large amount_ yes, we're wait for the new pods are ready. So here we're using the same logic for the pods calculation as for the basic canary (based on `maxUnavailable` and `maxSurge`)

Awesome, thanks!

harikrongali:
as per the current design, dynamicStableScale & setCanaryScale are mutually exclusive and shouldn't be used together. I would propose to make a validation check to not allow both configurations together.

bpoland:
> as per the current design, dynamicStableScale & setCanaryScale are mutually exclusive and shouldn't be used together. I would propose to make a validation check to not allow both configurations together.

Interesting, can you share the design doc?  It seems to me that keeping the `stable` ReplicaSet at or above the percentage of traffic that it's supposed to be receiving (not necessarily the size of the `canary` ReplicaSet which can be modified by `setCanaryScale`) would allow both options to be used together.

perenesenko:
Going to cancel this PR and we will implement the feature by providing `setStableScale` configuration. More details in a ticket: [here](https://github.com/argoproj/argo-rollouts/issues/1029#issuecomment-899518035)

jessesuen:
I would not abandon this PR yet because I think the syntax for controlling dynamic stable scaling is correct, but the implementation needs to simply calculate the stable scale differently:
https://github.com/argoproj/argo-rollouts/issues/1029#issuecomment-900545326

Issue #1383, opened 2021-07-30
Promote full did not work against BlueGreen with previewReplicaCount


alexmt:
## Summary 

Rollout stuck in degrade state during blue-green rollout. 

## Diagnostics

Rollout got paused (as expected) during blue-green. It was not approved within `progressDeadlineSeconds` timeout and switched into Degraded state. Promote was executed using Argo CD action but nothing happened - rollout stayed in Degraded state.

Based on logs rollout got unpaused but controller was trying to set `previewReplicaCount` to the new replicaset instead of `spec.replicas`.

```yaml
  strategy:
    blueGreen:
      activeService: paaf-active-service
      antiAffinity:
        requiredDuringSchedulingIgnoredDuringExecution: {}
      autoPromotionEnabled: false
      previewReplicaCount: 1
      previewService: paaf-preview-service
      scaleDownDelayRevisionLimit: 2
      scaleDownDelaySeconds: 300
```
status:

```
status:
  HPAReplicas: 3
  availableReplicas: 3
  blueGreen:
    activeSelector: 77954699c5
    previewSelector: 6599fdc6d
  canary: {}
  conditions:
    - lastTransitionTime: '2021-07-30T08:17:29Z'
      lastUpdateTime: '2021-07-30T08:17:29Z'
      message: RolloutCompleted
      reason: RolloutCompleted
      status: 'False'
      type: Completed
    - lastTransitionTime: '2021-07-30T08:45:37Z'
      lastUpdateTime: '2021-07-30T08:45:37Z'
      message: Rollout is paused
      reason: RolloutPaused
      status: 'False'
      type: Paused
    - lastTransitionTime: '2021-07-30T09:17:29Z'
      lastUpdateTime: '2021-07-30T09:17:29Z'
      message: Rollout has minimum availability
      reason: AvailableReason
      status: 'True'
      type: Available
    - lastTransitionTime: '2021-07-30T09:39:12Z'
      lastUpdateTime: '2021-07-30T09:39:12Z'
      message: ReplicaSet "reducted-rollout-6599fdc6d" has timed out progressing.
      reason: ProgressDeadlineExceeded
      status: 'False'
      type: Progressing
  currentPodHash: 6599fdc6d
  message: >-
    ProgressDeadlineExceeded: ReplicaSet "reducted-rollout-6599fdc6d" has timed out
    progressing.
  observedGeneration: '22'
  phase: Degraded
  promoteFull: true
  readyReplicas: 3
[...]

PR #1384, opened 2021-07-30
fix: Promote full did not work against BlueGreen with previewReplicaCount

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1383


alexmt (approved):
LGTM. 

Thanks for documenting `NewRSNewReplicas` as well - it is much easier to understand edge cases in this function now!

PR #1394, opened 2021-08-04
TrafficRouting ALB Step in Canary

khhirani:


jessesuen:
I'll take a look at this after https://github.com/argoproj/argo-rollouts/pull/1351 is merged. 

But one thing we need to do is fail validation for the other providers which do not support this feature.

khhirani:
@jessesuen #1351 is merged into master. Validation is taken care of in #1396

test/e2e/alb_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (s *ALBSuite) SetupSuite() {
+	s.E2ESuite.SetupSuite()
+}
+
+const actionTemplate = `{"Type":"forward","ForwardConfig":{"TargetGroups":[{"ServiceName":"%s","ServicePort":"%d","Weight":%d},{"ServiceName":"%s","ServicePort":"%d","Weight":%d}]}}`
+
+const actionTemplateWithExperiment = `{"Type":"forward","ForwardConfig":{"TargetGroups":[{"ServiceName":"%s","ServicePort":"%d","Weight":%d},{"ServiceName":"%s","ServicePort":"%d","Weight":%d},{"ServiceName":"%s","ServicePort":"%d","Weight":%d}]}}`
+
+func (s *ALBSuite) TestALBExperimentStep() {
```
jessesuen, line 32: Could you put this in `aws_test.go` instead? We don't need both aws_test.go and alb_test.go.

jessesuen (changes requested):
~`For the other providers which don't support additional destinations for traffic splitting (Istio, NGINX, Ambassador) could improve validation such that they fail when Experiment with traffic routing is used? ~

~Currently it appears we silently fail and do not honor the 3+ way split.~

Edit: nevermind I see that we are checking for this. Great!

Issue #1295, opened 2021-06-22
Rollout incorrectly manages 3 replicasets


gavin-db:
## Summary 

What happened/what you expected to happen?

In a dev region, we saw that one Rollout (graphql) was managing 3 replicasets for an extended period of time:

```
$kubectl get pods -n graphql
graphql-6c7865774c-h8mkl            2/2     Running   0          21h
graphql-6c7865774c-wbl78            2/2     Running   0          21h
graphql-6d588fb886-mpjtw            2/2     Running   0          3h22m
graphql-8547b7bb9c-zqc7f            2/2     Running   0          6h12m
```

```
$ kubectl get rollout -n graphql
NAME   DESIRED  CURRENT  UP-TO-DATE  AVAILABLE
graphql  2     4           4
$ kubectl get rollout -n graphql
...
Status:
  HPA Replicas:        4
  Available Replicas:  4
  Blue Green:
  Canary:
  Conditions:
    Last Transition Time:  2021-05-25T19:17:48Z
    Last Update Time:      2021-05-25T19:17:48Z
    Message:               Rollout has minimum availability
    Reason:                AvailableReason
    Status:                True
    Type:                  Available
    Last Transition Time:  2021-06-08T00:21:23Z
    Last Update Time:      2021-06-08T00:21:23Z
    Message:               RolloutCompleted
    Reason:                RolloutCompleted
    Status:                False
    Type:                  Completed
    Last Transition Time:  2021-06-08T00:43:50Z
    Last Update Time:      2021-06-08T00:43:50Z
    Message:               Rollout is paused
    Reason:                RolloutPaused
    Status:                False
    Type:                  Paused
    Last Transition Time:  2021-06-08T00:43:50Z
    Last Update Time:      2021-06-08T21:19:03Z
    Message:               Created new replica set "graphql-59495b85d4"
    Reason:                NewReplicaSetCreated
    Status:                True
    Type:                  Progressing
  Current Pod Hash:        59495b85d4
  Current Step Hash:       8cdbf64bf
  Current Step Index:      0
  Message:                 more replicas need to be updated
  Observed Generation:     46
  Phase:                   Progressing
  Ready Replicas:          4
  Replicas:                4
  Selector:                app=graphql
  Stable RS:               6c7865774c
Events:
  Type    Reason          Age   From                 Message
  ----    ------          ----  ----                 -------
  Normal  RolloutUpdated  27m   rollouts-controller  Rollout updated to revision 43
```

[...]

PR #1397, opened 2021-08-05
feat: ability to abort an update when exceeding progressDeadlineSeconds

### LINKED ISSUE #1376: Option to abort when exceeding progressDeadlineSeconds
# Summary

Currently, we only automatically abort an update if analysis fails. But if analysis is not used and a rollout exceeds progressDeadlineSeconds (e.g. because the new pods don't become ready/available or CrashLoopBackoff), the user should have an option to also automatically abort and rollback in this scenario.

# Use Cases

I want automated aborts but do not use analysis, or I want automated abort to complement analysis.

---

huikang:
- Add AbortExceedProgressDeadline to rollout spec
- when rollout doesn't use analysis, set AbortExceedProgressDeadline
  to true to scale down the new RS.
- e2e test added

close #1376 
Partial fix: #1295 

Checklist:

### DIFF
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -60,6 +60,10 @@ spec:
   # Defaults to 600s
   progressDeadlineSeconds: 600
 
+  # Whether to abort the update when ProgressDeadlineSeconds
+  # is exceeded if analysis is not used. Optional and default is false.
+  AbortExceedProgressDeadline: false
+
   # UTC timestamp in which a Rollout should sequentially restart all of
   # its pods. Used by the `kubectl argo rollouts restart ROLLOUT` command.
   # The controller will ensure all pods have a creationTimestamp greater
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -63,7 +63,7 @@ spec:
   # Whether to abort the update when ProgressDeadlineSeconds
   # is exceeded if analysis or experiment is not used.
   # Optional and default is false.
-  ProgressDeadlineAbort: false
+  progressDeadlineAbort: false
 
   # UTC timestamp in which a Rollout should sequentially restart all of
   # its pods. Used by the `kubectl argo rollouts restart ROLLOUT` command.
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -61,8 +61,9 @@ spec:
   progressDeadlineSeconds: 600
 
   # Whether to abort the update when ProgressDeadlineSeconds
-  # is exceeded if analysis is not used. Optional and default is false.
-  AbortExceedProgressDeadline: false
+  # is exceeded if analysis or experiment is not used.
+  # Optional and default is false.
+  ProgressDeadlineAbort: false
 
   # UTC timestamp in which a Rollout should sequentially restart all of
   # its pods. Used by the `kubectl argo rollouts restart ROLLOUT` command.
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -69,10 +69,10 @@ type RolloutSpec struct {
 	// Note that progress will not be estimated during the time a rollout is paused.
 	// Defaults to 600s.
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,8,opt,name=progressDeadlineSeconds"`
-	// AbortExceedProgressDeadline to whether abort the update when ProgressDeadlineSeconds
+	// ProgressDeadlineAbort is whether to abort the update when ProgressDeadlineSeconds
 	// is exceeded if analysis is not used. Default is false.
 	// +optional
-	AbortExceedProgressDeadline *bool `json:"abortExceedProgressDeadline,omitempty" protobuf:"varint,12,opt,name=abortExceedProgressDeadline"`
+	ProgressDeadlineAbort bool `json:"progressDeadlineAbort,omitempty" protobuf:"varint,12,opt,name=progressDeadlineAbort"`
 	// RestartAt indicates when all the pods of a Rollout should be restarted
 	RestartAt *metav1.Time `json:"restartAt,omitempty" protobuf:"bytes,9,opt,name=restartAt"`
 	// Analysis configuration for the analysis runs to retain
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -69,6 +69,10 @@ type RolloutSpec struct {
 	// Note that progress will not be estimated during the time a rollout is paused.
 	// Defaults to 600s.
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,8,opt,name=progressDeadlineSeconds"`
+	// AbortExceedProgressDeadline to whether abort the update when ProgressDeadlineSeconds
+	// is exceeded if analysis is not used. Default is false.
+	// +optional
+	AbortExceedProgressDeadline *bool `json:"abortExceedProgressDeadline,omitempty" protobuf:"varint,12,opt,name=abortExceedProgressDeadline"`
 	// RestartAt indicates when all the pods of a Rollout should be restarted
 	RestartAt *metav1.Time `json:"restartAt,omitempty" protobuf:"bytes,9,opt,name=restartAt"`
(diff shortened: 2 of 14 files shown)


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -69,6 +69,10 @@ type RolloutSpec struct {
 	// Note that progress will not be estimated during the time a rollout is paused.
 	// Defaults to 600s.
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,8,opt,name=progressDeadlineSeconds"`
+	// AbortExceedProgressDeadline to whether abort the update when ProgressDeadlineSeconds
+	// is exceeded if analysis is not used. Default is false.
+	// +optional
+	AbortExceedProgressDeadline *bool `json:"abortExceedProgressDeadline,omitempty" protobuf:"varint,12,opt,name=abortExceedProgressDeadline"`
```
jessesuen, line 75: Since:
1. the default of this feature is false
2. false booleans will be omitempty
3. and we don't need to distinguish unspecified vs. false

I think it would be better to change this to be a `bool` instead of a `*bool`

jessesuen: I propose we rename this to `progressDeadlineAbort` so that field names are prefixed consistently. i.e.

```yaml
spec:
    progressDeadlineSeconds: 600
    progressDeadlineAbort: true
```

rollout/sync.go
```diff
@@ -617,6 +617,11 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	// Check for progress. Only do this if the latest rollout hasn't completed yet and it is not aborted
 	if !isCompleteRollout && !isAborted {
 		switch {
+		case conditions.RolloutTimedOut(c.rollout, &newStatus) && defaults.GetAbortExceedProgressDeadlineOrDefault(c.rollout):
+			msg := fmt.Sprintf(conditions.RolloutTimeOutMessage, c.rollout.Name)
+			c.pauseContext.AddAbort(msg)
+			abortCondition := conditions.NewRolloutCondition(v1alpha1.RolloutPaused, corev1.ConditionTrue, conditions.TimedOutReason, msg)
+			conditions.SetRolloutCondition(&newStatus, *abortCondition)
```
jessesuen, line 624: Shouldn't this logic go into the latter case statement:

```golang
case !isIndefiniteStep(c.rollout) && conditions.RolloutTimedOut(c.rollout, &newStatus):
```

We do not want to abort if we are at an indefinite pause step.

utils/defaults/defaults.go
```diff
@@ -116,6 +118,13 @@ func GetExperimentProgressDeadlineSecondsOrDefault(e *v1alpha1.Experiment) int32
 	return DefaultProgressDeadlineSeconds
 }
 
+func GetAbortExceedProgressDeadlineOrDefault(rollout *v1alpha1.Rollout) bool {
+	if rollout.Spec.AbortExceedProgressDeadline != nil {
+		return *rollout.Spec.AbortExceedProgressDeadline
+	}
+	return DefaultAbortExceedProgressDeadline
+}
+
```
jessesuen, line 127: This will be unnecessary if we make progressDeadlineAbort a bool and not a bool pointer.

jessesuen (changes requested):

pkg/apis/rollouts/v1alpha1/types.go
huikang: Thanks @jessesuen , will update the PR.

utils/defaults/defaults.go
huikang: Removed

pkg/apis/rollouts/v1alpha1/types.go
huikang: Done

rollout/sync.go
huikang: The logic is moved to the suggested case. Please take another look. Thanks.

rollout/sync.go
```diff
@@ -656,7 +656,15 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
...
 			}
-			condition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionFalse, conditions.TimedOutReason, msg)
+
+			var condition *v1alpha1.RolloutCondition
+			// When ProgressDeadlineAbort is set, abort the update
+			if c.rollout.Spec.ProgressDeadlineAbort {
+				c.pauseContext.AddAbort(msg)
+				condition = conditions.NewRolloutCondition(v1alpha1.RolloutPaused, corev1.ConditionTrue, conditions.TimedOutReason, msg)
+			} else {
+				condition = conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionFalse, conditions.TimedOutReason, msg)
```
jessesuen, line 666: 1. I don't think it's correct to conditionally add the Progressing=False. If we get to line 652 case block, we should **_always_** set Progressing to be false.

2. It doesnt look like we will emit an event about the abort, like we do in line 607. Can we structure the code such that if we abort here, we also emit the K8s event about the abort?

3. The setting of the condition `condition = conditions.NewRolloutCondition(v1alpha1.RolloutPaused, corev1.ConditionTrue, conditions.TimedOutReason, msg)` seems incorrect. Why would we be considered paused?

huikang: Thanks. I was mistaken about the Paused and Progressing state. Please checkout the updated PR.

rollout/sync.go
```diff
@@ -648,6 +648,12 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 			if c.newRS != nil {
 				msg = fmt.Sprintf(conditions.ReplicaSetTimeOutMessage, c.newRS.Name)
 			}
+
+			// When ProgressDeadlineAbort is set, abort the update
+			if c.rollout.Spec.ProgressDeadlineAbort {
+				c.pauseContext.AddAbort(msg)
+				c.recorder.Warnf(c.rollout, record.EventOptions{EventReason: conditions.RolloutAbortedReason}, msg)
```
jessesuen, line 655: I think this will cause the Abort event to get emitted every reconciliation. We need to only have it emit the Abort event when there is a state change. I think this needs to be predicated by if we changed the condition, *and* c.rollout.Spec.ProgressDeadlineAbort, then we emit the event. 

```golang
if conditions.SetRolloutCondition(&newStatus, *condition) && c.rollout.Spec.ProgressDeadlineAbort {
    c.recorder.Warnf(c.rollout, record.EventOptions{EventReason: conditions.RolloutAbortedReason}, msg)
}
```

agill17:
So I have a question, maybe I am misunderstanding the changes in this PR. Would this change allow us to abort an update even if we are using analysisTemplate?
Lets say a new canary starts ( the rollout obj does have analysisTemplate(s) ) — but the new RS never becomes healthy because the probe keeps on failing… Does `ProgressDeadlineAbort: true`  apply in cases like this ( were probe is failing, which eventually hits progressDeadlineSeconds, but analysisTemplates are also part of it ? )

jessesuen:
Yes, using this flag the rollout will now abort if **_either one_** of these conditions become true:
1. analysis template fails
2. progress deadline exceeded

The second condition is turned on or off with `progressDeadlineAbort: true`. It allows a Rollout to abort/rollback to stable even if analysis is not used.

docs/features/specification.md
```diff
@@ -60,6 +60,11 @@ spec:
   # Defaults to 600s
   progressDeadlineSeconds: 600
 
+  # Whether to abort the update when ProgressDeadlineSeconds
+  # is exceeded if analysis or experiment is not used.
+  # Optional and default is false.
+  ProgressDeadlineAbort: false
```
cronik, line 66:
```suggestion
  progressDeadlineAbort: false
```

jessesuen:
@huikang the logic looks good, but could you cover the new code with unit tests?

huikang:
> @huikang the logic looks good, but could you cover the new code with unit tests?

Hi, @jessesuen , sure, will add a new test case.

huikang:
> @huikang the logic looks good, but could you cover the new code with unit tests?

Hi, @jessesuen , the unit tests have been added. Please take another look. Thanks.

jessesuen (approved):

jessesuen:
Great work!

PR #1420, opened 2021-08-14
feat: Adding support for defining max destination weights in Istio VS

agrawroh:
### Change Description

This PR adds the support for defining an optional max weight for the destinations in Istio VirtualService. If the `maxWeight` parameter is defined and has a valid value i.e. `0 < maxWeight <= 100` then it'll be used as an upper bound for the traffic sent to stable service + canary service.

The use cases for defining `maxWeight` includes:
- Having another set of fixed destination(s) which the user doesn't want to be affected during the Argo Rollouts (For example, a legacy service).

---

Signed-off-by: Rohit Agrawal <rohit.agrawal@databricks.com>

**Checklist:**

### DIFF
--- a/docs/getting-started/istio/index.md
+++ b/docs/getting-started/istio/index.md
@@ -33,6 +33,8 @@ spec:
           virtualService:
             # Reference to a VirtualService which the controller updates with canary weights
             name: rollouts-demo-vsvc
+            # Optional if there are only two destinations in your routes or if you want to split 100% traffic between stable and canary services. If specified, this will be used as an upper bound for traffic between canary + stable services.
+            maxWeight: 80
             # Optional if there is a single HTTP route in the VirtualService, otherwise required
             routes:
             - http-primary
@@ -71,11 +73,14 @@ spec:
   - name: http-primary  # Should match spec.strategy.canary.trafficRouting.istio.virtualService.routes
     route:
     - destination:
-        host: rollouts-demo-stable  # Should match spec.strategy.canary.stableService
-      weight: 100
+        host: rollouts-demo-stable  # Should match rollout.spec.strategy.canary.stableService
+      weight: 80
     - destination:
         host: rollouts-demo-canary  # Should match spec.strategy.canary.canaryService
       weight: 0
+    - destination:
+        host: rollouts-demo-legacy
+      weight: 20
   tls:
   - match:
     - port: 3000  # Should match the port number of the route defined in spec.strategy.canary.trafficRouting.istio.virtualService.tlsRoutes
@@ -84,11 +89,14 @@ spec:
       - localhost
     route:
     - destination:
-        host: rollouts-demo-stable  # Should match spec.strategy.canary.stableService
-      weight: 100
+        host: rollouts-demo-stable  # Should match rollout.spec.strategy.canary.stableService
+      weight: 80
     - destination:
         host: rollouts-demo-canary  # Should match spec.strategy.canary.canaryService
       weight: 0
+    - destination:
+        host: rollouts-demo-legacy
+      weight: 20
 ```
 
 Run the following commands to deploy:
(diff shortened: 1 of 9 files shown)


pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -368,6 +368,8 @@ type IstioTrafficRouting struct {
 	VirtualService IstioVirtualService `json:"virtualService" protobuf:"bytes,1,opt,name=virtualService"`
 	// DestinationRule references an Istio DestinationRule to modify to shape traffic
 	DestinationRule *IstioDestinationRule `json:"destinationRule,omitempty" protobuf:"bytes,2,opt,name=destinationRule"`
+	// Max weight that will be split between canary and stable services. If unset, it defaults to 100.
+	MaxWeight int64 `json:"maxWeight,omitempty" protobuf:"bytes,3,opt,name=maxWeight"`
```
huikang, line 372: missing `// +optional`

docs/getting-started/istio/index.md
```diff
@@ -33,6 +33,8 @@ spec:
           virtualService:
             # Reference to a VirtualService which the controller updates with canary weights
             name: rollouts-demo-vsvc
+            # Optional if there are only two destinations in your routes or if you want to split 100% traffic between stable and canary services. If specified, this will be used as an upper bound for traffic between canary + stable services.
+            maxWeight: 80
```
huikang, line 37: Feels like we are overusing the example in the getting-started page. Can we move this advanced use to https://argoproj.github.io/argo-rollouts/features/traffic-management/istio/ and leave the getting-started example straightfoward. Thanks.

huikang (commented):
Please update the specification doc: https://github.com/argoproj/argo-rollouts/blob/master/docs/features/specification.md

docs/getting-started/istio/index.md
agrawroh: Make sense. Will update!

wentwog, line 37: This PR adds maxWeight prop at the level of the virtualService, therefore it will affect all the routes ArgoRollouts manage - which are specified in strategy.canary.trafficRouting.istio.virtualService.routes

While in our use case we'd want only specific routes to have less then 100 weight managed by ArgoRollouts.
Our Virtual Service has such routes
- a route for the traffic originating in Mesh. Here we want e.g. 80% to be split between stable and canary in Mesh, and 20% to go to the legacy system.
- a route for the traffic coming to Mesh Ingress Gateway from the Legacy system. Here we want 100% to be split between stable and canary in Mesh. (we don't want to bounce the traffic back to the legacy)

What do you think? 

(maybe we can move the maxWeight to the level of the strategy.canary.trafficRouting.istio.virtualService.routes. Or remove maxWeight  and configure ArgoRollouts to manage only the first 2 destinations in founds in a route...)

wentwog (changes requested):

docs/getting-started/istio/index.md
agrawroh: There is no straightforward way to do it at the routes level. Currently, "routes" is nothing but an array of string which takes the HTTP route names. For us to do it at the route-level, we need a struct/object similar to "tlsRoutes" where we can add an optional "maxWeight" prop.

One proposal is to add "httpRoute" which will be an array of object similar to "tlsRoutes" and gradually deprecate the use of "routes". I proposed this to @jessesuen as well.

@jessesuen @huikang Any suggestions on what would be the best thing to do here?

dnguy078: Hi @agrawroh, I am interested in coming up with a solution for this as well. I would like to reopen this PR or something similar to support this feature. 

@jessesuen @huikang are there any suggestions for moving forward ?

agrawroh: @dnguy078 Sure, that'd be great. Please feel free to take it over or let me know how I can help.

dnguy078: Thank you, @agrawroh I am trying to solve an issue related to istio subset level in another PR as well. https://github.com/argoproj/argo-rollouts/pull/3126 I was wondering if you could take a quick look :)

bakayolo: @dnguy078 any update? This is a pretty big blocker for argo-rollouts adoption on our end.

Issue #1415, opened 2021-08-12
Canary deployment scaledDown stable pods during HPA scale event 


harikrongali:
## Summary 

When canary deployment is progressing and when HPA scale event happened during the time. Rollout controller scaled down stable pods to zero
```
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  labels:
    app: qb-company-service-appd
    app.kubernetes.io/instance: company-qbcompanyservice-usw2-prd
    env: prd-usw2-eks
  name: qb-company-service-appd-rollout
  namespace: company-qbcompanyservice-usw2-prd
spec:
  progressDeadlineSeconds: 1200
  selector:
    matchLabels:
      app: qb-company-service-appd
  strategy:
    canary:
      canaryService: qb-company-service-appd-desired-service
      maxSurge: 100%
      maxUnavailable: 1
      stableService: qb-company-service-appd-stable-service
  template:
    metadata:
      annotations:
        iam.amazonaws.com/role: k8s-company-qbcompanyservice-usw2-prd
    spec:
      containers:
          image: >-
            docker.artifactory.a.intuit.com/###
          name: app
          ports:
            - containerPort: 8443
              name: service
            - containerPort: 8490
              name: metrics
          readinessProbe:
            failureThreshold: 3
            httpGet:
              path: /health/full
              port: 8443
              scheme: HTTPS
            initialDelaySeconds: 400
            periodSeconds: 5
            successThreshold: 3
            timeoutSeconds: 1
          resources:
            limits:
              cpu: 1
              memory: 4096M
            requests:
              cpu: 1
              memory: 4096M
          volumeMounts:
            - mountPath: /etc/secrets
              name: secrets
    
```

## Diagnostics

1.0.2
```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME>
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

Issue #738, opened 2020-09-26
Scaling events mid-update may not match Deployment behavior


jessesuen:
## Summary 

Rollouts scale differently than a Deployment, when the scale event happens in the middle of an update:

To reproduce:

1. Apply canary rollout with no steps and:

```yaml
spec:
  replicas: 4
  strategy:
    canary:
      maxSurge: 2
      maxUnavailable: 0
```

2. Update rollout with a new image/spec will crashloop backoff (this allows us to halt progress of an update)
3. Rollout will surge up, such that there are 6 replicas (2 new, 4 old)
4. `kubectl scale` the rollout to 8 replicas
5. Rollout will then have 10 replicas (6 new, 4 old). Compared to the deployment which will have 3 new, 7 old.

The reason seems to be related to the fact that the canary weight is calculated as 100%, when we do not have any steps. After step 3, when we are at 6 replicas (2 new, 4 old), and then suddenly want to scale to 8 replicas, the controller calculates it can get to 8 replicas there by solely adding new versions of pods.

Also after step 5, if you try to scale the rollout back down to 4, it doesn't do anything, which is another bug.

NOTE: the scaling logic may need to remain fundamentally different than Deployment since Rollout has a concept of a "stable" ReplicaSet which it decides to balance weights against. For example, when applying a revision 3 manifest in this scenario (after step 5), unlike Deployment, rollouts would be able to balance the canary weights against revision 1 and 3.

## Diagnostics

What version of Argo Rollouts are you running? v0.8.3

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #1429, opened 2021-08-19
fix: canary scaledown with maxsurge

harikrongali:
Signed-off-by: hari rongali <hari_rongali@intuit.com>

fixes https://github.com/argoproj/argo-rollouts/issues/1415

Checklist:

### DIFF
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -211,9 +211,9 @@ spec:
 		ExpectCanaryStablePodCount(6, 4).
 		When().
 		ScaleRollout(4).
-	    WaitForRolloutReplicas(8).
+	    WaitForRolloutReplicas(6).
 		Then().
-		ExpectCanaryStablePodCount(4, 4)
+		ExpectCanaryStablePodCount(4, 2)
 }
 
 // TestReduceWeightAndHonorMaxUnavailable verifies we honor maxUnavailable when decreasing weight or aborting
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -208,6 +208,8 @@ spec:
 		ScaleRollout(8).
 		WaitForRolloutReplicas(10).
 		Then().
+		// NOTE: the numbers below may change in the future.
+		// See: https://github.com/argoproj/argo-rollouts/issues/738
 		ExpectCanaryStablePodCount(6, 4).
 		When().
 		ScaleRollout(4).
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -211,7 +211,7 @@ spec:
 		ExpectCanaryStablePodCount(6, 4).
 		When().
 		ScaleRollout(4).
-	    WaitForRolloutReplicas(4)
+	    WaitForRolloutReplicas(8)
 }
 
 // TestReduceWeightAndHonorMaxUnavailable verifies we honor maxUnavailable when decreasing weight or aborting
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -149,19 +149,19 @@ spec:
       containers:
       - name: updatescaling
         command: [/bad-command]`).
-		WaitForRolloutReplicas(6).
+		WaitForRolloutReplicas(7).
 		Then().
-		ExpectCanaryStablePodCount(2, 4).
+		ExpectCanaryStablePodCount(4, 3).
 		When().
 		ScaleRollout(8).
-		WaitForRolloutReplicas(10).
+		WaitForRolloutReplicas(11).
 		Then().
-		ExpectCanaryStablePodCount(6, 4).
+		ExpectCanaryStablePodCount(8, 3).
 		When().
 		ScaleRollout(4).
-		WaitForRolloutReplicas(4).
+		WaitForRolloutReplicas(7).
 		Then().
-		ExpectCanaryStablePodCount(6, 4)
+		ExpectCanaryStablePodCount(4, 3)
 }
 
 // TestRolloutScalingDuringUpdate verifies behavior when scaling a rollout up/down in middle of update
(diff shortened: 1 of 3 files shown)


harikrongali:
@alexmt / @jessesuen  Please review.

test/e2e/canary_test.go
```diff
@@ -156,12 +208,12 @@ spec:
...
-		// NOTE: the numbers below may change in the future.
-		// See: https://github.com/argoproj/argo-rollouts/issues/738
 		ExpectCanaryStablePodCount(6, 4).
 		When().
-		ScaleRollout(4)
-	// WaitForRolloutReplicas(4) // this doesn't work yet (bug)
+		ScaleRollout(4).
+	    WaitForRolloutReplicas(8).
+		Then().
+		ExpectCanaryStablePodCount(4, 4)
```
jessesuen, line 216: I don't understand the fix. Why after scaling down the spec.replicas, do we still have 8 pods?

harikrongali: fixed the issue. Thanks for the comment

harikrongali:
@jessesuen Please review

jessesuen (changes requested):
CalculateReplicaCountsForCanary is arguably the most critical part of rollout code and the new code is not covered in unit tests. I will insist that we test this with additional unit tests.

test/e2e/canary_test.go
```diff
@@ -112,6 +112,58 @@ func (s *CanarySuite) TestRolloutScalingWhenPaused() {
...
+		HealthyRollout(`
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
+  name: updatescaling
+spec:
+  replicas: 4
+  strategy:
+    canary:
+      maxSurge: 100
```
jessesuen, line 128: I think you mean for this to be "100%" and not `100`. The latter is an absolute number.

utils/replicaset/canary.go
```diff
@@ -180,39 +180,44 @@ func CalculateReplicaCountsForCanary(rollout *v1alpha1.Rollout, newRS *appsv1.Re
...
+		if allowedAvailableReplicaCount < newRSReplicaCount {
+			// ScaleDown canaryAllowedSet to allowed value to adhere to maxSurge
+			newRSReplicaCount = allowedAvailableReplicaCount
+		}
+	}
+	return newRSReplicaCount, stableRSReplicaCount
+}
 
-	if newRS != nil && *newRS.Spec.Replicas > desiredNewRSReplicaCount {
+func calculateScaleDownReplicaCount(replicaSet *appsv1.ReplicaSet, desireRSReplicaCount int32, maxReplicaCountAllowed int32, scaleDownCount int32, updatedReplicaCount int32) (int32, int32) {
```
jessesuen, line 209: Could you provide a function description for this, and what it returns?

test/e2e/canary_test.go
```diff
@@ -160,8 +212,10 @@ spec:
 		// See: https://github.com/argoproj/argo-rollouts/issues/738
 		ExpectCanaryStablePodCount(6, 4).
 		When().
-		ScaleRollout(4)
-	// WaitForRolloutReplicas(4) // this doesn't work yet (bug)
+		ScaleRollout(4).
+	    WaitForRolloutReplicas(6).
+		Then().
+		ExpectCanaryStablePodCount(4, 2)
```
jessesuen, line 218: I still think this is not correct. According to the sequence of events:

1. At T-0, we were at: 6 pods (2 canary, 4 stable, 4 available)
2. At T-1, we had a scale-up event and went to: 10 pods (6 canary, 4 stable, 4 available).
3. At T-2, we had a scale-down event and went back down to 6 pods (4 canary, 2 stable, 2 available)

Notice that at T-2, we have now violated maxUnavailable because we only have 2 pods available when the spec (maxUnavailable: 0) requires 4 to be available. 

Although this is a day-0 issue, I think the proper fix may be to address https://github.com/argoproj/argo-rollouts/issues/738. If we fix https://github.com/argoproj/argo-rollouts/issues/738, then this problem may also be fixed.

test/e2e/canary_test.go
harikrongali: yes..meant 100%

test/e2e/canary_test.go
harikrongali: Yes, you are correct that it requires changes to adhere to both maxSurge and minAvailabile. But bug referred in #738 is to fix weights inline to deployment.  The current I am trying to fix is not scaleDown completely on either stable or canary which I will address as a part of this PR and will take #738 to do refactor to fix the whole calculation.

harikrongali: Can you label #738 to 1.2?

jessesuen: > But bug referred in #738 is to fix weights inline to deployment

The two are closely related, if not the same issue. The behavior you saw in #1415 is a manifestation of the same underlying problem but with different parameters of maxSurge/minAvailable. 

This fix appears to addressing the part of symptom, whereas #738 is actually the root problem. I believe if we fix #738, then the incident in #1415 would not have happened.

I do recognize that this fix improves the current behavior, but I'm pretty sure the code in this PR will need to be thrown away in order to address #738

harikrongali: Agree, #738 would. be the fix we need to act on. My concern currently is current fix will stop the downtime in case of bad image versions. 
I wanted to take #738 for 1.2 to refactor the calculation and I don't want to refactor at the current stage given the complexity.

jessesuen (approved):
@harikrongali great work! 

So with this fix, we will now fix the bug where scale down could violate maxUnavailable. But we still have issue that scale up could violate maxSurge (https://github.com/argoproj/argo-rollouts/issues/738), correct?

harikrongali:
@jessesuen yes. issue #738 still exists where scale-up could violate.

PR #1430, opened 2021-08-19
feat: support dynamic scaling of stable ReplicaSet as inverse of canary weight

jessesuen:
Fixes https://github.com/argoproj/argo-rollouts/issues/1029

1. Introduces a `canary.dynamicStableScale` flag which allows the stable ReplicaSet to dynamically scale down/up depending on canary traffic weight. This feature is needed to run less pods during an update, but at the cost of slower aborts (since the stable will need to scale back up).

2. The feature introduces a new `weights` field into the rollout status `status.canary.weights`. e.g.:

```yaml
status:
  canary:
    weights:
      canary:
        podTemplateHash: 68978bfc96
        serviceName: dynamic-stable-scale-canary
        weight: 0
      stable:
        podTemplateHash: 5cc54cc5f4
        serviceName: dynamic-stable-scale-stable
        weight: 100
```

The tracking of canary weights into status is needed for us to understand the weight which was set in the traffic router. This allows us to calculate the amount of replicas which we can **_safely_** reduce the scale the stable (or canary during abort). For example, it is *not* safe to scale down the stable, until we know traffic has shifted away from it. It also avoids us to have all traffic routers implement a GetCurrentWeight() function because we instead remember what we set last.

We could also use this information to show in the Rollout dashboard in the future, or the CLI. Currently the actual weight field in the CLI is not very accurate.

3. During an abort, we scale stable to 100% immediately. As the stable pods become available, we start shifting traffic to the stable ReplicaSet (whatever is available), and start scaling down the canary. Users may opt-out of scaling down the canary upon abort by explicitly setting a value for `canary,abortScaleDownDelaySeconds`.


harikrongali:
@perenesenko can you please review and validate

perenesenko (commented):
I tested different scenarios of promotion and abortion with Istio and Nginx. All works fine
Just have two questions...

perenesenko:
Q1 what about using the `setCanaryScale`. Previously we discussed with @harikrongali Probably we should prevent to use of `dynamicStableScale` and `setCanaryScale` together.

In case of steps
```
      steps:
        - setCanaryScale:
            weight: 20
        - setWeight: 80
        - pause: {}
```
After the `setCanaryScale` the canary no longer be autoscaling by the dynamicStableScale flag.
Also we will have a situation when 20% of canary pods serving 80% of traffic

perenesenko:
Q2 The `scaleDownDelaySeconds` is not working with the `dynamicStableScaling` as once we scale canary to 100 the stable will have 0 pods. So the `scaleDownDelaySeconds` will be ignored. Should we mention this somewhere?

harikrongali:
For Q1, it is confirmed that https://github.com/argoproj/argo-rollouts/pull/1382#issuecomment-900547091 setCanaryScale and dynamicStableScale will work together. I would like to see what is the expected behavior if both values set. @jessesuen  can you please provide us the behavior when both are configured. @perenesenko when you validated, 20%of pods are taking 80% of traffic? and can you provide what is the stable replica set count at that time?

perenesenko:
@harikrongali
Yes, my configuration has 5 pods. During update the canary was set to 20% which is 1 pod and after sending 80 to canary it still has 1 pod.

perenesenko (approved):

jessesuen:
> Q1 what about using the `setCanaryScale`. Previously we discussed with @harikrongali Probably we should prevent to use of `dynamicStableScale` and `setCanaryScale` together.
> 
> In case of steps
> 
> ```
>       steps:
>         - setCanaryScale:
>             weight: 20
>         - setWeight: 80
>         - pause: {}
> ```
> 
> After the `setCanaryScale` the canary no longer be autoscaling by the dynamicStableScale flag.
> Also we will have a situation when 20% of canary pods serving 80% of traffic

Yes, this is a risk with the `setCanaryScale` feature. But that is a risk unrelated to the dynamicStableScale. I'm adding some documentation to highlight the concern.

> I would like to see what is the expected behavior if both values set. @jessesuen can you please provide us the behavior when both are configured.

They independent choices and do not affect each other. Both can be configured at the same time. dynamicStableScaling can be simply understood as: "whatever the current canary **traffic weight** is, the stable will be scaled to be:
`inverse canary weight * rollout.spec.replicas`

> Q2 The scaleDownDelaySeconds is not working with the dynamicStableScaling as once we scale canary to 100 the stable will have 0 pods. So the scaleDownDelaySeconds will be ignored. Should we mention this somewhere?

correct. scaleDownDelaySeconds is not expected to work with dynamicStableScaling and is ignored.

jessesuen:
> correct. scaleDownDelaySeconds is not expected to work with dynamicStableScaling and is ignored.

I added validation to make sure these two options are not used together.

docs/features/canary.md
```diff
@@ -109,11 +109,56 @@ spec:
...
+When using traffic routing, by default the stable ReplicaSet is left scaled to 100% during the update.
+This has the advantage that if an abort occurs, traffic can be immediately shifted back to the
+stable ReplicaSet without delay. However, it has the disadvantage that during the update, there will
+eventually be double the number of replica pods running (similar to in blue-green deployment) since
+the stable ReplicaSet is left scaled up for the full duration of the update.
+
+It is possible to reduce the scale of the stable ReplicaSet during an update such that it scales down
+as the traffic weight increases to canary. This would be desirable in scenarios where the Rollout has
+a high replica count and resource cost is a concern, or in bare-metal situations where it is not
+possible to create additional node capacity to accommodate double t
```
bpoland, line 143: Looks like this text got cut off :)

jessesuen: Oops thanks for catching that!

alexmt (approved):
LGTM

PR #1465, opened 2021-09-02
fix: Revert canary calculations

jessesuen:
Recent changes to canary calculation (https://github.com/argoproj/argo-rollouts/pull/1449 and https://github.com/argoproj/argo-rollouts/pull/1429) have caused Rollout update to become unstable.

This change reverts those until we have a proper fix.


jessesuen:
Actually even after reverting I am able to reproduce the flapping problem. So the bug may have existed for some time

PR #1479, opened 2021-09-08
fix: replicaset count would flap when interrupting update with new pod spec

jessesuen:
This fixes an issue (that exists only in master) where a ReplicaSet count would flap from 0 to N when an update was interrupted by an apply of a third pod spec. The scenario is:

1. Traffic routed canary Rollout is in middle of updating from V1 -> V2
2. V3 version of Rollout was applied
3. The ReplicaSet for V2 would flap from 0 to N on different reconciliations

The fix is that we should choose to leave V2 up and running until we believe it is safe to scale it down (because V1 and V3 have reached full readiness).


alexmt (approved):
LGTM

PR #1485, opened 2021-09-10
fix: target group verification did not work with named ports

jessesuen:
The v1.1 AWS target group verification feature was not working when the service was using named instead of numeric ports.


alexmt (approved):
Thank you @jessesuen !

PR #1515, opened 2021-09-15
feat: export currentPodHash in rollout_info. Fixes #1512

### LINKED ISSUE #1512: Export rollout.Status.CurrentPodHash in rollout_info metric
# Summary

Currently by using metric `rollout_info` we have no way to check which version of the rollout caused a specific event.
The metric `rollout_info` exposes information about the rollout namespace and name, but we don't have any dimension which refers to the version of the rollout/fleet where a specific state happened.
To export this dimension as part of `rollout_info`, we could expose the `rollout.Status.CurrentPodHash` since it refers to the hash of the `ReplicaSet` in change.

# Use Cases

Having the ability to identify which version caused a specific event would be beneficial to target releases responsible for a specific event (like a failed or aborted deployment) as well as to generate better reporting about the events.

---

jddcarreira:
## Summary

Currently by using the metric `rollout_info` we have no way to check which version of the rollout caused a specific event.
The metric `rollout_info` exposes information about the rollout namespace and name, but we don't have any dimension which refers to the version of the rollout/fleet where a specific state happened.
Having the ability to identify which version caused a specific event would be beneficial to target releases responsible for a specific event (like a failed or aborted deployment) as well as to generate better reporting about the events.

## Changes
This PR adds a new dimension, `current_pod_hash`, to the metric rollout_info which refers to the value of `rollout.Status.CurrentPodHash`.

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/1512) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.


huikang:
LGTM

jessesuen (changes requested):
Unfortunately, `current_pod_hash` is a bad candidate for a prometheus label. It's an unbounded value that is pretty much unique to *every* rollout in the cluster. Meaning there will essentially be a time series for every single rollout. This will result in an explosion the cardinality (memory requirements) of the prometheus database monitoring the controller.

See docs on labels:
> https://prometheus.io/docs/practices/naming/
> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

jddcarreira:
Hi @jessesuen 

Yes, that can happen since this is unique.
Since displaying this kind of detailed metrics can be abusive for some systems but useful for other users, do a think we could have an argument or config option, for example `detailed-metrics=true|false`, that would expose this kind of dimensions depending of the needs of the user?

Thanks

PR #1246, opened 2021-06-02
feat(controller): HA Leader election support on rollouts-controller

caoyang001:
Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not. 
* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

Closes [#727](https://github.com/argoproj/argo-rollouts/issues/727)

PR #1519, opened 2021-09-16
feat:update argo rollouts clusterrole for leader election

huikang:
Signed-off-by: caoyang001 <caoyang001@foxmail.com>

Duplicate from https://github.com/argoproj/argo-rollouts/pull/1246 with conflict resolved.

Checklist:

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -1,12 +1,15 @@
 package controller
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"time"
 
 	notificationapi "github.com/argoproj/notifications-engine/pkg/api"
 	notificationcontroller "github.com/argoproj/notifications-engine/pkg/controller"
+
 	"github.com/pkg/errors"
 	smiclientset "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/split/clientset/versioned"
 	log "github.com/sirupsen/logrus"
@@ -14,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
@@ -24,6 +28,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/util/workqueue"
 
 	"github.com/argoproj/argo-rollouts/analysis"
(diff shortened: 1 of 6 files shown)


controller/controller.go
```diff
@@ -62,8 +68,41 @@ const (
 
 	// DefaultIngressThreads is the default number of ingress worker threads to start with the controller
 	DefaultIngressThreads = 10
+
+	// DefaultLeaderElect is the default true leader election should be enabled
+	DefaultLeaderElect = true
```
harikrongali, line 73: to keep the backward compatibility, default should be false

controller/controller.go
```diff
@@ -321,3 +399,15 @@ func (c *Manager) Run(rolloutThreadiness, serviceThreadiness, ingressThreadiness
...
+
+func (c *Manager) startLeading(ctx context.Context, rolloutThreadiness, serviceThreadiness, ingressThreadiness, experimentThreadiness, analysisThreadiness int) {
+	defer runtime.HandleCrash()
+	// Start the informer factories to begin populating the informer caches
+	log.Info("Starting Controllers")
+	go wait.Until(func() { c.rolloutController.Run(rolloutThreadiness, ctx.Done()) }, time.Second, ctx.Done())
+	go wait.Until(func() { c.serviceController.Run(serviceThreadiness, ctx.Done()) }, time.Second, ctx.Done())
+	go wait.Until(func() { c.ingressController.Run(ingressThreadiness, ctx.Done()) }, time.Second, ctx.Done())
+	go wait.Until(func() { c.experimentController.Run(experimentThreadiness, ctx.Done()) }, time.Second, ctx.Done())
+	go wait.Until(func() { c.analysisController.Run(analysisThreadiness, ctx.Done()) }, time.Second, ctx.Done())
```
harikrongali, line 411: notification controller is missing here

controller/controller.go
```diff
@@ -62,8 +68,41 @@ const (
...
 	DefaultIngressThreads = 10
+
+	// DefaultLeaderElect is the default true leader election should be enabled
+	DefaultLeaderElect = true
+
+	// DefaultLeaderElectionNamespace is the default namespace used to perform leader election. Only used if leader election is enabled
+	DefaultLeaderElectionNamespace = "kube-system"
+
+	// DefaultLeaderElectionLeaseDuration is the default time in seconds that non-leader candidates will wait to force acquire leadership
+	DefaultLeaderElectionLeaseDuration = 15 * time.Second
```
harikrongali, line 79: can you provide proposal/discussion on how this works?

controller/controller.go
huikang: Fixed. Thanks for catching this!

controller/controller.go
huikang: Hi, @harikrongali , thanks for your review. After second thought, I think it is safe to enable leader elect by default. The reasons are
1. `Defaulting to true` is backward compatible. This is because when there is 1 instance, this node will be elected as the master node. So there is no impact on the existing deployment manifest.
2. `Defaulting to true` makes it easier to enable HA because the only change to the install manifest is to increase the replica from 1 to > 1. Otherwise, user has to add `--leader-elect` to the CMD of the manifest.

Does it make sense to you? Thanks.

controller/controller.go
huikang: I added some [description](https://github.com/argoproj/argo-rollouts/pull/1519/files#diff-0bb42a85d04c8b7c9b659307979fd43e8f6e084283a300bfff959adfcbddfd17) to the PR about how the implement work . Please let me know if more detail is needed. Thanks.

controller/controller.go
blkperl: argo-workflows defaults to LeaderElection enabled as well. 

https://github.com/argoproj/argo-workflows/blob/master/docs/environment-variables.md

Looking forward to this getting merged!  Thanks for writing the PR @huikang

docs/FAQ.md
```diff
@@ -37,6 +37,10 @@ Argo Rollouts is a Kubernetes controller that will react to any manifest change
 by a Git commit, an API call, another controller or even a manual `kubectl` command. You can use Argo Rollouts with any traditional CI/CD
 solution that does not follow the GitOps approach.
 
+### Can we run the Argo Rollouts controller in HA mode?
+
+Yes. A k8s cluster can run multiple replicas of Argo-rollouts controllers to achieve HA. To enable this feature, run the controller with `--leader-elect` flag and increase the number of replicas in the controller's deployment manifest. The implementation is based on the [k8s client-go's leaderelection package](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#section-documentation). This implemetnatioin is tolerant to *arbitrary clock skew* among replicas. The level of tolerance to skew rate can be configured by setting `--leader-election-lease-duration` and `--leader-election-renew-deadline` appropriately. Please refer to the [package documentation](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#pkg-overview) for details.
```
blkperl, line 42:
```suggestion
Yes. A k8s cluster can run multiple replicas of Argo-rollouts controllers to achieve HA. To enable this feature, run the controller with `--leader-elect` flag and increase the number of replicas in the controller's deployment manifest. The implementation is based on the [k8s client-go's leaderelection package](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#section-documentation). This implementation is tolerant to *arbitrary clock skew* among replicas. The level of tolerance to skew rate can be configured by setting `--leader-election-lease-duration` and `--leader-election-renew-deadline` appropriately. Please refer to the [package documentation](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#pkg-overview) for details.
```

huikang: @blkperl, thanks for your review. Your suggestion has been added to the PR.

fblgit:
nice feature, well done. looks very aligned to sigs controller-runtime but despite that, existing e2e's will validate a solo scenario but won't validate an HA :(

Just an idea, but you may wanna try to manipulate the election resource, which may allow you to run at once a test that validates existing leader and the taking over leader case.

harikrongali (approved):

harikrongali:
@alexmt @jessesuen this PR is ready to merge..if you can take final review & approve, that would be great.

manifests/base/argo-rollouts-deployment.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     app.kubernetes.io/name: argo-rollouts
     app.kubernetes.io/part-of: argo-rollouts
 spec:
+  replicas: 1
```
alexmt, line 10: Should we change strategy to rolling update as well?

alexmt (changes requested):

Issue #1272, opened 2021-06-11
Re-deploying the manifests after v1.0.0-rc1 is marking the crds as configured instead of unchanged 


rmorampudi:
## Summary 

What happened/what you expected to happen?

When we re-deploy the manifests for rollouts, for version later to v1.0.0-rc1, the manifests are expected to be marked as unchanged, but it's marked as configured instead, for the crds

```
1st time deployment
customresourcedefinition.apiextensions.k8s.io/analysisruns.argoproj.io created
customresourcedefinition.apiextensions.k8s.io/analysistemplates.argoproj.io created
customresourcedefinition.apiextensions.k8s.io/clusteranalysistemplates.argoproj.io created
customresourcedefinition.apiextensions.k8s.io/experiments.argoproj.io created
customresourcedefinition.apiextensions.k8s.io/rollouts.argoproj.io created
serviceaccount/argo-rollouts created
clusterrole.rbac.authorization.k8s.io/argo-rollouts-aggregate-to-admin created
clusterrole.rbac.authorization.k8s.io/argo-rollouts-aggregate-to-edit created
clusterrole.rbac.authorization.k8s.io/argo-rollouts-aggregate-to-view created
clusterrole.rbac.authorization.k8s.io/argo-rollouts-clusterrole created
clusterrolebinding.rbac.authorization.k8s.io/argo-rollouts-clusterrolebinding created
service/argo-rollouts-metrics created
deployment.apps/argo-rollouts created

2nd time deployment 

customresourcedefinition.apiextensions.k8s.io/analysisruns.argoproj.io configured
customresourcedefinition.apiextensions.k8s.io/analysistemplates.argoproj.io configured
customresourcedefinition.apiextensions.k8s.io/clusteranalysistemplates.argoproj.io configured
customresourcedefinition.apiextensions.k8s.io/experiments.argoproj.io configured
customresourcedefinition.apiextensions.k8s.io/rollouts.argoproj.io configured
serviceaccount/argo-rollouts unchanged
clusterrole.rbac.authorization.k8s.io/argo-rollouts-aggregate-to-admin unchanged
clusterrole.rbac.authorization.k8s.io/argo-rollouts-aggregate-to-edit unchanged
clusterrole.rbac.authorization.k8s.io/argo-rollouts-aggregate-to-view unchanged
clusterrole.rbac.authorization.k8s.io/argo-rollouts-clusterrole unchanged
[...]

PR #1524, opened 2021-09-20
fix(manifests): Removed preserveUnknownFields from the CRD(s) which is disallowe…

flaviolemos78:
Removed preserveUnknownFields from the CRD(s) which is disallowed on the v1 API (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122)

Checklist:


jessesuen:
This was actually intentionally left because we need to provide an upgrade path from older rollout versions. It's still premature to remove this field because people upgrading will not be able to. There is an issue for tracking when we can remove this:

https://github.com/argoproj/argo-rollouts/issues/1272#issuecomment-859734976

PR #1551, opened 2021-09-28
fix(datadog): allow empty datadog result. Fixes #1548

### LINKED ISSUE #1548: DataDog provider should allow dealing with no data points returned
## Summary 

When there are no data points in a DataDog query, we automatically return error: `Datadog returned no value` with no ability to change this behavior. However, some users may want to consider this scenario a success. The code which prevents users from deciding how to handle this is here:

https://github.com/argoproj/argo-rollouts/blob/f5e2d82cb3be3d0e7f51b35f81c8f34c0bbf929e/metricproviders/datadog/datadog.go#L142-L150

```golang
	if len(res.Series) < 1 || len(res.Series[0].Pointlist) < 1 {
		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Datadog returned no value: %s", string(bodyBytes))
	}

	series := res.Series[0]
	datapoint := series.Pointlist[len(series.Pointlist)-1]
	if len(datapoint) < 1 {
		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Datadog returned no value: %s", string(bodyBytes))
	}
```

The above logic does not allow a query to return with no datapoints because we error if the array length is 0.

One possible solution, is to delete this code and allow the result to return as `nil`. Then, users could decide how to deal with this by writing an expression like:

```yaml
successCondition: result == nil || result <= 0.05
```

A caveat is that I forget if expr library allows for polymorphic values where result may be a float or nil and compared using built-in operators. But if it doesn't, we could always add a convenience function, `IsNil()` to allow users to deal with it.

```yaml
successCondition: IsNil(result) || result <= 0.05
```

## Diagnostics

What version of Argo Rollouts are you running? v1.1.0-rc1

---

rms1000watt:
Checklist:


metricproviders/datadog/datadog.go
```diff
@@ -142,18 +142,24 @@ func (p *Provider) parseResponse(metric v1alpha1.Metric, response *http.Response
...
+		}
 	}
 
-	series := res.Series[0]
-	datapoint := series.Pointlist[len(series.Pointlist)-1]
-	if len(datapoint) < 1 {
-		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Datadog returned no value: %s", string(bodyBytes))
+	status, err := evaluate.EvaluateResult(value, metric, p.logCtx)
+	if value == nil {
+		return "0", status, err
```
rms1000watt, line 159: this is a magic number, ie. why `0` picked arbitrarily?

if there's a consistent convention we can tap into.. like.. "Argo Rollout queries default to 0 for metric interpolation," then this will be OK.

metricproviders/datadog/datadog_test.go
```diff
@@ -25,6 +25,19 @@ func TestRunSuite(t *testing.T) {
 
 	unixNow = func() int64 { return 1599076435 }
 
+	ddProviderIntervalDefault := v1alpha1.MetricProvider{
```
rms1000watt, line 28: sonarcloud failed with 40.5% duplication in this file, so I only did this part to tidy it all up

metricproviders/datadog/datadog.go
jessesuen: The value returned here is what we claim was returned by the provider in json form. It's actually not used for evaluation, only to help users debug why their metric failed. I think it would be better to either:
* return `null` 
* return `res.Series` array serialized to json

This is better than a value 0, which will mislead the user in that the provider actually returned a response.

rms1000watt: So, there are 2 parts that actually return `0`.

1. The spot you caught, here in `datadog.go` that doesn't actually get used for calculation
2. Inside `utils/evaluate/evaluate.go`.
Screen Shot 2021-09-29 at 2 45 11 PM

So the question is, if a nil pointer is provided (if an empty `result` is passed into `EvaluateResult`) what should happen?

The painful part is the datatype has to match whatever is provided in the expression. For example

```yaml
successCondition: isNil(result) || result < 0.05
```

The datatype of `result` in `result < 0.05` has to be `float`.

rms1000watt: I can create another function like.. 

```yaml
successCondition: compareWithNil(result, <, 0.05, true)
```

That allows `result` and `0.05` to be different datatypes (`*float` and `float`). If `result` is nil, fallback to the 4th argument to the default return value.

rms1000watt: Or, instead, a function like..

```yaml
successCondition: expressionFallback("result < 0.05", true)
```

That checks if `result` is nil. If nil, return the second argument. If not nil, run the expr library again.. (calls the expr lib twice in this case). I think this is heavier from a code/computation side, but maybe nicer from a usability perpsective

rms1000watt: Or, maybe...

```yaml
successCondition: default(result, 0) < 0.05
```

If result is nil, then use the default value for the calculation. This is probably the best approach honestly...

rms1000watt: I'm resolving this conversation with the latest push which includes `default()` functionality in the unit tests and the docs.

jessesuen:
Great idea to introduce default()! This will come in handy for other providers

utils/evaluate/evaluate.go
```diff
@@ -185,3 +187,49 @@ func Equal(a, b []string) bool {
...
+	case reflect.Ptr, reflect.Map, reflect.Array, reflect.Chan, reflect.Slice:
+		out = reflect.ValueOf(in).IsNil()
+	}
+
+	return
+}
+
+// valueFromPointer allows pointers to be passed in from the provider, but then extracts the value from
+// the pointer if the pointer is not nil, else returns nil
+func valueFromPointer(in interface{}) (out interface{}) {
```
rms1000watt, line 223: @jessesuen there's a design implementation question here.

Right now, it's coded as `valueFromPointer()` which allows `resultValue` to be:
- a value
- a pointer
- a nil pointer

Although it's already coded, the downsides are:
- The use of the `reflect` package
  - In general it should be used sparingly--only if necessary
  - Although all unit tests for all providers succeed, there might be other data types, functions, etc. that could break the logic. I'm decently comfortable with the `reflect` package logic here, but I'm not a wizard with the `reflect` package--there could be pitfalls I'm missing.
  
Alternatively, I can remove `valueFromPointer` and create `valueOrNil` function--which would reduce the overall functionality and only allow `resultValue` to be one of two datatypes:
- a value
- a nil pointer

The downside with this approach, is the providers won't be allowed to pass in `a pointer` where the value is parsed and evaluated in the condition. If `a pointer` was provided, the evaluation would error because of a datatype mismatch. This would require a developer to redo this work I have here if `a pointer` was required to be passed in as `resultValue` from the provider.

TBH, my vote is to leave it as is. **But**, I wanted to bring this up to everyone so its documented and everyone is made aware.

rms1000watt:
My bad, lagging on getting all the tests green

utils/evaluate/evaluate.go
jessesuen: This seems fine. In the end, I don't think this matters much because responses from providers will likely not have pointer representations of their payload responses. So I think this is an implementation detail that doesn't otherwise affect the overall approach. We can stick with what you have.

rms1000watt: thx, resolving

PR #1578, opened 2021-10-13
feat: add healthz probe port and update the install.yaml

### LINKED ISSUE #1522: Add a health probe endpoint to the controller
# Summary

What change needs making?

Add a health checkpoint like `:8081/healthz` to the controller. Currently we rely on the metrics server for health probe.

https://github.com/argoproj/argo-rollouts/blob/01879e474d7973ced4b963fc19276a594e93bdfc/manifests/base/argo-rollouts-deployment.yaml#L26-L38

# Use Cases

When would you use this?

When HA is enabled, the non-lead instance should not start the metrics server due to incomplete data; only the lead node can report metrics. Therefore, we need to distinguish between metrics listen port and the health probe port.

---

huikang:
- distinguish between exposed healthz port and metrics port
   - liveness probe uses healthz port
      readiness probe uses metrics port
      In HA mode, only the primary instance is ready to accept request
      and the metrics queries
    
close #1522 

Checklist:

### DIFF
--- a/analysis/controller_test.go
+++ b/analysis/controller_test.go
@@ -94,7 +94,7 @@ func (f *fixture) newController(resync resyncFunc) (*Controller, informers.Share
 	metricsServer := metrics.NewMetricsServer(metrics.ServerConfig{
 		Addr:               "localhost:8080",
 		K8SRequestProvider: &metrics.K8sRequestsCountProvider{},
-	})
+	}, true)
 
 	c := NewController(ControllerConfig{
 		KubeClientSet:        f.kubeclient,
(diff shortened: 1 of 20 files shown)


huikang:
@harikrongali @jessesuen , per our discussion during the contributor meeting, the new commit made the secondary controller expose empty metrics.

In addition, I still keep the healthz endpoint as the liveness check in the deployment manifest to reduce the load on liveness check. Please take a look when you have a chance. Thanks.

alexmt (approved):

PR #1627, opened 2021-11-06
feat(analysis): Add Dry-Run Mode

agrawroh:
# Background

We are now using Argo Rollouts in production and have a few additional requirements to extend it to all our backend services. One of the top ask/requirements in the checklist is to have an ability to evaluate new analyses for their real behavior, prior to actually rolling back based on those analyses. One possible solution would be to add a `Dry-Run` mode to the **AnalysisTemplate** so that after analyzing the metrics which are marked as `Dry-Run` the controller does not result in the rollback/failures.

# Use Cases

When we add a new M3/PROM metric to the **AnalysisTemplate**, users . Adding a `Dry-Run` mode can help us identify the issues (if any) in our metrics and resolve them before we start evaluating them to make the actual rollout decisions. 

For those metrics which are marked to be in the `Dry-Run` mode, we need a mechanism to understand/answer the question: _”How would Argo have evaluated this query?”_, but explicitly take no action.

# Changes

In this PR, we are adding a Dry-Run mode at the per-metric level.

`dryRun` can be used on a metric to control whether or not to evaluate that metric in a dry-run mode. A metric running 
in the dry-run mode won't impact the final state of the rollout or experiment even if it fails or the evaluation comes 
out as inconclusive.

The following example queries prometheus every 5 minutes to get the total number of **4XX** and **5XX** errors, and even if the
evaluation of the metric to monitor the 5XX error-rate fail, the analysis run will pass.

```yaml hl_lines="1 2"
  dryRun:
  - metricName: total-5xx-errors
  metrics:
  - name: total-5xx-errors
    interval: 5m
    failureCondition: result[0] >= 10
    failureLimit: 3
    provider:
      prometheus:
        address: http://prometheus.example.com:9090
        query: |
          sum(irate(
            istio_requests_total{reporter="source",destination_service=~"{{args.service-name}}",response_code~"5.*"}[5m]
          ))
  - name: total-4xx-errors
    interval: 5m
    failureCondition: result[0] >= 10
    failureLimit: 3
    provider:
      prometheus:
        address: http://prometheus.example.com:9090
        query: |
          sum(irate(
            istio_requests_total{reporter="source",destination_service=~"{{args.service-name}}",response_code~"4.*"}[5m]
          ))
```

A wildcard '*' can be used to make all the metrics run in the dry-run mode. In the following example, even if one or 
both metrics fail, the analysis run will pass.

```yaml hl_lines="1 2"
  dryRun:
  - metricName: .*
  metrics:
  - name: total-5xx-errors
    interval: 5m
    failureCondition: result[0] >= 10
    failureLimit: 3
    provider:
      prometheus:
        address: http://prometheus.example.com:9090
        query: |
          sum(irate(
            istio_requests_total{reporter="source",destination_service=~"{{args.service-name}}",response_code~"5.*"}[5m]
          ))
  - name: total-4xx-errors
    interval: 5m
    failureCondition: result[0] >= 10
    failureLimit: 3
    provider:
      prometheus:
        address: http://prometheus.example.com:9090
        query: |
          sum(irate(
            istio_requests_total{reporter="source",destination_service=~"{{args.service-name}}",response_code~"4.*"}[5m]
          ))
```

### Dry-Run Summary

If one or more metrics are running in the dry-run mode, the summary of the dry-run results gets appended to the analysis 
run message. Assuming that the `total-4xx-errors` metric fails in the above example but, the `total-5xx-errors` 
succeeds, the final dry-run summary will look like this.

```yaml hl_lines="2 3 4 5 6 7"
Message: Run Terminated
Dry Run Summary: 
  Count: 4
  Successful: 1
  Failed: 1
  Inconclusive: 1
  Error: 1
Dry Run Summary: 
  Count: 2
  Successful: 1
  Failed: 1
  Inconclusive: 0
  Error: 0
Metric Results:
...
```

### Dry-Run Rollouts

If a rollout wants to dry run its analysis, it simply needs to specify the `dryRun` field to its `analysis` stanza. If a
rollout wants to dry run its analysis, it simply needs to specify the `dryRun` field to its `analysis` stanza. In the 
following example, all the metrics from `random-fail` and `always-pass` get merged and executed in the dry-run mode.

```yaml hl_lines="9 10"
kind: Rollout
spec:
...
  steps:
  - analysis:
      templates:
      - templateName: random-fail
      - templateName: always-pass
      dryRun:
      - metricName: .*
```

# Metrics

The `analysis_run_metric_phase` metrics will have an additional `"dry_run"` label to indicate whether or not the metric is running in a Dry-Run mode.

# Testing

Screen Shot 2021-11-07 at 8 18 54 PM

# Checklist

---


pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -106,6 +106,8 @@ type Metric struct {
 	ConsecutiveErrorLimit *intstrutil.IntOrString `json:"consecutiveErrorLimit,omitempty" protobuf:"bytes,9,opt,name=consecutiveErrorLimit"`
 	// Provider configuration to the external system to use to verify the analysis
 	Provider MetricProvider `json:"provider" protobuf:"bytes,10,opt,name=provider"`
+	// Whether to evaluate this metric in a Dry-Run mode
+	DryRun bool `json:"dryRun,omitempty" protobuf:"varint,11,opt,name=dryRun"`
```
jessesuen, line 110: Do we anticipate users will want to parameterize this? If so we may need a different datatype than `bool` since bools cannot be string replaced from arguments.

Alternatively, we might want to rethink the syntax to perform dry runs. For example, it could be at the spec level:

Dryrun sucess-rate and latency, but not saturation.
```yaml
spec:
  metrics:
  - name: success-rate
    interval: 10m
...
  - name: latency
    interval: 10m
...
  - name: saturation
    interval: 10m
...
  dryRun: "success-rate,latency"
```

Dry run everything:
```yaml
spec:
  metrics:
  - name: success-rate
    interval: 10m
...
  - name: latency
    interval: 10m
  dryRun: "*"
```

And we could improve the handling such that `dryRun` could be replaced with arguments.

docs/features/analysis.md
```diff
@@ -541,6 +541,53 @@ encountered.
           ))
 ```
 
+## Dry-Run Mode
+
+`dryRun` can be used on a metric to control whether or not to evaluate that metric in a dry-run mode. A metric running 
+in the dry-run mode won't impact the final state of the rollout or experiment even if it fails or the evaluation comes 
+out as inconclusive.
+
```
jessesuen, line 549: We will need a doc stating this is available since v1.2 (since we don't have versioend docs yet)

pkg/apis/rollouts/v1alpha1/analysis_types.go
agrawroh: @jessesuen I was thinking of it this way:

**1).** It could be done at metric level like this:
```json
metrics:
  - name: total-5xx-errors
    interval: 5m
    dryRun: true
    failureCondition: result[0] >= 10
    failureLimit: 3
    provider:
      ...
```

**2).** It could be done at the **AnalysisRun** level like this:
```json
spec:
  metrics:
  - name: success-rate
    interval: 10m
...
  - name: latency
    interval: 10m
  dryRun: "*"
```

```json
spec:
  metrics:
  - name: success-rate
    interval: 10m
...
  - name: latency
    interval: 10m
  dryRun: "success-rate, latency"
```

In this PR, I have implemented **Case 1**. I'll create a follow-up PR with docs etc. for **Case 2**.

What do you think?

docs/features/analysis.md
agrawroh: Added, Thanks!

pkg/apis/rollouts/v1alpha1/analysis_types.go
jessesuen: After thinking about this more, I think a spec level AnalysisRun setting is the more appropriate place for dry-run decisions because tightly coupling with the metric does not allow flexibility for different rollouts/users to decide if they want to opt-in to the metric. If we were to have it at the metric level, it makes the decision part of the Analysis**_Template_** instead of the Analysis**_Run_**. For example, with ClusterAnalysisTemplates, an operator should not unilaterally decide to make a metric dry-run for all rollouts in the cluster. This is a decision that should be reserved to Rollout users.

But I have one improvement to the syntax which I think will make it more explicit and canonically kubernetes, which is to use a list instead of a comma-delimited string.

Dry-Run success-rate and latency but not saturation:

```yaml
spec:
  metrics:
  - name: success-rate
    interval: 10m
...
  - name: latency
    interval: 10m
...
  - name: saturation
    interval: 10m
...
  dryRun:
  - metricName: success-rate
  - metricName: latency
```

The list would support wildcards if they want to dry run everything:

```yaml
  dryRun:
  - metricName: *
```

Also notice that the dryRun list is a list of structured objects instead of primitives. I suggest this because it may allow us to expand on the dryRun capabilities in the future:

```yaml
  dryRun:
  - metricName: success-rate
    futureFeature: true
```

When a Rollout wants to dry run their analysis, they simply need to specify the dryRun flag to the analysis stanza. e.g.:

```yaml
      - analysis:
          templates:
          - templateName: random-fail
          - templateName: always-pass
          dryRun:
          - metricName: *
```

Which we would simply/blindly carry over when constructing the AnalysisRun resource.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -316,6 +331,22 @@ type AnalysisRunStatus struct {
...
+	// This is equal to the sum of Successful, Failed, Inconclusive
+	Count int32 `json:"count,omitempty" protobuf:"varint,1,opt,name=count"`
+	// Successful is the number of times the metric was measured Successful
+	Successful int32 `json:"successful,omitempty" protobuf:"varint,2,opt,name=successful"`
+	// Failed is the number of times the metric was measured Failed
+	Failed int32 `json:"failed,omitempty" protobuf:"varint,3,opt,name=failed"`
+	// Inconclusive is the number of times the metric was measured Inconclusive
+	Inconclusive int32 `json:"inconclusive,omitempty" protobuf:"varint,4,opt,name=inconclusive"`
+	// Error is the number of times an error was encountered during measurement
+	Error int32 `json:"error,omitempty" protobuf:"varint,5,opt,name=error"`
```
jessesuen, line 349: This information appears to be aggregate information from the metric result list, but just provided as a convenience for humans? If so, I think it is better to remove this since:

1) I don't think we should have two sources of truth in the status. This information can be recaluclated from `status.metricResults` if necessary
2) the information is specific to dry run, so even if we were to provide summary information, I would think it should be for both dry runs and wet runs.

controller/metrics/prommetrics.go
```diff
@@ -111,14 +111,14 @@ var (
...
-		append(namespaceNameLabels, "metric", "type"),
+		append(namespaceNameLabels, "metric", "type", "dryRun"),
 		nil,
 	)
 
 	MetricAnalysisRunMetricPhase = prometheus.NewDesc(
 		"analysis_run_metric_phase",
 		"Information on the duration of a specific metric in the Analysis Run",
-		append(namespaceNameLabels, "metric", "type", "phase"),
+		append(namespaceNameLabels, "metric", "type", "dryRun", "phase"),
```
jessesuen, line 121: prometheus metrics aren't typically camelCase so I feel this should be `dry_run`

controller/metrics/prommetrics.go
```diff
@@ -111,14 +111,14 @@ var (
 	MetricAnalysisRunMetricType = prometheus.NewDesc(
 		"analysis_run_metric_type",
 		"Information on the type of a specific metric in the Analysis Runs",
-		append(namespaceNameLabels, "metric", "type"),
+		append(namespaceNameLabels, "metric", "type", "dryRun"),
```
jessesuen, line 114: TBH, i think `MetricAnalysisRunMetricType` is redundant to `MetricAnalysisRunMetricPhase` so we can stop adding to this metric and consider deprecating in the future. So you can remove the dry run label here.

utils/analysis/helpers.go
```diff
@@ -92,6 +92,20 @@ func IsTerminating(run *v1alpha1.AnalysisRun) bool {
...
+	wildcardPresent := false
+	for _, dryRunObject := range dryRunArray {
+		if dryRunObject.MetricName == "*" {
+			wildcardPresent = true
+			continue
+		}
+		metricNames[dryRunObject.MetricName] = true
+	}
+	return metricNames, wildcardPresent
+}
```
jessesuen, line 107: I think this method and usage of `GetDryRunMetricNames()` is a bit odd, in that it returns a **_tuple_** of the list of metrics which are supposed to be dry-runned **_or_** a boolean which indicates the user specified wildcard.

To simplify things and make it more generic, I think the method should be changed such that this accepts the `spec.metrics` as a second argument. Then this helper will a subset of `spec.metric` names which are intended to be dry-runned. If the user specified global wildcard `*`, then the returned list of metric names is everything under `spec.metrics`.

Secondly, this would also allow us to support partial wildcards like `foo*bar` in addition to the global wildcard `*`.

jessesuen (changes requested):

controller/metrics/prommetrics.go
agrawroh: Changed, Thanks!

controller/metrics/prommetrics.go
agrawroh: Removed.

utils/analysis/helpers.go
agrawroh: Great point! I changed it accordingly.

pkg/apis/rollouts/v1alpha1/analysis_types.go
agrawroh: As discussed offline, I have changed it to a `summary` instead and `dry-run summary` is part of this.

pkg/apis/rollouts/v1alpha1/analysis_types.go
```diff
@@ -316,6 +331,38 @@ type AnalysisRunStatus struct {
...
```
jessesuen, line 365: Instead of embedding dryRunSummary into runSummary i think this will be cleaner:

1) Have a single RunSummary type which re-used for both `status.runSummary` and `status.dryRunSummary`

```golang
	// RunSummary contains the final status of the metric results collected from the run
	RunSummary RunSummary `json:"runSummary,omitempty" protobuf:"bytes,5,opt,name=runSummary"`

	// DryRunSummary contains the final status of the metric results collected from the run
	DryRunSummary *RunSummary `json:"dryRunSummary,omitempty" protobuf:"bytes,6,opt,name=runSummary"`
```

Also, `dryRunSummary` should be a pointer because it will be omitted in the case that dryRun is not used.


```yaml
status:
  summary:
    count: 1
    ...
  dryRunSummary:
    count: 1
    ...
```

jessesuen (changes requested):
I think last change request about the status type structure and I think this will be good to go.

pkg/apis/rollouts/v1alpha1/analysis_types.go
agrawroh: Changed, Thanks!

jessesuen (approved):
Really great feature. Thank you!

eilonmonday:
@jessesuen amazing feature!
when is the release?

Issue #1242, opened 2021-06-02
Allow analysis arguments to get valueFrom Rollout status


jessesuen:
# Summary

Currently, we can pass arguments to AnalysisTemplates from `metadata` labels and annotations, e.g.

```yaml
  strategy:
    canary:
      analysis:
        templates:
        - templateName: args-example
        args:
        - name: region
          valueFrom:
            fieldRef:
              fieldPath: metadata.labels['region']
```

We should also allow passing arguments to AnalysisTemplates whose value is derived from fields in the rollout status.

# Use Cases

As part of https://github.com/argoproj/argo-rollouts/issues/1241, we will be surfacing ALB information under the rollout status. For example:

```yaml
status:
  alb:
    loadBalancer:
      name: k8s-devdeplo-testdele-b9061572ad
      arn: arn:aws:elasticloadbalancing:us-west-2:1234567891011:loadbalancer/app/k8s-devdeplo-testdele-abcdef1234/3a47c96b75571d0e
    canaryTargetGroup:
      name: k8s-devdeplo-testdele-ab72c5b132
      arn: arn:aws:elasticloadbalancing:us-west-2:1234567891011:targetgroup/k8s-devdeplo-testdele-ab72c5b132/df2957821c1214fc
    stableTargetGroup:
      name: k8s-devdeplo-testdele-8832d6519a
      arn: arn:aws:elasticloadbalancing:us-west-2:1234567891011:targetgroup/k8s-devdeplo-testdele-8832d6519a/df2957821c1214fc 
```

Once that is available, it would then be useful to pass that information to an AnalysisTemplate, so that argo rollouts could perform CloudWatch queries against the canary targetgroup vs. stable targetgroup, like so:

```yaml
      analysis:
        templates:
        - templateName: cloud-watch-error-rate
        args:
        - name: env
          valueFrom:
            fieldRef:
              fieldPath: status.alb.canaryTargetGroup.name
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
To make this feature more useful, instead of single field paths, we should support JSON string objects and enhance analysis to allow dot notation referencing when the supplied argument is a json string.

noam-codefresh:
I am not sure I understand this last comment - do you mean something more complex than the example above? Or just having the ability to reference any field in the status, but some dot notation like `status.alb.canaryTargetGroup.name`?

[...]

PR #1629, opened 2021-11-08
feat(analysis): Allow analysis arguments to get valueFrom Rollout status (#1242)

noam-codefresh:
Checklist:

### DIFF
--- a/docs/features/analysis.md
+++ b/docs/features/analysis.md
@@ -426,8 +426,9 @@ spec:
           valueFrom:
             podTemplateHashValue: Latest
 ```
-Analysis arguments also support valueFrom for reading metadata fields and passing them as arguments to AnalysisTemplate.
-An example would be to reference metadata labels like env and region and passing them along to AnalysisTemplate.
+Analysis arguments also support valueFrom for reading any Rollout fields and passing them as arguments to AnalysisTemplate.
+An example would be to reference metadata labels like env and region and passing them along to AnalysisTemplate, or any field
+from the Rollout status
 ```yaml
 apiVersion: argoproj.io/v1alpha1
 kind: Rollout
@@ -457,6 +458,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.labels['region']
+        - name: canary-hash
+          valueFrom:
+            fieldRef:
+              fieldPath: status.canary.weights.canary.podTemplateHash
 ```
 
 ## BlueGreen Pre Promotion Analysis
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/argoproj/argo-rollouts
 go 1.16
 
 require (
+	github.com/PaesslerAG/jsonpath v0.1.1
 	github.com/antonmedv/expr v1.8.9
 	github.com/argoproj/notifications-engine v0.2.1-0.20210525191332-e8e293898477
 	github.com/argoproj/pkg v0.9.0
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/argoproj/argo-rollouts
 go 1.16
 
 require (
-	github.com/PaesslerAG/jsonpath v0.1.1
 	github.com/antonmedv/expr v1.8.9
 	github.com/argoproj/notifications-engine v0.2.1-0.20210525191332-e8e293898477
 	github.com/argoproj/pkg v0.9.0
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/argoproj/argo-rollouts
 go 1.16
 
 require (
+	github.com/PaesslerAG/jsonpath v0.1.1
```
jessesuen, line 6: Hmmm, this library uses "BSD 3-Clause "New" or "Revised" License" which is not what we typically use (Apache 2 or MIT). I'm not sure this is okay to use. Is there something else we can use or write our own?

jessesuen (changes requested):
Could you provide documentation on this? It's not clear how this will work and how to use it.

go.mod
noam-codefresh: I tried using a different library called `objx`, which was similar. But the problem there was that using a non-existing path (`path: "some.none.existing[4].path.in.rollout"`) just returns an empty string. So it's not possible to display an error when the user used an invalid path or not (unless we assume an empty string is always an error).

If you know of any other possible libraries, I'd be happy to give them a shot as well.
I can also write a simple parser by splitting on `"."`. I don't think it will be too complicated, given our very specific use-case (specifically with field names that do not contain dots themselves)

noam-codefresh:
I removed the dependency on `jsonPath`, and now the code is parsing the path directly. It will support dot notation and array indexers, so something like `status.canary.weights.canary.podTemplateHash` (the original request), or `status.conditions[0].status`, but not anything more complex

utils/analysis/factory.go
```diff
@@ -217,3 +227,42 @@ func ValidateMetric(metric v1alpha1.Metric) error {
...
+	}
+
+	if m == nil {
+		return "", fmt.Errorf("invalid path %s in rollout", path)
+	}
+
+	var isArray, isMap bool
+	_, isArray = m.([]interface{})
+	_, isMap = m.(map[string]interface{})
+	if isArray || isMap {
```
harikrongali, line 263: what is this check is for? can you provide an example?

noam-codefresh: If the user puts in something like `status.canary.weights.canary` - the resulting value would be a map, and not a primitive value (string/number/bool).
I guess we could support that as well, and just return it as a JSON string, but is it a reasonable usecase? It seems like it might be a future ticket.

utils/analysis/factory.go
```diff
@@ -217,3 +227,42 @@ func ValidateMetric(metric v1alpha1.Metric) error {
 	}
 	return nil
 }
+
+func extractValueFromRollout(r *v1alpha1.Rollout, path string) (string, error) {
```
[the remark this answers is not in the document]
noam-codefresh: I added some tests for invalid paths.
I also removed the error check after the json.Marshal - I assume that a valid Rollout object can always be marshalled to json.

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/soheilhy/cmux v0.1.4
 	github.com/spaceapegames/go-wavefront v1.8.1
 	github.com/spf13/cobra v1.1.3
+	github.com/stretchr/objx v0.3.0 // indirect
```
harikrongali, line 35: are we using this library?

harikrongali:
LGTM..except the comment about lib objx

go.mod
noam-codefresh: you are right. fixing now.

harikrongali (approved):

alexmt (approved):

magec:
This is exactly what I am looking for, I need to make prometheus queries using `status.currentPodHash`, my life would be way easier with this feature. Looking forward for the merge.

Thanks @noam-codefresh!

alexmt:
@jessesuen can you take another look please ?

docs/features/analysis.md
```diff
@@ -426,8 +426,9 @@ spec:
           valueFrom:
             podTemplateHashValue: Latest
 ```
-Analysis arguments also support valueFrom for reading metadata fields and passing them as arguments to AnalysisTemplate.
-An example would be to reference metadata labels like env and region and passing them along to AnalysisTemplate.
+Analysis arguments also support valueFrom for reading any Rollout fields and passing them as arguments to AnalysisTemplate.
+An example would be to reference metadata labels like env and region and passing them along to AnalysisTemplate, or any field
+from the Rollout status
```
jessesuen, line 431: Since we don't have versioned docs yet, could we add a note stating v1.2 supports valueFrom rollout status?

jessesuen (approved):
Minor nit about documentation but otherwise LGTM

magec:
Can this be merged? I have run E2E tests locally and they seem to work well.

PR #1646, opened 2021-11-11
fix: Modify Experiment collision naming from dot-notation to dash

### LINKED ISSUE #1631: Faulty naming convention for Experiments with collision handling
We currently use dot-convention to handle collisions in event that there are multiple Experiments with same name and Spec. This is an issue because the name is not DNS compatible and causes problems in the AWS console.

Here is the problematic code:

image

https://github.com/argoproj/argo-rollouts/blob/master/rollout/experiment.go#L208-#L236

khhirani:
Fix #1631 


harikrongali (approved):

harikrongali:
@khhirani can you rebase to latest and rerun e2e & unit tests

harikrongali:
@khhirani can you check e2e tests?

alexmt (approved):

Issue #1624, opened 2021-11-04
`progressDeadlineAbort` causes degraded Rollout outside of a deployment


bpoland:
## Summary 

It seems that with `progressDeadlineAbort: true` that some Rollouts randomly get marked as degraded when no deployment is in progress.  This has been happening repeatedly since I turned on `progressDeadlineAbort` (and stops happening when I turn that back off).

When the issue happens, the Rollout is left like this:
```
Name:            name
Namespace:       default
Status:          ✖ Degraded
Message:         RolloutAborted: Rollout aborted update to revision 216
Strategy:        Canary
  Step:          20/20
  SetWeight:     0
  ActualWeight:  100
Images:          image:tag (stable)
Replicas:
  Desired:       3
  Current:       3
  Updated:       3
  Ready:         3
  Available:     3

NAME                                    KIND         STATUS        AGE  INFO
⟳ name                                 Rollout      ✖ Degraded    22d
├──# revision:216
│  ├──⧉ name-7bc889bb44                ReplicaSet   ✔ Healthy     64m  stable
│  │  ├──□ name-7bc889bb44-79cvk       Pod          ✔ Running     63m  ready:2/2
│  │  ├──□ name-7bc889bb44-5nwqw       Pod          ✔ Running     59m  ready:2/2
│  │  └──□ name-7bc889bb44-nqqt8       Pod          ✔ Running     53m  ready:2/2
│  └──α name-7bc889bb44-216            AnalysisRun  ✔ Successful  64m  ✔ 92
├──# revision:215
│  ├──⧉ name-5f5d69cb8                 ReplicaSet   • ScaledDown  98m
│  └──α name-5f5d69cb8-215             AnalysisRun  ✔ Successful  98m  ✔ 108
└──# revision:214
   ├──⧉ name-99f85845f                 ReplicaSet   • ScaledDown  21h
   └──α name-99f85845f-214             AnalysisRun  ✔ Successful  21h  ✔ 103
```

Luckily with the changes in 1.1 having the `SetWeight: 0` does not actually disrupt traffic but it is still concerning.  I would expect `progressDeadlineAbort` not to abort after the deployment is complete.

## Diagnostics

Controller 1.1

Here are the redacted logs for a rollout that had the issue.  You can see that the rollout gets aborted due to progress at 2021-11-04T14:46:54Z, nearly half an hour after the rollout completed at 2021-11-04T14:18:15Z

[...]

PR #1649, opened 2021-11-11
fix: reset the progress condition when a pod is restarted

huikang:
- if the progress condition is not reset, the timeout check
  returns true for a healthy rollout

fix: #1624 

Checklist:

### DIFF
--- a/rollout/restart.go
+++ b/rollout/restart.go
@@ -2,11 +2,11 @@ package rollout
 
 import (
 	"context"
-	"fmt"
+	// "fmt"
 	"sort"
 	"time"
 
-	"github.com/argoproj/argo-rollouts/utils/conditions"
+	// "github.com/argoproj/argo-rollouts/utils/conditions"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1beta1"
@@ -125,14 +125,17 @@ func (p *RolloutPodRestarter) Reconcile(roCtx *rolloutContext) error {
 		canRestart -= 1
 		restarted += 1
 
-		// Everytime a pod is restarted, updating the rollout's progressingCondition
-		existProgressingCondition := conditions.GetRolloutCondition(roCtx.rollout.Status, v1alpha1.RolloutProgressing)
-		if existProgressingCondition != nil {
-			conditions.RemoveRolloutCondition(&roCtx.rollout.Status, v1alpha1.RolloutProgressing)
-		}
-		msg := fmt.Sprintf("pod %s restarted", pod.Name)
-		newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.NewRSAvailableReason, msg)
-		conditions.SetRolloutCondition(&roCtx.rollout.Status, *newProgressingCondition)
+		// TODO: remove
+		/*
+			// Everytime a pod is restarted, updating the rollout's progressingCondition
+			existProgressingCondition := conditions.GetRolloutCondition(roCtx.rollout.Status, v1alpha1.RolloutProgressing)
+			if existProgressingCondition != nil {
+				conditions.RemoveRolloutCondition(&roCtx.rollout.Status, v1alpha1.RolloutProgressing)
+			}
+			msg := fmt.Sprintf("pod %s restarted", pod.Name)
+			newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.NewRSAvailableReason, msg)
+			conditions.SetRolloutCondition(&roCtx.rollout.Status, *newProgressingCondition)
+		*/
 	}
 	remaining := needsRestart - restarted
 
--- a/rollout/restart.go
+++ b/rollout/restart.go
@@ -2,9 +2,11 @@ package rollout
 
 import (
 	"context"
+	"fmt"
 	"sort"
 	"time"
 
+	"github.com/argoproj/argo-rollouts/utils/conditions"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1beta1"
@@ -122,6 +124,15 @@ func (p *RolloutPodRestarter) Reconcile(roCtx *rolloutContext) error {
 		}
 		canRestart -= 1
 		restarted += 1
+
+		// Everytime a pod is restarted, updating the rollout's progressingCondition
+		existProgressingCondition := conditions.GetRolloutCondition(roCtx.rollout.Status, v1alpha1.RolloutProgressing)
+		if existProgressingCondition != nil {
+			conditions.RemoveRolloutCondition(&roCtx.rollout.Status, v1alpha1.RolloutProgressing)
+		}
+		msg := fmt.Sprintf("pod %s restarted", pod.Name)
+		newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.NewRSAvailableReason, msg)
+		conditions.SetRolloutCondition(&roCtx.rollout.Status, *newProgressingCondition)
 	}
 	remaining := needsRestart - restarted
 
--- a/rollout/restart.go
+++ b/rollout/restart.go
@@ -2,11 +2,9 @@ package rollout
 
 import (
 	"context"
-	// "fmt"
 	"sort"
 	"time"
 
-	// "github.com/argoproj/argo-rollouts/utils/conditions"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1beta1"
@@ -124,18 +122,6 @@ func (p *RolloutPodRestarter) Reconcile(roCtx *rolloutContext) error {
 		}
 		canRestart -= 1
 		restarted += 1
-
-		// TODO: remove
-		/*
-			// Everytime a pod is restarted, updating the rollout's progressingCondition
-			existProgressingCondition := conditions.GetRolloutCondition(roCtx.rollout.Status, v1alpha1.RolloutProgressing)
-			if existProgressingCondition != nil {
-				conditions.RemoveRolloutCondition(&roCtx.rollout.Status, v1alpha1.RolloutProgressing)
-			}
-			msg := fmt.Sprintf("pod %s restarted", pod.Name)
-			newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.NewRSAvailableReason, msg)
-			conditions.SetRolloutCondition(&roCtx.rollout.Status, *newProgressingCondition)
-		*/
 	}
 	remaining := needsRestart - restarted
 
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -586,14 +586,15 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
 			changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
 
-			// if any rs status changes (e.g., pod restarted, evicted -> recreated ) to a previous completed rollout,
-			// we need to reset the progressCondition to avoid timeout
+			// If the ReplicaSet status changes (e.g., one of the pod restarts, evicted -> recreated) for a previously
+			// completed rollout, we'll need to reset the rollout's condition to `PROGRESSING` to avoid any timeouts.
 			if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS)) {
 				existProgressingCondition := conditions.GetRolloutCondition(newStatus, v1alpha1.RolloutProgressing)
 				if existProgressingCondition != nil {
 					conditions.RemoveRolloutCondition(&newStatus, v1alpha1.RolloutProgressing)
 				}
 				msg := fmt.Sprintf("stable rs %s progressing", c.stableRS.Name)
+				c.log.Infof("Fully promoted rollout %s becomes progressing due to: %s", c.rollout.Name, msg)
 				newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.NewRSAvailableReason, msg)
 				conditions.SetRolloutCondition(&newStatus, *newProgressingCondition)
 			}
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -586,16 +586,16 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
 			changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
 
-			// If the ReplicaSet status changes (e.g., one of the pod restarts, evicted -> recreated) for a previously
-			// completed rollout, we'll need to reset the rollout's condition to `PROGRESSING` to avoid any timeouts.
-			if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS)) {
+			// When a fully promoted rollout becomes Incomplete, e.g., due to the ReplicaSet status changes like
+			// pod restarts, evicted -> recreated, we'll need to reset the rollout's condition to `PROGRESSING` to
+			// avoid any timeouts.
+			if changed {
 				existProgressingCondition := conditions.GetRolloutCondition(newStatus, v1alpha1.RolloutProgressing)
 				if existProgressingCondition != nil {
 					conditions.RemoveRolloutCondition(&newStatus, v1alpha1.RolloutProgressing)
 				}
-				msg := fmt.Sprintf("stable rs %s progressing", c.stableRS.Name)
-				c.log.Infof("Fully promoted rollout %s becomes progressing due to: %s", c.rollout.Name, msg)
-				newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.NewRSAvailableReason, msg)
+				c.log.Infof(conditions.RolloutBecomesIncompleteMessage)
+				newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.RolloutBecomesIncomplete, conditions.RolloutBecomesIncompleteMessage)
 				conditions.SetRolloutCondition(&newStatus, *newProgressingCondition)
 			}
 		}
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -584,7 +584,19 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	} else {
 		if completeCond != nil {
 			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
-			conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+			changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+
+			// if any rs status changes (e.g., pod restarted, evicted -> recreated ) to a previous completed rollout,
+			// we need to reset the progressCondition to avoid timeout
+			if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS)) {
+				existProgressingCondition := conditions.GetRolloutCondition(newStatus, v1alpha1.RolloutProgressing)
+				if existProgressingCondition != nil {
+					conditions.RemoveRolloutCondition(&newStatus, v1alpha1.RolloutProgressing)
+				}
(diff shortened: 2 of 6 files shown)


jessesuen (commented):
@huikang Do you know if this problem happen in the general sense? If so, I think this won't solve the problem in the generic case (e.g. when pods restart out-of-band from a rollout restart action) because this code is only in the pod restarter `Reconcile`

If we can reproduce this **_outside_** of the restart action, then we will need to fix this in a more generic way. I think one change that might be needed is: at the time when we transition from Healthy to Progressing, we also to also reset the progress counter so that the deadline is reset after we transition out of Healthy.

agrawroh:
Pods can transition to a non-running state for a variety of reasons. For example, due to k8s liveness probe starts failing.
I think we should think a little bit more about the behavior. Do we expect an already completed rollout to transition to a "Progressing" state if one of the stable replicas go from "Running" to "Pending" state? In my mind, I wouldn't care if anything happens to the stable replicas once the rollout is completed.
cc @huikang @jessesuen What do you think?

agrawroh (commented):
Pods can transition to a non-running state for a variety of reasons. For example, due to k8s liveness probe starts failing.
I think we should think a little bit more about the behavior. Do we expect an already completed rollout to transition to a "Progressing" state if one of the stable replicas go from "Running" to "Pending" state? In my mind, I wouldn't care if anything happens to the stable replicas once the rollout is completed.
cc @huikang @jessesuen What do you think?

huikang:
@jessesuen , you are right. This only fixes the `restart pod` case by resetting the progress counter, but not for a general case.

@agrawroh , if the pods of a completed rollout becomes unready (e.g., restart, failed liveness probe), we should mark the rollout from healthy to `progressing` because the rollout controller is reconciling the state.

So a generic solution as @jessesuen mentioned is to reset the progress counter so that conditions.RolloutTimedOut doesn't return true based on an old `LastUpdateTime`. Will update the PR.

agrawroh:
> @jessesuen , you are right. This only fixes the `restart pod` case by resetting the progress counter, but not for a general case.
> 
> @agrawroh , if the pods of a completed rollout becomes unready (e.g., restart, failed liveness probe), we should mark the rollout from healthy to `progressing` because the rollout controller is reconciling the state.
> 
> So a generic solution as @jessesuen mentioned is to reset the progress counter so that conditions.RolloutTimedOut doesn't return true based on an old `LastUpdateTime`. Will update the PR.

@huikang Thanks for the explanation. This might be a naive question - Do we expect the rollout to create another AnalysisRun if it gets from "Completed" -> "Progressing" state? Also, if somehow the `progressDeadlineAbort` gets fired while the rollout is still in the "Progressing" state, will we mark it as degraded?

jessesuen:
> Do we expect the rollout to create another AnalysisRun if it gets from "Completed" -> "Progressing" state

No. We only trigger analysis when we are in the middle of an update. Transitioning to progressing does not qualify in that regard. But if user "retries" the update, analysis will start again. We know we are in the middle of an update when the stable hash != desired hash.

> Also, if somehow the progressDeadlineAbort gets fired while the rollout is still in the "Progressing" state, will we mark it as degraded

The intended way progressDeadlineAbort is meant to work is: if we are in the middle of an update, we will *abort* and go back to stable if we fail to make progress. This does not apply when we are in a fully rolled out state (stable is already desired). That said, if we exceed the progressDeadlineSeconds when we are fully promoted (stable == desired), the rollout should be in a degraded status (just like Deployment).

huikang:
```
t1:   Complete and healthy Rollout (stable RS == descired RS)
t2:   Something happened to the stable RS, e.g., a pod is evicted.
       The rollout becomes progressing.

    ...  After some time

t3:  The evicted pod becomes ready so the stable RS becomes available again
```

The logic is as follows:

```go
if complete rollout becomes incomplete and not in the middle of an update {
    reset the progress LastUpdate time to t2
}

If (t3 - t2) > progressDeadlineSeconds (i.e., the stable RS of a fully promoted rollout becomes ready after `progressDeadlineSeconds`), {
     the rollout is degraded
} else {
    the rollout is healthy
}

```

rollout/sync.go
```diff
@@ -584,7 +584,19 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	} else {
 		if completeCond != nil {
 			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
-			conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+			changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+
+			// if any rs status changes (e.g., pod restarted, evicted -> recreated ) to a previous completed rollout,
+			// we need to reset the progressCondition to avoid timeout
+			if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS)) {
+				existProgressingCondition := conditions.GetRolloutCondition(newStatus, v1alpha1.RolloutProgressing)
```
agrawroh, line 592: **Minor:** `existProgressingCondition` --> `isRolloutConditionInProgressing`

rollout/sync.go
```diff
@@ -584,7 +584,19 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	} else {
 		if completeCond != nil {
 			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
-			conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+			changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+
+			// if any rs status changes (e.g., pod restarted, evicted -> recreated ) to a previous completed rollout,
+			// we need to reset the progressCondition to avoid timeout
```
agrawroh, line 590:
```suggestion
			// If the ReplicaSet status changes (e.g., one of the pod restarts, evicted -> recreated) for a previously
			// completed rollout, we'll need to reset the rollout's condition to `PROGRESSING` to avoid any timeouts.
```

**Minor:** Making the comment a little bit more clear for any future maintainers who are not very familiar to this logic.

rollout/sync.go
```diff
@@ -584,7 +584,19 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
...
+			if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS)) {
+				existProgressingCondition := conditions.GetRolloutCondition(newStatus, v1alpha1.RolloutProgressing)
+				if existProgressingCondition != nil {
+					conditions.RemoveRolloutCondition(&newStatus, v1alpha1.RolloutProgressing)
+				}
+				msg := fmt.Sprintf("stable rs %s progressing", c.stableRS.Name)
```
agrawroh, line 596: Maybe also log something about the Rollout existing condition being changed from `existProgressingCondition` --> `PROGRESSING`? It doesn't have to be a part of this log (Can be a separate log).

rollout/restart.go
```diff
@@ -122,6 +124,18 @@ func (p *RolloutPodRestarter) Reconcile(roCtx *rolloutContext) error {
 		}
 		canRestart -= 1
 		restarted += 1
+
+		// TODO: remove
```
agrawroh, line 128: Do you still want to keep this?

agrawroh (commented):
Looks good overall. Lets some nits.

rollout/sync.go
huikang: Done

rollout/sync.go
huikang: isRolloutConditionInProgressing sounds a bool variable. existProgressingCondition is consistent with other places where `GetRolloutCondition` is called. What do you think?

rollout/restart.go
huikang: Removed. Thanks.

agrawroh (approved):

rollout/sync.go
```diff
@@ -584,7 +584,20 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
 	} else {
 		if completeCond != nil {
 			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
-			conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+			changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
+
+			// If the ReplicaSet status changes (e.g., one of the pod restarts, evicted -> recreated) for a previously
+			// completed rollout, we'll need to reset the rollout's condition to `PROGRESSING` to avoid any timeouts.
+			if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS)) {
```
jessesuen, line 591: It seems we only do this when we are fully promoted. But why isnt' this check just:

```golang
if changed {
```

In all cases, if we move from Completed to not Completed, don't we want to reset the progressing condition timestamp?

huikang: True. When a rollout becomes incomplete from complete, we always want to reset the progressing condition.

I made the change and tested successfully with `plugin rollout restart` and out-of-band pod recreation.

huikang: @jessesuen , 
I remember the reason of using `if changed && c.stableRS != nil && c.newRS != nil && (replicasetutil.GetPodTemplateHash(c.stableRS) == replicasetutil.GetPodTemplateHash(c.newRS))` is to only reset the progress when (rollout becomes complete to incomplete) AND (not in the middle of update).

During an update, we don't need to reset the progressing condition here, because it has already been reset. 
So using the only condition `if changed` causes duplicated resetting. WDYK?

huikang: Reverted to the condition of  (rollout becomes incomplete from complete) AND (not in the middle of update) due to the failed `TestFunctionalSuite/TestWorkloadRef` in e2e test using only `if changed {`

```
ERRO[2021-11-13T00:20:14-05:00] Recovered from panic: runtime error: invalid memory address or nil pointer dereference
goroutine 185 [running]:
runtime/debug.Stack(0xc0006be678, 0x2f60a40, 0x472d5e0)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/debug/stack.go:24 +0x9f
github.com/argoproj/argo-rollouts/utils/controller.processNextWorkItem.func1.1.1(0xc000ba6070, 0xc0006bfaf0)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/utils/controller/controller.go:149 +0x5b
panic(0x2f60a40, 0x472d5e0)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/panic.go:965 +0x1b9
github.com/argoproj/argo-rollouts/rollout.(*rolloutContext).calculateRolloutConditions(0xc000bd3c00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc0008e58e4, 0xa, 0x0, ...)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/sync.go:597 +0x13c0
github.com/argoproj/argo-rollouts/rollout.(*rolloutContext).syncRolloutStatusBlueGreen(0xc000bd3c00, 0x0, 0xc000b30280, 0x0, 0x0)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/bluegreen.go:312 +0x538
github.com/argoproj/argo-rollouts/rollout.(*rolloutContext).rolloutBlueGreen(0xc000bd3c00, 0x32fa72f, 0x17)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/bluegreen.go:35 +0x2f5
github.com/argoproj/argo-rollouts/rollout.(*rolloutContext).reconcile(0xc000bd3c00, 0xc000026c00, 0xc000bd3c00)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/context.go:82 +0x1e5
github.com/argoproj/argo-rollouts/rollout.(*Controller).syncHandler(0xc000c12000, 0xc00094a040, 0x1e, 0x0, 0x0)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/controller.go:396 +0x630
github.com/argoproj/argo-rollouts/utils/controller.processNextWorkItem.func1.1(0x0, 0x0)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/utils/controller/controller.go:153 +0x7c
github.com/argoproj/argo-rollouts/utils/controller.processNextWorkItem.func1(0x36dec60, 0xc000b6ed80, 0x32df9ad, 0x7, 0xc00149de60, 0xc000490e00, 0x2e4c200, 0xc001192850, 0x0, 0x0)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/utils/controller/controller.go:157 +0x326
github.com/argoproj/argo-rollouts/utils/controller.processNextWorkItem(0x36dec60, 0xc000b6ed80, 0x32df9ad, 0x7, 0xc00149de60, 0xc000490e00, 0x1)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/utils/controller/controller.go:171 +0x9a
github.com/argoproj/argo-rollouts/utils/controller.RunWorker(...)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/utils/controller/controller.go:104
github.com/argoproj/argo-rollouts/rollout.(*Controller).Run.func1()
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/controller.go:326 +0xa5
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc000e4a050)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155 +0x5f
k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0xc000e4a050, 0x366f0e0, 0xc0005d6c00, 0x1, 0xc0005a0300)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156 +0x9b
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000e4a050, 0x3b9aca00, 0x0, 0x1, 0xc0005a0300)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133 +0x98
k8s.io/apimachinery/pkg/util/wait.Until(0xc000e4a050, 0x3b9aca00, 0xc0005a0300)
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:90 +0x4d
created by github.com/argoproj/argo-rollouts/rollout.(*Controller).Run
	/Users/hui.kang/go/src/github.com/huikang/argo-rollouts-backup2/rollout/controller.go:325 +0xac  namespace=default rollou

```

jessesuen: Hmmm. I was under the impression that we would not enter into a completed condition when the rollout was in the middle of an update, which is why I suggested the change. Are you finding that is not the case

huikang: > Hmmm. I was under the impression that we would not enter into a completed condition when the rollout was in the middle of an update, which is why I suggested the change. Are you finding that is not the case

There might be some bug for `workloadRef`, which causes entering a completed condition in the middle of an update. Let me dig deep to the case.

jessesuen: OK. Let me know what you find. I think we might be masking some underlying issue by having the second check. If it turns out it is really necessary, I think a better option is to use the convenience method:

`conditions.RolloutComplete(c.rollout, &newStatus)`

huikang: @jessesuen , it turns out the panic was due to the nil pointer of c.StableRS of workloadRef. `if changed` works fine and I also added a new progressing reason and its associated msg.

rollout/sync.go
```diff
@@ -584,7 +584,20 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
...
+			// pod restarts, evicted -> recreated, we'll need to reset the rollout's condition to `PROGRESSING` to
+			// avoid any timeouts.
+			if changed {
+				existProgressingCondition := conditions.GetRolloutCondition(newStatus, v1alpha1.RolloutProgressing)
+				if existProgressingCondition != nil {
+					conditions.RemoveRolloutCondition(&newStatus, v1alpha1.RolloutProgressing)
+				}
+				c.log.Infof(conditions.RolloutBecomesIncompleteMessage)
+				newProgressingCondition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.RolloutBecomesIncomplete, conditions.RolloutBecomesIncompleteMessage)
+				conditions.SetRolloutCondition(&newStatus, *newProgressingCondition)
```
jessesuen, line 599: @huikang i think this message and reason will get clobbered by the switch statement that comes later.

Instead of updating the Progressing condition here (where we detect that we became incomplete), can we update the Progressing condition at the current place in the code where we are managing the Progressing condition (in the switch statement)? This will require some new boolean (e.g. becameIncomplete) set in the completed check:

```golang
	var becameIncomplete bool // remember if we transitioned from completed:
	if !isPaused && conditions.RolloutComplete(c.rollout, &newStatus) {
...
	} else {
		if completeCond != nil {
			updateCompletedCond := conditions.NewRolloutCondition(v1alpha1.RolloutCompleted, corev1.ConditionFalse, conditions.RolloutCompletedReason, conditions.RolloutCompletedReason)
			becameIncomplete = conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
changed := conditions.SetRolloutCondition(&newStatus, *updateCompletedCond)
...

	if !isCompleteRollout && !isAborted {
		switch {
...
		case conditions.RolloutProgressing(c.rollout, &newStatus):
...
			// Give a more accurate reason for the Progressing condition
			if newStatus.StableRS == newStatus.CurrentPodHash {
				reason = conditions.ReplicaSetNotAvailableReason
				msg = conditions.NotAvailableMessage // re-use existing message
			} else {
				reason = conditions.ReplicaSetUpdatedReason
			}
			condition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, reason, msg)

			if currentCond != nil  || becameIncomplete {
				if currentCond.Status == corev1.ConditionTrue {
					condition.LastTransitionTime = currentCond.LastTransitionTime
				}
				conditions.RemoveRolloutCondition(&newStatus, v1alpha1.RolloutProgressing)
			}
			conditions.SetRolloutCondition(&newStatus, *condition)
```

utils/conditions/conditions.go
```diff
@@ -86,6 +86,15 @@ const (
...
 
+	// RolloutBecomesIncomplete is added when a fully promoted rollout becomes incomplete, e.g.,
+	// due to  pod restarts, evicted -> recreated. In this case, we'll need to reset the rollout's
+	// condition to `PROGRESSING` to avoid any timeouts.
+	RolloutBecomesIncomplete = "RolloutBecomesIncomplete"
+	// RolloutBecomesIncompleteMessage is added when a fully promoted rollout becomes incomplete, e.g.,
+	// due to  pod restarts, evicted -> recreated. In this case, we'll need to reset the rollout's
+	// condition to `PROGRESSING` to avoid any timeouts.
+	RolloutBecomesIncompleteMessage = "Fully promoted rollout becomes incomplete"
+
```
jessesuen, line 97: Lets create a new reason:

```golang
ReplicaSetNotAvailableReason = "ReplicaSet is not available"
```

and instead of `RolloutBecomesIncompleteMessage`, we can use existing `NotAvailableMessage` message

rollout/sync.go
huikang: Thanks for the suggestion. Separating the processing of progressing and complete condition sounds better. Let me update the PR to test the logic.

rollout/sync.go
```diff
@@ -619,14 +620,24 @@ func (c *rolloutContext) calculateRolloutConditions(newStatus v1alpha1.RolloutSt
...
 			if c.newRS != nil {
 				msg = fmt.Sprintf(conditions.ReplicaSetProgressingMessage, c.newRS.Name)
 			}
-			condition := conditions.NewRolloutCondition(v1alpha1.RolloutProgressing, corev1.ConditionTrue, conditions.ReplicaSetUpdatedReason, msg)
+			reason := conditions.ReplicaSetUpdatedReason
+
+			// When a fully promoted rollout becomes Incomplete, e.g., due to the ReplicaSet status changes like
+			// pod restarts, evicted -> recreated, we'll need to reset the rollout's condition to `PROGRESSING` to
+			// avoid any timeouts.
+			if becameIncomplete {
```
jessesuen, line 635: @huikang i have a feeling that this check will cause the Progressing reason to be `ReplicaSetUpdatedReason` after two reconciliations. Which is why in my suggestion i had the check:

```
if newStatus.StableRS == newStatus.CurrentPodHash {
```

Do you agree?

jessesuen (approved):

jessesuen:
Looks like my suggestion caused unit test to fail because change in message:
```
        	            	-{"status":{"blueGreen":{"previewSelector":"68f45df865"},"conditions":[{"lastTransitionTime":"2021-11-18T23:43:38Z","lastUpdateTime":"2021-11-18T23:43:38Z","message":"ReplicaSet \"foo-68f45df865\" is progressing.","reason":"ReplicaSetUpdated","status":"True","type":"Progressing"},{"lastTransitionTime":"2021-11-18T23:43:38Z","lastUpdateTime":"2021-11-18T23:43:38Z","message":"Rollout does not have minimum availability","reason":"AvailableReason","status":"False","type":"Available"}],"message":"more replicas need to be updated","observedGeneration":"123","phase":"Progressing","selector":"foo=bar","stableRS":"68f45df865"}}
        	            	+{"status":{"blueGreen":{"previewSelector":"68f45df865"},"conditions":[{"lastTransitionTime":"2021-11-18T23:43:38Z","lastUpdateTime":"2021-11-18T23:43:38Z","message":"Rollout does not have minimum availability","reason":"ReplicaSetNotAvailable","status":"True","type":"Progressing"},{"lastTransitionTime":"2021-11-18T23:43:38Z","lastUpdateTime":"2021-11-18T23:43:38Z","message":"Rollout does not have minimum availability","reason":"AvailableReason","status":"False","type":"Available"}],"message":"more replicas need to be updated","observedGeneration":"123","phase":"Progressing","selector":"foo=bar","stableRS":"68f45df865"}}
```

huikang:
Hi, @jessesuen , added a new unit test. Please take a look. Thanks.

alexmt (approved):
LGTM

PR #1654, opened 2021-11-12
refactor: stop using mpatch

alexmt:
Signed-off-by: Alexander Matyushentsev <AMatyushentsev@gmail.com>

Unit tests are failing on mac :( Looks like https://github.com/undefinedlabs/go-mpatch does not work well on the latest mac os versions. There is no ARM support as well: https://github.com/undefinedlabs/go-mpatch/issues/2 . The ARM support was requested in 2020, so unlikely it will be supported. 

I suggest to stop using https://github.com/undefinedlabs/go-mpatch. PR removes the usage in one place. We still use the library in `controller_test.go#newFixture`

### DIFF
--- a/analysis/analysis.go
+++ b/analysis/analysis.go
@@ -7,17 +7,17 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/utils/pointer"
-
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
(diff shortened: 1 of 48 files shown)


jessesuen (commented):
You removed one instance, but there are 3 others so we haven't removed the dependency completely since the go.mod/go.sum haven't been affected.

I think you either have to modify the other tests to remove it or:

If you are emboldened, we could switch our code to use a `util/time.Now()` helper which is can be overwritten by unit tests

alexmt:
Good suggestion @jessesuen .

I've introduced `util/time` package and using it instead of `time.Now`/`metav1.Now`. I was planning to make minimal possible changes. It appears that I had to change a lot to make tests pass. The PR is big but all changes just additional import and switch to `timeutil.Now`/`timeutil.MetaNow` function.

PTAL

jessesuen (approved):

PR #1663, opened 2021-11-18
fix: Promote-full with TrafficRouting waits for all canary pods to come up

### LINKED ISSUE #1580: Promote-full with traffic splitting doesn't wait for new pods to be ready
## Summary 

When using the promote-full option with traffic splitting, 100% of traffic is immediately sent to the new ReplicaSet, even if it is not scaled up enough to handle the traffic.

Instead, it should immediately scale up the new RS to full size, but wait to adjust the traffic split as the new pods become ready.  I had thought this was the difference between set weight and actual weight, but that doesn't seem to be the case (maybe someone could help me understand what the difference is then?)

As a side effect when using the new `dynamicStableScale` feature in 1.1, this means that the old RS gets immediately scaled down and we could be left with very few running/ready pods.  I think this is a symptom of the above root cause though, and I guess the traffic split is set to send all traffic to the new RS anyway.

## Diagnostics

Rollouts 1.1

---

khhirani:
Fixes #1580 


harikrongali:
@khhirani can you add unit test?

alexmt (approved):

Issue #1581, opened 2021-10-14
Dashboard should allow promote/promote-full at any time


bpoland:
# Summary

Currently the rollout dashboard disables the promote and promote-full buttons at certain points (seems to be when scaling canary/setting weight requires pods to be started and become ready).  The CLI options do not have this restriction, so I think the dashboard should also allow promoting at any time.

# Use Cases

When I want to promote/promote-full while pods are starting

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
Yes, this seems reasonable. Promote and promote-full should be available anytime during an update (but not when fully rollouted out at a steady state).

For reference, here is when Argo CD presents them as options (resume is similar to promote in the UI).
https://github.com/argoproj/argo-cd/blob/master/resource_customizations/argoproj.io/Rollout/actions/discovery.lua

PR #1669, opened 2021-11-24
fix: dashboard promote buttons disabled during deploy

bpoland:
- keep promote and abort buttons enabled while deploying
- disable them at all other times
- also, only enable retry button when Rollout is degraded

Fixes https://github.com/argoproj/argo-rollouts/issues/1581

Signed-off-by: Ben Poland <ben.poland@faire.com>

Checklist:

### DIFF
--- a/ui/src/app/components/rollout-actions/rollout-actions.tsx
+++ b/ui/src/app/components/rollout-actions/rollout-actions.tsx
@@ -18,6 +18,7 @@ export const RolloutActionButton = (props: {action: RolloutAction; rollout: Roll
     const namespaceCtx = React.useContext(NamespaceContext);
 
     const restartedAt = formatTimestamp(props.rollout.restartedAt || '');
+    const isDeploying = props.rollout.status === RolloutStatus.Progressing || props.rollout.status === RolloutStatus.Paused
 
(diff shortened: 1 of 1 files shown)


ui/src/app/components/rollout-actions/rollout-actions.tsx
```diff
@@ -36,6 +37,7 @@ export const RolloutActionButton = (props: {action: RolloutAction; rollout: Roll
                 label: 'RETRY',
                 icon: 'fa-redo-alt',
                 action: api.rolloutServiceRetryRollout,
+                disabled: props.rollout.status !== RolloutStatus.Degraded,
```
harikrongali, line 40: why is this added? you can retry degraded rollout as well

bpoland: Yes I think the button should only be enabled when the rollout is degraded, right?  The double negative is a little confusing :)

harikrongali: i misread the line..yes you are correct

harikrongali (approved):

alexmt (approved):

PR #1679, opened 2021-12-02
feat: increase default QPS/Burst to 40/80. Allow values to be tunable

jessesuen:

### DIFF
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -48,6 +48,8 @@ func newCommand() *cobra.Command {
 		metricsPort          int
 		healthzPort          int
 		instanceID           string
+		qps                  float32
+		burst                int
 		rolloutThreads       int
 		experimentThreads    int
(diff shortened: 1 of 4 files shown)


huikang (approved):
LGTM

cmd/rollouts-controller/main.go
```diff
@@ -48,6 +48,8 @@ func newCommand() *cobra.Command {
 		metricsPort          int
 		healthzPort          int
 		instanceID           string
+		qps                  float32
```
agrawroh, line 51: Does it make sense to have QPS also as an `int` instead of a `float`?

huikang: I think this is to make it consistent with the client-go type: https://github.com/kubernetes/client-go/blob/d85d0d914a32b2cd96ebfd5f65f2087d6272f70c/rest/config.go#L45

rbreeze (approved):

Issue #1681, opened 2021-12-03
Promote-full with Traffic Splitting sends traffic to non-ready ReplicaSet (still)


bpoland:
## Summary

This is a follow-up to https://github.com/argoproj/argo-rollouts/issues/1580 and the discussion at the end.  I am still seeing issues with promote-full particularly when it's triggered near the middle/end of the canary steps.

Here's a timeline of a test that I just performed.  Note that 7:00 here corresponds with 12:00 in the controller log below.

7:13 - deployment initiated
- one new pod starts up with 1% of traffic
- over the next 12 minutes, we progress through the defined canary steps

7:25 - 60% of traffic is being sent to canary, corresponding with the defined canary steps

7:26 - I trigger a promote-full
- Traffic split is immediately updated to send 100% of traffic to the stable revision (which is still the old ReplicaSet)
- All of the old/stable RS pods start terminating and 4 new pods are started (I haven't seen this happen before 😟 )
- This means there are no ready pods to serve traffic and you can see nearly nothing gets through until the replacement old pods start

7:28 - replacement pods in the old RS become ready
- they start serving traffic again, but still nearly 100% of traffic is going to the old version

7:31 - all pods in the new canary RS become ready
- canary RS is promoted to stable
- all traffic is flipped immediately from old to new

Here is a graph of traffic during this test:
![image](https://user-images.githubusercontent.com/24273909/144603513-d8d0af0a-5a91-48cb-a664-7b2e60ec68ee.png)

So there are 2 problems here:
1. When using promote-full, all traffic is shifted back to the old stable RS until all the new canary RS pods start and it gets promoted to be the new stable RS.  Based on [the discussion in the previous issue](https://github.com/argoproj/argo-rollouts/issues/1580#issuecomment-974855441), the traffic split should be frozen where it is until the promotion happens (e.g. in this case, 60% of traffic going to the new RS, then jump up to 100% when the rest of the pods are ready).  As I said in the other issue: it would be awesome if instead of freezing the traffic split, it would progressively increase based on the percentage of pods available in the new RS but I understand that's more work.
2. The old/stable RS pods should not have been restarted when I used promote-full.  I have not seen this problem happen before but it's pretty scary that traffic was totally dropped from 7:26 to 7:28 while the pods came back up.  It's a good thing this was just a test environment.

[...]

Issue #1677, opened 2021-11-30
Rollout stuck in progressing status with an old revision in progressing status


huikang:
## Summary 

What happened/what you expected to happen?

The rollout is stuck in the progressing status and eventually becomes timeout.

```
Name:            app1
Namespace:       my-namespace
Status:          ◌ Progressing
Message:         waiting for all steps to complete
Strategy:        Canary
  Step:          0/3
  SetWeight:     0
  ActualWeight:  0
Images:          image:72 (canary)
                 image:v1 (stable)
Replicas:
  Desired:       3
  Current:       6
  Updated:       3
  Ready:         3
  Available:     3

NAME                                                   KIND        STATUS              AGE    INFO
⟳ app1                            Rollout     ◌ Progressing       14d    
├──# revision:4                                                                               
│  └──⧉ app1-7494db6cb6           ReplicaSet  ✔ Healthy           7m46s  canary
│     ├──□ app1-7494db6cb6-wgz8s  Pod         ✔ Running           7m44s  ready:3/3
│     ├──□ app1-7494db6cb6-pzzkx  Pod         ✔ Running           7m43s  ready:3/3
│     └──□ app1-7494db6cb6-zj4fl  Pod         ✔ Running           7m42s  ready:3/3
├──# revision:3                                                                               
│  └──⧉ app1-5679f6cfdb           ReplicaSet  ◌ Progressing       6d     stable
│     ├──□ app1-5679f6cfdb-4pjxd  Pod         ⚠ ImagePullBackOff  6d     ready:2/3
│     ├──□ app1-5679f6cfdb-hz9z4  Pod         ⚠ ImagePullBackOff  6d     ready:2/3
│     └──□ app1-5679f6cfdb-l6crl  Pod         ⚠ ImagePullBackOff  6d     ready:2/3
└──# revision:2                                                                               

```

Retrying the rollout or updatin doesn't resolve the issue. 
The only way to remedy the problem is to delete the replicaset app1-5679f6cfdb (`revision: 3`). After that, the rollout becomes healthy.
My double is that the rollout progress is blocked due to the Progressing state of the replicaset of `revision:3`.

Is there a bug in the controller that may cause this issue?

## Diagnostics

What version of Argo Rollouts are you running?

v1.1.0

```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
[...]

PR #1683, opened 2021-12-06
fix: Promote-full with TrafficRouting will dynamically increase weight according to available canary pods. Fixes: #1681

nissy34:
On promote-full calculate the weight by (100 * canaryAvailablereplicas) / Replica.
This will make sure we will not have an outage when we do promote full.

fixes: https://github.com/argoproj/argo-rollouts/issues/1681

Signed-off-by: Nissan Hefetz <nissy34@gmail.com>

### DIFF
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -129,7 +129,12 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 		} else if c.rollout.Status.PromoteFull {
 			// on a promote full, desired stable weight should be 0 (100% to canary),
 			// But we can only increase canary weight according to available replica counts of the canary.
-			desiredWeight = (100 * c.newRS.Status.AvailableReplicas) / *c.rollout.Spec.Replicas
+			// we will need to set the desiredWeight to 0 when the newRS is not available.
+			newRSAvailableReplicas := int32(0)
+			if c.newRS != nil {
+				newRSAvailableReplicas = c.newRS.Status.AvailableReplicas
+			}
+			desiredWeight = (100 * newRSAvailableReplicas) / *c.rollout.Spec.Replicas
 		} else if c.newRS == nil || c.newRS.Status.AvailableReplicas == 0 {
 			// when newRS is not available or replicas num is 0. never weight to canary
 			weightDestinations = append(weightDestinations, c.calculateWeightDestinationsFromExperiment()...)
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -130,11 +130,11 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 			// on a promote full, desired stable weight should be 0 (100% to canary),
 			// But we can only increase canary weight according to available replica counts of the canary.
 			// we will need to set the desiredWeight to 0 when the newRS is not available.
-			newRSAvailableReplicas := int32(0)
-			if c.newRS != nil {
-				newRSAvailableReplicas = c.newRS.Status.AvailableReplicas
+			if c.rollout.Spec.Strategy.Canary.DynamicStableScale {
+				desiredWeight = (100 * c.newRS.Status.AvailableReplicas) / *c.rollout.Spec.Replicas
+			} else if c.rollout.Status.Canary.Weights != nil {
+				desiredWeight = c.rollout.Status.Canary.Weights.Canary.Weight
 			}
-			desiredWeight = (100 * newRSAvailableReplicas) / *c.rollout.Spec.Replicas
 		} else if c.newRS == nil || c.newRS.Status.AvailableReplicas == 0 {
 			// when newRS is not available or replicas num is 0. never weight to canary
 			weightDestinations = append(weightDestinations, c.calculateWeightDestinationsFromExperiment()...)
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -126,6 +126,9 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 			if c.rollout.Spec.Strategy.Canary.DynamicStableScale {
 				desiredWeight = 100 - ((100 * c.stableRS.Status.AvailableReplicas) / *c.rollout.Spec.Replicas)
 			}
+		} else if c.newRS == nil || c.newRS.Status.AvailableReplicas == 0 {
+			// when newRS is not available or replicas num is 0. never weight to canary
+			weightDestinations = append(weightDestinations, c.calculateWeightDestinationsFromExperiment()...)
 		} else if c.rollout.Status.PromoteFull {
 			// on a promote full, desired stable weight should be 0 (100% to canary),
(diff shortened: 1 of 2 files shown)


rollout/trafficrouting.go
```diff
@@ -126,6 +126,15 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
...
 			}
+		} else if c.rollout.Status.PromoteFull {
+			// on a promote full, desired stable weight should be 0 (100% to canary),
+			// But we can only increase canary weight according to available replica counts of the canary.
+			// we will need to set the desiredWeight to 0 when the newRS is not available.
+			newRSAvailableReplicas := int32(0)
+			if c.newRS != nil {
+				newRSAvailableReplicas = c.newRS.Status.AvailableReplicas
+			}
+			desiredWeight = (100 * newRSAvailableReplicas) / *c.rollout.Spec.Replicas
```
jessesuen, line 137: When promote full happens for canary with traffic routing, the desired behavior is we go from the current percentage to 100% (instead of a gradual). This is to support use case for users who are using the canary strategy like a blue-green, going from 0% to 100%, because their API cannot tolerate two versions. Unfortunately, this technique will break that experience so I think we will need to find a different way to fix the bug.

nissy34: > When promote full happens for canary with traffic routing, the desired behavior is we go from the current percentage to 100% (instead of a gradual). This is to support use case for users who are using the canary strategy like a blue-green, going from 0% to 100%, because their API cannot tolerate two versions. Unfortunately, this technique will break that experience so I think we will need to find a different way to fix the bug.

Users who are using the canary strategy like a blue-green it doesn't make sense for them to use the `dynamicStableScale` feature,  as before promotion we will need the canary and stable to be 100% scaled until the rollout is fully promoted in order that when we do cut off the traffic we will not have an outage.

Therefore my suggestion is 

- if  `dynamicStableScale` is `true` then we will go with the dynamic traffic routing. 
- if `dynamicStableScale` is `false`  then leave the current behavior, the canary and stable will be 100% scaled and the traffic will go from 0% to 100%.

@jessesuen does this seem reasonable?

bpoland: If someone wants a blue-green deployment then shouldn't they be using the blue-green strategy?  If we're going from 50% to 100%, it makes sense to get as much traffic on the new version as we can (given pod readiness)

derjust: According to https://github.com/argoproj/argo-rollouts/issues/1677#issuecomment-985796488 - `promote -full` is a break-glass operation that should shift all the traffic no matter what.
This change implies that there is only a subsection of the whole traffic sent to the new canary (or maybe i'm misreading the code)?

nissy34: > According to #1677 (comment) - promote -full is a break-glass operation that should shift all the traffic no matter what.
This change implies that there is only a subsection of the whole traffic sent to the new canary (or maybe i'm misreading the code)?

Correct, the traffic will be shifted by the percentage of the available canary pods until the new canary pods will be 100% scaled then it will get 100% of the traffic, this guarantees that the new canary pods are scaled enough to handle the traffic.

jessesuen: > Therefore my suggestion is
> if dynamicStableScale is true then we will go with the dynamic traffic routing.
> if dynamicStableScale is false then leave the current behavior, the canary and stable will be 100% scaled and the traffic will go from 0% to 100%.

@nissy34 Does the bug only happen with `dynamicStableScale: true`? If so, I agree with your suggestion. In fact, your suggestion is something I considered myself but wasn't sure if we were still leaving a bug open for `dynamicStableScale: false` users. 

> If someone wants a blue-green deployment then shouldn't they be using the blue-green strategy? If we're going from 50% to 100%, it makes sense to get as much traffic on the new version as we can (given pod readiness)

There are valid reasons to accomplish blue-green using mesh/ingress provider features rather than service level cutovers because there are implications for changing the service selectors. 

For example, when speaking with AWS team, they recommended to us that blue-green be achieved using weighted target groups instead of service selector cutovers when using AWS load balancer controller. This is because service label modifications do not allow pod readiness gates to be injected properly (since aws-load-balancer-controller relies on service labels to determine the correct pods to inject).

Another reason is, when we have more advanced traffic shaping features like shadow or header-based traffic shaping, users will still want to use the canary strategy instead of blue-green, because the blue-green strategy won't be able to shadow or canary traffic based on header, while still doing a 0% to 100% cutover for production users.

nissy34: > Does the bug only happen with `dynamicStableScale: true`?

I played around with `dynamicStableScale: true` and the bug is still present.

The root cause is when it takes more than a single reconciliation until the canary replica sets are available

As at the end of the first reconcile the currentStepIndex is set to equal the number of steps  
https://github.com/argoproj/argo-rollouts/blob/master/rollout/canary.go#L353-L359
```go
	if c.rollout.Status.PromoteFull {
		c.pauseContext.ClearPauseConditions()
		c.pauseContext.RemoveAbort()
		if stepCount > 0 {
			currentStepIndex = &stepCount
		}
	}
```
And in the next reconciliation in reconcileTrafficRouting
https://github.com/argoproj/argo-rollouts/blob/master/rollout/trafficrouting.go#L134-L153

because that `!atDesiredReplicaCount && !c.rollout.Status.PromoteFull` is `false` and `*index != int32(len(c.rollout.Spec.Strategy.Canary.Steps))` is also `false` because in the prev reconcile the currentStepIndex  was set to `int32(len(c.rollout.Spec.Strategy.Canary.Steps))`

So the desiredWeight will be 0 even so we didn't yet promote the newRS to be stable ( as not all pods are ready yet).

Can we use ` c.rollout.Status.Canary.Weights.Canary.Weight` to Freeze the weigts?
if so we can do in https://github.com/argoproj/argo-rollouts/blob/master/rollout/trafficrouting.go#L132
```go
if c.rollout.Status.PromoteFull {
	// on a promote full, desired stable weight should be 0 (100% to canary),
	// But we can only increase canary weight according to available replica counts of the canary.
	// we will need to set the desiredWeight to 0 when the newRS is not available.
	if c.rollout.Spec.Strategy.Canary.DynamicStableScale {
		desiredWeight = (100 * c.newRS.Status.AvailableReplicas) / *c.rollout.Spec.Replicas
	} else {
		desiredWeight = c.rollout.Status.Canary.Weights.Canary.Weight
	}
}
```

harikrongali:
@nissy34 code looks good to me. Let me talk to @jessesuen tomorrow if there is any other use cases that are missing. But freezing the desired weight is correct until the pods are ready.

jessesuen:
@nissy34 we discussed the approach in this PR and we think it looks good. Could you rebase your PR to re-run e2e tests?

nissy34:
> @nissy34 we discussed the approach in this PR and we think it looks good. Could you rebase your PR to re-run e2e tests?

@jessesuen I rebased the PR.
(The e2e tests needs a maintainer approvel)
Thnaks

jessesuen (approved):

Issue #913, opened 2020-12-17
E2E tests timeout with `waiting for rollout spec update to be observed`


jessesuen:
## Summary 

E2E tests flake and timeout randomly with error: "waiting for rollout spec update to be observed". Example:

```
2020-12-17T05:55:07.6038186Z time="2020-12-17T05:55:07Z" level=info msg="Deleting e2e-test-name=TestRolloutAbortRetryPromote"
2020-12-17T05:55:07.9659474Z time="2020-12-17T05:55:07Z" level=info msg="rollout.argoproj.io/basic created\n" rollout=basic
2020-12-17T05:55:07.9667983Z time="2020-12-17T05:55:07Z" level=info msg="Waiting for condition: status=Healthy" rollout=basic
2020-12-17T05:55:09.5565956Z time="2020-12-17T05:55:09Z" level=info msg="Condition 'status=Healthy' met after 1s" rollout=basic
2020-12-17T05:55:09.5567086Z time="2020-12-17T05:55:09Z" level=info msg="Updated rollout pod spec: 2020-12-17T05:55:09Z" rollout=basic
2020-12-17T05:55:09.5666892Z time="2020-12-17T05:55:09Z" level=info msg="Waiting for condition: status=Paused" rollout=basic
2020-12-17T05:56:09.5706252Z     when.go:287: timeout after 1m0s waiting for condition status=Paused
2020-12-17T05:56:09.6818120Z Name:            basic
2020-12-17T05:56:09.6819518Z Namespace:       default
2020-12-17T05:56:09.6821338Z Status:          [94m◌[0m Progressing
2020-12-17T05:56:09.6826252Z Message:         waiting for rollout spec update to be observed
2020-12-17T05:56:09.6827847Z Strategy:        Canary
2020-12-17T05:56:09.6828685Z   Step:          2/2
2020-12-17T05:56:09.6829499Z   SetWeight:     100
2020-12-17T05:56:09.6830309Z   ActualWeight:  100
2020-12-17T05:56:09.6832269Z Images:          nginx:1.19-alpine ([32mstable[0m)
2020-12-17T05:56:09.6833516Z Replicas:
2020-12-17T05:56:09.6834332Z   Desired:       1
2020-12-17T05:56:09.6834890Z   Current:       1
2020-12-17T05:56:09.6835695Z   Updated:       1
2020-12-17T05:56:09.6836477Z   Ready:         1
2020-12-17T05:56:09.6837254Z   Available:     1
[...]

PR #1698, opened 2021-12-11
fix: missing rollout informer writeback

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/913

I finally encountered the e2e flakiness https://github.com/argoproj/argo-rollouts/issues/913 in local development and spent some time studying the sequence of events and logs. I believe the root cause to be a bug in the controller with the informer cache becoming stale, because we don't write back the rollout in the event where `spec.replicas` is defaulted. 

Update: this will only fix one source of e2e flakiness, but there is much more to be done


alexmt (approved):
🎉

PR #1714, opened 2021-12-15
fix: get rollout selects objects by ro's selector

### LINKED ISSUE #1711: `get` command is slow with a large number of resources
## Summary

The running time of the `kubectl argo rollouts get` command is proportional to the number of Rollouts, ReplicaSets, Pods, Experiments, Analyses in the cluster. With a large number of these resources it can make the command inconvenient to run. For example with roughly 10,000 of these resources, the command takes 40 seconds to run in our cluster.

The get command [calls `controller.GetRolloutInfo`](https://github.com/argoproj/argo-rollouts/blob/5e188f94ac5f678d27774249100078d88c8225ee/pkg/kubectl-argo-rollouts/cmd/get/get_rollout.go#L53). This method will [fetch _all_ resources mentioned above](https://github.com/argoproj/argo-rollouts/blob/5e188f94ac5f678d27774249100078d88c8225ee/pkg/kubectl-argo-rollouts/viewcontroller/viewcontroller.go#L172-L195).

## Diagnostics

```
kubectl-argo-rollouts: v1.1.0+ff3471a
  BuildDate: 2021-10-11T20:20:08Z
  GitCommit: ff3471a2fc3ccb90dbb1f370d7e399ff3064043a
  GitTreeState: clean
  GoVersion: go1.16.3
  Compiler: gc
  Platform: darwin/amd64
```

---

huikang:

close #1711 

Checklist:


huikang:
@jessesuen , I have a question about adding the `selector` filter for listing the 4 resources (rs, pod, experiment, and ar) in 

https://github.com/argoproj/argo-rollouts/blob/061efd4a527dafe74d42387ad2874ca24448b5c3/pkg/kubectl-argo-rollouts/viewcontroller/viewcontroller.go#L177-L192

I think using the `ro.spec.selector` only works for rs because the rs uses the selector labels from the rollout. For the other 3 resources, they don't have ro.spec.selector labels. For example, when I check the labels of a pod created a an analysisrun, its labels only contain

```
  labels:
    controller-uid: dbda5a43-22ca-4eb0-8edb-7617411ff171
    job-name: 23f00574-f022-4e8a-9939-6972a7ca78f2.sleep-job.1
```

To fix this performance issue, it seems we need to propagate the rollout's label to the analysisrun and it's pods. Please correct me if I miss anything. Thanks.

jessesuen:
> To fix this performance issue, it seems we need to propagate the rollout's label to the analysisrun and it's pods. Please correct me if I miss anything. Thanks.

You're right. Also, Experiment pods as well.  The issue is that there isn't a single consistent label we could query for that works for both Rollout Pods vs. AnalysisRun Pods.

For Rollout ReplicaSets and Pods, we could use `rollout-pod-template-hash` as the label selector. But for Experiment and AnalysisRun pods, we would need to introduce and inject a new label like:

```yaml
metadata:
  labels:
    rollout: foo
```

Perhaps, the AnalysisRun controller and ExperimentController should inject this `rollout: foo` label to Pods, if it sees that the owner reference to the AnalysisRun/Experiment is a Rollout.

Then the RolloutViewController would have to have two Pod listers, one based on `rollout-pod-template-hash` and the other based on `rollout`.

huikang:
@jessesuen , thanks for the suggestions. Let me try implementing this and get back to you.

huikang:
@jessesuen , after some investigation, I realized that experiment and analysisrun have already have the `rollout-pod-template-hash` label. The only objects that misses the label is the analysisrun created by an experiment in a rollout.
So the PR adds the label to those analysisrun.

jessesuen:
> I realized that experiment and analysisrun have already have the rollout-pod-template-hash label. 

I think the issue is that:

1) Pods which AnalysisRuns create (i.e. from the Job) are not labeled. And so we are forced to query *all* pods in the cluster. Also, we do not want to carry over `rollout-pod-template-hash` because this is at the discretion of the user.

2) Similarly, Experiments can be created without `rollout-pod-template-hash` if users explicitly sets labels. So for the pods created by Experiments would also be missing `rollout-pod-template-hash`.

I think the only way to have an efficient query on pods, is by injecting the rollout name into the label

huikang:
@jessesuen , thanks for your comments. Please see my replies below
> 
>     1. Pods which AnalysisRuns create (i.e. from the Job) are not labeled. And so we are forced to query _all_ pods in the cluster. Also, we do not want to carry over `rollout-pod-template-hash` because this is at the discretion of the user.

There are two reasons I think we don't need to add labels to the pods of the analysisruns:

1. The current implementation of `get rollout/experiment` doesn't visualize pods in the output

https://github.com/argoproj/argo-rollouts/blob/a74ad8eb9c3e076e40b9b35d2693f3d9c1a874f8/pkg/kubectl-argo-rollouts/info/rollout_info.go#L40

2. The job metric provider doesn't have direct access the the created pods and it seems adding selector labels to the job and its pods is not recommended by k8s (https://kubernetes.io/docs/concepts/workloads/controllers/job/#specifying-your-own-pod-selector)

Therefore, we don't need query for pods of the analysisruns.

> 
>     2. Similarly, Experiments can be created without `rollout-pod-template-hash` if users explicitly sets labels. So for the pods created by Experiments would also be missing `rollout-pod-template-hash`.
> 

The PR firsts compute the hash of the replicaset of the experiments and then uses the hash to query for the pods. 
Since the rs created by experiment always have the label `rollout-pod-template-hash` as the selector labels, we will not miss any pods in the query.

https://github.com/argoproj/argo-rollouts/blob/a74ad8eb9c3e076e40b9b35d2693f3d9c1a874f8/experiments/replicaset.go#L168-L187

> 
>      I think the only way to have an efficient query on pods, is by injecting the rollout name into the label

huikang:
@jessesuen , after some thought, I think it makes sense to introduce the rollout labels to the experiments and analysisrun created from rollout as the podTemplatehas is established. Please take another look at the PR when you have time. Thanks.

pkg/kubectl-argo-rollouts/viewcontroller/viewcontroller.go
```diff
@@ -182,26 +185,55 @@ func (c *RolloutViewController) GetRolloutInfo() (*rollout.RolloutInfo, error) {
...
+	allReplicaSets, err := c.replicaSetLister.List(selector)
 	if err != nil {
 		return nil, err
 	}
 
-	allExps, err := c.experimentLister.List(labels.Everything())
+	// Get experiments and their rs
+	selector = labels.SelectorFromSet(map[string]string{
+		v1alpha1.DefaultRolloutLabelKey:          ro.Name,
+		v1alpha1.DefaultRolloutNamespaceLabelKey: ro.Namespace,
```
jessesuen, line 202: I don't think we need `v1alpha1.DefaultRolloutNamespaceLabelKey` at all. The c.experimentLister c.rolloutLister, c.replicaSetLister are all already namespaced lister and so it wouldn't return anything outside of the current namespace.

pkg/kubectl-argo-rollouts/info/testdata/experiment-analysis/canary-analysis.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
   labels:
     rollout-type: Background
     rollouts-pod-template-hash: 6f646bf7b7
+    rollout: rollout-experiment-analysis
+    rollout-namespace: jesse-test
```
jessesuen, line 13: Can we remove these from the examples?

huikang: Done

leoluz:
@huikang We are thinking about cutting the 1.3-rc1 and this PR is listed in 1.3 milestone. Would you be able to address @zachaller 's comments so we can merge and include this in the release candidate?

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -152,6 +152,8 @@ const (
 	// LabelKeyControllerInstanceID is the label the controller uses for the rollout, experiment, analysis segregation
 	// between controllers. Controllers will only operate on objects with the same instanceID as the controller.
 	LabelKeyControllerInstanceID = "argo-rollouts.argoproj.io/controller-instance-id"
+	// DefaultRolloutLabelKey is the label of rollout name added to experiments, analysisrun
+	DefaultRolloutLabelKey string = "rollout"
```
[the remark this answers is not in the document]
leoluz: Regarding

> I also feel that just rollout is also a bit to generic.

What will be the label value? Rollout UUID? The hash? If so, I agree that the label key could be more specific with something like `rolloutUUID` for example

leoluz (changes requested):

huikang:
> What will be the label value? Rollout UUID? The hash? If so, I agree that the label key could be more specific with something like rolloutUUID for example

The label value will be the rollout name, so I think DefaultRolloutLabelKey is fine because we want to list the pods by the rollout name.

Rollout uuid has already been in those pods, e.g., https://github.com/argoproj/argo-rollouts/pull/1714/files#diff-1665985c7b9b43001c8296cb941f35ef473f29e841b29437edd2aa23ee154cedL44

zachaller:
I still think re-using `ManagedByRolloutsKey = "argo-rollouts.argoproj.io/managed-by-rollouts"` makes more sense then adding a new label that is just `rollout` it gets used the same way in other places as well https://github.com/argoproj/argo-rollouts/blob/master/rollout/trafficrouting/istio/istio.go so I was just asking why not re-use that?

huikang:
> 

I think I have updated according to the above suggestions. Please let me know if I have missed anything in the new commit.

zachaller:
@huikang not seeing any changes yet are they pushed? I still see it using https://github.com/argoproj/argo-rollouts/pull/1714/files#diff-7f320c5085c7046aa6445744c7dee388eb6fde191e315213c305c2880843f932R460 with the key value of `rollout`

huikang:
@zachaller , my bad; I thought I have pushed the code. Please take another look. Thanks.

leoluz (approved):

leoluz:
@huikang There is a unit-test failing. 
Can you pls take a look so I can merge?

```
=== FAIL: pkg/kubectl-argo-rollouts/cmd/get TestGetAbortedRollout (0.11s)
    get_test.go:560: 
        	Error Trace:	/home/runner/work/argo-rollouts/argo-rollouts/pkg/kubectl-argo-rollouts/cmd/get/get_test.go:26
        	            				/home/runner/work/argo-rollouts/argo-rollouts/pkg/kubectl-argo-rollouts/cmd/get/get_test.go:560
        	Error:      	Not equal: 
        	            	expected: "Name:            rollout-background-analysis\nNamespace:       default\nStatus:          ✖ Degraded\nMessage:         RolloutAborted: metric \"web\" assessed Failed due to failed (1) > failureLimit (0)\nStrategy:        Canary\n  Step:          0/2\n  SetWeight:     0\n  ActualWeight:  0\nImages:          argoproj/rollouts-demo:blue (stable)\nReplicas:\n  Desired:       1\n  Current:       1\n  Updated:       0\n  Ready:         1\n  Available:     1\n\nNAME                                                     KIND         STATUS        AGE  INFO\n⟳ rollout-background-analysis                            Rollout      ✖ Degraded    7d\n├──# revision:2\n│  ├──⧉ rollout-background-analysis-db976bc44            ReplicaSet   • ScaledDown  7d   canary\n│  └──α rollout-background-analysis-db976bc44-2          AnalysisRun  ✖ Failed      7d   ✖ 1\n└──# revision:1\n   └──⧉ rollout-background-analysis-7d84d44bb8           ReplicaSet   ✔ Healthy     7d   stable\n      └──□ rollout-background-analysis-7d84d44bb8-z5wps  Pod          ✔ Running     7d   ready:1/1\n"
        	            	actual  : "Name:            rollout-background-analysis\nNamespace:       default\nStatus:          ✖ Degraded\nMessage:         RolloutAborted: metric \"web\" assessed Failed due to failed (1) > failureLimit (0)\nStrategy:        Canary\n  Step:          0/2\n  SetWeight:     0\n  ActualWeight:  0\nImages:          argoproj/rollouts-demo:blue (stable)\nReplicas:\n  Desired:       1\n  Current:       1\n  Updated:       0\n  Ready:         1\n  Available:     1\n\nNAME                                                     KIND        STATUS        AGE  INFO\n⟳ rollout-background-analysis                            Rollout     ✖ Degraded    7d\n├──# revision:2\n│  └──⧉ rollout-background-analysis-db976bc44            ReplicaSet  • ScaledDown  7d   canary\n└──# revision:1\n   └──⧉ rollout-background-analysis-7d84d44bb8           ReplicaSet  ✔ Healthy     7d   stable\n      └──□ rollout-background-analysis-7d84d44bb8-z5wps  Pod         ✔ Running     7d   ready:1/1\n"
```

PR #1715, opened 2021-12-15
feat(analysis): Add Measurements Length Option for Dry-Run Metrics

agrawroh:
# Description

It could be very useful to retain the maximum data points while running an analysis in the dry-run mode to understand what caused a particular metric to error out or why the final result came out as inconclusive. 

Currently, the controller only retains the ten latest measurements and it's not possible to configure/change this value as its hard-coded in the code.

# Changes

This PR adds a new option called `measurementsLength` in the `dryRun` which can be used to retain other than the latest ten results for the metrics running in the dry-run mode. Setting this option to `0` would disable it and, the controller will revert back to the existing behavior of retaining the latest ten measurements.

# Checklist

---

### DIFF
--- a/analysis/analysis.go
+++ b/analysis/analysis.go
@@ -101,7 +101,7 @@ func (c *Controller) reconcileAnalysisRun(origRun *v1alpha1.AnalysisRun) *v1alph
 		}
 	}
 
-	err = c.garbageCollectMeasurements(run, DefaultMeasurementHistoryLimit)
+	err = c.garbageCollectMeasurements(run, DefaultMeasurementHistoryLimit, dryRunMetricsMap)
 	if err != nil {
 		// TODO(jessesuen): surface errors to controller so they can be retried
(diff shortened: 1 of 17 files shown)


jessesuen (commented):
I agree the measurement retention limit should be modifiable. But I also don't think we should limit it to dryRun. So I suggest the following changes:

1. suggest the field name to be: `measurementRetention` (which is more clear than length)
2. add this to AnalysisRun spec instead of spec.dryRun.
3. Have a similar UX for configuring the retention for different metrics. e.g.:

Retain 20 measurements for each metric:

```yaml
spec:
  measurementRetention:
  - metricName: .*
    limit: 20
```

retain 20 success-rate measurements:

```yaml
spec:
  measurementRetention:
  - metricName: success-rate
    limit: 20
```

These would apply to either dry runs or wet runs.

jessesuen:
Closing since superseded by https://github.com/argoproj/argo-rollouts/pull/1729

PR #1717, opened 2021-12-15
fix: leaderelection uses the lock in the same ns as the controller

### LINKED ISSUE #1704: Leader election lease namespace should be controller namespace
## Summary 

I noticed that the upcoming 1.2 leader election feature will try to use `kube-system` as the namespace to manage the Lease object. However, I think it should manage it in the same namespace as the controller (e.g. argo-rollouts). @huikang is there a reason `kube-system` is used as the default?

## Diagnostics

What version of Argo Rollouts are you running? v1.2 (unreleased)

---

huikang:
- by default, the lock is in argo-rollouts
- remove leader namespace from cli arguments

close #1704 

Checklist:


jessesuen (approved):

jessesuen (changes requested):
It looks like every e2e test all time out. I suspect the controller is not actually starting properly after this change for some reason.

huikang:
> It looks like every e2e test all time out. I suspect the controller is not actually starting properly after this change for some reason.

Yeah, I think it is because the e2e cluster doesn't have the argo-rollouts` ns. Pushing a new commit to create the ns in e2e.

huikang:
@jessesuen , I think the failed tests are due to flakiness.

jessesuen (approved):

PR #1724, opened 2021-12-20
feat: add measurment results to event log

### LINKED ISSUE #1719: Feat: Add query result value to logs
# Summary

I'd like to see the value of the AnalysisRun result put into the logs.  This would be similar to the AnalysisRun object in k8s, but would make it easier to analyze bugs via digging through the logs rather than relying on k8s recycling objects.  Specifically, I want the value from the query that Argo received.

# Use Cases

Tracking down bugs where Rollouts fail or succeed for unusual reasons.

---

huikang:

fix: #1719 

Checklist:


analysis/analysis.go
```diff
@@ -375,6 +375,12 @@ func (c *Controller) runMeasurements(run *v1alpha1.AnalysisRun, tasks []metricTa
 					metricResult.ConsecutiveError++
 					logger.Warnf("Measurement had error: %s", newMeasurement.Message)
 				}
+				eventType := corev1.EventTypeNormal
+				switch newMeasurement.Phase {
+				case v1alpha1.AnalysisPhaseError, v1alpha1.AnalysisPhaseFailed:
```
leoluz, line 380: why not incorporating this switch in the main one above (line 360)?

Issue #1707, opened 2021-12-13
Total replica count should be capped during a rollout


johnniee:
# Summary

According to the 'important' section [here](https://argoproj.github.io/argo-rollouts/features/canary/
) - Argo Rollouts makes the best effort to achieve the percentage listed in the last setWeight step.

> In the case where the setWeight is 15%, the Rollout attempts to get there by rounding up the calculation (i.e. the new ReplicaSet has 2 pods since 15% of 10, rounds up to 2 and the old ReplicaSet has 9 pods since 85% of 10, rounds up to 9).

Is there any other way I can achieve a situation where the rollout will always respect the number of the replica count I provided in the deployment file? Using traffic management won't help here since It's a producer-consumer platform.

# Use Cases
I'm using Argo-Rollouts for my Kafka consumers. And the number of my replicas should be equal and constant to the number of Kafka partitions. In case the rollout will exceed the number of the desired replicas in my manifest it might cause issues in my system. So it should be `maxOldRevision + newRevision == totalReplicaCount`.

---

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

johnniee:
Hi @huikang,
I created a PR that should handle this use case:
https://github.com/argoproj/argo-rollouts/pull/1743
I'll be glad to hear if you have any feedback!

Vignesh-Au:
@johnniee Can you please share an example configuration for argo rollout canary release for a kafka consumer application? I am new to argo rollout and would like to know how weight distribution happens for a kafka consumer app.

johnniee:
Hi @Vignesh-Au, here's a snippet for our canary strategy we are using for our Kafka consumers:
```yaml
canary:
  maxSurge: 0
  maxUnavailable: 1
  stableMetadata:
    labels:
      role: "stable"
      canary: "false"
  canaryMetadata:
    labels:
      role: "canary"
      canary: "true"
  steps:
  - setWeight: 1
  - pause: { }
```

The `maxSurge: 0` will make sure there won't be any 'spare' replicas during the deployment.

Vignesh-Au:
Thanks @johnniee.

PR #1743, opened 2022-01-04
feat(controller): Capping the replica count during a canary deployment 

johnniee:
This PR is following an issue we encountered during our adoption of ArgoRollouts.
There is an open issue I created here:
#1707

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -28,3 +28,4 @@ Organizations below are **officially** using Argo Rollouts. Please send a PR wit
 1. [Twilio SendGrid](https://sendgrid.com)
 1. [Ubie](https://ubie.life/)
 1. [VISITS Technologies](https://visits.world/en)
+1. [ironSource](https://is.com)
--- a/test/e2e/canary_test.go
+++ b/test/e2e/canary_test.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 package e2e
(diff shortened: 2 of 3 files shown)


liorfranko:
> 

Hi,

The use-case we're trying to solve is performing a Canary rollout to kafka consumers.
And during a Canary, the total number of pods must remain the same.

In Kafka, the number of replicas of pods (consumers) should be in correlation to the number of partitions configured on the Kafka topic.
And choosing the number of partitions on a Kafka topic should be correlated to the number of Kafka brokers in the cluster.
So changing the number of replicas of the pods (consumers) has many dependencies and it's not an easy thing to do.

I guess this change in behavior will pop up again for other companies as well.

jessesuen:
> The use-case we're trying to solve is performing a Canary rollout to kafka consumers.
> And during a Canary, the total number of pods must remain the same.

I do understand the use case with respect to kafka consumers and ensuring it does not exceed replica count. This is normally achieved by setting maxSurge to `0`.  However, I think the issue is that Argo Rollotus does not respect maxSurge if it feels like it should achieve a certain canary weight via rounding. I think there is room to improve this where we can round up but only if it does not exceed max surge.

jessesuen:
I have created PR https://github.com/argoproj/argo-rollouts/pull/1759 which reworks the basic canary calculation such that `maxSurge: 0` will now be honored. With the fix, if a user sets `maxSurge: 0` it will ensure that stable + canary replica counts do not exceed `spec.replicas`. My PR should supersede this one.

PR #1758, opened 2022-01-12
fix: Add MissingTrafficRoutingMessage validation

Eslam-mohammed-anwar:
Checklist:

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -28,3 +28,4 @@ Organizations below are **officially** using Argo Rollouts. Please send a PR wit
 1. [Twilio SendGrid](https://sendgrid.com)
 1. [Ubie](https://ubie.life/)
 1. [VISITS Technologies](https://visits.world/en)
+1. [Talabat](http://talabat.com/)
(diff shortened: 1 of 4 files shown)


jessesuen (changes requested):
Actually, there's no reason we should be enforcing this. It's a perfectly valid use case to allow canary and stable services to be defined/modified with the basic canary strategy (e.g. I want a way to only reach canary or stable pods).

I don't think this should be merged.

Eslam-mohammed-anwar:
@jessesuen Let me try to explain the downtime we have due to that
- We are using traefik ingress and we set canary and stable services without setting TrafficRouting as traefik ingress is not supported by ArgoRollout.
- The issue here is that during canary steps , ArgoRollout will inject podHash of new pod in the canary service spec.selector and old podHash in the stable service spec.selector , but the traefik ingress is only sending the traffic to stable service so all the traffic load is going to stable service and no load goes to canary at all in our case.
- In that case the traffic will increase in the old pods and old pods are terminated and then will end to a downtime until the canary steps are completed
- So if you define stable and canary you should define TrafficRouting
Thank you

PR #1772, opened 2022-01-13
chore: make ci/local codegen consistent

jessesuen:
CI system is not currently catching PRs which fail to run `make codegen`. Also, the local vs. CI codegen was not consistent or reproducible.


alexmt (approved):
LGTM

Issue #1731, opened 2021-12-26
Downtime On initial deployment


nissy34:
## Summary 

On first migration from deployment to rollouts I get a few seconds of downtime when using trafficRouting.
I assume that argo rollouts doesn't wait for the StableRS to be ready before adding the pod hash to the stable service selector.

This can be seen from the logs and events.

In the events, we can see that the switch happened before the scaling up.
```
Events:
  Type    Reason                 Age   From                 Message
  ----    ------                 ----  ----                 -------
  Normal  RolloutUpdated         25m   rollouts-controller  Rollout updated to revision 1
  Normal  NewReplicaSetCreated   25m   rollouts-controller  Created ReplicaSet vod-stitcher-f7cc77d48 (revision 1)
  Normal  CreatingCanaryIngress  25m   rollouts-controller  Creating canary ingress `vod-stitcher-vod-stitcher-canary` with weight `0`
  Normal  RolloutCompleted       25m   rollouts-controller  Rollout completed update to revision 1 (f7cc77d48): Initial deploy
  Normal  SwitchService          25m   rollouts-controller  Switched selector for service 'vod-stitcher' from '' to 'f7cc77d48'
  Normal  ScalingReplicaSet      25m   rollouts-controller  Scaled up ReplicaSet vod-stitcher-f7cc77d48 (revision 1) from 0 to 1
```
and in the logs first we have 
```
time="2021-12-26T17:03:37Z" level=info msg="Switched selector for service 'linear-stitcher-public' from '' to '78cf8cfb69'" event_reason=SwitchService namespace=dai rollout=linear-stitcher
```
and a few logs line after we see
```
time="2021-12-26T17:03:37Z" level=info msg="Scaled up ReplicaSet linear-stitcher-78cf8cfb69 (revision 1) from 0 to 1" event_reason=ScalingReplicaSet namespace=dai rollout=linear-stitcher
time="2021-12-26T17:03:37Z" level=info msg="Not finished reconciling new ReplicaSet 'linear-stitcher-78cf8cfb69'" namespace=dai rollout=linear-stitcher
```
## Diagnostics

What version of Argo Rollouts are you running? 1.1.1

# Paste the logs from the rollout controller

```
time="2021-12-26T16:03:20Z" level=info msg="rollout deleted" namespace=dai rollout=linear-stitcher
time="2021-12-26T16:03:20Z" level=info msg="syncing service" namespace=dai rollout=linear-stitcher service=linear-stitcher-canary
[...]

PR #1777, opened 2022-01-15
fix: delay service injection of selector labels until ReplicaSet available

jessesuen:
This change delays the updating of the stable service selector unless the stable replicaset is ~ready~ available. This helps ensure we don't prematurely point the stable service to nothing in the event we are taking over a Service in the migration case, potentially causing downtime.

Resolves https://github.com/argoproj/argo-rollouts/issues/1731

### DIFF
--- a/rollout/service.go
+++ b/rollout/service.go
@@ -240,16 +240,15 @@ func (c *rolloutContext) reconcileStableAndCanaryService() error {
 	if err != nil {
 		return err
 	}
-
-	if replicasetutil.IsReplicaSetReady(c.newRS) {
-		err = c.ensureSVCTargets(c.rollout.Spec.Strategy.Canary.CanaryService, c.newRS)
-		if err != nil {
-			return err
-		}
+	err = c.ensureSVCTargets(c.rollout.Spec.Strategy.Canary.CanaryService, c.newRS)
+	if err != nil {
+		return err
 	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


rollout/service.go
```diff
@@ -258,8 +257,16 @@ func (c *rolloutContext) ensureSVCTargets(svcName string, rs *appsv1.ReplicaSet)
 	if err != nil {
 		return err
 	}
-	if svc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey] != rs.Labels[v1alpha1.DefaultRolloutUniqueLabelKey] {
-		err = c.switchServiceSelector(svc, rs.Labels[v1alpha1.DefaultRolloutUniqueLabelKey], c.rollout)
+	currSelector := svc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey]
+	desiredSelector := rs.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
+	if currSelector != desiredSelector {
+		// ensure we have available replicas, otherwise we will point the service to nothing
+		if rs.Status.AvailableReplicas == 0 {
```
MarkSRobinson, line 264: If you only check against zero then you can have the situation where 1 pod is available but the service needs 100 causing an outage.

This should check that AvailableReplicas == DesiredReplicas.

jessesuen: Good catch. Will check for full availability.

alexmt (approved):
LGTM

Issue #1537, opened 2021-09-23
plugin set image does not work with workloadRef


jessesuen:
## Summary 

Spawned from https://github.com/argoproj/argo-rollouts/issues/1525. 

If using workloadRef, the plugin `set image` command does not work. We made it work for `undo` but not for set image.

## Diagnostics

What version of Argo Rollouts are you running? v1.1.0-rc1

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #1787, opened 2022-01-19
fix: plugin did not set deployment image when using workloadRef

jessesuen:
When running: `kubectl argo rollouts set image` against a Rollout which is using the workloadRef feature, the CLI does not properly set the image on the workload (the Deployment). This change detects if the rollout is referencing a Deployment, then sets the image on the referenced Deployment.

Resolves https://github.com/argoproj/argo-rollouts/issues/1537


rbreeze (approved):
LGTM!

PR #1790, opened 2022-01-19
feat(experiment): Add Measurements Retention Limit Option for Metrics

agrawroh:
# Description

Thins PR extends the `measurementRetention` functionality to the experiments.
https://argoproj.github.io/argo-rollouts/features/analysis/#measurements-retention

# Checklist

---


docs/features/analysis.md
```diff
@@ -714,7 +714,7 @@ spec:
 ## Measurements Retention
 
 !!! important
-Available since v1.2
+    Available since v1.2
```
huikang, line 717: nit: is this an intended change?

docs/features/analysis.md
```diff
@@ -804,6 +804,36 @@ spec:
         limit: 20
 ```
 
+### Measurements Retention for Experiments
+
+If an experiment wants to retain more results of its analysis metrics, it simply needs to specify the 
+`measurementRetention` field under its specs. In the following example, all the metrics from `analyze-job` matching the 
+RegEx rule `test.*` will have their latest twenty measurements get retained instead of the default ten.
```
huikang, line 811: qq: if an measurementRetention is defined both in analysistemplate and an experiment that refers that analysis template, which one takes priority?

huikang (commented):
LGTM

Just two minor questions.

docs/features/analysis.md
agrawroh: It'll be merged just like other fields (args etc.) and then validated (no duplicates etc.)

agrawroh: @huikang It will throw an exception as while merging "test.*" will get shadowed by ".*" defined in the analysis template.
Basically, the metric name RegEx should be unique for the definitions to get merged or otherwise it'll throw an exception.

huikang (approved):

PR #1792, opened 2022-01-20
fix(ui): Show container images in dashboard for rollouts with a WorkloadRef

### LINKED ISSUE #1525: Rollouts Dashboard doesn't fully support using workloadRef
## Summary 

When using a `Rollout` which uses `workloadRef`:
- the "Containers" section is not populated (it just says "Add more containers to fill this space")
- the "Rollback" buttons on previous revisions don't work -- clicking them doesn't do anything

I think either those features should be removed when using `workloadRef`, or they should be updated to work.

## Diagnostics

Recent master build

---

rbreeze:
Fixes #1525 

Signed-off-by: Remington Breeze <remington@breeze.software>

Checklist:

### DIFF
--- a/pkg/kubectl-argo-rollouts/info/rollout_info.go
+++ b/pkg/kubectl-argo-rollouts/info/rollout_info.go
@@ -89,8 +89,6 @@ func NewRolloutInfo(
 		}
 	}
 
-	
-
 	if ro.Status.RestartedAt != nil {
 		roInfo.RestartedAt = ro.Status.RestartedAt.String()
 	} else {
--- a/pkg/kubectl-argo-rollouts/info/rollout_info.go
+++ b/pkg/kubectl-argo-rollouts/info/rollout_info.go
@@ -77,16 +77,15 @@ func NewRolloutInfo(
 	roInfo.Icon = rolloutIcon(roInfo.Status)
 	roInfo.Containers = []*rollout.ContainerInfo{}
 
+	var containerList []corev1.Container
 	if workloadRef != nil {
-		for c := range workloadRef.Spec.Template.Spec.Containers {
-			curContainer := workloadRef.Spec.Template.Spec.Containers[c]
-			roInfo.Containers = append(roInfo.Containers, &rollout.ContainerInfo{Name: curContainer.Name, Image: curContainer.Image})
-		}
+		containerList = workloadRef.Spec.Template.Spec.Containers
 	} else {
-		for c := range ro.Spec.Template.Spec.Containers {
-			curContainer := ro.Spec.Template.Spec.Containers[c]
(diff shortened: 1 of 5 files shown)


pkg/kubectl-argo-rollouts/info/rollout_info.go
```diff
@@ -75,11 +76,21 @@ func NewRolloutInfo(
...
+		}
+	} else {
+		for c := range ro.Spec.Template.Spec.Containers {
+			curContainer := ro.Spec.Template.Spec.Containers[c]
+			roInfo.Containers = append(roInfo.Containers, &rollout.ContainerInfo{Name: curContainer.Name, Image: curContainer.Image})
+		}
 	}
 
+	
+
```
huikang, line 93: nit: remove extract space

pkg/kubectl-argo-rollouts/info/rollout_info.go
```diff
@@ -75,11 +76,21 @@ func NewRolloutInfo(
...
+	if workloadRef != nil {
+		for c := range workloadRef.Spec.Template.Spec.Containers {
+			curContainer := workloadRef.Spec.Template.Spec.Containers[c]
+			roInfo.Containers = append(roInfo.Containers, &rollout.ContainerInfo{Name: curContainer.Name, Image: curContainer.Image})
+		}
+	} else {
+		for c := range ro.Spec.Template.Spec.Containers {
+			curContainer := ro.Spec.Template.Spec.Containers[c]
+			roInfo.Containers = append(roInfo.Containers, &rollout.ContainerInfo{Name: curContainer.Name, Image: curContainer.Image})
+		}
```
huikang, line 89: I guess we can improve the code by

```
var containerList []corev1.container
if workloadRef != nil {
    containerList = workloadRef.Spec.Template.Spec.Containers
} else {
    containerList = ro.Spec.Template.Spec.Containers
}

for _, c := range ro.Spec.Template.Spec.Containers {
	roInfo.Containers = append(roInfo.Containers, &rollout.ContainerInfo{Name: c.Name, Image: c.Image})
}
```

rbreeze: Ah, I did try this, but then there is a panic at `for _, c := range ro.Spec.Template.Spec.Containers {` if using a workloadRef, because the rollout has no containers.

I also considered setting a "workload" variable to either the rollout or deployment depending on if using workloadRef, but the issue there is they are not the same type.

huikang: My mistake. The for loop should be `for _, c := range containerList {`. WDYT?

rbreeze: Ah I see good suggestion, changing now! Thanks

jessesuen (approved):
LGTM!

Issue #1747, opened 2022-01-05
kubectl-argo-rollouts panic while watching progress


bpoland:
## Summary 

We were running the kubectl-argo-rollouts plugin to wait for a rollout to complete.  Right near the end of the rollout, it paniced even though the rollout had completed successfully.

The script waits for two rollouts to complete, e.g.
```
kubectl argo rollouts status rollout1
kubectl argo rollouts status rollout2
```

So I'm not 100% sure if it was at the end of watching rollout1 that it happened, or when it tried to watch rollout2.

This has only happened once and isn't a huge issue, but I just thought I'd share in case something jumps out at you.  Let me know if I can provide any other information

## Diagnostics

What version of Argo Rollouts are you running?

Rollout controller 1.1.1

Plugin was just the latest from a few months ago:
```
 BuildDate: 2021-10-11T20:19:55Z
  GitCommit: ff3471a2fc3ccb90dbb1f370d7e399ff3064043a
  GitTreeState: clean
  GoVersion: go1.16.3
  Compiler: gc
  Platform: linux/amd64
```

Plugin output: https://gist.github.com/bpoland/020e37d181115abd43d4c2c42bd61767

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

Temikus:
Can confirm this started happening only after upgrading from `1.1.0` to `1.1.1`

Temikus:
One more addition here - something is wrong with the controller as well. It looks like to be a combination of things. 1.1.0 Kubectl + 1.1.1 controller seem to have the same issue.

```
Paused - CanaryPauseStep
--
  | Paused - CanaryPauseStep
  | Paused - CanaryPauseStep
  | Progressing - waiting for all steps to complete
  | Healthy
  | E0118 17:58:27.419321     245 runtime.go:78] Observed a panic: "send on closed channel" (send on closed channel)
  | goroutine 42 [running]:
  | k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1c95ea0, 0x2897310)
  | /go/pkg/mod/k8s.io/apimachinery@v0.21.0-alpha.0/pkg/util/runtime/runtime.go:74 +0x95
  | k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
[...]

PR #1796, opened 2022-01-20
fix: plugin panic while watching progress

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/1747

Fixes a panic introduced in v1.1.0 plugin which closed a channel prematurely while the sender (view controller) was still sending on it. This change prevents this by deferring the channel close, and also niling out the view controller's callback list when the view controller stops, so that it will not attempt to perform any callbacks when finished.

As part of this fix, I made the plugin less chatty and only showed status changes. Previously, it would print out the phase and message on every rollout change, even if it was the same as before. Example output:

```
$ kubectl-argo-rollouts status my-rollout -w
Progressing - more replicas need to be updated
Paused - CanaryPauseStep
Progressing - more replicas need to be updated
Progressing - updated replicas are still becoming available
Healthy
```


rbreeze (approved):
LGTM

PR #1798, opened 2022-01-21
fix: client can detect if rollout is in the process of unpausing

jessesuen:
While debugging e2e flakiness, I determined that **_some_** of the flakiness is actually caused by a bug where the client does not properly detect that the rollout is still in the process of being unpaused. This is probably a rare event but is exacerbated when the client is aggressively monitoring rollout status (such as in e2e tests which are performing watches against the rollout). The scenario is this:

1. rollout becomes automatically paused (e.g. canary pause step). Status becomes:

```yaml
status:
  phase: Paused
  pauseConditions:
  - reason: CanaryPauseStep
  controllerPause: true
```

2. client (plugin) unpauses the rollout (by nilling out pauseConditions). Status becomes:

```yaml
status:
  phase: Paused
  controllerPause: true
```

3. Before the controller has time to update status to `Progressing`, the client examines the rollout and sees that the phase is `Paused`, despite the fact that the controller is unpausing the rollout and hasn't yet updated the status. In this situation checking `metadata.generation` and `status.observedGeneration` does not help because Kubernetes status changes to not increment `metadata.generation`.

To fix this, we need a client side check similar to `observedGeneration` such that the client looks at both the `controllerPause` and `pauseConditions` to detect if the controller has not yet reacted to the unpause, and then considers the rollout as `Progressing` despite the `status.phase` being `Paused`.

### DIFF
--- a/pkg/apiclient/rollout/rollout.swagger.json
+++ b/pkg/apiclient/rollout/rollout.swagger.json
@@ -1300,7 +1300,7 @@
           "items": {
             "$ref": "#/definitions/github.com.argoproj.argo_rollouts.pkg.apis.rollouts.v1alpha1.PauseCondition"
           },
-          "title": "PauseConditions indicates why the rollout is currently paused"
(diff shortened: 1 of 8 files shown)


huikang (approved):
Great findings!

Issue #1355, opened 2021-07-16
Dashboard: display the scaledown delay second if a replicaset is scheduled to be scaled down


huikang:
# Summary

What change needs making?

The argo-rollouts UI displays the scaledown delay second if a replicaset is scheduled to be scaled down.

# Use Cases

When would you use this?

Some useful information for user to understand the replicaset status.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

schakrad:
The delay button is added on the dashboard to display the scaleDownDelay associated with the scaled down revision.

PR #1804, opened 2022-01-25
feat: Added delay button in the scaled down revision (#1355)

schakrad:
fix for https://github.com/argoproj/argo-rollouts/issues/1355

Before the addition of button.
Screen Shot 2022-02-07 at 11 52 30 PM

After the addition of button to show the scaleDownDelay in seconds
Screen Shot 2022-02-08 at 12 00 49 AM

Signed-off-by: “schakradari” <“saisindhu_chakradari@intuit.com”>

Checklist:


harikrongali:
@schakrad can you cleanup the PR and have diff show only required changes

ui/package.json
```diff
@@ -60,7 +60,7 @@
         "sass": "^1.32.8",
         "ts-loader": "^8.0.17",
         "webpack-bundle-analyzer": "^4.4.0",
-        "webpack-cli": "^4.5.0",
```
harikrongali, line 63: is this change required?

ui/yarn.lock
```diff
@@ -2256,22 +2256,22 @@
     "@webassemblyjs/wast-parser" "1.9.0"
     "@xtuc/long" "4.2.2"
 
-"@webpack-cli/configtest@^1.0.3":
```
harikrongali, line 2259: not sure if lock file needs to be committed. @alexmt is it fine to commit this file?

harikrongali:
Please add before and after snapshots as a part of the PR description.

Please fix CI checks: DCO & build
For DCO, you can squash commits and sign off on the last one

ui/package.json
rbreeze: +1

ui/yarn.lock
rbreeze: @harikrongali Lock file does need to be committed, but we should avoid bumping versions in any of these packages if possible. @schakrad I don't see any code changes that should require updating package versions, can you confirm?

rbreeze (changes requested):
Code changes look good, my only request is to avoid changing package versions

ui/yarn.lock
harikrongali: Thanks, @rbreeze for confirmation. @schakrad please revert the package.json update

schakrad: Sure Hari, I will check and revert back.

harikrongali (approved):

ui/src/app/components/Timer/timer.tsx
```diff
@@ -0,0 +1,27 @@
...
+           setTime(start);
+       }, 1000);   
+       return () => {
+         if(timer) {
+            clearInterval(timer);
+         }
+       }
+   }, [props.date]);
+   return (
+      time === 0  ? null: <ActionButton label={`delay: ${time}s`} />
```
alexmt, line 25: Looks like this component is not just Timer. It has a button and `delay` label so it should be renamed accordingly. 

The `argo-ui` has `Ticker` component. So might consider just using it and "inline" implementation into `pods.tsx`. E.g.: 


pods.tsx
```tsx
import {Ticker} from 'argo-ui';

...
  <Ticker>
  {now => {
    // calculate time here
    return time === 0 ? null: <ActionButton label={`delay: ${time}s`} />;
  }}
...
```

alexmt (changes requested):

ui/src/app/components/Timer/timer.tsx
schakrad: Thanks Alex, I will check and update.

ui/src/app/components/pods/pods.tsx
```diff
@@ -107,6 +110,16 @@ export const ReplicaSet = (props: {rs: RolloutReplicaSetInfo; showRevision?: boo
                 <ThemeDiv className='pods__header'>
                     <span style={{marginRight: '5px'}}>{rsName}</span> <ReplicaSetStatusIcon status={props.rs.status as ReplicaSetStatus} />
                     {props.showRevision && <div style={{marginLeft: 'auto'}}>Revision {props.rs.revision}</div>}
+                    {props.rs.scaleDownDeadline && (
+                        <div style={{marginLeft: 'auto'}}>
+                            <Ticker>
+                                {(now) => {
+                                    const time = moment(props.rs.scaleDownDeadline).diff(now, 'second');
+                                    return time === 0 ? null : <ActionButton label={`delay: ${time}s`} />;
```
alexmt, line 118: I don't think it makes sense to use `ActionButton` here because user cannot really click anything and trigger any action. Can we use info item instead:

image

revision.tsx
```typescript
const ImageItems = (props: {images: ImageInfo[]}) => {
    return (
        <div>
            {props.images.map((img) => {
                let imageItems = img?.tags?.map((t) => {
                    return {content: t, icon: IconForTag(t)} as InfoItemProps;
                });
                if (imageItems.length === 0) {
                    imageItems = [];
                }
                imageItems.push({content: <Tooltip content='Scaledown in 3 hours 5 mins'><span>3 hours 5 mins</span></Tooltip>, icon: 'fa fa-clock'} as any);
                return <InfoItemRow key={img.image} label={<ThemeDiv className={`image image--${img.color || 'unknown'}`}>{img.image}</ThemeDiv>} items={imageItems} />;
            })}
        </div>
    );
};

```

alexmt (changes requested):

alexmt (approved):

alexmt:
LGTM. 


@rbreeze can you please take a look one more time?

rbreeze (changes requested):
Just one last comment, can we revert `package.json` and `yarn.lock` changes?

alexmt:
@rbreeze changes in package.json are actually required. Apparently `argo-ui`/`argo-cd/ui` and `argo-rollouts/ui` were using different `classnames` versions. I temporary downgraded `classnames` to unblock this PR.

rbreeze:
@alexmt Oh I see, thank you!

rbreeze (approved):
LGTM!

PR #1818, opened 2022-01-28
feat: Add logformat flag to rollouts-controller

Karol-Szlachta-DM:
Checklist:

### DIFF
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	smiclientset "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/split/clientset/versioned"
@@ -36,7 +37,9 @@ import (
 
 const (
 	// CLIName is the name of the CLI
-	cliName = "argo-rollouts"
+	cliName    = "argo-rollouts"
+	jsonFormat = "json"
+	textFormat = "text"
 )
 
 func newCommand() *cobra.Command {
@@ -44,6 +47,7 @@ func newCommand() *cobra.Command {
 		clientConfig         clientcmd.ClientConfig
 		rolloutResyncPeriod  int64
 		logLevel             string
+		logFormat            string
 		klogLevel            int
 		metricsPort          int
 		healthzPort          int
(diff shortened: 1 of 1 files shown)


RaviHari (approved):
LGTM

cmd/rollouts-controller/main.go
```diff
@@ -267,6 +269,22 @@ func setLogLevel(logLevel string) {
...
+	var formatType log.Formatter
+	switch strings.ToLower(logFormat) {
+	case jsonFormat:
+		formatType = &log.JSONFormatter{}
+	case textFormat:
+		formatType = &log.TextFormatter{
+			FullTimestamp: true,
+		}
+	default:
+		formatType = &log.TextFormatter{}
```
jessesuen, line 282: Logrus has its own default behavior to set the output to json or text or depending on a TTY. So changing the default explicitly to text will cause a change in behavior to everyone. I think it's best **not** to set a logformat in the default case and rely on logrus' default behavior.

cmd/rollouts-controller/main.go
```diff
@@ -213,6 +214,7 @@ func newCommand() *cobra.Command {
 	command.Flags().Int64Var(&rolloutResyncPeriod, "rollout-resync", controller.DefaultRolloutResyncPeriod, "Time period in seconds for rollouts resync.")
 	command.Flags().BoolVar(&namespaced, "namespaced", false, "runs controller in namespaced mode (does not require cluster RBAC)")
 	command.Flags().StringVar(&logLevel, "loglevel", "info", "Set the logging level. One of: debug|info|warn|error")
+	command.Flags().StringVar(&logFormat, "logformat", "info", "Set the logging format. One of: text|json")
```
jessesuen, line 217: This is set to `info` which is not a valid option. I think default should be empty string so that behavior is whatever logrus' default behavior is.

jessesuen (changes requested):

cmd/rollouts-controller/main.go
Karol-Szlachta-DM: Good catch! I have set it to an empty string.

cmd/rollouts-controller/main.go
Karol-Szlachta-DM: I have implemented the suggested change.

jessesuen (approved):
Nice improvement!

Issue #1755, opened 2022-01-10
Adding InfluxDB as a metrics provider for Argo Rollouts


jaymebrd:
# Summary

Add influxDB as a metrics provider to Argo Rollouts

# Use Cases

InfluxDB is a popular open source TSBD, it's make a great addition to the current metric providers available.

I already have a basic implementation of this, and if you're happy I'll go ahead and clean it up and make a PR for it.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

perenesenko:
Hi @jaymebrd 
Please go ahead with your PR.

jaymebrd:
Thanks @perenesenko I've created a PR for this.

jaymebrd:
Hi @jessesuen @huikang - Just wanted to reach out again, would love to get the above PR in soon if possible - would appreciate reviews

jaymebrd:
This has been implemented, merged and released so I think we can close this.

PR #1839, opened 2022-02-02
feat: add support for influxdb as a metrics provider

jaymebrd:
This PR addresses the enhancement proposal here #1755 - to add InfluxDB, a popular time series database as a metrics to provider for Argo Rollouts.

Checklist:


metricproviders/influxdb/influxdb.go
```diff
@@ -0,0 +1,135 @@
...
+	"errors"
+	"fmt"
+	"github.com/argoproj/argo-rollouts/utils/defaults"
+	"github.com/argoproj/argo-rollouts/utils/evaluate"
+	influxdb2 "github.com/influxdata/influxdb-client-go/v2"
+	influxapi "github.com/influxdata/influxdb-client-go/v2/api"
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"time"
```
huikang, line 14: Could you format the imports to resolve the link error?

huikang (commented):
could you run `make codegen` to resolve the failed CI job? Thanks.

jaymebrd:
> could you run `make codegen` to resolve the failed CI job? Thanks.

Thanks for the review @huikang - I've fixed the codegen and goimport format issues.

metricproviders/influxdb/influxdb.go
```diff
@@ -0,0 +1,136 @@
...
+func (p *Provider) processResponse(metric v1alpha1.Metric, result *influxapi.QueryTableResult) (string, v1alpha1.AnalysisPhase, error) {
+	var res []interface{}
+	if result == nil {
+		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("no results returned from flux query")
+	}
+	for result.Next() {
+		res = append(res, result.Record().Value())
+	}
+	if len(res) == 0 {
+		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("no results returned from flux query")
```
huikang, line 98: Shall we distinguish the error message from `result==nil` (line 91)?

huikang (dismissed):
LGTM

huikang (commented):
Could you resolve the security issue report by 
`security/snyk (Argoproj) `

jaymebrd:
> Could you resolve the security issue report by `security/snyk (Argoproj) `

Thanks @huikang - I can't access the `security/snyk (Argoproj)` results to see what the issue is - could you post here what's failing? I clicked through and hit the request access button incase that's possible.

jessesuen:
It's complaining about `github.com/deepmap/oapi-codegen`. I've attached the screenshot

image

docs/analysis/influxdb.md
```diff
@@ -0,0 +1,41 @@
...
+  args:
+  - name: application-name
+  metrics:
+  - name: error-rate
+    # NOTE: To be consistent with the prometheus metrics provider InfluxDB query results are returned as an array.
+    # In the example we're looking at index 0 of the returned array to obtain the value we're using for the success condition
+    successCondition: result[0] <= 0.01
+    provider:
+      influxdb:
+        profile: my-influxdb-secret  # optional, defaults to 'influxdb'
```
jessesuen, line 20: Do we imagine the need for multiple influxDB profiles within the same cluster? And would different users of the cluster specify different influxdb profiles?

My thought process is that the argo-rollouts admin would configure a single influxdb at a system-level, used by all rollouts in the cluster. In which case, the end-users would never need to specify a profile in their templates and we could eliminate this field in the analysis template completely.

jaymebrd:
> It's complaining about `github.com/deepmap/oapi-codegen`. I've attached the screenshot
> 
> image

Thanks @jessesuen - Looks like an issue with a transitive dependency - `moq` is a dependency of `github.com/deepmap/oapi-codegen` which is a dependency of the Influx Go client, so tricky to update.

I've made a PR on the upstream project to update it https://github.com/deepmap/oapi-codegen/pull/516  - would this be a blocker until resolved?

docs/analysis/influxdb.md
jaymebrd: The use case I had in mind is that you may have multiple orgs that you want to query within a single Influx database, or multiple Influx cloud accounts that you may want to query from a single cluster - It's probably not necessary as each cluster rollouts is running in would likely have a single org/database to query - happy to simplify it there.

jessesuen: Thanks for explaining the use case. I'm asking because we made a misstep with the prometheus provider by making the prometheus server address a required field in the template (which is not something dev teams are typically concerned with) as opposed to a central configuration. And so I wanted to avoid a similar mistake here. I actually just filed an issue to allow the prometheus address to be optional/centrally defined: https://github.com/argoproj/argo-rollouts/issues/1846.

When comparing the metric provider integrations there are a couple of different approaches of where connection information is configured:

Prometheus: analysis template level (too fined grain)
DataDog: single central definition
NewRelic: multiple central definitions (with a default)
Wavefront: multiple central definitions (with no default)
CloudWatch: single central definition

Your implementation is like NewRelic, which IMO the most flexible, so let's stick with your approach.

jessesuen:
> Thanks @jessesuen - Looks like an issue with a transitive dependency - moq is a dependency of github.com/deepmap/oapi-codegen which is a dependency of the Influx Go client, so tricky to update.

Can't we fix this with a `require` statement in the go.mod, forcing the transitive dependency to be at a higher version?

jaymebrd:
> > Thanks @jessesuen - Looks like an issue with a transitive dependency - moq is a dependency of github.com/deepmap/oapi-codegen which is a dependency of the Influx Go client, so tricky to update.
> 
> Can't we fix this with a `require` statement in the go.mod, forcing the transitive dependency to be at a higher version?

I've ended up doing a `go mod edit -replace` to force the new version, trying to just go get -u a specific version and using the `require` in `go.mod` just ends up being removed by `go mod tidy`. I'm no expert here, but I think this is the right way to force it.

metricproviders/influxdb/influxdb.go
jaymebrd: Done here 878fe31196303ca9fe97f4e7b5321d7c4cade27b

jessesuen:
> I've ended up doing a go mod edit -replace to force the new version

Oops! I said `require` but I meant to say `replace`. Glad you worked it out.

docs/analysis/influxdb.md
jaymebrd: That makes sense - thanks for the context and yeah I modelled after NewRelic. I also found it to be the most flexible - so I'll leave it as is.

jaymebrd:
@jessesuen no rush, but I think this is ready for review again

jaymebrd:
@jessesuen @huikang Just checking in - I think this PR is good to go, let me know if you want me to do anything else to get this in.

jessesuen:
Sorry for the delay. I'll take a look again

jaymebrd:
> Sorry for the delay. I'll take a look again

No problem @jessesuen, I'm merging master in the branch periodically to make sure it doesn't get out of date

zachaller:
This PR looks good overall some minor changes I think worth doing are just getting some of the sonarcloud duplication changes a little lower. I think if you add a function like this to the test case then called it instead of duplicating it.

```
func setupProvider(t *testing.T, csvTable string, query string) *Provider {
	e := log.Entry{}
	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte(csvTable))
	}))
	defer ts.Close()

	client := influxdb2.NewClient(ts.URL, "x")
	pt := client.QueryAPI("fake-org")
	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
	defer cancel()
	result, err := pt.Query(ctx, query)
	if err != nil {
		t.Fatal(err)
	}
	mock := &mockAPI{response: result}
	p := NewInfluxdbProvider(mock, e)
	return p
}
```

It would clean up a lot of the test duplication other than that it LGTM.

jaymebrd:
> This PR looks good overall some minor changes I think worth doing are just getting some of the sonarcloud duplication changes a little lower. I think if you add a function like this to the test case then called it instead of duplicating it.
> 
> ```
> func setupProvider(t *testing.T, csvTable string, query string) *Provider {
> 	e := log.Entry{}
> 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
> 		w.Write([]byte(csvTable))
> 	}))
> 	defer ts.Close()
> 
> 	client := influxdb2.NewClient(ts.URL, "x")
> 	pt := client.QueryAPI("fake-org")
> 	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
> 	defer cancel()
> 	result, err := pt.Query(ctx, query)
> 	if err != nil {
> 		t.Fatal(err)
> 	}
> 	mock := &mockAPI{response: result}
> 	p := NewInfluxdbProvider(mock, e)
> 	return p
> }
> ```
> 
> It would clean up a lot of the test duplication other than that it LGTM.

Thanks @zachaller for the review! I actually added an upstream change https://github.com/influxdata/influxdb-client-go/pull/304 to the influx go client to expose an exported constructor to simplify the mocking in the tests, that was recently released so I've gone ahead an updated the PR to use that. This removes that duplicated code completely.

huikang (approved):

harikrongali:
@jaymebrd can you please rebase your PR against master so that we can merge?

zachaller (approved):

leoluz (approved):
I see all comments addressed and 2 approvals.
Will merge once the build is green.

@jaymebrd Tks for the contribution!

jaymebrd:
> I see all comments addressed and 2 approvals. Will merge once the build is green.
> 
> @jaymebrd Tks for the contribution!

Thanks @[harikrongali](https://github.com/harikrongali) and @[leoluz](https://github.com/leoluz)! It looks like a flaky E2E test failed, do we want to re-run it until we get a green e2e build?

harikrongali:
not required as we are aware of an issue with failed test.

PR #1907, opened 2022-03-11
feat: Traefik support. Fixes #516

### LINKED ISSUE #516: Traefik Support
In addition to the current controllers, it would be great to support Traefik!

The [TraefikService](https://docs.traefik.io/routing/providers/kubernetes-crd/#kind-traefikservice) object supports the ability for weighted round robin load balancing and traffic mirroring when using Traefik as ingress.

As a quick pass, an `IngressRoute` is configured to send traffic to a `TraefikService`, which then provides configuration for sending traffic to one or more Kubernetes `Service` objects. While I haven't dived into Argo's architecture, I imagine it wouldn't be too complicated to own and manage the TraefikService object.

While I may not be the best to contribute code, I could try or help with docs and testing the implementation. Or if someone wants to chat on Slack and give me direction on how/where to get started, I can take a crack at it.

PhilippPlotnikov:
Fixes #516

Checklist:


harikrongali:
@zachaller Please review

pkg/apis/rollouts/v1alpha1/types.go
```diff
@@ -361,6 +361,13 @@ type RolloutTrafficRouting struct {
 	Ambassador *AmbassadorTrafficRouting `json:"ambassador,omitempty" protobuf:"bytes,5,opt,name=ambassador"`
 	// AppMesh holds specific configuration to use AppMesh to route traffic
 	AppMesh *AppMeshTrafficRouting `json:"appMesh,omitempty" protobuf:"bytes,6,opt,name=appMesh"`
+	// Traefik holds specific configuration to use Traefik to route traffic
+	Traefik *TraefikTrafficRouting `json:"traefik,omitempty" protobuf:"bytes,7,opt,name=traefik"`
+}
+
+// TraefikTrafficRouting defines the configuration required to use Traefik as traffic router
+type TraefikTrafficRouting struct {
+	Service string `json:"service" protobuf:"bytes,1,name=service"`
```
leoluz, line 370: I think we should make it more evident that this expects a `TraefikService` name and not a Kubernetes service. Proposal:
```go
TraefikServiceName string `json:"traefikService" protobuf:"bytes,1,name=service"`
```
Please add field documentation as well.

leoluz:
Thank you for the PR!
Traefik support would require adding documentation like what we have for all existing traffic routers. Are you planning to add the docs in this PR?

PhilippPlotnikov:
> Hi, I think add documentation in seperate PR

harikrongali:
@PhilippDeveloper Norm that is followed is to add documentation along with features as it is easy to review based on the documentation.

PhilippPlotnikov:
@harikrongali ok

rollout/trafficrouting/traefik/traefik.go
```diff
@@ -0,0 +1,164 @@
...
+	return nil
+}
+
+func (r *Reconciler) SetWeight(desiredWeight int32, additionalDestinations ...v1alpha1.WeightDestination) error {
+	ctx := context.TODO()
+	rollout := r.Rollout
+	traefikServiceName := rollout.Spec.Strategy.Canary.TrafficRouting.Traefik.TraefikServiceName
+	traefikService, err := r.Client.Get(ctx, traefikServiceName, metav1.GetOptions{})
+	if err != nil || traefikService == nil {
+		return err
```
leoluz, line 75: If `traefikService == nil` it will return a nil error for this exception. It's better to handle this error separately and check for the proper error type in case it isn't found. 
Example: `k8s.io/apimachinery/pkg/api/errors.IsNotFound(err)`

rollout/trafficrouting/traefik/traefik.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	canaryServiceName := rollout.Spec.Strategy.Canary.CanaryService
+	stableServiceName := rollout.Spec.Strategy.Canary.StableService
+	services, isFound, err := unstructured.NestedSlice(traefikService.Object, "spec", "weighted", "services")
+	if err != nil || !isFound {
+		return err
+	}
+	canaryService, err := getService(canaryServiceName, services)
+	if err != nil || canaryService == nil {
```
leoluz, line 84: if `canaryService == nil` it will return a nil error. Better handle this case separately and return proper error messages.

rollout/trafficrouting/traefik/traefik.go
```diff
@@ -0,0 +1,164 @@
...
+
+func getService(serviceName string, services []interface{}) (map[string]interface{}, error) {
+	var selectedService map[string]interface{}
+	for _, service := range services {
+		typedService, ok := service.(map[string]interface{})
+		if !ok {
+			return nil, errors.New("Failed type assertion setting weight for traefik service")
+		}
+		nameOfCurrentService, isFound, err := unstructured.NestedString(typedService, "name")
+		if err != nil || !isFound {
```
leoluz, line 115: if `!isFound` it will return a nil error. Better handle this case separately and return proper error messages.

rollout/trafficrouting/traefik/traefik.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	err = unstructured.SetNestedField(stableService, int64(100-desiredWeight), "weight")
+	if err != nil {
+		return err
+	}
+	err = unstructured.SetNestedSlice(traefikService.Object, services, "spec", "weighted", "services")
+	if err != nil {
+		return err
+	}
+	_, err = r.Client.Update(ctx, traefikService, metav1.UpdateOptions{})
```
leoluz, line 103: Consider sending events for exceptional cases while interacting with kube API.
[Example](https://github.com/argoproj/argo-rollouts/blob/4333b4f08d613d881dc891170e23774df248c7ce/rollout/trafficrouting/ambassador/ambassador.go#L174-L179)

rollout/trafficrouting/traefik/traefik.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+// Type holds this controller type
+const Type = "Traefik"
+
+var (
+	apiGroupToResource = map[string]string{
+		"traefik.containo.us": "traefikservices",
```
leoluz, line 22: You can use the `defaults.DefaultTraefikAPIGroup` for the key. It would be good to have the value (`traefikservices`) defined as a constant as well

leoluz (changes requested):
Nice PR!
Please address the problems with error handling.

.github/workflows/go.yml
```diff
@@ -20,8 +20,10 @@ jobs:
         uses: actions/setup-go@v3
```
leoluz, line 20: There is no real change in this file. Can you please revert?

PhilippPlotnikov: I changed this file as it is the fix of lint. (I couldnt pass it)

rollout/trafficrouting/traefik/traefik.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		if nameOfCurrentService == serviceName {
+			selectedService = typedService
+			break
+		}
+	}
+	return selectedService, nil
+}
+
+func (r *Reconciler) VerifyWeight(desiredWeight int32, additionalDestinations ...v1alpha1.WeightDestination) (*bool, error) {
```
perenesenko, line 126: Probably there is no need to implement this method in that way. You're accessing the same traefic service object (which you just modified in a SetWeight method)

Right now it's implemented for ALB only. The reason whay we have this implementation for the ALB is delays in traffic shifting.
For the ALB once you modified the Ingress object it take some time to really shift the traffic by creating/modifying target group. So here we need to make validation against the target group objects to see the traffic was really updated.

perenesenko (commented):
have one comment for the VerifyWeight

rollout/trafficrouting/traefik/traefik.go
PhilippPlotnikov: I checked that if it cant find the needing resource it return corresponding error so I think I can remove the check traefikService

leoluz (approved):
LGTM

perenesenko (approved):
LGTM

ayanhamza:
@PhilippPlotnikov Hello !
I am using Traefik ingress for Canary Deploy and used kind TraefikService for trafficRouting. Everything works fine when i start deploy canary release. Traffic for my services splits by weight for Canary pods by percentage. BUT when Canary version of my app is Degraded (in case of failure analysis) it's rollback to old version and requests for my Stable version not go. Because argo rollout after Degraded don't sends traffic via canary service.  How to modify code to send traffic via canary service also after Degraded?? 
Cause in IngressRoute object of Traefik I declare in spec service "TraefikService". And TraefikService dont't work's properly when one of the services can't receive traffic to pods.


```
  strategy:
    canary:
      canaryService: colorapi-canary
      stableService: colorapi-stable
      trafficRouting:
        traefik:
          weightedTraefikServiceName: colorapi
      analysis:
        templates:
        - templateName: web
      steps:
      - setWeight: 10
      - pause:
          duration: 50s
      - setWeight: 50
      - pause:
          duration: 50s
```

```
apiVersion: traefik.containo.us/v1alpha1
kind: TraefikService
metadata:
  name: colorapi
  namespace: devops-services
spec:
  weighted:
    services:
      - name: colorapi-canary
        port: 5000
      - name: colorapi-stable
        port: 5000
```
```
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  annotations:
    kubernetes.io/ingress.class: traefik
  creationTimestamp: null
  name: colorapi
  namespace: devops-services
spec:
  entryPoints: []
  routes:
  - kind: Rule
    match: Host("colorapi.k8s-dev.kar-tel.local") && PathPrefix("/")
    middlewares: []
    priority: 0
    services:
    - kind: TraefikService
      name: colorapi
```

ayanhamza:
I found that after "Degraded" selector of kind:Service canary not return to old ReplicaSet Hash. And still remains in new ReplicaSet Hash. That's why canary-service doesn't have endpoints. And TraefikService not serve traffic to stable version,
Please help. How to modify code to change Hash of selector to old ReplicaSet if new canary version of app don't updated and Degraded ?

PR #1911, opened 2022-03-15
chore: fix ui-proto make target

crenshaw-dev:
Looks like the ui-proto make target has been broken for a while and api.ts is pretty badly out of date.


zachaller:
Already fixed in master

PR #1919, opened 2022-03-22
chore: Improve image build speed

34fathombelow:

Disabled cache, it is configured to only cache the export layers that are in the final build stages.  Not taking advantage of the cache on disk for the second run of the `kubectl-argo-rollouts` image  This will reduce build times by **half**.  I recommend a second PR to cleanup the existing cache steps from the docker-publish.yml as they are not needed anymore.

I have tested these builds locally.  I'm expecting the complete Workflow should be complete in approximately 20-25 minutes on GitHub runners.

The flowchart below shows how the Dockerfile is processed using cross-compilation. Orange nodes are x86_64 and opaque orange nodes are arm64.

```mermaid
flowchart TB

id1([builder]) --> id2([argo-rollouts-ui]) --> id3([argo-rollouts-build]) --> id4([kubectl-argo-rollouts]) --> id5([x86_64 Final_image])
id3([argo-rollouts-build]) --> id6([kubectl-argo-rollouts]) --> id7([arm64 Final_Image])


style id6 fill:#5C74C6,stroke:#333,stroke-width:2px,color:#ffffff
classDef x86 fill:#EF7B4D,stroke:#EF7B4D,stroke-width:2px,color:#ffffff
classDef arm64 fill:#FCE5DB,stroke:#EF7B4D,stroke-width:2px,color:#EF7B4D
class id1,id2,id3,id4,id5 x86
class id6,id7,id8 arm64

```

Checklist:

* [x] Either ~~(a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or~~ (c) this is a chore.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -87,8 +87,6 @@ jobs:
           platforms: ${{ steps.platform-matrix.outputs.platform-matrix }}
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.controller-meta.outputs.tags }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache
 
       - name: Build and push (plugin-image)
         uses: docker/build-push-action@v2
@@ -97,8 +95,6 @@ jobs:
           platforms: ${{ steps.platform-matrix.outputs.platform-matrix }}
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.plugin-meta.outputs.tags }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache-new
 
         # Temp fix
         # https://github.com/docker/build-push-action/issues/252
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -23,13 +23,13 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
-      - name: Cache Docker layers
```
harikrongali, line 26: is this intentional to remove cache?

34fathombelow: Yes, the cache is only caching the exported layers in the final build stage.  Which provides no real benefit. This could be changed to mode=max to export all interminate layers but GitHub has a limit of 10GB across a single repo.   Also the time to prossess approximately 20gb of cache and compress it really outweighs the benefits.  A full cache will also not prossess RUN commands in the dockerfile.

34fathombelow: Also this cache is being accessed during pull requests.  It can create conflicts and slow the processes down.

harikrongali (approved):

harikrongali:
thanks for the PR

jessesuen (approved):

34fathombelow:
@harikrongali @jessesuen Thank you for the review and merge.   Image build times are much better,  but the plugin-image should have been built from the runners cache of the controller-image.  I will look into this and have a solution by Monday.    We should be able to cut off an additional 20+ minutes.

PR #1920, opened 2022-03-22
chore: Pin golang to 1.17 to avoid CVEs in docker image

bduverger:
### Context:
I found few CVE about golan, and update it.
- critical CVE-2022-23806 about Go fixed in: 1.17.7
- few high CVE about Go fixed in: 1.17.7 and 1.17.8 ([CVE-2022-24921](https://nvd.nist.gov/vuln/detail/CVE-2022-24921), [CVE-2022-23772](https://nvd.nist.gov/vuln/detail/CVE-2022-23772), [CVE-2022-23773](https://nvd.nist.gov/vuln/detail/CVE-2022-23773))

### Testing:
I did a "make test" with go 1.17.8, and build the docker images.

### Checklist:


alexec:
In workflows we pin to v1.17, no patch version. No patch updates are ever needed.

bduverger:
Hello,

I have exactly the same issue reported in https://github.com/argoproj-labs/argocd-vault-plugin/issues/320
The CVEs are on /bin/rollouts-controller even with the release v1.2.0. You're right, the workflow 'Go' in pinned to 1.17 and when the v1.2.0 was builded, it was golang 1.17.7 from hostedcache:
https://github.com/argoproj/argo-rollouts/actions/runs/2018668325
But in my report, I can see the binary was builded with 1.17.6 and have some CVEs, so it should be modified in the Dockerfile and everywhere 1.17.6 is.

By the way, I have the same CVEs in argo-cd.

alexec:
I'm suggesting this change

```
          go-version: 1.17.6
```

Becomes:

```
           go-version: 1.17
```

This solves the problem both for today (1.17.6) but also tomorrow (>=1.17.8).

Are you saying we need to consider caches too?

bduverger:
So, I replaced 1.17.6 in Dockerfile by 1.17 and I don't have anymore the CVEs that I mentionned. Initially I didn't want to change the behavior but I don't have problem to pinned everywhere to 1.17.
The commit will follow.

harikrongali (approved):

jessesuen (approved):

PR #1935, opened 2022-03-27
Issue #1779: add MinReplicas option to SetCanaryScale

### LINKED ISSUE #1779: Provide a way to specify a minimum number of pods per Replicaset
# Summary

The issue that I was trying to solve with https://github.com/argoproj/argo-rollouts/issues/1738 is when a horizontal pod autoscaler is controlling the #replicas in the rollout. I want the #canary pods to match the weight of the rollout, but I also want to ensure a minimum number pods (e.g. 2) to ensure high availability. Currently, I can only use setCanaryScale to set either a fixed number of pods or a percentage of the total #replicas I want to be able to set a percentage, but apply a minimum as well.

Also, if I am using Dynamic Stable Scale, I want a way to ensure that the stable replicaset never goes below a minimum (e.g. 2) number of pods to ensure high availability.

# Use Cases

Assume a Horizontal Pod Autoscaler is controlling the #replicas in the rollout, with a min of 10 pods and a max of 2000. I do not know up-front how many pods will set in the rollout. So, lets say I am running a step with the canary weight of 5 percent. If the total number of pods is less than 20, that would only give me one canary pod and I will not have high availability. However, I can't use the 'replicas' option of setCanaryScale to set a fixed number of replicas, because if the HPA does scale up to 2000 pods, I want to have 100 canary pods.

A second, but related, use case is if I am using Dynamic Stable Scale, so the stable replicaset scales down when the canary scales up. If we have 4 replicas and the canary is at 50% we are OK (2 pods each for canary and stable). However, if we then go to 75% canary, we will be left with only one pod in the stable ReplicaSet and again lose high availability. Again, the total #replicas may be controlled by an HPA, so we don't know up front when creating this step how any replicas there will be,

---

ssanders1449:
Fixes #1779 

Adds an optional 'MinReplicas' field to SetCanaryScale

See https://github.com/argoproj/argo-rollouts/issues/1779 for use cases


ssanders1449:
Hmm. It seems that the last force-push (to add the signoff) brought too many changes into the PR, I am going to redo the PR

PR #1938, opened 2022-03-28
feat: Implement Issue #1779: Add MinReplicas option to SetCanaryScale

ssanders1449:
Implements issue #1779
See https://github.com/argoproj/argo-rollouts/issues/1779 for Use Cases


ssanders1449:
@jessesuen - can you please review this? We need this in order to ensure high availability when using TrafficRoutedCanary

@nissy34 - FYI

ssanders1449:
I'm surprised that no one else has this issue. We worked around it so far, by adding a default-backend service to the ingress which routes to all of pods (both stable and canary), so that if one of the replicasets has only one POD and it goes down (e.g. it is being moved to another node), the requests will get routed to the other pods via the default-backend.

However, we can't always have a default-backend service (for example, if we want to protect the backend services by only forwarding requests with a give path prefix)

ssanders1449:
@zachaller - can you take a look at this? We are currently building from a fork, and would like this to be eventually included

ssanders1449:
Based on https://github.com/argoproj/argo-rollouts/issues/1779#issuecomment-1314690329, this can be replaced by

https://github.com/argoproj/argo-rollouts/pull/2410

ssanders1449:
closing in favor of https://github.com/argoproj/argo-rollouts/pull/2448

PR #1948, opened 2022-04-02
chore: Improve image build speed #1919

34fathombelow:

This is a follow-up to PR#1919.  Cache created from the `controller-image` should be used to build the `plugin-image`.  Buildx uses Buildkit under the hood and uses "Automatic garbage collection" which has a [default policy of 10% of free disk space](https://github.com/moby/moby/blob/master/builder/builder-next/worker/gc_unix.go).  Disabling the garbage collection ensures that the cache is temporarily stored for the duration of the runner.  

I also made a small change to the Dockerfile to provide support for older versions of Docker and Podman which do not support multiple args on a single line.

Checklist:

* [x] Either ~~(a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or~~ (c) this is a chore.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -22,14 +22,10 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-
-      # - name: Cache Docker layers
-      #   uses: actions/cache@v2
-      #   with:
-      #     path: /tmp/.buildx-cache
-      #     key: ${{ runner.os }}-buildx-${{ github.sha }}
-      #     restore-keys: |
-      #       ${{ runner.os }}-buildx-
+        with:
+          config-inline: |
+            [worker.oci]
+            gc = false
 
       - name: Docker meta (controller)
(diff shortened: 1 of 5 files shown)


34fathombelow:
I noticed line# 66 of the Dockerfile we were compiling `controller plugin plugin-linux plugin-darwin plugin-windows`.  There is no need to compile `plugin-linux plugin-darwin plugin-windows` binaries each time we build images, as they are not in any of the final images.   With two different architectures being built, we were compiling six additional unneeded binaries.  They are added back into the release workflow by including them as a --build-arg in `/hack/build-release-plugins.sh`

The release workflow now also supports all of the optimizations. Image build times should be about 10 minutes now!!!

**Please add a label  of "test-arm-image" to this PR.**

34fathombelow:
@harikrongali can you please review this when you get a chance?

harikrongali:
@crenshaw-dev   can you review?

alexmt (approved):
thank you!

Issue #1846, opened 2022-02-04
Make prometheus server address optional / centrally configured


jessesuen:
# Summary

AnalysisTemplate authors using the prometheus integration are currently required to specify a prometheus server address as a mandatory field in the template. In retrospect, this server address is something that operators would want to set up, and namespace level users would not care/need to know about this address.

This enhancement proposal is to make the prometheus server address an optional field, and allow an administrator to define the "default" server used by the cluster.

# Use Cases

When would you use this?

It could be painful to migrate the server address URL for hundreds of AnalysisTemplates that propagated in a cluster or set of clusters. So central control of this is likely the more common use case.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #1956, opened 2022-04-05
feat: Make prometheus server address optional/centrally configured

RaviHari:
Signed-off-by: Ravi Hari <ravireliable@gmail.com>

Fixes: #1846 

```
- add a new cli option to pass prometheusAddress that can be centrally configured.
- if analysis template has the prometheus address that is preferred.
- it wouldn't accept conflicting addresses as the arguments are combined when analysis run is created with templates.
- if analysis template does not have the prometheus address defined the address from the cli argument is taken for analysisrun.
```

Checklist:

### DIFF
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -69,6 +69,7 @@ func newCommand() *cobra.Command {
 		awsVerifyTargetGroup bool
 		namespaced           bool
 		printVersion         bool
+		prometheusAddress    string
 	)
 	electOpts := controller.NewLeaderElectionOptions()
 	var command = cobra.Command{
@@ -94,6 +95,7 @@ func newCommand() *cobra.Command {
 			defaults.SetAmbassadorAPIVersion(ambassadorVersion)
 			defaults.SetSMIAPIVersion(trafficSplitVersion)
 			defaults.SetAppMeshCRDVersion(appmeshCRDVersion)
+			defaults.SetPrometheusAddress(prometheusAddress)
 
 			config, err := clientConfig.ClientConfig()
 			checkError(err)
@@ -245,6 +247,7 @@ func newCommand() *cobra.Command {
 	command.Flags().DurationVar(&electOpts.LeaderElectionLeaseDuration, "leader-election-lease-duration", controller.DefaultLeaderElectionLeaseDuration, "The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled.")
 	command.Flags().DurationVar(&electOpts.LeaderElectionRenewDeadline, "leader-election-renew-deadline", controller.DefaultLeaderElectionRenewDeadline, "The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable if leader election is enabled.")
 	command.Flags().DurationVar(&electOpts.LeaderElectionRetryPeriod, "leader-election-retry-period", controller.DefaultLeaderElectionRetryPeriod, "The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled.")
+	command.Flags().StringVar(&prometheusAddress, "prometheus-address", "", "Set the  prometheus address that analysis controller uses when querying for metrics.")
 	return &command
 }
 
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -69,7 +69,6 @@ func newCommand() *cobra.Command {
 		awsVerifyTargetGroup bool
 		namespaced           bool
 		printVersion         bool
-		prometheusAddress    string
 	)
 	electOpts := controller.NewLeaderElectionOptions()
 	var command = cobra.Command{
@@ -95,7 +94,6 @@ func newCommand() *cobra.Command {
 			defaults.SetAmbassadorAPIVersion(ambassadorVersion)
 			defaults.SetSMIAPIVersion(trafficSplitVersion)
 			defaults.SetAppMeshCRDVersion(appmeshCRDVersion)
-			defaults.SetPrometheusAddress(prometheusAddress)
 
 			config, err := clientConfig.ClientConfig()
 			checkError(err)
@@ -247,7 +245,6 @@ func newCommand() *cobra.Command {
 	command.Flags().DurationVar(&electOpts.LeaderElectionLeaseDuration, "leader-election-lease-duration", controller.DefaultLeaderElectionLeaseDuration, "The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled.")
 	command.Flags().DurationVar(&electOpts.LeaderElectionRenewDeadline, "leader-election-renew-deadline", controller.DefaultLeaderElectionRenewDeadline, "The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable if leader election is enabled.")
 	command.Flags().DurationVar(&electOpts.LeaderElectionRetryPeriod, "leader-election-retry-period", controller.DefaultLeaderElectionRetryPeriod, "The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled.")
-	command.Flags().StringVar(&prometheusAddress, "prometheus-address", "", "Set the  prometheus address that analysis controller uses when querying for metrics.")
 	return &command
 }
 
--- a/metricproviders/prometheus/prometheus.go
+++ b/metricproviders/prometheus/prometheus.go
@@ -26,7 +26,6 @@ const (
 	// ResolvedPrometheusQuery is used as the key for storing the resolved prometheus query in the metrics result
 	// metadata object.
 	ResolvedPrometheusQuery = "ResolvedPrometheusQuery"
-	TokensSecretName        = "prometheus"
 	Address                 = "PROMETHEUS_ADDRESS"
 )
 
@@ -177,7 +176,7 @@ func IsUrl(str string) bool {
 	if err != nil {
 		log.Errorf("Error in parsing url: %v", err)
 	}
-	log.Infof("Parsed url: %v", u)
+	log.Debugf("Parsed url: %v", u)
 	return err == nil && u.Scheme != "" && u.Host != ""
 }
 
@@ -186,10 +185,10 @@ func lookupKeysInEnv(keys []string) map[string]string {
 	for i := range keys {
 		key := keys[i]
 		formattedKey := strings.ToUpper(strings.ReplaceAll(key, "-", "_"))
-		log.Infof("formattedKey: %v", formattedKey)
+		log.Debugf("formattedKey: %v", formattedKey)
 		if value, ok := os.LookupEnv(fmt.Sprintf("%s", formattedKey)); ok {
 			valuesByKey[key] = value
-			log.Infof("PROMETHEUS_ADDRESS: %v", value)
+			log.Debugf("PROMETHEUS_ADDRESS: %v", value)
 		}
 	}
 	return valuesByKey
--- a/metricproviders/prometheus/prometheus.go
+++ b/metricproviders/prometheus/prometheus.go
@@ -144,8 +144,8 @@ func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
 		Address: metric.Provider.Prometheus.Address,
 	})
 	if err != nil {
+		log.Errorf("Error in getting prometheus client: %v", err)
 		return nil, err
 	}
-
 	return v1.NewAPI(client), nil
 }
--- a/metricproviders/prometheus/prometheus.go
+++ b/metricproviders/prometheus/prometheus.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/argoproj/argo-rollouts/utils/defaults"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	"net/url"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/argoproj/argo-rollouts/utils/defaults"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/prometheus/client_golang/api"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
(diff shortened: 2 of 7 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -23,13 +23,13 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
-      - name: Cache Docker layers
-        uses: actions/cache@v2
```
harikrongali, line 27: can you rebase to master? these changes are not related to your PR

harikrongali, line 27: can you rebase to master? these changes are not related to your PR

Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Initial stage which pulls prepares build dependencies and CLI tooling we need for our final image
 # Also used as the image in CI jobs so needs all dependencies
 ####################################################################################################
-FROM golang:1.17.6 as builder
+FROM --platform=$BUILDPLATFORM golang:1.17.6 as builder
```
harikrongali, line 6: same here

harikrongali, line 6: same here

cmd/rollouts-controller/main.go
```diff
@@ -245,6 +247,7 @@ func newCommand() *cobra.Command {
...
 	command.Flags().DurationVar(&electOpts.LeaderElectionRenewDeadline, "leader-election-renew-deadline", controller.DefaultLeaderElectionRenewDeadline, "The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable if leader election is enabled.")
 	command.Flags().DurationVar(&electOpts.LeaderElectionRetryPeriod, "leader-election-retry-period", controller.DefaultLeaderElectionRetryPeriod, "The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled.")
+	command.Flags().StringVar(&prometheusAddress, "prometheus-address", "", "Set the  prometheus address that analysis controller uses when querying for metrics.")
```
harikrongali, line 250: what is the default address? do we set anywhere?

harikrongali, line 250: what is the default address? do we set anywhere?

pkg/kubectl-argo-rollouts/info/testdata/canary/canary-rollout5.yaml
```diff
@@ -0,0 +1,96 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
```
harikrongali, line 3: is this related to your change?

harikrongali, line 3: is this related to your change?

utils/defaults/defaults.go
```diff
@@ -49,6 +49,7 @@ const (
 	DefaultSMITrafficSplitVersion       = "v1alpha1"
 	DefaultTargetGroupBindingAPIVersion = "elbv2.k8s.aws/v1beta1"
 	DefaultAppMeshCRDVersion            = "v1beta2"
+	DefaultPrometheusAddress            = "http://127.0.0.1:9090"
```
harikrongali, line 52: i see default here.

harikrongali, line 52: i see default here.

harikrongali:
can you rebase to latest? i see code diff not related to prometheus change.

also, can you confirm if it is backward compatible? 

if the template has a Prometheus address, do you accept the URL?

harikrongali:
can you rebase to latest? i see code diff not related to prometheus change.

also, can you confirm if it is backward compatible? 

if the template has a Prometheus address, do you accept the URL?

harikrongali:
there seems to be rebase issue, i see commits listed in the diff rather master rebase.

harikrongali:
there seems to be rebase issue, i see commits listed in the diff rather master rebase.

RaviHari:
> there seems to be rebase issue, i see commits listed in the diff rather master rebase.

Fixed and rebased with master.. only changes from my pr are made available now.

RaviHari:
> can you rebase to latest? i see code diff not related to prometheus change.
> 
> also, can you confirm if it is backward compatible?
> 
> if the template has a Prometheus address, do you accept the URL?

The cli is optional if we do not pass anything it will use from the templates so yes it is backward compatible.
If the template has prometheus address it will accept that.. if it is not passed it will pickit up from the cli address provided from cli.

harikrongali:
got it. Thanks for the explanation. LGTM

harikrongali (approved):

harikrongali:
@agrawroh can you review as well?

cmd/rollouts-controller/main.go
```diff
@@ -69,6 +69,7 @@ func newCommand() *cobra.Command {
 		awsVerifyTargetGroup bool
 		namespaced           bool
 		printVersion         bool
+		prometheusAddress    string
```
jessesuen, line 72: We should configure this via configmap/secret instead of CLI flag (just like we configure datadog, new relic, wavefront). Could we switch to that model instead?

agrawroh: Hmmm. What do you think about configuring this using an environment variable instead? ConfigMap or Secret doesn't look like a good place to configure this. Maybe I'm missing some background/context.

cmd/rollouts-controller/main.go
```diff
@@ -94,6 +95,7 @@ func newCommand() *cobra.Command {
 			defaults.SetAmbassadorAPIVersion(ambassadorVersion)
 			defaults.SetSMIAPIVersion(trafficSplitVersion)
 			defaults.SetAppMeshCRDVersion(appmeshCRDVersion)
+			defaults.SetPrometheusAddress(prometheusAddress)
```
agrawroh, line 98: Is it possible to do some validation inside `SetPrometheusAddress()` to make sure that the passed value is a URI in the expected format (Scheme, Port Number, ...)?

cmd/rollouts-controller/main.go
RaviHari: Looking a some k8s references: https://kubernetes.io/docs/reference/command-line-tools-reference I thought cli paramter option was good. I can update with configmap / secret.. however would like to understand why they are preferred?

RaviHari:
@jessesuen and @agrawroh can you please review this pr updated as per the suggestions to pick the address from secret or env if not specified in template. Also validating the url format is correct.

agrawroh:
@jessesuen Thoughts on whether we also need to pick this up from K8S Secrets? It doesn't look right to me but if you think it's good to have it for consistency then it's okay. I do think that ENV is probably the right way to define these settings.

jessesuen:
> @jessesuen Thoughts on whether we also need to pick this up from K8S Secrets? It doesn't look right to me but if you think it's good to have it for consistency then it's okay. I do think that ENV is probably the right way to define these settings.

I discussed with @harikrongali. Looking back, I think we originally made a mistake on how datadog, wavefront, etc.. were configured via a Kubernetes Secret when they could have been done with environment variables (which happen to be backed by a secret).

Going forward (starting with this integration), I agree we should move to a model where metric provider connection information can be configured via environment variables, which will eliminate the need for the rollout controller to look up the kubernetes secret.

@RaviHari -- could we change the implementation once again to remove the option to look up the kubernetes secret and rely solely on env vars? The user could choose to store this as a secret or not (e.g. when we later add support for prometheus authentication).

agrawroh:
> > @jessesuen Thoughts on whether we also need to pick this up from K8S Secrets? It doesn't look right to me but if you think it's good to have it for consistency then it's okay. I do think that ENV is probably the right way to define these settings.
> 
> I discussed with @harikrongali. Looking back, I think we originally made a mistake on how datadog, wavefront, etc.. were configured via a Kubernetes Secret when they could have been done with environment variables (which happen to be backed by a secret).
> 
> Going forward (starting with this integration), I agree we should move to a model where metric provider connection information can be configured via environment variables, which will eliminate the need for the rollout controller to look up the kubernetes secret. Later we can improve the other implementations such that they could also be configured via env vars.
> 
> I just filed #1979 as a future improvement
> 
> @RaviHari -- could we change the implementation once again to remove the option to look up the kubernetes secret and rely solely on env vars? The user could choose to store this as a secret or not. Currently, since this is just a prometheus address, it there's no need for it to be a secret. Later, if/when we later add support for prometheus authentication, they would want it to be a secret, obviously.

@jessesuen Thanks for chiming in. I can probably take the improvement you filed if I get some free time next week.

RaviHari:
@jessesuen and @agrawroh thanks.. I made the changes to make read it from env variable. Can you please review.

metricproviders/prometheus/prometheus.go
```diff
@@ -22,6 +26,8 @@ const (
 	// ResolvedPrometheusQuery is used as the key for storing the resolved prometheus query in the metrics result
 	// metadata object.
 	ResolvedPrometheusQuery = "ResolvedPrometheusQuery"
+	TokensSecretName        = "prometheus"
```
agrawroh, line 29: Do we still need this now?

metricproviders/prometheus/prometheus.go
```diff
@@ -140,12 +146,51 @@ func NewPrometheusProvider(api v1.API, logCtx log.Entry) *Provider {
...
-
 	return v1.NewAPI(client), nil
 }
+
+func IsUrl(str string) bool {
+	u, err := url.Parse(str)
+	if err != nil {
+		log.Errorf("Error in parsing url: %v", err)
+	}
+	log.Infof("Parsed url: %v", u)
```
agrawroh, line 180: `info` might be too verbose. Should we change it to `debug` instead?

metricproviders/prometheus/prometheus.go
```diff
@@ -140,12 +146,51 @@ func NewPrometheusProvider(api v1.API, logCtx log.Entry) *Provider {
...
+	return err == nil && u.Scheme != "" && u.Host != ""
+}
+
+func lookupKeysInEnv(keys []string) map[string]string {
+	valuesByKey := make(map[string]string)
+	for i := range keys {
+		key := keys[i]
+		formattedKey := strings.ToUpper(strings.ReplaceAll(key, "-", "_"))
+		log.Infof("formattedKey: %v", formattedKey)
```
agrawroh, line 189: Same as above. I feel that we should be a little conservative while using `info` level.

metricproviders/prometheus/prometheus.go
```diff
@@ -140,12 +146,51 @@ func NewPrometheusProvider(api v1.API, logCtx log.Entry) *Provider {
...
+		if value, ok := os.LookupEnv(fmt.Sprintf("%s", formattedKey)); ok {
+			valuesByKey[key] = value
+			log.Infof("PROMETHEUS_ADDRESS: %v", value)
```
agrawroh, line 192: I'd evaluate where we can switch `info` -> `debug` to not pollute the logs.

metricproviders/prometheus/prometheus_test.go
```diff
@@ -58,6 +62,44 @@ func TestRunSuccessfully(t *testing.T) {
...
 
+func TestRunSuccessfullyWithEnv(t *testing.T) {
+	e := log.Entry{}
+	mock := mockAPI{
+		value: newScalar(10),
+	}
+	address := "http://127.0.0.1:9090"
+	useEnvVarForKeys := true
+	os.Setenv("PROMETHEUS_ADDRESS", address)
+	tokenSecret := &corev1.Secret{
```
agrawroh, line 73: Do we still need this now?

agrawroh (approved):
LGTM. I'd leave the final approval to @jessesuen and @harikrongali

Thank you so much for making this change!

RaviHari:
> LGTM. I'd leave the final approval to @jessesuen and @harikrongali
> 
> Thank you so much for making this change!

Thanks for the review and suggestions @agrawroh @harikrongali @jessesuen

metricproviders/prometheus/prometheus.go
```diff
@@ -140,12 +145,51 @@ func NewPrometheusProvider(api v1.API, logCtx log.Entry) *Provider {
 
 // NewPrometheusAPI generates a prometheus API from the metric configuration
 func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
+	secretKeys := []string{Address}
+	envValuesByKey := lookupKeysInEnv(secretKeys)
+	log.Infof("envValuesByKey: %v, len: %v", envValuesByKey, len(envValuesByKey))
```
jessesuen, line 150: This is a high-frequency portion of code. I prefer that we don't logging this every time we construct a prometheus API client.

metricproviders/prometheus/prometheus.go
```diff
@@ -140,12 +145,51 @@ func NewPrometheusProvider(api v1.API, logCtx log.Entry) *Provider {
...
+		key := keys[i]
+		formattedKey := strings.ToUpper(strings.ReplaceAll(key, "-", "_"))
+		log.Debugf("formattedKey: %v", formattedKey)
+		if value, ok := os.LookupEnv(fmt.Sprintf("%s", formattedKey)); ok {
+			valuesByKey[key] = value
+			log.Debugf("PROMETHEUS_ADDRESS: %v", value)
+		}
+	}
+	return valuesByKey
+}
```
jessesuen, line 195: I don't understand the purpose of this function. Why aren't we just simply looking up an env var by its well-known, documented env var name? Can we delete this?

metricproviders/prometheus/prometheus.go
```diff
@@ -22,6 +26,7 @@ const (
 	// ResolvedPrometheusQuery is used as the key for storing the resolved prometheus query in the metrics result
 	// metadata object.
 	ResolvedPrometheusQuery = "ResolvedPrometheusQuery"
+	Address                 = "PROMETHEUS_ADDRESS"
```
jessesuen, line 29: 1. `PROMETHEUS_ADDRESS` could use a better configuration convention. I think we can go with `ARGO_ROLLOUTS_PROMETHEUS_ADDRESS`. In the future, we will have more of such environment variables for configuration and the `ARGO_ROLLOUTS` is a prefix to group rollouts specific env vars together. We do the same with `ARGOCD_`

2. nit: The global variable name `Address` is not very specific. Commonly if something is used as an env var, you can prefix it with `EnvVar<keyname>`. e.g. `EnvVarArgoRolloutsPrometheusAddress`

metricproviders/prometheus/prometheus.go
```diff
@@ -130,7 +135,7 @@ func (p *Provider) processResponse(metric v1alpha1.Metric, response model.Value)
 	}
 }
 
-// NewPrometheusProvider Creates a new Prometheus client
+//NewPrometheusProvider Creates a new Prometheus client
```
jessesuen, line 138: add back space?

jessesuen (changes requested):

metricproviders/prometheus/prometheus.go
RaviHari: This function is useful when we want to extract multiple strings from env variables such as auth headers, api keys other than just the address.. this is also converting formats to "_" in case there was a misconfiguration with "-"  in env variable .. but I agree in this case just getting from env variable is fine.

jessesuen (approved):
Could you please follow up this PR with doc changes?

PR #1990, opened 2022-04-21
feat: Allow Traffic shaping through header based routing. Fixes #474

### LINKED ISSUE #474: Allow Traffic shaping through header based routing
Instead of routing traffic to different ReplicaSet by using percentage, some developers want to route traffic based on the headers of the request. For example, the developer may want their Firefox users to go to the canary version or the client adds a special canary header that users want to route to the canary ReplicaSet. Many traffic management tools (like Istio, linkerd via SMI, and Nginx ingress controller) offer support for header-based routing, but Argo Rollouts currently cannot automate the process.

Some questions that need to be addressed by the design:
- How does this integrate with setWeight steps?
- How do users decide how many replicas to run for ReplicaSet receiving the rerouted traffic
- Are there any difference between any of the traffic management providers?
- Does the controller keep the headers when the Rollout is in a steady-state?

perenesenko:
Checklist:


perenesenko:
Fix for https://github.com/argoproj/argo-rollouts/issues/474

rollout/mocks/TrafficRoutingReconciler.go
```diff
@@ -14,7 +14,7 @@ type TrafficRoutingReconciler struct {
 }
 
 // SetWeight provides a mock function with given fields: desiredWeight, additionalDestinations
-func (_m *TrafficRoutingReconciler) SetWeight(desiredWeight int32, additionalDestinations ...v1alpha1.WeightDestination) error {
+func (_m *TrafficRoutingReconciler) SetWeight(desiredWeight int32, headerWeight *v1alpha1.SetHeaderWeight, additionalDestinations ...v1alpha1.WeightDestination) error {
```
leoluz, line 17: Just confirming, is this adding the ability to specify the header weight in the existing `setWeight` step? Can `WeightDestination` be used instead of adding an additional param in the signature?

leoluz (changes requested):
Added a comment regarding the SetWeight signature. I think it is better to leverage the `v1alpha1.WeightDestination` instead of adding an additional param to the function.

Another point is that I believe all traffic routers support header based routing. This PR is currently just enabling it for Istio. Shouldn't the header based routing be implemented for all traffic routers?

rollout/mocks/TrafficRoutingReconciler.go
perenesenko: @leoluz Yes. It's adding the new route with a header based traffic.
BTW there is a mistake here: should be `SetHeaderRoute`.

We can not use the **WeightDestination** as **SetHeaderRoute** has different fields

harikrongali:
@leoluz PR will focus on enabling for istio and will expand to other providers. It would be better to split for better code reviews. Given that, the expectation is to write generic code whenever possible.

perenesenko:
@leoluz Yes, this PR is for the Istio only.

zachaller:
> Added a comment regarding the SetWeight signature. I think it is better to leverage the `v1alpha1.WeightDestination` instead of adding an additional param to the function.
> 
> Another point is that I believe all traffic routers support header based routing. This PR is currently just enabling it for Istio. Shouldn't the header based routing be implemented for all traffic routers?

I think this also make sense and allows for possible more future proofing for other things such as routing via path and or ip. Does something like this make sense for v1alpha1.WeightDestination I added path as just an example we would not implement this right now though.

```
{
  "weight": 50,
  "serviceName": "someService",
  "podTemplateHash": "fa3f3ae",
  "rules": {
    "headers":  {
      "match": [{
        "headerName": "",
        "headerValue": "",
        "headerRegex": ""
    }]},
    "paths": {
      "match": [{
        "method": "GET",
        "path": "/some/path"
    }]}
  }
}
```

If we don't like the idea of using v1alpha1.WeightDestination then I would say maybe adding another method to the interface would be slightly better than adding another parameter to SetWeight maybe?

leoluz (commented):
One additional comment added. Besides that, I think it is a good idea to improve the unit test coverage in the istio.go to validate more scenarios.

zachaller:
I wanted to leave one note here for myself to come back to this but I think this also needs to take into account additionalWeights for when we are doing header routing during an experiments which istio is one of the few traffic routers that support experiments .

harikrongali:
please add more unit tests, I see code coverage check red

zachaller:
We should add at the minimum some documentation on the spec for this feature at https://argoproj.github.io/argo-rollouts/features/specification/

pragmaticivan:
Howdy! not sure if that's too late but the same features can be supported by Nginx. Thoughts on implementing the Nginx part of it as well?

Ref: https://github.com/fluxcd/flagger/blob/main/pkg/router/ingress.go#L137-L144

pragmaticivan:
Also, the same applies to AppMesh which was added recently: https://github.com/fluxcd/flagger/blob/main/pkg/router/appmesh.go#L249-L293

pragmaticivan:
Howdy! Will that implementation work with cookies?

```
curl -H 'X-Canary: insider' http://app.example.com
curl -b 'canary=always' http://app.example.com
```

leoluz:
> Howdy! Will that implementation work with cookies and user-agent? (I'm guessing yes once minimum documentation is added)
> 
> ```
> curl -H 'X-Canary: insider' http://app.example.com
> curl -b 'canary=always' http://app.example.com
> ```

@pragmaticivan 
This is PR is implementing the proposal described by https://github.com/argoproj/argo-rollouts/issues/474 which about header based routing, not cookies.

leoluz:
@perenesenko There are some pending issues that needs fixing before merging:
- Fix failing e2e test
- Fix codecov CI issues
- Add documentation about the new feature

pragmaticivan:
> > Howdy! Will that implementation work with cookies and user-agent? (I'm guessing yes once minimum documentation is added)
> > ```
> > curl -H 'X-Canary: insider' http://app.example.com
> > curl -b 'canary=always' http://app.example.com
> > ```
> 
> @pragmaticivan This PR is implementing the proposal described by #474 which about header based routing, not cookies.

In this context, would not cookie always be interpreted as a header? 

The implementation might not be that off: https://github.com/fluxcd/flagger/blob/main/pkg/router/ingress.go#L179-L248

harikrongali:
@perenesenko can you add documentation as well?

zachaller:
LGTM as long as we hold on a 1.3 release until after mirroring is also merged.

leoluz (approved):

zachaller (approved):

jessesuen (approved):

PR #2056, opened 2022-05-27
fix: Failed to process: Object 'Kind' is missing in Errors with rollouts notification

RaviHari:
Checklist:


pkg/apiclient/rollout/rollout.proto
```diff
@@ -71,6 +71,7 @@ message VersionInfo {
 }
 
 message RolloutInfo {
+  k8s.io.apimachinery.pkg.apis.meta.v1.TypeMeta typeMeta = 21;
```
leoluz, line 74: nit: I know it looks a bit better to group those fields together at the top of the file but I would prefer to maintain the proto standard and put this new attribute at the bottom of the `RolloutInfo` message (bellow line 98). This would avoid someone duplicating the attribute 21 because he/she didn't pay attention that it was already defined at the top of the message.

leoluz (changes requested):
One really small nit. Otherwise LGTM

leoluz (approved):
LGTM

leoluz:
@RaviHari There is an e2e test failing. Is it a flaky or an issue that could have been introduced by this PR?

RaviHari:
> @RaviHari There is an e2e test failing. Is it a flaky or an issue that could have been introduced by this PR?

Hi @leoluz, Thanks for the review. It is flaky as this test case that failed has no impact with the changes made.

jessesuen:
@RaviHari to elaborate on our discussed, my main concern with this approach is how the `updatedRollout = rolloutUpdate(updatedRollout)` called is needed multiple times in non-obvious portions of the code, and also in critical paths when I believe the fix should only occur in the notification area of the code. Could you point me to a version of this PR that you mentioned that resolved part of the issues but not all? I'd like to see if we can go with an approach that does not require these lower-level changes.

Also, could you explain why the proto needs to be changed as well?

RaviHari:
> @RaviHari to elaborate on our discussion in the contributors meeting, my main concern with this approach is how the `updatedRollout = rolloutUpdate(updatedRollout)` called is needed multiple times in non-obvious portions of the code, and also in critical paths when I believe the fix should only occur in a single place in the notification area of the code. Could you point me to a version of this PR that you mentioned that resolved part of the issues but not all? I'd like to see if we can go with an approach that does not require these lower-level changes.
> 
> Also, could you explain why the proto needs to be changed as well?

Hi @jessesuen ,
I confirm that rolloutinfo object in rollout.proto doesnot need change to add typeMeta field..
so we have two solution to solve:

- with updates to writetoInformer: https://github.com/argoproj/argo-rollouts/compare/master...RaviHari:argo-rollouts:issue-1720-3
or
- https://github.com/argoproj/argo-rollouts/pull/2056

the overhead is going to be the same in either of the cases.. but option 1 seems more clean.. the change to critical path is required as the notification engine shares the rollout Informer.

RaviHari:
closing this in favor of: https://github.com/argoproj/argo-rollouts/pull/2150

Issue #2065, opened 2022-06-01
Checksum mismatch downloading argo-rollouts v1.2.0, v1.2.1


kevinburkesegment:
Hi, did you happen to push different content to the same Git tag, or change the released content at the URL? I had a checksum mismatch in my local go.sum for this module and version.

```
verifying github.com/argoproj/argo-rollouts@v1.2.0: checksum mismatch
	downloaded: h1:+4a2BRD+Fjox5rcLEKmcUFmpwPhR5oaWFe2CrQZK31o=
	go.sum:     h1:3CxEe2tyeZd9I6qWQo7RUAYyGb3qAnXo4SgSVBJBEBs=
```

kevinburkesegment:
Aha, I figured it out - the module in proxy.golang.org works fine but the contents behind the proxy - ie if you do GOPROXY=direct - do not match up.

kevinburkesegment:
Here's the diff I see between what's available for download from proxy.golang.org and what's available if I check out v1.2.0 of the source directly, with this command (which is what Go runs):

```
git -c core.autocrlf=input -c core.eol=lf archive --format=zip --prefix=prefix/ 08cf10e554fe99c24c8a37ad07fadd9318e4c8a1
```

```
$ diff -r ./github.com/argoproj/argo-rollouts@v1.2.0 ./prefix
diff -r ./github.com/argoproj/argo-rollouts@v1.2.0/Dockerfile ./prefix/Dockerfile
15,17c15
< RUN wget https://install.goreleaser.com/github.com/golangci/golangci-lint.sh  && \
<     chmod +x ./golangci-lint.sh && \
<     ./golangci-lint.sh -b $GOPATH/bin && \
---
> RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.44.0 && \
```

kevinburkesegment:
It also looks like there were two different CI builds for v1.2.0, and I am guessing for v1.2.1 as well 

https://github.com/argoproj/argo-rollouts/runs/5634127235?check_suite_focus=true
https://github.com/argoproj/argo-rollouts/runs/5634458780?check_suite_focus=true

jessesuen:
Ohh I think you may be right about this. The release process is such that the tag comes first, and then the build. This sometimes leads to situation where even though the tip of the branch passed, the release action fails and needs a minor fixing and retagging.

[...]

PR #2077, opened 2022-06-07
chore: set version to 1.2.2; updates #2065

kevinburkesegment:
Versions v1.2.0 and v1.2.1 both featured tags that were pushed,
deleted, and then re-pushed. Because proxy.golang.org cached the first
version of each, this means the source code downloaded from
proxy.golang.org, and directly from Github, is different.

This means that anyone who does _not_ want to use proxy.golang.org
(like me) cannot import the source code, because it fails checksum
verification.

Creating a new version tag (and ensuring that the tag only gets pushed
once) should resolve this issue, since users can upgrade to the new
version.

Updates #2065.

Checklist:

* [ ] I have written unit and/or e2e tests for my change (not sure these are
* necessary)


huikang (approved):

zachaller:
@kevinburkesegment can we drop the version change now that 1.3 has been released? Do you still want to add Twilio Segment to users?

Issue #1979, opened 2022-04-13
Configure metric provider connection information vi env vars


jessesuen:
# Summary

Currently Argo Rollouts is configured with metric provider connection information via Kubernetes Secrets which it has to look up directly.

Instead, we could do a more traditional approach of using env vars to configure this, which would still allow them to be stored as secrets, but just using an `envSource` in the pod.

# Use Cases

We could eliminate the need for rollout controller to look up secrets (although not yet until we deprecate the old methods) as well as standardize the configuration of metric providers.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #2080, opened 2022-06-07
feat: Configure metric provider connection information via env vars

RaviHari:
fixes: https://github.com/argoproj/argo-rollouts/issues/1979

Checklist:

### DIFF
--- a/docs/analysis/datadog.md
+++ b/docs/analysis/datadog.md
@@ -32,10 +32,34 @@ Datadog api and app tokens can be configured in a kubernetes secret in argo-roll
 apiVersion: v1
 kind: Secret
 metadata:
-  name: datadog
+  name: datadog-secret
 type: Opaque
 data:
-  address: https://api.datadoghq.com
-  api-key: <datadog-api-key>
-  app-key: <datadog-app-key>
+  ARGO_ROLLOUTS_DD_ADDRESS: https://api.datadoghq.com
+  ARGO_ROLLOUTS_DD_API_KEY: <datadog-api-key>
+  ARGO_ROLLOUTS_DD_APP_KEY: <datadog-app-key>
 ```
+
+In Rollout Deployment add the follow env source
+
+```yaml
+spec:
+  containers:
+  - env:
+    - name: ARGO_ROLLOUTS_DD_ADDRESS
+      valueFrom:
+        secretKeyRef:
+          key: ARGO_ROLLOUTS_DD_ADDRESS
+          name: datadog-secret
+    - name: ARGO_ROLLOUTS_DD_APP_KEY
+      valueFrom:
+        secretKeyRef:
+          key: ARGO_ROLLOUTS_DD_APP_KEY
+          name: datadog-secret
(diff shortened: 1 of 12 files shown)


zachaller (commented):
Instead of informing people in the docs to add individual env vars to the deployment. I wonder if it make more sense to setup the deployment with say a metrics-config-env configmap then using `envFrom:` to include it into the deployment https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables

RaviHari:
> Instead of informing people in the docs to add individual env vars to the deployment. I wonder if it make more sense to setup the deployment with say a metrics-config-env configmap then using `envFrom:` to include it into the deployment manifests https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables
> 
> I would also probably use a k8s Secret instead of a configmap or at the least setup both.

Here is the discussion on why this approach is chosen: https://github.com/argoproj/argo-rollouts/pull/1956#issuecomment-1098460062

zachaller:
> > Instead of informing people in the docs to add individual env vars to the deployment. I wonder if it make more sense to setup the deployment with say a metrics-config-env configmap then using `envFrom:` to include it into the deployment manifests https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables
> > I would also probably use a k8s Secret instead of a configmap or at the least setup both.
> 
> Here is the discussion on why this approach is chosen: [#1956 (comment)](https://github.com/argoproj/argo-rollouts/pull/1956#issuecomment-1098460062)

Yup I am totally for that method I just think the way it is documented here is not ideal. Let me try to be more direct with the info from the k8s doc. Instead of having to add a whole bunch of env vars to the deployment manifest for each configuration what I am suggesting is using k8s `envFrom:` as a quick example from the docs page.

```
apiVersion: v1
kind: Secret
metadata:
  name: datadog-secret
type: Opaque
data:
  ARGO_ROLLOUTS_DD_ADDRESS: https://api.datadoghq.com
  ARGO_ROLLOUTS_DD_API_KEY: <datadog-api-key>
  ARGO_ROLLOUTS_DD_APP_KEY: <datadog-app-key>
---
spec:
  containers:
    - name: test-container
      image: k8s.gcr.io/busybox
      command: [ "/bin/sh", "-c", "env" ]
      envFrom:
      - configMapRef:
           name: datadog-secret
      - secretRef:
           name: datadog-secret
```

That is one option that is very specific still you could take that one step farther and make it very generic and we could include both an empty secret and configmap and include it in the manifests and then setup the deployment to mount both of them then end users would just need to patch or add their own config to the config map and not touch the deployment at all.

agrawroh (commented):
Thank you so much for moving these flags to env. Left some very high level comments but it's in the right direction.

PR #2081, opened 2022-06-07
chore: improve openapi schema

mvgmb:
### Changes

I want to be able to patch an AnalysisTemplate using Kustomize, however, the generated OpenAPI schema file [rollout_cr_schema.json](https://github.com/argoproj/argo-rollouts/blob/master/docs/features/kustomize/rollout_cr_schema.json) only include the Rollout CRD.

I removed the filter that only included the Rollout in the OpenAPI schema, allowing all the other CRDs to be generated.

---

Another problem with this hack is the naming convention when setting the `definitions` map values. It uses `version.name` as its key:
```json
{
    "definitions": {
        "v1alpha1.Rollout": {...}
    }
}
```

The kube-openapi, however, uses another convention. This difference makes it hard to merge my cluster's generated OpenAPI schema with Argo Rollouts'. Therefore, I changed the definitions keys to use the same naming convention as K8s':
```json
{
    "definitions": {
        "io.argoproj.v1alpha1.AnalysisRun": {...},
        "io.argoproj.v1alpha1.AnalysisTemplate": {...},
        "io.argoproj.v1alpha1.ClusterAnalysisTemplate": {...},
        "io.argoproj.v1alpha1.Experiment": {...},
        "io.argoproj.v1alpha1.Rollout": {...}
    }
}
```

### Checklist


zachaller:
LGTM, however I am not really sure of the value of adding `AnalysisRun` I do think the others make sense. I am also not sure if there is any harm in adding `AnalysisRun` so probably ok to keep unless there is some weird size issue somewhere.

PR #2131, opened 2022-07-07
fix(analysis): Fix Analysis Terminal Decision For Dry-Run Metrics

agrawroh:
## Description

Metrics running in the Dry-Run mode shouldn't impact the terminal decision even if they error out or fail but currently we ignore the metric state which results in the analysis run being terminated earlier in case there are any Dry-Run failures.

Expected behavior is to take into account the status of the metric i.e. Dry-Run or Wet-Run while making the terminal decision inside `IsTerminating(...)`.

## Checklist

### DIFF
--- a/analysis/analysis_test.go
+++ b/analysis/analysis_test.go
@@ -1500,7 +1500,7 @@ func TestAssessRunStatusErrorMessageAnalysisPhaseFail(t *testing.T) {
 
 func TestAssessRunStatusErrorMessageAnalysisPhaseFailInDryRunMode(t *testing.T) {
 	status, message, dryRunSummary := StartAssessRunStatusErrorMessageAnalysisPhaseFail(t, true)
-	assert.Equal(t, v1alpha1.AnalysisPhaseSuccessful, status)
+	assert.Equal(t, v1alpha1.AnalysisPhaseRunning, status)
 	assert.Equal(t, "", message)
 	expectedDryRunSummary := v1alpha1.RunSummary{
 		Count:        2,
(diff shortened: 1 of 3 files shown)


utils/analysis/helpers.go
```diff
@@ -87,7 +87,9 @@ func IsTerminating(run *v1alpha1.AnalysisRun) bool {
 	for _, res := range run.Status.MetricResults {
 		switch res.Phase {
 		case v1alpha1.AnalysisPhaseFailed, v1alpha1.AnalysisPhaseError, v1alpha1.AnalysisPhaseInconclusive:
-			return true
+			// If this metric is running in the dryRun mode then we don't care about the failures and hence the terminal
+			// decision shouldn't be affected.
+			return !res.DryRun
```
harikrongali, line 92: so if analysisRun is completed, then dryRun will keep it in running mode (argoCD UI shows as spinning and that will confuse users).
Don't we need default state or ignore analysis run state when dryRun is enabled

harikrongali: maybe I am wrong here.. eventually, analysis run will be completed

agrawroh: @harikrongali Yes, it'll only keep running if it's a background analysis where the `count` is not specified. Think of it as the case where there are no failures/errors and the AR keeps running in background until the rollout finishes.

analysis/analysis_test.go
```diff
@@ -1500,7 +1500,7 @@ func TestAssessRunStatusErrorMessageAnalysisPhaseFail(t *testing.T) {
 
 func TestAssessRunStatusErrorMessageAnalysisPhaseFailInDryRunMode(t *testing.T) {
 	status, message, dryRunSummary := StartAssessRunStatusErrorMessageAnalysisPhaseFail(t, true)
-	assert.Equal(t, v1alpha1.AnalysisPhaseSuccessful, status)
+	assert.Equal(t, v1alpha1.AnalysisPhaseRunning, status)
```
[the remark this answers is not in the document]
agrawroh: "Running" is actually the correct state here as these are the background runs which would keep running if there is no `count` specified.

harikrongali (approved):

huikang (approved):

Issue #1939, opened 2022-03-30
Is v1.2.0 compatible with Kubernetes 1.16? If not, can the project provide a compatibility matrix?


hcelaloner:
# Summary

Previously, some of our clusters had Argo Rollouts v1.1.1 installed via helm charts located in the argo-helm repository. When we tried to upgrade to v1.2.0, we encountered the following error 

```
error validating data: ValidationError(CustomResourceDefinition.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.template.properties.spec.properties.nodeSelector): unknown field "x-kubernetes-map-type" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps
```

I guess `x-kubernetes-map-type` was added in k8s 1.17. If Argo Rollouts is not compatible with k8s 1.16 clusters, it may be good to state this information somewhere? Maybe release notes if that is the case or a compatibility matrix for recent versions (like: https://github.com/open-telemetry/opentelemetry-operator#opentelemetry-operator-vs-kubernetes-vs-cert-manager)? Or with some sort of skip validation option can someone install it to a v1.16 cluster?

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

harikrongali:
Every ArgoRollout release supports the latest & latest-2 versions only. But you are correct, it is not mentioned anywhere in the docs and we will update them

thatsmydoing:
I think this ticket is related to ingress versions https://github.com/argoproj/argo-rollouts/issues/1507 (there's a flag for it)

> Every ArgoRollout release supports the latest & latest-1 versions only. But you are correct, it is not mentioned anywhere in the docs and we will update them

I'm not sure that's true as only 1.2.0 supports kubernetes 1.22 and above. And 1.1.0 was released when 1.22 was latest.

rcontreras-te:
So, I bumped into this just today.  We're running K8S `v1.16.13` / argo-rollouts `v1.1.1` and I was unable to upgrade to argo-rollouts `v1.2.1` because of this.  

So what is the minimum version of K8S required in order to use argo-rollouts `v1.2.0`?  I didn't see minimum requirement mention in the `v1.2.0` release notes.  @harikrongali , you mentioned that there might be an update to the docs for this... is it possible to get a link to that?

hcelaloner:
[...]

PR #2163, opened 2022-07-26
docs: mention supported versions

kostis-codefresh:
Signed-off-by: Kostis Kapelonis <kostis@codefresh.io>

Solves https://github.com/argoproj/argo-rollouts/issues/1939

Checklist:

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -86,3 +86,12 @@ You can run it like any other Docker image or use it in any CI platform that sup
 docker run quay.io/argoproj/kubectl-argo-rollouts:master version
 ```
 
+## Supported versions
+
+At any point in time the officially supported version of Argo Rollouts is the latest released one, on Kubernetes versions N and N-1 (as supported by the Kubernetes project itself).
+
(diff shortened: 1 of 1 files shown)


rcontreras-te (commented):
Thanks for the docs update @kostis-codefresh !

zachaller:
This looks good I do think it makes sense to possibly do a table with versions that are supported as well so we can track the history of support, thoughts? If we do this we should also make note in the release docs https://github.com/argoproj/argo-rollouts/blob/master/docs/releasing.md about updating the table until we can possibly automate it.

kostis-codefresh:
@zachaller FYI this is the same PR as Argo CD https://argo-cd.readthedocs.io/en/stable/operator-manual/installation/#supported-versions

It is bulletproof in the sense that it needs no updates at all.

zachaller:
Yup that is fair it just makes it hard for the end users to know which versions where supported in the past without doing a whole bunch of date comparisons and looking at when argo rollouts was released and comparing it to when kubernetes was released to be able to say something like versions 1.1 of rollouts supported kubernetes versions 1.xx if we find that is to much a burden on the release to keep up to date I am fine either way nothing should block this PR. Just wanted feedback on if we felt that was useful.

zachaller (approved):

leoluz (approved):

Issue #2151, opened 2022-07-19
AnalysisRun gets stuck in Running state for dryrun + successful job metric


chrisplim:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've inclued the version of argo rollouts.

**Describe the bug**

AnalysisRun gets stuck in `Running` state if the metric is listed as a dryrun and is a Job that completes Successfully with exit code 0.

**To Reproduce**

1. Apply this manifests file: https://gist.github.com/chrisplim/c864a22a5ea7ba788a6a86ca13e54807
   ```
   kubectl apply -f dryrun-manifests.yaml
   ```
2. Get/watch the rollout status
    ```
    kubectl argo rollouts get rollout dryrun-rollout --watch
    ```
3. In a separate terminal, update the rollout's pod's image to cause a rollout
    ```
    kubectl argo rollouts set image dryrun-rollout helloworld=hashicorp/http-echo:0.2.3
    ```
4. Notice the AnalysisRun's Job completes successfully with exit code 0, but the AnalysisRun never transitions from the `Running` state to `Successful`

**Expected behavior**

The AnalysisRun should transition to the `Successful` state after a Job completes successfully with exit code 0, even though the metric was listed as a dryRun.

Note: We can verify that the dryrun with a *failed* job works as expected, which is to transition the AnalysisRun to be `Successful`, despite the Job failing.
steps:
1. Use the same dryrun-manifests.yaml file, except editing the rollout's `spec.strategy.blueGreen.prePromotionAnalysis.templates[0].templateName` to be `dryrun-analysis-failure`
2. `kubectl apply -f dryrun-manifests.yaml`
3. Change the rollout image to force a rollout
   ```
   kubectl argo rollouts set image dryrun-rollout helloworld=hashicorp/http-echo:0.2.3
   ```
4. Notice that a failed job + dryrun metric works as intended.
<img width="796" alt="Screen Shot 2022-07-19 at 12 44 07 PM" src="https://user-images.githubusercontent.com/4838817/179835250-dad5f57c-856d-4280-bdd5-6c84513e64bc.png">

**Screenshots**

Screenshot showing the job completed successfully, but AnalysisRun gets stuck in Running state
(the output of ` kubectl argo rollouts get rollout dryrun-rollout --watch`)
[...]

PR #2230, opened 2022-09-05
fix(analysis): Make AnalysisRun end when only Dry-Run metrics are defined. Fixes: #2151

agrawroh:
## Description

This PR fixes the issue defined in https://github.com/argoproj/argo-rollouts/issues/2151 where `AnalysisRun` doesn't transition to `Successful` from `Running` when only Dry-Run metrics are defined and being evaluated. As per the current logic, we don't set the decision while evaluating the Dry-Run metrics but the proper fix is to set it to `Successful` at least if nothing is set.

## Checklist

### DIFF
--- a/analysis/analysis_test.go
+++ b/analysis/analysis_test.go
@@ -1500,7 +1500,7 @@ func TestAssessRunStatusErrorMessageAnalysisPhaseFail(t *testing.T) {
 
 func TestAssessRunStatusErrorMessageAnalysisPhaseFailInDryRunMode(t *testing.T) {
 	status, message, dryRunSummary := StartAssessRunStatusErrorMessageAnalysisPhaseFail(t, true)
-	assert.Equal(t, v1alpha1.AnalysisPhaseRunning, status)
+	assert.Equal(t, v1alpha1.AnalysisPhaseSuccessful, status)
 	assert.Equal(t, "", message)
 	expectedDryRunSummary := v1alpha1.RunSummary{
 		Count:        2,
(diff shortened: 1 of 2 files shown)


analysis/analysis_test.go
```diff
@@ -1613,6 +1613,115 @@ func StartAssessRunStatusWorstMessageInReconcileAnalysisRun(t *testing.T, isDryR
...
+func TestAssessRunStatusWithOnlyDryRunMetrics(t *testing.T) {
+	f := newFixture(t)
+	defer f.Close()
+	c, _, _ := f.newController(noResyncPeriodFunc)
+
+	run := v1alpha1.AnalysisRun{
+		Spec: v1alpha1.AnalysisRunSpec{
+			Metrics: []v1alpha1.Metric{
+				{
+					Name: "failed-metric",
```
chrisplim, line 1627: Is it worth also testing when there's a single dry-run metric that succeeds, like in the original issue filed? For whatever reason, a single failed metric in the dry-run metric was able to transition the analysis to Successful before this change. 

For my own understanding, can you explain the behavior before the fix of `worstStatus = v1alpha1.AnalysisPhaseSuccessful` that a single failed metric was able to transition the analysis to successful, but the single successful metric was not able to transition the analysis to successful?

agrawroh: Isn't this Test Case doing exactly that? Here, it's just a single metric which is defined as a `Dry-Run`.

agrawroh: Make sense. Changed it to `Success` instead.

chrisplim: Thanks for updating. Do you think we need both tests, one test for the single dry-run metric that succeeds. And then separately another test for the single dry-run metric that fails?

Otherwise, I think this lgtm

harikrongali (approved):

chrisplim:
@agrawroh thanks for the fix and @harikrongali thanks for the review!

Issue #2228, opened 2022-09-05
Notification not sent when the conditions are not set


RaviHari:
The issues is related to: https://github.com/argoproj/argo-rollouts/pull/2066

Notifications are working when run on v1.2.2 but when upgraded v1.3.0-rc versions Notifications are not sent when conditions are not mentioned.

The logs are being generated though:
```
INFO[0115] Rollout has waited the duration of the pause step  namespace=rollout-canary rollout=rollouts-demo
DEBU[0115] No configured template for trigger: on-rollout-step-completed  namespace=rollout-canary rollout=rollouts-demo
INFO[0115] Event(v1.ObjectReference{Kind:"Rollout", Namespace:"rollout-canary", Name:"rollouts-demo", UID:"b3fbd661-423f-4e1f-a15b-175e9fca5f35", APIVersion:"argoproj.io/v1alpha1", ResourceVersion:"21533943", FieldPath:""}): type: 'Normal' reason: 'RolloutStepCompleted' Rollout step 7/7 completed (pause: 5) 
DEBU[0115] Notification sent                             namespace=rollout-canary rollout=rollouts-demo time_ms=0.041305
INFO[0115] Rollout step 7/7 completed (pause: 5)         event_reason=RolloutStepCompleted namespace=rollout-canary rollout=rollouts-demo
DEBU[0116] No configured destinations for trigger: on-rollout-resumed  namespace=rollout-canary rollout=rollouts-demo
DEBU[0116] Notification sent                             namespace=rollout-canary rollout=rollouts-demo time_ms=0.062244999999999995
```

PR #2231, opened 2022-09-07
fix: enable notifications without when condition

RaviHari:
fixes: #2228 
Signed-off-by: Ravi Hari <ravireliable@gmail.com>

Checklist:

### DIFF
--- a/utils/record/record.go
+++ b/utils/record/record.go
@@ -2,6 +2,8 @@ package record
 
 import (
 	"context"
+	"crypto/sha1"
+	"encoding/base64"
 	"encoding/json"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


utils/record/record.go
```diff
@@ -278,16 +278,28 @@ func (e *EventRecorderAdapter) sendNotifications(object runtime.Object, opts Eve
...
+						if err != nil {
+							log.Errorf("notification error: %s", err.Error())
+							return err
+						}
+					}
+				}
+			}
+		} else {
+			for _, dest := range destinations {
 				err = notificationsAPI.Send(objMap, triggerActions[0].Send, dest)
```
zachaller, line 303: Why are we always sending triggerAction[0] in this loop of trigger actions?

RaviHari: not necessary we can ta.. but from the event notification design it is always going to have only once event for which the trigger runs at any point so it would work. However it would be good to replace it with the particular triggeraction at that index as we are anyway looping through the list will update the pr.

utils/record/record.go
```diff
@@ -278,17 +278,29 @@ func (e *EventRecorderAdapter) sendNotifications(object runtime.Object, opts Eve
...
+			res, err := notificationsAPI.RunTrigger(trigger, objMap)
+			if err != nil {
+				log.Errorf("Failed to execute condition of trigger %s: %v", trigger, err)
+				return err
+			}
+			log.Infof("Trigger %s result: %v", trigger, res)
+
+			for _, dest := range destinations {
+				for _, c := range res {
+					if c.Triggered == true {
```
leoluz, line 292: Im missing where is the connection between the RunTrigger results (`res`) and the current triggerAction (`ta`). It seems that if `res` has multiple entries, it will send duplicated notifications. Can you please confirm this logic?

leoluz (commented):
Asked a question as I would like to understand this logic a bit better.

utils/record/record.go
RaviHari: `Runtrigger` is used only for validating when condition.. if the condition doesnot match it returns with `res` with `Triggered` in it as false and that will not send notification. `res` here holds the template that needs to be sent where as trigger actions holds the condition.

beyond that the logic of execution is same as earlier.

leoluz: @RaviHari Ok. 
But what should happen if `res` has 2 entries: one with `Triggered == false` and the other with `Triggered == true`?
It seems to me that it should only send the notification if all entries are `true` isn't it?

RaviHari: ok so I validated it with multiple when conditions and a corresponding send template for each condition..
Then the result object has twice as many entries/elements in the list that of the conditions.. and the result object is going awry.. and the send notification fails.

trigger_action
Res1
Res2
Res3
Res4

as it sees the third element in the list is true it tries to send notification but fails with:
```
{"ok":false,"error":"no_text"}  service=slack
ERRO[0504] notification error: no_text                  
DEBU[0504] Notification sent                             namespace=rollout-canary rollout=rollouts-demo time_ms=439906.596329
```
as the template for that entry is nil.  I will need to discuss this further to see the way forward.

leoluz (approved):

PR #2301, opened 2022-10-07
fix(webmetrics): Support jsonBody for web metric provider Fixes #2275

### LINKED ISSUE #2275: Suport jsonBody for web metric provider
# Summary

Add support to web metric provider to make posting json easier. We would do this by adding a new field called `jsonBody`.

Today in order to post json to a web metric provider to POST data you end up with a metric configuration like this
```
metrics:
  - name: webmetric
    successCondition: result == true
    provider:
      web:
        method: POST # valid values are GET|POST|PUT, defaults to GET
        url: "http://my-server.com/api/v1/measurement?service={{ args.service-name }}"
        timeoutSeconds: 20 # defaults to 10 seconds
        headers:
          - key: Authorization
          - key: Content-Type # if body is a json, it is recommended to set the Content-Type
            value: "application/json"
        body: "{\"key\": \"string value\"}"
        jsonPath: "{$.data.ok}"
```

The body field being just a string that gets posted as the body in the http request.

I propose to add something like `jsonBody` which would make constructing json payloads cleaner

```
metrics:
  - name: webmetric
    successCondition: result == true
    provider:
      web:
        method: POST # valid values are GET|POST|PUT, defaults to GET
        url: "http://my-server.com/api/v1/measurement?service={{ args.service-name }}"
        timeoutSeconds: 20 # defaults to 10 seconds
        headers:
          - key: Authorization
          - key: Content-Type # if body is a json, it is recommended to set the Content-Type
            value: "application/json"
        jsonBody:  # If using jsonBody we could also auto set Content-Type headers
          key1: value_1
          key2:
            nestedObj: nested value
            key3: {{ args.api-token }}
        jsonPath: "{$.data.ok}"
```

I am not sure if today we support parsing templated args into body or not but it would be very valuable to do that if we do not as well.

# Use Cases

When would you use this?

---

premadk:
Signed-off-by: Prema devi Kuppuswamy <premadk@gmail.com>

Checklist:


premadk:
opened new pr: https://github.com/argoproj/argo-rollouts/pull/2312

Issue #2233, opened 2022-09-07
Experiment create service with hardcoded port. 


jeromeguiard:
**Describe the bug**

Using experiment in rollout step, it is expected to have a service created to route traffic to it. as describe in this doc: https://github.com/argoproj/argo-rollouts/blob/master/docs/features/experiment.md 

``` 
When a weighted experiment step with traffic routing is used, a service is auto-created for each experiment template. The traffic routers use this service to send traffic to the experiment pods.
```

Created service related to the experiment have port 8080 as targetPort

**To Reproduce**

Start an experiment, in a rollout canary step, set weight for stable and canary. Once experiment start describe created services.

**Expected behavior**

We should have the ability to add service configuration, to have, on ports, the ability to specify Port, TargetPort, Name, multiple ports.

**Screenshots**

from experiments/service.go 

<img width="1279" alt="Screenshot 2022-09-05 at 21 54 32" src="https://user-images.githubusercontent.com/555535/188919531-e4b440a9-3733-401f-876a-3e56280ee783.png">

**Version**

1.2, 1.3

**Logs**

Controller doesn't show error

jessesuen:
Yes, this is a bit of an embarrassing bug. I am pretty sure @zachaller is right and `TemplateService` was intended to be used to give full control on the service that was created.

alexef:
@jessesuen / @zachaller I assume that the fix for this would be to follow up on https://github.com/argoproj/argo-rollouts/pull/1158 and infer the servicePorts from the Rollout.spec.template.containers, or additionally allow the user to specify the port.

We have an additional use case for this - we use experiments to create a separate workload (different pod labels) that does not receive production traffic (therefore we don't set weight). We would like Argo Rollouts to generate the service in that case as well.

I can start the implementation of the bug fix. Please let me know what you think about also implementing the additional functionality to generate the service on request (via step definition) and not only when weight is set.

zachaller:
So we have talked about this a bit and it is a bug however I think the fix we want to implement is to embed the whole service spec into the TemplateService something like:

```
type TemplateService struct{
	corev1.ServiceSpec
}
```

I think we might want to maintain control of the metadata like auto generating names etc. Do you see any issues with this approach? I feel it will be a lot more future proof this way and provide full customization.

alexef:
Thanks for the link to #2257, I missed it.

[...]

PR #2357, opened 2022-10-21
feat(controller): allow setting the Experiment's service name; don't hardcode ports

alexef:
Fixes: #2233 

Stop hardcoding service port.
Allow user to set the experiment service name.
Create a service even though a weight is not set (when the name is provided)

Example:
```
...
steps:
- experiment:
     templates: 
     - name: experiment-baseline
       specRef: stable
       service: 
          name: my-experiment # optional
          ports:
          - port: 80
            targetPort: 8080
            protocol: TCP
```

This is a follow up of https://github.com/argoproj/argo-rollouts/pull/2356, see also [this comment](https://github.com/argoproj/argo-rollouts/issues/2233#issuecomment-1286900573) for context.

Checklist:


alexef:
@zachaller can you have a look at this?

experiments/service.go
```diff
@@ -73,12 +72,8 @@ func (ec *experimentContext) CreateService(serviceName string, template v1alpha1
...
 		},
 		Spec: corev1.ServiceSpec{
-			Selector: selector,
-			Ports: []corev1.ServicePort{{
-				Protocol:   "TCP",
-				Port:       int32(80),
-				TargetPort: intstr.FromInt(8080),
-			}},
+			Selector: template.Service.Selector,
+			Ports:    template.Service.Ports,
```
zachaller, line 76: Why are we not setting the Spec whole templates.Service here?

alexef: Exactly, I couldn’t figure out a way to do it. Happy to learn

zachaller (approved):

zachaller:
One last thing then I think this is good could you add some documentation and update docs about when a service is generated etc I think mainly around this page https://argoproj.github.io/argo-rollouts/features/experiment/ and maybe on the spec https://argoproj.github.io/argo-rollouts/features/specification/

alexef:
Updated the docs, not sure if it's okay: https://github.com/argoproj/argo-rollouts/pull/2357/commits/0f5ed7afb764354148e3842d923e348dfb2f696a

Regarding service spec, we should probably also mark this in the docs as available from next release?

alexef:
superseeded by: https://github.com/argoproj/argo-rollouts/pull/2397

Issue #2015, opened 2022-05-03
shell completion: `kubectl-argo-rollouts get rollout TAB` does not auto completes the rollout names


thomas-riccardi:
## Summary 

I'm following the getting started guide on linux with auto completion on bash (https://argoproj.github.io/argo-rollouts/getting-started/).

When I hit TAB after `kubectl-argo-rollouts get rollout`, it does not auto completes the rollout names as I would expect. (for example `kubectl get pod TAB` does auto completes with existing pod names.)

It would be really useful for discovery of what's inside the cluster.

## Diagnostics

What version of Argo Rollouts are you running?

v1.2.0 server and kubectl plugin, on linux, bash.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

zachaller:
I will take this

thomas-riccardi:
I've started working on it in #2379, with several questions there

PR #2379, opened 2022-10-27
WIP feat(cli): PoC dynamic shell completion for 'kubectl-argo-rollouts get rollout [TAB]'

thomas-riccardi:
works toward #2015

use ValidArgsFunction from cobra to list candidates for completion: https://github.com/spf13/cobra/blob/main/shell_completions.md#dynamic-completion-of-nouns
- filtering on 'toComplete' doesn't seem necessary, at least on bash (most probably bash-completion handles that itself), not sure for the others; I could not find cobra documentation about that

use v2 of GenBashCompletion from cobra: https://github.com/spf13/cobra/blob/main/shell_completions.md#bash-completion-v2
- I chose to disable descriptions for completion (as a bash user I'm not used to that), but it's enabled for fish it seems, we can enable it if desired, I have no strong opinion on it

If this is validated, I will try to generalize this to all commands.
Do you have commands to implement in priority?

Regarding unit-tests, should I go the way it was done in kubectl when they migrated to go completion https://github.com/kubernetes/kubernetes/pull/96087? (they were able to factorize most `ValidArgsFunction`s and do some test on those, but not on custom ones it seems).
Are such unit-tests blocking for merging this PR?

Checklist:


zachaller:
> filtering on 'toComplete' doesn't seem necessary, at least on bash (most probably bash-completion handles that itself), not sure for the others; I could not find cobra documentation about that

I am maybe missing what exactly this would get us. I am a fish user and things seem to also be working as I would expect with the first completion you added in your WIP PR so I think it would be ok to not do this as well.


>I chose to disable descriptions for completion (as a bash user I'm not used to that), but it's enabled for fish it seems, we can enable it if desired, I have no strong opinion on it

Yes for fish this make sense to leave on and I think it is generally expected and nice as far as bash I am also impartial as a non bash user, I am fine leaving it to false and seeing if a request comes in later to change etc.

>Regarding unit-tests, should I go the way it was done in kubectl when they migrated to go completion https://github.com/kubernetes/kubernetes/pull/96087? (they were able to factorize most ValidArgsFunctions and do some test on those, but not on custom ones it seems).
Are such unit-tests blocking for merging this PR?

I took a quick look at the kubectl testing and this would also work well here. With some of the stances within ArgoProj coming up around PR acceptance criteria having unit tests would be ideal as well.

Also thank you for contributing and working on this feature it will be really nice to have!

thomas-riccardi:
> > filtering on 'toComplete' doesn't seem necessary, at least on bash (most probably bash-completion handles that itself), not sure for the others; I could not find cobra documentation about that
> 
> I am maybe missing what exactly this would get us. I am a fish user and things seem to also be working as I would expect with the first completion you added in your WIP PR so I think it would be ok to not do this as well.

The cobra spec I could find is unclear: maybe it's mandatory we do the filtering in go. But it doesn't seem to be the case, thanks to your fish test.
=> good for me


> 
> > I chose to disable descriptions for completion (as a bash user I'm not used to that), but it's enabled for fish it seems, we can enable it if desired, I have no strong opinion on it
> 
> Yes for fish this make sense to leave on and I think it is generally expected and nice as far as bash I am also impartial as a non bash user, I am fine leaving it to false and seeing if a request comes in later to change etc.

=> good for me

> 
> > Regarding unit-tests, should I go the way it was done in kubectl when they migrated to go completion [kubernetes/kubernetes#96087](https://github.com/kubernetes/kubernetes/pull/96087)? (they were able to factorize most ValidArgsFunctions and do some test on those, but not on custom ones it seems).
> > Are such unit-tests blocking for merging this PR?
> 
> I took a quick look at the kubectl testing and this would also work well here. With some of the stances within ArgoProj coming up around PR acceptance criteria having unit tests would be ideal as well.

It will take me more time to complete that, I'm not sure I can finish before the end of hacktoberfest.
For a first PR, would you prefer a dynamic completion for a restricted subset of the commands, with unit tests, or more commands but no unit-tests? Or neither are acceptable for merge?

thomas-riccardi:
Current state:
- factorized the completion functions
  - in new util hierarchy: `pkg/kubectl-argo-rollouts/util/completion/completion.go`, following what was done in kubectl; let me know if I should move that elsewere
  - per resource type; I could not find an easy way to factorize all types (I'm new in golang)
- implemented completion and used them everywhere
  - by grepping into the auto-generated CLI doc to find all the usages (so maybe I missed some if the `Use` strings are not all correct)
  - for 3 types: Rollout, Experiment, AnalysisRun
    - missing 2 types related to notifications, but for that I need to patch in another git repo: https://github.com/argoproj/notifications-engine/blob/master/pkg/cmd/trigger.go => this will be for another PR
- no unit-test for now, although with the partial factorization it should be tractable
- it seems the cli `Use` strings are not showing when the command supports multiple resources, at least for `abort`: for now only the first resource name is completed; I could support more but we should probably start by documenting those cases in the `Use` and maybe `Short`?
- no bash default completion on file disabled on the other commands (cf `kubectl-argo-rollouts __complete lint ""`: `Completion ended with directive: ShellCompDirectiveDefault`): I haven't yet read the doc/usages to know if we can disable it by default or if we need to add it everywhere.
- no dynamic completion on flags for now, such as `create -f` or `lint -f`
- no dynamic completion on `set image` second arg: `CONTAINER=IMAGE` for now

zachaller:
I think doing a dynamic completion for a restricted subset of the commands with tests would be prefered for hacktoberfest. Even if you don't finish the PR 100% completely but it is close I don't mind adding the label to it if you need a bit more time to finish up after as well.

>in new util hierarchy: pkg/kubectl-argo-rollouts/util/completion/completion.go, following what was done in kubectl; let me know if I should move that elsewere

I think this is a good location for this package

thomas-riccardi:
> I think doing a dynamic completion for a restricted subset of the commands with tests would be prefered for hacktoberfest. Even if you don't finish the PR 100% completely but it is close I don't mind adding the label to it if you need a bit more time to finish up after as well.

I started looking into the tests (I'm thinking about reusing `pkg/kubectl-argo-rollouts/info/testdata`), but I'm not sure I'll have a ready PR before end of month. I'd like the label please :)

thomas-riccardi:
I've pushed unit tests, directly in the new `completion` pacakge, like it was done in kubectl (https://github.com/kubernetes/kubernetes/pull/96087, as well as a later refactor (https://github.com/kubernetes/kubernetes/pull/108493/commits/c4f8c57b43843abd13bae677b73c622e1c30b46f)).

- fake cmd to avoid circular import
- reuse info/testdata but we could build our own data for more isolated tests if needed
- I don't really understand what I'm doing with all the initializations in the tests (mainly `tf`): it's a mix of what I found in argo-rollouts and kubectl tests , but it seems to work...

thomas-riccardi:
I've completed the PR description: the PR is ready for review from my PoV.

zachaller:
Could you please try to just trigger tests again with a command something like `git commit -s -S --allow-empty -m "github trigger re-run"` or just push some commit I am not sure why e2e tests where failing want to make sure its a true failure and not just a github issue etc etc.

zachaller:
I actually think the e2e errors are because of not pinning a k8s version in e2e tests, I will work on getting that fixed.

thomas-riccardi:
> I actually think the e2e errors are because of not pinning a k8s version in e2e tests, I will work on getting that fixed.

`TestFunctionalSuite/TestRolloutPDBRestart` also failed in my other PR #2375

when working on the multi k8s version e2e tests PR (https://github.com/argoproj/argo-rollouts/pull/2380) I also have some tests failing, but different ones, and only on k8s 1.23 (not sure if those are hard or flaky tests..)

zachaller:
There are quite a bit of flaky tests but they do retry, I got the multi k8s versions PR merged can you trigger a retest on this PR

zachaller (approved):

PR #2394, opened 2022-11-03
feat: rollback windows. Fixes #574

### LINKED ISSUE #574: Rollback windows for fast-tracked rollbacks
Spawned from https://github.com/argoproj/argo-rollouts/issues/557.

Currently we perform a "fast-tracked rollback" (which skips pauses, steps, analysis) in two circumstances:
1. we detect if we are moving back to a blue-green ReplicaSet which exists and is still scaled up (within its scaleDownDelay)
2. if we are moving back to the canary's "stable" ReplicaSet and the upgrade has **not yet completed**.

I think we should add more controls for intelligent, fast-tracked rollbacks for both the blue-green and canary strategies. For example, a use case is to have a fast-tracked rollback happen when moving to an older blue-green ReplicaSet even if it's scaled down (and not just if it's in the scaleDownDelay).

I think this can either be time based:

```yaml
spec:
  rollbackWindow:
    duration: 24h
  strategy:
    blueGreen:
...
```

Or it could possibly be revision based (e.g. fast-rollback if we are moving to an n-2 revision):

```yaml
spec:
  rollbackWindow:
    revisions: 2
  strategy:
    canary:
...
```

I propose to even place the `rollbackWindow` stanza outside strategy since both blue-green and canary would benefit from it.

alexef:
fixes: #574 

Checklist:


alexef:
@jessesuen here's my first attempt at this. I went for the revision count option, the time based one can also be added later on

zachaller:
Could you also add some e2e tests for this change? I would like to get the behavior tested there as well.

alexef:
> add some e2e tests for this change

will do. I also need to make sure unit tests cover changes so that build is green

rollout/sync.go
```diff
@@ -857,6 +857,34 @@ func (c *rolloutContext) resetRolloutStatus(newStatus *v1alpha1.RolloutStatus) {
...
+	if c.newRS == nil || c.stableRS == nil {
+		return false
+	}
+	// first check if this is a rollback
+	if c.newRS.CreationTimestamp.Before(&c.stableRS.CreationTimestamp) {
+		// then check if we are within window
+		if c.rollout.Spec.RollbackWindow != nil {
+			if c.rollout.Spec.RollbackWindow.Revisions > 0 {
+				var windowSize int32
+				for _, rs := range c.allRSs {
```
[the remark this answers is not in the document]
alexef: good question. I'll check now if experiments' RSs are part of allRSs and if yes, I need to filter them out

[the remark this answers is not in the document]
alexef: yup, seems like it. I'll add a test and exclude them

zachaller:
Can we also get a little bit of documentation on this as well? I will get this reviewed this week as well.

alexef:
@zachaller / @jessesuen can you have another look at this?
We identified the issue that was preventing the windows to work for us, updated the code and the e2e and I believe now the feature is ready.

rollout/sync.go
```diff
@@ -857,6 +857,38 @@ func (c *rolloutContext) resetRolloutStatus(newStatus *v1alpha1.RolloutStatus) {
 	newStatus.CurrentStepIndex = replicasetutil.ResetCurrentStepIndex(c.rollout)
 }
 
+func (c *rolloutContext) isRollbackWithinWindow() bool {
+	if c.newRS == nil || c.stableRS == nil {
+		return false
+	}
+	// first check if this is a rollback
+	if c.newRS.CreationTimestamp.Before(&c.stableRS.CreationTimestamp) {
```
zachaller, line 865: In playing around with this there is interesting behavior when going back and forward within a couple of replica sets because if we are rolling back to a replicaset that exists already it does not update the created timestamp. I wonder if it makes more sense to use this annotation `rollout.argoproj.io/revision: "9"`  to determine if we are within range instead. It dose change the behavior slightly in that it would not let you jump back multiple "revisions" which also could be a good thing.

alexef: We had a similar question internally, but I would prefer to stick with the creation timestamp, because it normally follows the build order. Jumping back too many builds is not an issue, as:
- this is only meant to be used during "incidents", when it's critical to get to a previously known good known version as fast as possible
- we are anyway limited by the revisionHistoryLimit which is a number between 5-10

If we used the revision annotation instead of the creation timestamp, then the build order is not taken into account any more, and I can imagine a scenario where we fast rollback to a build that is "newer" than the current stable one (because the revision is mutated every time a build is promoted).

For instance, imagine:
- build 1 - rev 1
- build 2 - rev 2
- build 3 - rev 3
- build 4 - rev 4
window size: 2

with Revision Annotation:
- rollback to 2 (fast), now build 2 - rev 5 (history becomes: 1, 3, 4, 5)
- rollback to 3 (fast), now build 3 - rev 6 (history becomes: 1, 4, 5, 6) - this is already wrong, as build 3 is "newer" than build 2, so it may contain changes that broke the deployment

with Creation timestamp:
- rollback to 2 (fast)
- rollback to 3 (regular deployment)

zachaller: Yup that makes sense and thank you for the detailed explanation.

zachaller (approved):

PR #2397, opened 2022-11-04
feat(controller): don't hardcode experiment ports; always create service

alexef:
superseeds: #2357 

fixes: #2233
Checklist:


docs/features/experiment.md
```diff
@@ -243,5 +245,9 @@ to `experiment-baseline`, leaving the remaining 90% of traffic to the old stack.
 
 !!! note
     When a weighted experiment step with traffic routing is used, a
-    service is auto-created for each experiment template. The traffic routers use
-    this service to send traffic to the experiment pods.
\ No newline at end of file
+    Service is auto-created for each experiment template. The traffic routers use
+    this service to send traffic to the experiment pods.
+
+By default, the generated Service has the name of the ReplicaSet and inherits
```
alexef, line 251: needs this https://github.com/argoproj/argo-rollouts/pull/2396/files to be merged first

experiments/service.go
```diff
@@ -74,11 +73,7 @@ func (ec *experimentContext) CreateService(serviceName string, template v1alpha1
 		},
 		Spec: corev1.ServiceSpec{
 			Selector: selector,
-			Ports: []corev1.ServicePort{{
```
alexef, line 77: this was #2233

docs/features/experiment.md
```diff
@@ -6,6 +6,8 @@ The Experiment CRD allows users to have ephemeral runs of one or more ReplicaSet
 running ephemeral ReplicaSets, the Experiment CRD can launch AnalysisRuns alongside the ReplicaSets.
 Generally, those AnalysisRun is used to confirm that new ReplicaSets are running as expected.
 
+A Service routing traffic to the Experiment ReplicaSet is also generated.
```
alexef, line 9: FIXME: this doesn't hold true

docs/features/experiment.md
zachaller: has been merged

zachaller:
In regards to the e2e tests, can you confirm for me that if we modify the code to not create the service if there is no ports defined on the pod spec. I think that should fix all the e2e tests and not require a port on the container which I think we should support

alexef:
>  if we modify the code to not create the service if there is no ports defined on the pod spec

correct, I forgot about that. let me do this change

zachaller (approved):

rollout/experiment.go
```diff
@@ -64,9 +64,7 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
 			Name:     templateStep.Name,
 			Replicas: templateStep.Replicas,
 		}
-		if templateStep.Weight != nil {
```
alexef, line 67: Feels like we can use weight: 0 to keep creating the service, if automated creation is rolled back

PR #2407, opened 2022-11-11
fix: add patch verb to deployment resource

tukak:
Dashboard fails with ```failed restoring revision 21: deployments.apps "ourapp" is forbidden: User "system:serviceaccount:argo-rollouts:argo-rollouts-dashboard" cannot patch resource "deployments" in API group "apps" in the namespace "ournamespace"``` when trying to rollback `Rollout` defined using existing `Deployment` and `workloadRef` config. Adding `patch` verb for `deployments` should fix it.

Checklist:

### DIFF
--- a/manifests/dashboard-install.yaml
+++ b/manifests/dashboard-install.yaml
@@ -59,6 +59,7 @@ rules:
   - update
   - list
   - watch
+  - patch
 - apiGroups:
   - apps
   resources:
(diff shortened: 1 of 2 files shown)


zachaller:
Can you run `make codegen` that is why the test is not passing

tukak:
@zachaller thanks for the hint, it passes now

PR #2481, opened 2022-12-15
chore: fix checksum generation

34fathombelow:

This PR fixes the checksum generation done during a release.  Also included SBOM to be signed. Cosign public key added for public record.

Checklist:


zachaller (approved):

zachaller:
Just an fyi, I removed the public cosing key, when I talked to Jesse about this and I think it makes sense as well we did not want to really publish a pub key outside of the release. The reason for this is if we ever need to rotate the private key if users are used to having to pull public keys from the release assets only it makes it really easy to rotate without having to have a notice of the leak and breaking newer releases from being checked if they are always just pulling the pub key from the release they will always have the correct key even if we switch. Do you see any issue with that?

PR #2489, opened 2022-12-16
feat: Adding SigV4 Option for Prometheus Metric Analysis

### LINKED ISSUE #2458: Prometheus Analysis SigV4 Support
# Summary

Prometheus analysis should support SigV4 when creating the prometheus client. This would allow it to query Amazon Managed Prometheus, which requires [SigV4](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html) authentication.

The prometheus client already supports a [SigV4 RoundTripper](https://pkg.go.dev/github.com/prometheus/common/sigv4). This can be used when constructing the client to allow a user to authenticate with Amazon Managed Prometheus using the default GO credential provider chain.

# Use Cases

When performing prometheus analysis using Amazon Managed Prometheus as a metric data store.

---

lewinkedrs:
Fixes #2458 

Checklist:

* [ a ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [  b ] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [ X ] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)
* [ N/A] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ X ] My builds are green. Try syncing with master if they are not. 
* [ Not a User, building for AMP Community ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).


metricproviders/prometheus/prometheus.go
```diff
@@ -162,21 +165,47 @@ func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
...
 	}
 	log.Debugf("Parsed url: %v", u)
 	return err == nil && u.Scheme != "" && u.Host != ""
 }
+
+//Create a new sigv4Client that uses GO default provider chain
+func CreateSigV4Client (address string) (api.Client, error) {
+	 var cfg *sigv4.SigV4Config
+	 var next http.RoundTripper
+	 sigv4RoundTripper, err := sigv4.NewSigV4RoundTripper(cfg,next)
```
zachaller, line 205: Can we fix this ineffectual assignment

metricproviders/prometheus/prometheus.go
```diff
@@ -4,14 +4,17 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
+	"strings"
```
zachaller, line 9: Can we run make lint

zachaller (changes requested):

metricproviders/prometheus/prometheus.go
```diff
@@ -162,21 +165,47 @@ func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
...
 	u, err := url.Parse(str)
 	if err != nil {
 		log.Errorf("Error in parsing url: %v", err)
 	}
 	log.Debugf("Parsed url: %v", u)
 	return err == nil && u.Scheme != "" && u.Host != ""
 }
+
+//Create a new sigv4Client that uses GO default provider chain
+func CreateSigV4Client (address string) (api.Client, error) {
```
zachaller, line 202: This probably dose not need to be public? Unless we need to for tests

metricproviders/prometheus/prometheus.go
```diff
@@ -162,21 +165,47 @@ func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
...
-	client, err := api.NewClient(api.Config{
-		Address: metric.Provider.Prometheus.Address,
-	})
-	if err != nil {
-		log.Errorf("Error in getting prometheus client: %v", err)
-		return nil, err
+
+	//Check if using Amazon Managed Prometheus if true build sigv4 client
+	promUrl := metric.Provider.Prometheus.Address
+	amznUrlSubstring := "aps-workspaces"
```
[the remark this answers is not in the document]
jeromeinsf: that's not something expected to change. Would break too many customers.
But you maybe be more precise in the check based on https://docs.aws.amazon.com/general/latest/gr/prometheus-service.html

> dataplane endpoints are in the form aps-workspaces.region.amazonaws.com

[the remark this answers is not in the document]
jeromeinsf: Alternatively, you could have the auth a configuration?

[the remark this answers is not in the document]
lewinkedrs: Changed substring to more specific endpoint "aps-workspaces.region.amazonaws.com"

metricproviders/prometheus/prometheus.go
lewinkedrs: Added handling for the error.

metricproviders/prometheus/prometheus.go
lewinkedrs: make lint ran

zachaller:
Can I have you update from master and fix conflicts?

metricproviders/prometheus/prometheus.go
```diff
@@ -162,14 +165,28 @@ func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
...
-	client, err := api.NewClient(api.Config{
-		Address: metric.Provider.Prometheus.Address,
-	})
-	if err != nil {
-		log.Errorf("Error in getting prometheus client: %v", err)
-		return nil, err
+
+	//Check if using Amazon Managed Prometheus if true build sigv4 client
+	promUrl := metric.Provider.Prometheus.Address
+	amznUrlSubstring := "aps-workspaces.region.amazonaws.com"
```
zachaller, line 171: Based on there docs I am seeing urls such as `aps-workspaces.${AWS_REGION}.amazonaws.com` so I think what you had before made more sense and just checking for aps-workspaces I was just hoping for something more robust however I am ok with just using `aps-workspaces` if that makes sense for now

zachaller: Just saw some of the urls from here https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-onboard-ingest-metrics-existing-Prometheus.html

lewinkedrs:
Merged updates from master and fixed merged conflicts. Set AMP substring match back to 'aps-workspaces'. Not sure why the codecov/patch CI check is failing, do we need to do anything more with that?

zachaller:
> Merged updates from master and fixed merged conflicts. Set AMP substring match back to 'aps-workspaces'. Not sure why the codecov/patch CI check is failing, do we need to do anything more with that?

Yes it because you added a bunch of code that has no unit tests to cover the newly added code. Looking at your changes a bit more you could improve the coverage a bit more by refactoring your if logic so that you do not duplicate so much code. Bad sudo code below but gets the idea across, would you mind trying to update the logic to follow that pattern?

```
client, err := api.NewClient(api.Config{
	Address: metric.Provider.Prometheus.Address,
})
if err != nil {
	log.Errorf("Error in getting prometheus client: %v", err)
	return nil, err
}

if strings.Contains(promUrl, amznUrlSubstring) {
		roundTripper, err := createSigV4Client(promUrl)
                client.RoundTripper = roundTripper
}
return v1.NewAPI(client), nil
```

lewinkedrs:
Refactored the code to improve test coverage. It looks like tests are now failing but the failure is in:

```
/home/runner/work/argo-rollouts/argo-rollouts/rollout/controller_test.go:1952
```

This is a completely different section of the code. Looking through the error I can't seem to trace it back to anything that was done by creating our prometheus sigv4 client. Any ideas?

zachaller (approved):

PR #2491, opened 2022-12-19
feat(analysis): add Apache SkyWalking as metrics provider

kezhenxu94:
Checklist:


kezhenxu94:
Tests seem to be unstable, it can pass in my local machine

harikrongali:
https://github.com/argoproj/argo-rollouts/pull/2514 solves supporting additional metrics providers without adding them to rollout source. @zachaller can confirm if we need this PR?

zachaller:
> #2514 solves supporting additional metrics providers without adding them to rollout source. @zachaller can confirm if we need this PR?

Even though plugin system has just been merged it also is not 100% complete yet though it is close. I still do plan and think that merging this is the correct thing to do for a few reasons. One being that a good majority of users of this metric provider will be APISIX traffic router users the two projects are both under the apache foundation and share the same user base. 

We have already merged APISIX before any plugin systems where in place and this PR was also submitted way before any plugin systems where in place as well. Therefore I do think it makes sense to allow this PR through however this will probably be the last one that goes straight into the main project. 

I do think though that even with plugins having some form of graduation type system in place where plugins can gain enough traction and or meet some requirements that they move into "core" project however all that has not really been decided yet.

zachaller (approved):

kezhenxu94:
@zachaller thank you for helping to resolve the conflicts and merging the pr!

PR #2545, opened 2023-01-27
docs: Add best practice for reducing memory usage

blkperl:
Signed-off-by: William Van Hevelingen <william.vanhevelingen@acquia.com>

Checklist:

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -52,3 +52,12 @@ spec:
 
 The above technique has the a benefit in that it would not incur additional cost of allocating
 additional load balancers.
+
+## Reducing operator memory usage
+
+On clusters with thousands of rollouts memory usage for the argo-rollouts
+operator can be reduced significantly by changing RevisionHistoryLimit from the
+default of 10 to a lower number. One user of Argo Rollouts saw a 27% reduction
+in memory usage for a cluster with 1290 rollouts by changing
+RevisionHistoryLimit from 10 to 0.
+


zachaller (approved):

blkperl:
@zachaller Thanks for the review! Let me know if you have questions or suggestions on changing the wording.  

We could also mention that operators may wish to do the same for Deployments as the argo-controller watches all ReplicaSets on the clusters. One way to avoid this is to use NewCache which I don't think the controller uses. I can open a new issue for this if you think that makes sense. The tigera operator uses NewCache with a filter for NetworkPolicies if you are looking for an example.

- https://github.com/tigera/operator/blob/master/main.go#L231-L235
- https://github.com/argoproj-labs/argocd-operator/blob/master/main.go#L123
- https://master.sdk.operatorframework.io/docs/best-practices/designing-lean-operators/

zachaller:
Yea I would love a new issue for looking into using that cache to get an informer that is more limited in scope that say all replicasets to reduce memory usage. Rollouts was created before a lot of the kubebuilder stuff so there is quite a few things done a little lower level but I took a quick peak at that cache package and it looks like I can get an informer still off of it so I would like to spend time looking into that and would use the issue to help track that.

zachaller:
I think rollouts might need to do something like this instead of the cache

```
kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(
				kubeClient,
				resyncDuration,
				kubeinformers.WithNamespace(namespace), kubeinformers.WithTweakListOptions(func(options *metav1.ListOptions) {
					options.LabelSelector = "app=nats-box"
				}))
```

The key being WithTweakListOptions

blkperl:
Thanks, I filed #2552

PR #2572, opened 2023-02-07
feat: Add name attribute to ServicePort

daniddelrio:
We encountered this error when we tried to run an Experiment for a service with multiple named ports:
```
Failed to create Service for template 'smoketest': Service
    is invalid: [spec.ports[0].name: Required value,
    spec.ports[1].name: Required value, spec.ports[2].name:
    Required value]
```

Argo Rollouts currently only copies the port number and protocol from the Rollout to the service it creates during experiments, but not the name. This PR addresses this issue by also including the port name when copying.

Relates to: https://github.com/argoproj/argo-rollouts/pull/2397

Checklist:


experiments/experiment.go
```diff
@@ -290,6 +290,7 @@ func (ec *experimentContext) createTemplateService(template *v1alpha1.TemplateSp
 	for _, ctr := range rs.Spec.Template.Spec.Containers {
 		for _, port := range ctr.Ports {
 			servicePort := corev1.ServicePort{
+				Name:       port.Name,
```
alexef, line 293: I believe ports can also be unnamed. what is the default value here, is it compatible? (ideally type's default which is "")

test/e2e/functional/experiment-with-service.yaml
```diff
@@ -26,5 +26,6 @@ spec:
                   memory: 16Mi
                   cpu: 1m
               ports:
-                - protocol: TCP
+                - name: testport
```
alexef, line 29: does the test fail when name is not present? can we rather have two tests, one with single, unnamed port, another one with two named ports?

experiments/experiment.go
daniddelrio: [Documentation](https://pkg.go.dev/k8s.io/api@v0.26.1/core/v1#ContainerPort) says it will be omitted if the name is not given. Should we make it by default an empty string if not given?

test/e2e/functional/experiment-with-service.yaml
daniddelrio: Yes, test fails when name isn't present. I'll try that out

alexef: okay, so let's make sure the test doesn't fails without name first

experiments/experiment.go
alexef: I would not set it at all

experiments/experiment_test.go
```diff
@@ -70,6 +70,7 @@ func setExperimentService(template *v1alpha1.TemplateSpec) {
 	template.Service = &v1alpha1.TemplateService{}
 	template.Template.Spec.Containers[0].Ports = []corev1.ContainerPort{
 		{
+			Name:          "testport",
```
alexef, line 73: this is not needed any more

alexef (approved):

test/e2e/experiment_test.go
```diff
@@ -99,6 +99,23 @@ func (s *ExperimentSuite) TestExperimentWithServiceAndScaleDownDelay() {
 		ExpectExperimentServiceCount("experiment-with-service", 0)
 }
 
+func (s *ExperimentSuite) TestExperimentWithMultiportServiceAndScaleDownDelay() {
+	g := s.Given()
+	g.ApplyManifests("@functional/experiment-with-multiport-service.yaml")
+	g.When().
+		WaitForExperimentPhase("experiment-with-multiport-service", "Running").
+		Sleep(time.Second*5).
```
zachaller, line 107:
```suggestion
                WaitForExperimentCondition("experiment-with-service", func(ex *rov1.Experiment) bool {
			return s.GetReplicaSetFromExperiment(ex, "test").Status.Replicas == 1
		}, "number-of-rs-pods-meet", fixtures.E2EWaitTimeout).
```

test/e2e/experiment_test.go
```diff
@@ -99,6 +99,23 @@ func (s *ExperimentSuite) TestExperimentWithServiceAndScaleDownDelay() {
...
+		Then().
+		ExpectExperimentTemplateReplicaSetNumReplicas("experiment-with-multiport-service", "test", 1).
+		ExpectExperimentServiceCount("experiment-with-multiport-service", 1).
+		When().
+		WaitForExperimentPhase("experiment-with-multiport-service", "Successful").
+		Sleep(time.Second*15).
```
zachaller, line 113:
```suggestion
```suggestion
                WaitForExperimentCondition("experiment-with-service", func(ex *rov1.Experiment) bool {
			return s.GetReplicaSetFromExperiment(ex, "test").Status.Replicas == 0
		}, "number-of-rs-pods-meet", fixtures.E2EWaitTimeout).
```
```

test/e2e/experiment_test.go
daniddelrio: @alexef what do you think about this?

alexef: the proposed change makes sense, most of the time rs gets ready in less than 5 seconds

zachaller (approved):

zachaller:
Thanks for contributing!

Issue #2549, opened 2023-01-30
fix: datadog metric provider should use v2 endpoint


alexef:
# Summary

Datadog is moving away from the [legacy v1 api](https://docs.datadoghq.com/api/latest/metrics/#query-timeseries-points) to [the GA v2 one](https://docs.datadoghq.com/api/latest/metrics/#query-timeseries-data-across-multiple-products). Rollouts metric provider should also follow this api change.

Note that the payload is slightly different, and the query is now part of `attributes.queries.*`.

# Use Cases

Datadog imposes stricter limits on v1 vs v2 endpoints. This change will mitigate the 429 one may receive when running multiple AnalysisRuns with datadog as the metric source.

PR #2580, opened 2023-02-10
feat: retryable http client for datadog metric provider

### LINKED ISSUE #2578: Use exponential backoff on ratelimits with datadog provider
# Summary
When running multiple analysis with the datadog provider it's possible to hit rate limits with the v1/query API. These limits are returned in the headers on requests -- what we did to mitigate this was to increase the rate limit with datadog directly -- but it'd also be good to automatically retry on status code 429 with exponential backoff.

Somewhat related issue: https://github.com/argoproj/argo-rollouts/issues/2549

# Use Cases
When using Datadog as a metric source for AnalysisTemplates.

---

kaldorn:
Adds a retryable http.client that automatically handles 429 rate limit responses with exponential backoff.

Not 100% sure this is the best approach for this, open to suggestions.

related to: https://github.com/argoproj/argo-rollouts/issues/2549
closes: https://github.com/argoproj/argo-rollouts/issues/2578

Signed-off-by: kaldorn [nate.kaldor@outreach.io]

Checklist:


zachaller:
@kaldorn I merged the pr to add datadog v2 to master. I am ok with also having retries if you want to add it will require a rebase and addition to both v1 and v2, if you do not and v2 is enough for your rate limiting/errors issues etc we can close this.

PR #2586, opened 2023-02-11
fix: remove deprecated GetValueFromIntOrPercent method

fengshunli:
Signed-off-by: fengshunli <1171313930@qq.com>

Checklist:

### DIFF
--- a/utils/defaults/defaults.go
+++ b/utils/defaults/defaults.go
@@ -21,9 +21,9 @@ const (
 	// DefaultAnalysisRunUnsuccessfulHistoryLimit default number of unsuccessful AnalysisRuns to keep if .Spec.Analysis.UnsuccessfulRunHistoryLimit is nil
 	DefaultAnalysisRunUnsuccessfulHistoryLimit = int32(5)
 	// DefaultMaxSurge default number for the max number of additional pods that can be brought up during a rollout
-	DefaultMaxSurge = "25"
+	DefaultMaxSurge = "25%"
 	// DefaultMaxUnavailable default number for the max number of unavailable pods during a rollout
-	DefaultMaxUnavailable = "25"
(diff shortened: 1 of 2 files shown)


zachaller:
Docs from deprecated function:

```
GetValueFromIntOrPercent was deprecated in favor of GetScaledValueFromIntOrPercent. This method 
was treating all int as a numeric value and all strings with or without a percent symbol as a
percentage value. Deprecate
```

This is a change in behavior that needs to be noted, it is still the correct thing to do because the old behavior is confusing but going to make a BREAKING CHANGE note in the sqash commit to note the behavior change.

kostis-codefresh:
[Related to https://github.com/argoproj/argo-rollouts/pull/4729](https://github.com/argoproj/argo-rollouts/issues/4567)

kostis-codefresh (changes requested):
The fact that this is a breaking change was already mentioned by Zach and this is obviously very important.

However, I have an additional comment:

While this PR solves the problem itself, it silently consumes the error and never reports back if the user has done a mistake and entered "1" (i.e. an actual string). Or am I missing something here?

I think we should at least follow [Argo CD and expose the error to the user](https://github.com/argoproj/argo-cd/blob/96506c592f3bd54f332908ab30e245c4550b4752/applicationset/progressivesync/progressive_sync.go#L489-L497)

Ideally we should modify [ValidateRollout](https://github.com/argoproj/argo-rollouts/blob/180300b62a426a80f93cb3475a3055189a0f0b4d/pkg/apis/rollouts/validation/validation.go#L101) and show a validation error (which is similar to what the standard K8s deployment controller will do)

Otherwise not only we break backwards compability, but we also don't notify existing users that something wrong is happening with their rollout spec.

PR #2592, opened 2023-02-15
feat: Allow switching between Datadog v1 and v2. Fixes #2549

### LINKED ISSUE #2549: fix: datadog metric provider should use v2 endpoint

daniddelrio:
Since Datadog will deprecate the legacy v1 api, Rollouts should now use the v2 API version. Datadog imposes stricter rate limits on v1 because of this. This PR introduces an `apiVersion` field in the Datadog configuration that can allow users to toggle between [v1](https://docs.datadoghq.com/api/latest/metrics/?code-lang=python#query-timeseries-points) and [v2](https://docs.datadoghq.com/api/latest/metrics/?code-lang=python#query-timeseries-data-across-multiple-products). 

Addresses: #2549 
Relates to: #2580 

Slack name: Dani Del Rio

Checklist:


zachaller:
I have a few questions the upgrade makes sense I am just not sure we want to break backwards compatibility by removing v1. Does datadog have an end date set for when they will stop supporting v1? If that date is still pretty far out do you think it would make sense to support both versions via a config option?

daniddelrio:
> I have a few questions the upgrade makes sense I am just not sure we want to break backwards compatibility by removing v1. Does datadog have an end date set for when they will stop supporting v1? If that date is still pretty far out do you think it would make sense to support both versions via a config option?

I can't find anything on a timeline that Datadog will follow for stopping support of v1 completely, but there are already some endpoints that they've deprecated, like for [Dashboards](https://docs.datadoghq.com/dashboards/guide/dashboard-lists-api-v1-doc/?tab=python) and [Usage Attribution](https://docs.datadoghq.com/account_management/guide/usage-attribution-migration/).

The primary benefit for this is to mitigate the rate limit errors that some have been encountering. We personally have encountered this rate limit and there's also this #2578. We can introduce a config option, but I'm not sure if that'll be the best idea given that we'll eventually move away from v1 anyway.

zachaller:
> The primary benefit for this is to mitigate the rate limit errors that some have been encountering. We personally have encountered this rate limit and there's also this #2578. We can introduce a config option, but I'm not sure if that'll be the best idea given that we'll eventually move away from v1 anyway.

Yup I agree 100% I like retries in general but I also agree moving to v2 is a good thing I just don't want to force users to have to upgrade we need to be mindful of backward compatibility and if users are not ready to migrate we need to support both still.

I am thinking something like `apiVersion: v2` and we default to v1 and then have documentation the we recommend v2 for x reasons and we can start deprecating v1 over time as well and eventually switch to v2 as a default, thoughts?

```
provider:
      datadog:
        apiVersion: v2
        interval: 5m
        query: |
          sum:requests.error.count{service:{{args.service-name}}} /
          sum:requests.request.count{service:{{args.service-name}}}
```

zachaller:
I would also maybe add a log line that v1 is deprecated as well

daniddelrio:
I think that would be a good idea. I can try to implement that

daniddelrio:
@zachaller it seems that the checks are failing because of the generated files. I did see this but I wasn't sure about how to update them. How can I fix these?

zachaller:
> @zachaller it seems that the checks are failing because of the generated files. I did see this but I wasn't sure about how to update them. How can I fix these?

run `make codegen`

metricproviders/datadog/datadog.go
```diff
@@ -72,16 +95,34 @@ func (p *Provider) Run(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric) v1alph
...
+		p.logCtx.Warn("Datadog will soon deprecate their API v1. Please consider switching to v2 soon.")
+	}
+
+	route := "/api/v1/query"
+	if apiVersion == "v2" {
+		route = "/api/v2/query/timeseries"
+	}
+
+	// Add endpoint after getting the API version
+	url, _ = url.Parse(endpoint + route)
```
zachaller, line 124: Can we handle this error

metricproviders/datadog/datadog.go
```diff
@@ -173,6 +250,56 @@ func (p *Provider) parseResponse(metric v1alpha1.Metric, response *http.Response
...
+func (p *Provider) parseResponseV2(metric v1alpha1.Metric, response *http.Response) (string, v1alpha1.AnalysisPhase, error) {
+
+	bodyBytes, err := io.ReadAll(response.Body)
+
+	if err != nil {
+		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Received no bytes in response: %v", err)
+	}
+
+	if response.StatusCode == http.StatusForbidden || response.StatusCode == http.StatusUnauthorized {
+		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("received authentication error response code: %v %s", response.StatusCode, string(bodyBytes))
```
zachaller, line 262: Will this log any credentials?

metricproviders/datadog/datadogV1_test.go
```diff
@@ -33,8 +33,9 @@ func TestRunSuite(t *testing.T) {
 
 	ddProviderInterval10m := v1alpha1.MetricProvider{
 		Datadog: &v1alpha1.DatadogMetric{
-			Query:    "avg:kubernetes.cpu.user.total{*}",
-			Interval: "10m",
+			Query:      "avg:kubernetes.cpu.user.total{*}",
+			Interval:   "10m",
+			ApiVersion: "v1",
```
zachaller, line 38: Can we have a test that tests the default value as well, so that when it changes tests will also need updating etc.

zachaller: I am ok even just dropping setting v1 hear and only testing via the default path.

metricproviders/datadog/datadog.go
daniddelrio: Sorry, can you elaborate further? Not sure what you mean by this

metricproviders/datadog/datadog.go
daniddelrio: This shouldn't as this is just the same logic for parsing as in v1: https://github.com/argoproj/argo-rollouts/blob/master/metricproviders/datadog/datadog.go#L141

metricproviders/datadog/datadog.go
zachaller: you are sending the err to _ I would like to handle it, even though both endpoint and route variables are static and controlled in code to future proof say they become user configurable in the future etc it would be nice just to have the error handheld incase Parse fails for any reason.

metricproviders/datadog/datadogV1_test.go
daniddelrio: Commit: https://github.com/argoproj/argo-rollouts/pull/2592/commits/79112123f09c1e60d96227e595cb85aa0679e0e5

metricproviders/datadog/datadog.go
daniddelrio: Commit: https://github.com/argoproj/argo-rollouts/pull/2592/commits/2799dc91318264296f553a85d818781ce6b4ba0b

metricproviders/datadog/datadog.go
```diff
@@ -129,7 +172,45 @@ func (p *Provider) Run(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric) v1alph
...
+	if apiVersion == "v1" {
+		q := url.Query()
+		q.Set("query", query)
+		q.Set("from", strconv.FormatInt(now-interval, 10))
+		q.Set("to", strconv.FormatInt(now, 10))
+		url.RawQuery = q.Encode()
+
+		return &http.Request{Method: "GET"}, nil
+	} else if apiVersion == "v2" {
+		queryBody, _ := json.Marshal(datadogQuery{
```
zachaller, line 185: Can we handle this error in case user supplied query breaks marshaling.

daniddelrio: Commit: https://github.com/argoproj/argo-rollouts/pull/2592/commits/1e446dd612e3bffacde4ccb7eeab787c8cee4593

zachaller (approved):

Issue #2622, opened 2023-02-23
Breaking change with v1.4: experiment k8s service created even if steps are not weighted


fwippe:
Checklist:

* [x] I've included steps to reproduce the bug.
* [X] I've included the version of argo rollouts.

**Describe the bug**
Even for simple experiments, i.e.,  without any weighted steps, Argo always creates corresponding k8s services.
This behavior was introduced with [this PR](https://github.com/argoproj/argo-rollouts/pull/2397). According to the [docs](https://github.com/argoproj/argo-rollouts/blob/master/docs/features/experiment.md#weighted-experiment-step-with-traffic-routing) a service should only be created if there are weighted experiment steps.

Can you maybe elaborate on the reasons for this breaking change? Why do you always create a k8s service even if it's not needed?

*Why is this a breaking change?*

After upgrading to version 1.4, suddenly experiments were failing with different error messages complaining about missing or invalid k8s service properties (like `service name too long (> 63 chars)` or `missing port name attribute`).

**Expected behavior**
- Argo rollouts should not create an additional experiment service if there are no weighted experiment steps

**Version**
v1.4.0

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

zachaller:
@fwippe Yea that was a big miss on my part sorry for the change in behavior, I have been talking with alex and dani who originally had the request for a use case they had which does make sense. We tried to do the change in a non spec changing way which resulted in this change in behavior. 

The plan is going forward will be to revert the change in behavior in a 1.4.1  release that I will be doing here soon to keep the expected behavior and look into possible other solutions to support always creating services for inline experiments all the time.

I will be noting in the release notes for 1.4.1 that there was an unintended change in behavior that was reverted for 1.4.1 as well.

Issue #2608, opened 2023-02-19
experiments: recovered from panic, on v1.4


Lykathia:
When running an experiment, a number of nil pointer exceptions are thrown - before the experiment eventually passes and moseys on its merry way.

We use istio in our environment.

It seems to resolve itself after a dozen+ failures.

**To Reproduce**

```
spec:
  analysis:
    successfulRunHistoryLimit: 1
    unsuccessfulRunHistoryLimit: 1
  progressDeadlineAbort: true
  replicas: 2
  revisionHistoryLimit: 5
  strategy:
    canary:
      steps:
        - experiment:
            analyses:
              - args:
                  - name: api-root-url
                    value: >-
                      http://example-preview.default.svc.cluster.local:8080
                  - name: request-timeout-milliseconds
                    value: '1000'
                name: example-smoke-test-analysis
                requiredForCompletion: true
                templateName: example-smoke-test-analysis
            duration: 5m
            templates:
              - metadata:
                  labels:
                    app.kubernetes.io/name: example-preview
                name: preview
                replicas: 1
                selector:
                  matchLabels:
                    app.kubernetes.io/name: example-preview
                specRef: canary
        - setWeight: 5
        - pause:
            duration: 10m
      trafficRouting:
        istio:
          destinationRule:
            canarySubsetName: canary
            name: example-mesh-destination
            stableSubsetName: stable
          virtualService:
            name: example-mesh-virtualservice
  workloadRef:
    apiVersion: apps/v1
    kind: Deployment
    name: example-svc-deployment
```

**Expected behavior**

The logs to not raise errors of nil pointer exceptions

**Version**

1.4

**Logs**

```
Recovered from panic: runtime error: invalid memory address or nil pointer dereference
goroutine 325 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
github.com/argoproj/argo-rollouts/utils/controller.processNextWorkItem.func1.1.1()
	/go/src/github.com/argoproj/argo-rollouts/utils/controller/controller.go:149 +0x58
panic({0x21c1260, 0x3c0ff10})
	/usr/local/go/src/runtime/panic.go:884 +0x212
github.com/argoproj/argo-rollouts/rollout.(*rolloutContext).calculateWeightDestinationsFromExperiment(0xc00198ac00)
	/go/src/github.com/argoproj/argo-rollouts/rollout/trafficrouting.go:333 +0x1f7
[...]

PR #2624, opened 2023-02-24
fix: Rollback change on service creation with weightless experiments

daniddelrio:
closes: #2622 

possible fix for: https://github.com/argoproj/argo-rollouts/issues/2608

This rolls back the change introduced in #2397 which allows experiments with no weights to still create services. We want it so that services won't be created if there is no weight set for the experiments.

Checklist:

### DIFF
--- a/rollout/experiment.go
+++ b/rollout/experiment.go
@@ -64,7 +64,9 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
 			Name:     templateStep.Name,
 			Replicas: templateStep.Replicas,
 		}
-		template.Service = &v1alpha1.TemplateService{}
+		if templateStep.Weight != nil {
+			template.Service = &v1alpha1.TemplateService{}
+		}
 		templateRS := &appsv1.ReplicaSet{}
 		switch templateStep.SpecRef {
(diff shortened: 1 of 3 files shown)


rollout/experiment.go
```diff
@@ -64,7 +64,9 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
 			Name:     templateStep.Name,
 			Replicas: templateStep.Replicas,
 		}
-		template.Service = &v1alpha1.TemplateService{}
+		if templateStep.Weight != nil {
```
alexef, line 67:
```suggestion
		if (templateStep.Weight != nil || templateStep.Service != nil) {
```
should we put this already in?

atm the spec doesn't allow `templateStep.Service` to be anything but:
- nil
- empty object

so this shouldn't be a breaking change (famous last words)

alexef: see: https://github.com/alexef/argo-rollouts/blob/master/pkg/apis/rollouts/v1alpha1/experiment_types.go#L92

alexef: nevermind, this is not possible without a Rollout spec change.

zachaller (approved):

PR #2633, opened 2023-03-01
feat: Add Service field to Rollout Experiment to allow service creation

daniddelrio:
Follow-up to: #2624 

#2624 rolled back the change to allow rollout experiments with no weights to still create services. However, we still want the allow the users to choose to create services for their experiments even with no weight/traffic routing enabled. This PR makes it so that users can set the Service attribute for their rollout experiments (with an optional `name` field), giving them the ability to still create services without setting `weight`.

Checklist:


rollout/experiment.go
```diff
@@ -64,8 +64,11 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
 			Name:     templateStep.Name,
 			Replicas: templateStep.Replicas,
 		}
-		if templateStep.Weight != nil {
+		if templateStep.Weight != nil || templateStep.Service != nil {
 			template.Service = &v1alpha1.TemplateService{}
+			if templateStep.Service != nil && templateStep.Service.Name != "" {
```
alexef, line 69:
```suggestion
			if templateStep.Service.Name != "" {
```

you already checked for service not being nil

rollout/experiment_test.go
```diff
@@ -820,3 +820,48 @@ func TestRolloutCreateExperimentWithService(t *testing.T) {
...
+	steps := []v1alpha1.CanaryStep{{
+		Experiment: &v1alpha1.RolloutExperimentStep{
+			Templates: []v1alpha1.RolloutExperimentTemplate{
+				// Service should be created for "stable-weightless-template"
+				{
+					Name:     "stable-weightless-template",
+					SpecRef:  v1alpha1.StableSpecRef,
+					Replicas: pointer.Int32Ptr(1),
+					Service: &v1alpha1.TemplateService{
+						Name: "test-service",
```
alexef, line 836: please add test without name (if it exists, please ignore comment)

alexef (changes requested):
hmm, shouldn't you pass template.Service.Name in CreateService call? instead of rs.Name 

https://github.com/daniddelrio/argo-rollouts/blob/service-in-experiment/experiments/experiment.go#L304

rollout/trafficrouting_test.go
```diff
@@ -248,6 +248,7 @@ func TestRolloutWithExperimentStep(t *testing.T) {
 					SpecRef:  "canary",
 					Replicas: pointer.Int32Ptr(1),
 					Weight:   pointer.Int32Ptr(5),
+					Service:  &v1alpha1.TemplateService{},
```
zachaller, line 251: Why did this have to be added?

rollout/experiment.go
daniddelrio: This `templateStep.Service != nil` is for the case where templateStep.Weight != nil and templateStep.Service == nil, else `templateStep.Service.Name` will return an invalid memory address error

rollout/trafficrouting_test.go
daniddelrio: Actually that didn't need to be. Thanks

rollout/experiment.go
alexef: but you already check it on line 67, so it can never be nil on line 69, right?

daniddelrio: Line 67 will pass even if Service is nil as long as Weight is not nil. Confirmed it by rerunning the unit tests as well

rollout/experiment_test.go
daniddelrio: Done. See: https://github.com/argoproj/argo-rollouts/pull/2633/files#diff-4ee6ea6eacb6c4ad54a87a51cdba981669b25bbd2d30a0b51bbd8fcd56bce155R870

alexef (approved):

zachaller (approved):

Issue #2471, opened 2022-12-12
Enhancement Proposal: Multiple ALB Ingress Support for Canary Traffic Management


n888:
# Enhancement Proposal: Multiple ALB Ingress Support for Canary Traffic Management

Enable defining multiple AWS ALB Ingress objects within a single Rollout object, with all Ingresses progressing in unison for each canary step.

# Summary

ArgoRollouts currently supports managing a single ALB Ingress object per Rollout object. This limitation poses a problem when a single Application is served using multiple ALB Ingress objects.

For example: Ingresses for an Internal and External network, when there are various EC2 / AWS Services that need to reach a service inside an EKS cluster.

# Workarounds

An example workaround is to create multiple Rollout objects in an Application, one for each ALB Ingress object. This option puts the onus on the operator to ensure all of the Rollout objects are progressing/resumed together in the correct order, increasing the chance of human error.

This workaround also unnecessarily increases the overall Application Pod count by at least 1, since each additional Rollout creates a separate replicaset.

Another workaround is potentially using AnalysisTemplates to trigger child rollout objects, but this seems a bit too clunky.

# Example Multi-ALB Patch

https://github.com/argoproj/argo-rollouts/compare/master...n888:argo-rollouts:master

The limitation was a deal-breaker for some of our internal organizations, and we have created a Multi-ALB patch to bridge the gap for now and to re-spark the idea to the community, which we have been running in production workloads for the past three months.

The logic is heavily modeled after the Multiple Nginx Ingress patch by @tperdue321 (Thank you!) https://github.com/rallyhealth/argo-rollouts/pull/1

## Components Modified

* `Traffic Routing`: Updates ALB data type to include a list of additionalIngresses
* SetWeight: SetWeight is now a stub function that calls SetWeightPerIngress which processes the `additionalIngresses` followed by the `ingress` 
  * Returns errors encountered during any of the ingress weight modifications
* VerifyWeight: VerifyWeight is now a stub function that calls VerifyWeightPerIngress which processes the `additionalIngresses` followed by the `ingress`
  * Returns errors encountered during any of the ingress weight verifications
* Validators: Validates all available ingresses
* Rollout CRD: Updated to include `trafficRouting.alb.additionalIngresses` field

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

n888:
Resolved via https://github.com/argoproj/argo-rollouts/pull/2639

PR #2467, opened 2022-12-08
feat: support N nginx ingresses

tperdue321:
Enhancement proposal: https://github.com/argoproj/argo-rollouts/issues/2001

## Modifications:

- update the `Nginx` datatype to have a field called `AdditionalStableIngresses` to allow for 1-N stable ingresses instead of just 1. This naming convention is modeled after `AdditionalIngressAnnotations`
- modify the `SetWeight` method to create a canary for each stable ingress and apply new weight to all canary ingresses.
- update validations to validate all stable ingresses.
- designed to be fully backwards compatible.

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not. 
* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

PR #2639, opened 2023-03-06
feat: Support Multiple ALB Ingresses

n888:
Enhancement proposal: https://github.com/argoproj/argo-rollouts/issues/2471

Changes:
- Update the ALB datatype to have a field called `AdditionalIngresses` to allow for multiple additional ingresses instead of just just the one

- This naming convention is modeled after AdditionalIngressAnnotations

- Modify the SetWeight method to create a canary for each ingress and apply new weight to all canary ingresses

- Update validations to validate all ingresses.

- Designed to be fully backwards compatible

- Modeled after the Nginx Multi Ingress patch: https://github.com/argoproj/argo-rollouts/pull/2467

- Open to all direction/comments/suggestions, Thank you

Checklist:


zachaller:
Can we switch to dropping the `additional` portion of the spec and just making it plural `ingresses` can you also add docs and e2e tests.

huynhsontung:
@zachaller Do we want to keep the singular `ingress` field? We named it `additionalIngresses` to exist alongside with `ingress`.  Removing it would make a breaking change on the spec. Having both `ingress` and `ingresses` would be confusing imo.

n888:
Thanks for the info @zachaller @huynhsontung, I will get on the `ingresses` change and e2e tests.

n888:
1. ALB traffic routing `ingress` and `ingresses` are implemented and mutually exclusive

2. Added e2e tests

3. Updated StickinessConfig types.go description as it seemed off

4. Currently when `ingresses` is in use, `status.alb` returns the status from the last ingress defined in the array.
* One potential backwards compatible solution is to create a `status.albs` array and return each ALB status to there when `ingresses` is in use. If `ingress` is in use, just continue to return the single alb to `status.alb`.

zachaller:
I like the idea of status.albs, also feel free to ping me. If this is ready for review etc

n888:
@zachaller, Please review when you have some time, thank you

zachaller:
@n888 will do I will need to spend a little extra time on this one because Intuit also uses alb's so will need to make sure nothing breaks for us internally as well

huynhsontung:
@zachaller Any updates on the review process? It's been a while and this feature is quite critical for us as well.

zachaller:
I will be focusing on open source here for a bit and this is one of my top priorities after I get some notification engine stuff done. I will also make sure this gets included in 1.6

zachaller (approved):

zachaller:
Sorry for the long review time, thanks for the contributions!

n-elkan-srca:
Thank you very much @zachaller @huynhsontung  & Argo Team :tada:

n888:
Thank you @zachaller @huynhsontung & Argo Team :100:

thecosmicfrog:
This is fantastic @n888 @zachaller @huynhsontung! As it happens, we are looking into Argo Rollouts for our canary rollout requirements and... you guessed it... we need support for multiple ALB ingresses. What excellent timing!

Can you advise when this feature will be released? We are using the Helm chart for installing the controller and CRDs. Thanks!

zachaller:
It will be in 1.6, rc should come out this week hopefully.

thecosmicfrog:
Great, thanks @zachaller!

PR #2656, opened 2023-03-13
docs: Include PagerDuty in service type listing

jess-belliveau:
Exists as an option in the sidebar, but not included in the list of "Service Types".

Apologies, I'm not exactly sure where these "generated" docs come from and if I should be modifying elsewhere.


jess-belliveau:
Based on the "verify codegen" step failing - I guess I have the wrong spot.

zachaller:
This file is auto generated and comes from https://github.com/argoproj/notifications-engine/blob/master/docs/services/overview.md

jess-belliveau:
Thanks for the info @zachaller. Will raise over there!

PR #2690, opened 2023-04-02
docs(example): interval requires count

meeech:
docs(example): moved other count up next to interval for visibility

fixes: #2691

Checklist:

* [ n/a ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ n/a ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,7 +11,7 @@ assignees: ''
 Checklist:
 
 * [ ] I've included steps to reproduce the bug.
-* [ ] I've inclued the version of argo rollouts.
+* [ ] I've included the version of argo rollouts.
 
 **Describe the bug**
 
--- a/examples/analysis-templates.yaml
+++ b/examples/analysis-templates.yaml
@@ -10,6 +10,7 @@ metadata:
 spec:
   metrics:
   - name: pass
+    count: 1
     interval: 5s
     failureLimit: 1
     provider:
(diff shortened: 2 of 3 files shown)


agrawroh (changes requested):
I don't think you need `count` when interval is specified. The docs clearly say:

_"If both interval and count are omitted, the effective count is 1. If only interval is specified, metric runs indefinitely."_

I would expect the analysis to keep running indefinitely in absence of the `count`.

meeech:
@agrawroh 1.4.1
Does it make sense for it to run indefinitely when it's an inline step? Then the rollout would never complete? So I suppose the error makes sense in that we wouldnt want people to accidentally put themselves in a rollout that would never progress? Or would the only way out be a manual promote?

agrawroh:
> @agrawroh 1.4.1 Does it make sense for it to run indefinitely when it's an inline step? Then the rollout would never complete? So I suppose the error makes sense in that we wouldnt want people to accidentally put themselves in a rollout that would never progress? Or would the only way out be a manual promote?
> 
> Could be the docs are a bit out of date - I could update the pr to update to the docs you pointed out - maybe we need them to be more specific - making the distinction between inline vs background analysis.
> 
> In any case, I would defer to the code over the docs - and assume docs need updating.

You definitely need one of `interval` or `count` but I'm not convinced that we need to specify both. For example, if I just specify the `interval` with some success/failure counts, then I'd like the analysis to run and take as many measurements as possible and either fail or succeed based on the specified conditions. I can optionally specify the `count` to only take N measurements but it shouldn't be required if I have already specified the `interval`.

zachaller (approved):

zachaller:
The reason count is needed with interval here is because in the examples the way this template is used is in an inline step which has to have an end, otherwise the step would never complete. In background analysis you need interval but not count because it will just run till the rollout is done. I agree docs could probably be updated around this.

zachaller:
This PR just fixes the case for how this template is used within the examples folder.

PR #2694, opened 2023-04-04
feat(trafficrouting): add role for contour/httpproxy

izturn:
Checklist:

Add the role for the plugin: https://github.com/argoproj-labs/rollouts-contour-trafficrouter-plugin

### DIFF
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -14782,6 +14782,15 @@ rules:
   - watch
   - get
   - update
+- apiGroups:
+  - projectcontour.io
+  resources:
+  - httpproxies
+  verbs:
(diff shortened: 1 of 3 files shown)


zachaller:
So I am guessing this is for a plugin? I am not sure we should be putting rbac rules in the main repo, I still think the plugin other should create the required config and role bindings you could just bind it to the default role defined in upstream manifest but also explain that if you changed the role that it needs these permissions. I actually also currently do not like that we blanket grant permissions for every traffic routers upstream I think we should allow users to only install rbac for the traffic routers they use.

izturn:
got it thx @zachaller

PR #2742, opened 2023-04-24
Add failure message

Sineaggi:
Checklist:


metricproviders/datadog/datadogV1_test.go
```diff
@@ -67,6 +69,7 @@ func TestRunSuite(t *testing.T) {
 		},
 		// Same test as above, but derive DD keys from env var instead of k8s secret
 		{
+			testName:          "Same test as above, but derive DD keys from env var instead of k8s secret",
```
meeech, line 72: nit: dont refer to above since order can change :D

utils/evaluate/evaluate.go
```diff
@@ -45,7 +45,13 @@ func EvaluateResult(result interface{}, metric v1alpha1.Metric, logCtx logrus.En
 	}
 
 	if failCondition {
-		return v1alpha1.AnalysisPhaseFailed, nil
+		if metric.SuccessCondition != "" {
```
meeech, line 48: I think it's possible to have a success and fail condition for 1 metric?  so wouldn't fail always win, even if it was the success condition that didn't pass?

Sineaggi:
I've added an enhancement proposal here https://github.com/argoproj/argo-rollouts/issues/3008 where we can discuss this feature

Sineaggi:
Is this still an interesting feature?

PR #2752, opened 2023-04-30
fix(analysis): Graphite query - remove whitespaces

mdsjip:
If metric query contains surrounding whitespace, then it's not trimmed and Graphite host returns `400 Bad Request`

Simplest example of analysis template:
```yaml
kind: ClusterAnalysisTemplate
metadata:
  name: error-rate
  namespace: argo-rollouts
spec:
  metrics:
    - name: error-rate
      failureLimit: 1
      # Note that the Argo Rollouts Graphite metrics provider returns result as an array of float64s with 6 decimal places.
      successCondition: 'len(result) > 0 and result[len(result)-1] < asFloat(10)'
      provider:
        graphite:
          address: https://some-graphite-host.com
          query: |
            from=-2&target=sumSeries(app.http.*.*.count)

```

When this is parsed from YAML, this query contains a newline in the end, so when it's "trimmed", it still contains a trailing space, so when put into the query, it becomes:

```
format=json&from=-2min&target=sumSeries%28app.http.%2A.%2A.count%29+
```

This leads to `400 Bad Request` response from Graphite, example cURL command:

```shell
$ curl 'https://some-graphite-host.com/render?format=json&from=-2min&target=sumSeries%28app.http.%2A.%2A.count%29+'
Bad Request: Bad Request

Target              : sumSeries(app.http.*.*.count) 
Parsed so far       : sumSeries(app.http.*.*.count)
Could not parse     :  
```

Proposed solution: **replace all whitespace in the query with an empty string**.

Checklist:

### DIFF
--- a/metricproviders/graphite/api.go
+++ b/metricproviders/graphite/api.go
@@ -77,7 +77,7 @@ func (api APIClient) Query(quer string) ([]dataPoint, error) {
 
 func (api APIClient) trimQuery(q string) string {
 	space := regexp.MustCompile(`\s+`)
-	return space.ReplaceAllString(q, " ")
+	return space.ReplaceAllString(q, "")
 }
 
 type dataPoint struct {
--- a/metricproviders/graphite/api.go
+++ b/metricproviders/graphite/api.go
@@ -27,9 +27,11 @@ type APIClient struct {
 	logCTX log.Entry
 }
 
+var spaceRegex = regexp.MustCompile(`\s+`)
+
 // Query performs a Graphite API query with the query it's passed
 func (api APIClient) Query(quer string) ([]dataPoint, error) {
-	query := api.trimQuery(quer)
+	query := api.sanitizeQuery(quer)
 	u, err := url.Parse(fmt.Sprintf("./render?%s", query))
 	if err != nil {
 		return []dataPoint{}, err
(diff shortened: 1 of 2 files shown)


metricproviders/graphite/api.go
```diff
@@ -73,7 +73,7 @@ func (api APIClient) Query(quer string) ([]dataPoint, error) {
 
 func (api APIClient) trimQuery(q string) string {
 	space := regexp.MustCompile(`\s+`)
-	return space.ReplaceAllString(q, " ")
+	return space.ReplaceAllString(q, "")
```
mdsjip, line 76: One can possibly argue that this function name doesn't reflect its purpose: it doesn't **just** trim the whitespace, but also can modify the input string content in the middle.
Also, compiling regular expression every time this function is called is sub-optimal.
Considering the above, it can be rewritten to something like this:

```golang
var space = regexp.MustCompile(`\s+`)

func (api APIClient) sanitizeQuery(q string) string {
	return space.ReplaceAllLiteralString(q, "")
}
```

zachaller: Yea I like this change.

mdsjip: > Yea I like this change.

Thank you, I've uploaded this as a separate commit.

zachaller (approved):

mdsjip:
@zachaller will these changes be included in the next release soon? :slightly_smiling_face:

Issue #2780, opened 2023-05-15
SetHeaderRoute not change VirtualService


huoyinghao:
### Discussed in https://github.com/argoproj/argo-rollouts/discussions/2776

<div type='discussions-op-text'>

<sup>Originally posted by **huoyinghao** May 12, 2023</sup>
Hi, team

here is my rollout yaml ↓

```yaml

---
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  annotations:
    rollout.argoproj.io/revision: "1"
    rollout.argoproj.io/workload-generation: "20"
  creationTimestamp: "2023-05-12T11:07:57Z"
  generation: 1
  name: test-rollout-150
  namespace: test-rollout
  resourceVersion: "23904088"
  uid: db29e57b-c939-41c0-bf8c-53b9fe38f06f
spec:
  replicas: 1
  strategy:
    canary:
      steps:
      - setCanaryScale:
          replicas: 1
      - setHeaderRoute:
          match:
          - headerName: header1
            headerValue:
              prefix: test1
          - headerName: header2
            headerValue:
              prefix: test2
          name: route-1
      - pause: {}
      workloadRef:
        apiVersion: apps/v1
        kind: Deployment
        name: test-rollout-150
      trafficRouting:
        istio:
          destinationRule:
            canarySubsetName: canary
            name: test-rollout-150
            stableSubsetName: stable
          virtualService:
            name: test-rollout-150
            routes:
            - primary
        managedRoutes:
        - name: route-1
---
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  creationTimestamp: "2023-05-12T08:44:58Z"
  generation: 14
  name: test-yaml-rollout
  namespace: test-rollout
  ownerReferences:
  - apiVersion: apps/v1
    kind: Deployment
    name: test-yaml-rollout
    uid: dbc63044-3d9a-44a8-8618-aa63bc369c4f
  resourceVersion: "23881171"
  uid: 21351a56-123e-4ba5-9942-3194f073347c
spec:
  hosts:
  - test-yaml-rollout
  http:
  - name: primary
    route:
    - destination:
        host: test-yaml-rollout
        subset: stable
      weight: 100
    - destination:
        host: test-yaml-rollout
        subset: canary
      weight: 0
  
```

when i promote it,i think VirtualService should be changed

but the fact doesn t seem like that, afterwards i went to check the relevant code,i found the code to set the header route.

```go
[...]

PR #2797, opened 2023-05-19
remove generation whil setHeaderRoute and setMirrorRoute

huoyinghao:
Checklist:

remove the generation check for set header route,  
When using Workload Ref to associate resources, we modify the mirror version of the resources, and the generation of rollouts will not be updated


rollout/trafficrouting.go
```diff
@@ -237,10 +237,7 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 				desiredWeight = 100
 			}
 		}
-
-		// We need to check for Generation > 1 because when we first install the rollout we run step 0 this prevents that.
-		// We could also probably use c.newRS == nil || c.newRS.Status.AvailableReplicas == 0
-		if currentStep != nil && c.rollout.ObjectMeta.Generation > 1 {
```
zachaller, line 243: I don't think this is the correct thing to do here these steps should not run when a rollout is first installed into a cluster. What is your reasoning for this change?

zachaller:
closes #https://github.com/argoproj/argo-rollouts/issues/2780

rollout/trafficrouting.go
huoyinghao: [while i](https://github.com/argoproj/argo-rollouts/issues/2780)  
When I followed the steps of this issue to run rollback, I found that VitrualService was not set correctly.  
Only when the generation of rollout is greater than 1 will VirtualService be correctly changed

zachaller: I think this is expected behavior though, when a rollout is first installed we do not run any of the steps because there is nothing to canary too because the rollouts is being installed for the very first time, any subsequent updates should will bump generation so that we run the `steps`

huoyinghao: But when I update the image of deployment  referenced by `WorkloadRef`, the generation of Rollout does not change. I think this is because we didn't update the Rollout resource directly. 

Maybe we should use the condition `c.newRS == nil || c.newRS.Status.AvailableReplicas == 0` as the comment indicates?

zachaller: ahh sure, this is with workloadRef that makes more sense. Using the other condition makes sense should also add test case for workloadRef as well

huoyinghao: I modify it use `currentStep instead= Nil&&(c.newRS!=nil&&c.newRS. Status. AvailableReplicas=c.newRS.Status. Replicas)` made a judgment and added some unit tests

huoyinghao: @zachaller please take a look, communicate with me if there are any problems, thanks

huoyinghao: @alexmt please take a look, thanks

rollout/trafficrouting.go
```diff
@@ -233,10 +234,10 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
...
+		// We need to check for revision > 1 because when we first install the rollout we run step 0 this prevents that.
+		// There is a bigger fix needed for the reasons on why we run step 0 on rollout install, that needs to be explored.
+		revision, revisionFound := annotations.GetRevisionAnnotation(c.rollout)
+		if currentStep != nil && (revisionFound && revision > 1) {
```
zachaller, line 240: @huoyinghao Would you mind double checking that this also fixes your use case, I think this is a better fix after having spent time looking into it?

huoyinghao: i also think so, thanks

zachaller (approved):

PR #2851, opened 2023-06-19
fix(controller): Remove name label from some k8s client metrics

SuperQ:
The `name` label in the `controller_clientset_k8s_request_total` metric
produce an excessive amount of cardinality for `events` and `replicasets`.
This can lead to hundreds of thousands of unique metrics over a couple 
weeks in a large deployment. Set the name to "N/A" for these client request 
types.

Checklist:

### DIFF
--- a/controller/metrics/client.go
+++ b/controller/metrics/client.go
@@ -27,7 +27,7 @@ func (m *K8sRequestsCountProvider) IncKubernetesRequest(resourceInfo kubeclientm
 	namespace := resourceInfo.Namespace
 	kind := resourceInfo.Kind
 	statusCode := strconv.Itoa(resourceInfo.StatusCode)
-	if resourceInfo.Verb == kubeclientmetrics.List {
+	if resourceInfo.Verb == kubeclientmetrics.List || kind == "events" || kind == "replicasets" {
 		name = "N/A"
 	}
 	if resourceInfo.Verb == kubeclientmetrics.Unknown {


zachaller:
I wonder if instead of dropping name from everything we just set name to something like "N/A" for the resources that have random strings in it or we strip the random parts off if that is an easy task as well.

It's nice to have the name of the rollout object that is causing a request to be made but agree the random fields are a bad anti-pattern and should probably be cleaned up.

SuperQ:
@zachaller The problem is that `name` causes huge cardinality leaks, since every rollout generates a bunch of new series. We end up with 50,000 new series _every week_ in one of our dev clusters. We currently drop the whole metric in order to reduce the load on our Prometheus server. If the argo-rollouts binary is up for 2 weeks, we end up with 100,000 metrics and it takes 3 seconds just to scrape the data.

SuperQ:
Let me look over some of the raw data to see if it's possible to filter out specific resources.

SuperQ:
Another source of cardinality problems is `namespace`. We generate ephemeral namespaces for PRs. So we end up with things like

```
controller_clientset_k8s_request_total{kind="rollouts",name="some-service-deployment",namespace="test-service-12345",status_code="409",verb="Update"}
```

Over a week we get 2000 of these metrics.

SuperQ:
For `kind="events"` we generate 20k/week.
For `kind="replicasets"` we generate 10k/week.

If you really want to keep the `name` label just for `kind="rollouts"`, it's not ideal, but not the worst cardinality problem.

zachaller:
One other thing what version of rollouts are you on because pre [this](https://github.com/argoproj/argo-rollouts/pull/2100) it would be even worse, however I do agree the cardinality here can get bad we have turned it off as well. if we remove name/namespace on the RO object though I think the whole stat actually becomes even less usefull because k8s itself has request metrics at the kind level

SuperQ:
We are running 1.5.x.

SuperQ:
Ok, I've simplified this to just the top two sources of cardinality leaks.

zachaller (approved):

PR #2898, opened 2023-07-22
chore: leave the validation of setHeaderRoute to the plugin when plugins is not empty.

andyliuliming:
leave the validation of setHeaderRoute to the plugin when plugins is not empty.
resolving #2899

Checklist:

### DIFF
--- a/pkg/apis/rollouts/validation/validation.go
+++ b/pkg/apis/rollouts/validation/validation.go
@@ -305,7 +305,7 @@ func ValidateRolloutStrategyCanary(rollout *v1alpha1.Rollout, fldPath *field.Pat
 
 		if step.SetHeaderRoute != nil {
 			trafficRouting := rollout.Spec.Strategy.Canary.TrafficRouting
-			if trafficRouting == nil || (trafficRouting.Istio == nil && trafficRouting.ALB == nil && trafficRouting.Apisix == nil) {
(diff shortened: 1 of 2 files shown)


zachaller:
Can you give an example of why you need this validation. Also, argo rollouts supports multiple traffic routers at once I am unsure if it is the right move to turn off validation if say istio and a plugin are used together.

andyliuliming:
> Can you give an example of why you need this validation removed. Also, argo rollouts supports multiple traffic routers at once I am unsure if it is the right move to turn off validation if say istio and a plugin are used together.

@zachaller we are trying to use the max traffic weight > 100.

so 2 ways for us:
1. use the set traffic header to do the header regex
2. add support for the max weight > 100 as I was trying to do it in this PR:
https://github.com/argoproj/argo-rollouts/pull/2900/files#diff-d4a6b587ffdebe4f22c2cc8b65d601c03c9eb897a06f36c1f633778db3fbe58e
but that PR is more complex than I expected, since looks like the max weight is not in the spec.

this PR is to try the #1 way.

andyliuliming:
@zachaller maybe add one field 
MaxTrafficWeight *int32 `json:"maxTrafficWeight,omitempty" protobuf:"varint,10,opt,name=maxTrafficWeight"` into RolloutTrafficRouting struct is accepatable?

PR #2903, opened 2023-07-22
feat(analysis): Adds rollout Spec.Selector.MatchLabels to AnalysisRun. Fixes #2888

### LINKED ISSUE #2888: Add Rollout match selector label to AnalysisRuns
# Summary

Add the Rollout match selector to AnalysisRuns as labels

# Use Cases

At the moment there is no easy to select all of the AnalysisRuns that belong to a specific rollout so if you want to review the past history with kubectl it's painful to select them. While the AnalysisRun does have the ownerReference set to the Rollout this is not selectable with a field-selector or any other mechanism AFAIK, typically some scripting would be needed to do so.

The idea here is to basically do the same thing with AnalysisRuns that already happens with ReplicaSets and Pods.

---

gnunn1:
fixes: https://github.com/argoproj/argo-rollouts/issues/2888

Checklist:

This is a small change to add the rollout matchLabels to the AnalysisRun when created as per the linked issue.

Note that I am not a golang developer, even though it's a tiny chnage that I've tested locally it's possible I've missed things so feel free to give me feedback and I'm happy to modify as needed. I'm also not sure how/were a unit test for this would need to be created as I'm not seeing existing tests for labels in analysis_test.go that I could use as a reference, happy to take feedback on that as well if one is needed.


rollout/analysis.go
```diff
@@ -460,6 +460,11 @@ func (c *rolloutContext) newAnalysisRunFromRollout(rolloutAnalysis *v1alpha1.Rol
 	for k, v := range rolloutAnalysis.AnalysisRunMetadata.Labels {
 		run.Labels[k] = v
 	}
+
+	for k, v := range c.rollout.Spec.Selector.MatchLabels {
```
zachaller, line 464: I think taking the labels on the rollout object vs the MatchLabels probably make more sense and follows what we do with ReplicaSets, with RS we take the pod template labels and put them on the RS as labels.

gnunn1: Thanks @zachaller, I'll have a look at the RS code and follow the same practice.

gnunn1: Sorry for the delay, as requested updated to use template labels.

zachaller (approved):

PR #2948, opened 2023-08-09
feat: Traefik mirror

Philipp-Plotnikov:
Checklist:


Philipp-Plotnikov:
@zachaller if we need to have ability to mirror to the service also, we need to add field for port
Also am I right that for creation or updating(to set the match rules) ingressRoute we will make plugin to not additional fields ?

zachaller:
This PR does not seem to offer the route matching feature this is needed in order to support controlling what routes to mirror this is a needed feature because you generally can not mirror write requests which would cause double updates etc etc.

```
- setMirrorRoute:
          name: mirror-route
          percentage: 35
          match:
            - method:
                exact: GET
              path:
                prefix: /
```

As you can see you can specify a match section.

Philipp-Plotnikov:
@zachaller yes, but for that I will need to change IngressRoute but for that we need to add field to specify the name of IngressRoute that we need to change. We talked about this and as I understood for that we will create plugin. I asked about this ealier in this PR

Philipp-Plotnikov:
To controll request match, user will need to create needing IngressRoute

Philipp-Plotnikov:
@zachaller If we decide that we need to control match from argo manifest we need to decide we will need to create or change IngressRoute ? If change we need field for IngressRoute name. If create we can create IngressRoute with some prefix as mirror TraefikService or specify with what name we would like to create IngressRoute

idurgakalyan:
@Philipp-Plotnikov I'm happy to take this up and move Traefik to stable version, let me know if there is anything I have to be aware of.

PR #2997, opened 2023-08-30
fix(metricprovider): support Datadog v2 API Fixes #2813

### LINKED ISSUE #2813: Datadog v2 API fails when using formulas in query
Checklist:

**Describe the bug**

Using formulas in AnalysisTemplate with Datadog v2 API fails with API error 

```received non 2xx response code: 400 {"errors":["Functions are not supported in metrics data source field. Use the formulas payload."]}```

**To Reproduce**

Use a query with formula in Datadog provider for AnalysisTemplate and set apiRevision to v2. The query works with v1.

```yaml
kind: AnalysisTemplate
metadata:
  name: rollout-test-success-rate
spec:
  metrics:
    - name: success-rate
      interval: 1m
      successCondition: default(result, 0) < 10
      provider:
        datadog:
          apiVersion: v2
          interval: 1m
          query: |
            sum:trace.rack.request.hits{env:staging AND service:service_name AND http.status_code:5*}.as_count() /
            sum:trace.rack.request.hits{env:staging AND service:service_name AND http.status_code:2*}.as_count() * 100
```

**Expected behavior**

It should work with v2 as well, or the note in https://argo-rollouts.readthedocs.io/en/stable/analysis/datadog/ should warn about using formulas (now it says that "If you switch to v2, you will not need to change any other field aside from apiVersion.")

**Version**

1.5.1

**Logs**

```
time="2023-05-26T10:02:02Z" level=info msg="Taking 1 Measurement(s)..." analysisrun=rollout-test-6b986bcd5b-7.1 namespace=default
time="2023-05-26T10:02:02Z" level=info msg="Measurement Completed. Result: Error" analysisrun=rollout-test-6b986bcd5b-7.1 metric=success-rate namespace=default
time="2023-05-26T10:02:02Z" level=warning msg="Measurement had error: received non 2xx response code: 400 {\"errors\":[\"Functions are not supported in metrics data source field. Use the formulas payload.\"]}" analysisrun=rollout-test-6b986bcd5b-7.1 metric=success-rate namespace=default
```

---

meeech:
fixes #2813 #2819 

Add support for Datadog v2 api. 
Add new properties `queries` and `formula`.
Update the docs with some more examples, more details around v2.

Update CRDs to express validation requirements, allowing us to remove some guardrails in the code. 

Rationale:
given the validation at the heart of k8s objects is held to a high standard
given the context the analysis Runs exec is k8s
given the the Analysis Template has a defined spec

eg: we can then make some assumptions
apiVersion _is_ empty, v1, v2 
set some defaults. 
```
(server dry run)
The AnalysisTemplate "scratch.example-v2-template" is invalid: spec.metrics[0].provider.datadog.apiVersion: Unsupported value: "v3": supported values: "v1", "v2"
```

**Why not use the DD go client**?

I explored it. There was no obvious benefit vs making the requests how we do now, with the drawback of having a new external dep. Which is in an unstable state - this is emitted when using the DD client api.
`WARNING: Using unstable operation 'v2.QueryTimeseriesData'" `

**v1 vs v2**
~~That being said, I do have a request out to Datadog to clarify the future of v1 api so we can definitively clear up this business where we recommend moving to v2 because of impending deprecation of v1, and more guidance around the rate limits.~~

(as of sept 20, 2023)

From DD re: v1 deprecation: " I received confirmation from our engineering team that the api endpoint, /api/v1/query, is not going to be deprecated in the near future."

From DD re: Rate limits:  "increasing the rate limit of /api/v1/query api endpoint is possible and reviewed on a case-to-case basis."

**Why scalar and not timeseries?**
Looking into timeseries vs scalar. I think since we only need 1 value, we do better to hit the  `v2/query/scalar` endpoint instead of `v2/query/timeseries`. A lot more data is being transferred for timeseries with no benefit- we throw it away.
Also confirmed with DD: From support ticket with DD: "...I have also tested the scalar api endpoint with the `last` aggregator as well as the timeseries api endpoint. They do indeed return the same values when retrieving the values via the api endpoints. Observing the time it takes to retrieve the values, they remain relatively the same..."

I have done similar tests and can confirm values returning remain relatively the same.

Checklist:


metricproviders/datadog/datadog.go
```diff
@@ -300,14 +329,8 @@ func (p *Provider) parseResponseV2(metric v1alpha1.Metric, response *http.Respon
...
-	timepoint := attributes.Times[len(attributes.Times)-1]
-	if timepoint == 0 {
-		return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Datapoint does not have a corresponding time value")
-	}
-
-	value := datapoint[len(datapoint)-1]
+	column := res.Data.Attributes.Columns[0]
+	value := column.Values[0]
 	status, err := evaluate.EvaluateResult(value, metric, p.logCtx)
 	return strconv.FormatFloat(value, 'f', -1, 64), status, err
```
meeech, line 335: should we set  to 6 for precision? timeseries results seemed to be max 6 decimals

metricproviders/datadog/datadog_test.go
```diff
@@ -0,0 +1,148 @@
...
+		versionEnums := ddSpec.Properties["apiVersion"].Enum
+
+		assert.Equal(t, 2, len(versionEnums), "Expecting 2 enum options")
+		assert.Equal(t, "\"v1\"", string(versionEnums[0].Raw), "\"v1\" expected, got %s", string(versionEnums[0].Raw))
+		assert.Equal(t, "\"v2\"", string(versionEnums[1].Raw), "\"v2\" is missing, got %s", string(versionEnums[1].Raw))
+	})
+
+	t.Run("interval: Validate default is 5m", func(t *testing.T) {
+		defaultInterval := string(ddSpec.Properties["interval"].Default.Raw)
+		assert.Equal(t, "\"5m\"", defaultInterval, "Default version should be v1")
```
meeech, line 54: fix error message

meeech:
@zachaller What do we do re: conflict with generated?
Any way we can get this reviewed anytime soon so a) don't have to keep rebasing and b) before the context around choices leaves my head :)
I am happy to do a paired pr review if that is helpful/makes it quicker.

zachaller:
@meeech I had reviewed this and it was good I see you made the only change with the error let me give it one quick go over again and merge it.

zachaller (approved):

PR #3068, opened 2023-09-28
fix: permit spec.template.metdata.name in rollout crd

### LINKED ISSUE #3067: Rollout `spec.template.spec.metadata.name` is an invalid field, breaking compatibility with apps/v1 Deployment
Checklist:

**Describe the bug**

Creating/Updating a rollout with `spec.template.spec.metadata.name` field set causes the following client side validation error:
```
Error: UPGRADE FAILED: error validating "": error validating data:  ValidationError(Rollout.spec.template.metadata): unknown field "name" in io.argoproj.v1alpha1.Rollout.spec.template.metadata
```

However setting the same field in the `apps/v1` `Deployment` causes no issue. Disabling client validation succeeds.

I understand setting the field has no effect on the final deployment object, but the field cannot be removed from our workflow as we set the same default `metadata` object (name, labels, annotations) on all resources in our clusters.

Docs state: 
> Template describes the pods that will be created. **Same as deployment.**

**To Reproduce**

1. `kubectl apply` the snippet and observe the valid Deployment
2. Change the apiVersion/kind to ArgoRollout, and try to apply. Observe the validation error
3. Comment out the `spec.template.spec.metadata.name` field and try to apply again, and observe success

```
apiVersion: apps/v1
kind: Deployment
# apiVersion: argoproj.io/v1alpha1
# kind: Rollout
metadata:
  name: test
spec:
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      name: test
      labels:
        app: test
    spec:
      containers:
      - name: test
        image: argoproj/rollouts-demo:blue
```

**Expected behavior**
Argo Rollout `spec.template.spec` should be fully interchangable with Deployment (POD) spec

**Version**

v1.5.0

**Logs**

Logs are N/A, this is a **client side validation** issue

---

jasondamour:
Checklist:

Closes #3067

### DIFF
--- a/manifests/crds/rollout-crd.yaml
+++ b/manifests/crds/rollout-crd.yaml
@@ -942,6 +942,8 @@ spec:
                 properties:
                   metadata:
                     properties:
+                      name:
+                        type: string
                       annotations:
                         additionalProperties:
                           type: string


meeech:
Hey Jason. To update CRDs you need to update the code that they are generated from. For an example, you can see this PR currently in flight - I had to update some CRDs as well. 

https://github.com/argoproj/argo-rollouts/pull/2997/files#diff-4c6b93aa75d5affde60dc3849606c9acd75ed444d52e99f3055fc0c7aa77e9e0R181

Hope that helps. Let me know if you have any Qs.

zachaller:
This can't be fixed this what this is all done via go types and auto generated code

jasondamour:
@meeech Ah thank you for that pointer. But it raises a new question for my change!
1. Argo Rollouts uses `k8s.io/api/core/v1` `PodTemplateSpec` [as seen here](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/v1alpha1/types.go#L50)...
2. `PodTemplateSpec` uses `k8s.io/apimachinery/pkg/apis/meta/v1` `ObjectMetadata`, [as seen here](https://pkg.go.dev/k8s.io/api/core/v1#PodTemplateSpec)...
3. `ObjectMetdata` has a `name` field, [as seen here](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta)!
So why does the generated CRD not have the `name` field, since it seems to be in the golang struct? @zachaller can you weigh in on where the disconnect might be?

zachaller:
That is a great question and also will get us down a rabbit hole into k8s crds/schema stuff, I don't have much of an idea on why it should not work

jasondamour:
Long shot, but tagging @rbreeze since you were the last person to touch this line and might know whats going on. Can you give any pointers as to why `spec.template.metdata.name` isn't in the generated CRD when it seems to be in the types.go?

PR #3076, opened 2023-10-04
fix: missing notification on error

zachaller:
Notifications would stop processing if any error happened on a single notification 'API' configuration, we should continue processing other notifications so that we do not miss sending a notification that is properly configured.

This also fixes an incorrect prometheus stat, we where only incrementing errors and success counts on api creations not at the sending and erroring of the actually execution of the notification.

### DIFF
--- a/utils/record/record.go
+++ b/utils/record/record.go
@@ -217,7 +217,7 @@ func (e *EventRecorderAdapter) defaultEventf(object runtime.Object, warn bool, o
 		for _, api := range apis {
 			err := e.sendNotifications(api, object, opts)
 			if err != nil {
-				logCtx.Errorf("notifications failed to send for eventReason %s with error: %s", opts.EventReason, err)
+				logCtx.Errorf("Notifications failed to send for eventReason %s with error: %s", opts.EventReason, err)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


utils/record/record.go
```diff
@@ -274,34 +272,46 @@ func (e *EventRecorderAdapter) sendNotifications(notificationsAPI api.API, objec
...
 		if err != nil {
-			log.Errorf("Failed to execute condition of trigger %s: %v", trigger, err)
-			return err
+			log.Errorf("failed to run trigger, trigger: %s, destination: %s, namespace config: %s : %v",
+				trigger, destination, notificationsAPI.GetConfig().Namespace, err)
+			errors = append(errors, err)
+			continue
 		}
-		log.Infof("Trigger %s result: %v", trigger, res)
+		log.Infof("trigger %s result: %v", trigger, res)
```
leoluz, line 290: nit: Argo CD logs we usually start with uppercase.

utils/record/record.go
```diff
@@ -274,34 +272,46 @@ func (e *EventRecorderAdapter) sendNotifications(notificationsAPI api.API, objec
...
 	objMap, err := toObjectMap(object)
 	if err != nil {
-		return err
+		return []error{err}
 	}
 
 	emptyCondition := hash("")
 
+	// We should not return in these loops because we want other configured notifications to still send if they can.
+	errors := []error{}
```
leoluz, line 281: Not sure if I am understanding this correctly but I can't find where this list of `errors` is returned..

leoluz (commented):
Pls check my comments

leoluz (approved):
LGTM

Issue #3020, opened 2023-09-06
Fast-tracked rollback to stable with dynamicStableScale went under maxUnavailable


jessesuen:
**Describe the bug**

Argo Rollouts v1.4.1

The following sequence of events occurred on a user's rollout, with:
* 21 replicas
* traffic routing
* dynamicStableScale

1. The rollout was in the middle of an update. Traffic was split between canary stable
2. Before the update completed, the stable pod spec was reapplied. When this happens, we treat this event as a fast-tracked rollback
3. As the rollout was updating to the new desired/stable RS, the total available pods went *under* the minAvailable and traffic was directed 100% to an undersized replicaset.

See the following abbreviated logs (with emphasis on problematic behavior). The logs are grepped by only Patch events and Kube Events:

```
time="2023-08-29T02:22:18Z" level=info msg="Event(): reason: 'TrafficWeightUpdated' Traffic weight updated from 85 to 90"
[...]

PR #3077, opened 2023-10-04
fix: rollback to stable with dynamicStableScale could overwhelm stable pods

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/3020

Changes made:

**1. Safer scaledown decisioning of "old" ReplicaSets**

When scaling down "old" ReplicaSets, we now additionally check if they are still referenced by services before allowing them to be scaled down. This is an overall safety improvement to make sure we scale down older ReplicaSets **_only after_** it doesn't matter anymore (nothing is pointing to them).

**2. Delay canary service selector switch, if we are rolling back to stable with dynamicStableScale and it is not fully available** 

Before this PR, when the user re-applied stable pod spec, the rollout controller would immediately change the service selector of the canary service back to the stable RS. This meant that all production traffic immediately hit the stable RS, even if it was undersized due to dynamicStableScale. This PR detects if `dynamicStableScale` is used, and if stableRS is not fully available, then we delay the service switch until stableRS is fully available. Now that we no longer touch the the canary service in this scenario, it allows the controller to continue to balance traffic between stable vs. previous desiredRS, so that traffic can be safely shifted away from it (improvement 3 below).

**3. Gradually traffic shift back to stable RS to avoid overwhelming stable pods**

A user can re-apply the stable manifest in the middle of an update. When this happens, stableRS == desiredRS. Before this PR, when this happened in the middle of an update, we would immediately shift 100% of the traffic back to stableRS (by setting weight to 0), **_even though it may have been undersized due to dyanamicStableScaling_**. With this PR, we will now follow similar logic as abort with dynamicStableScaling, and only increase traffic back to stable in accordance with stable's available replica count. A key difference between abort vs. stable rollback is that:
* With Abort: traffic is balanced between desired vs. stable
* With Stable Rollback: traffic is balanced between *previous desired* vs. stable

Because of this difference, the trafficrouting logic has special case logic for traffic splitting to happen between previous desired vs. stable in the event of a rollback to stable.

I will be adding an e2e test for this.

Checklist:


jessesuen:
Here is the new order of events after this PR:

1. First rollout is updated and reaches a pause step where 90% is to canary:

```
8s          Normal    Updated VirtualService   rollout/istio-canary                 VirtualService `istio-canary` set to desiredWeight '90'
8m33s       Normal    TrafficWeightUpdated     rollout/istio-canary                 Traffic weight updated from 0 to 90
8m33s       Normal    RolloutStepCompleted     rollout/istio-canary                 Rollout step 1/2 completed (setWeight: 90)
8m33s       Normal    ScalingReplicaSet        rollout/istio-canary                 Scaled down ReplicaSet istio-canary-f7d4dcd68 (revision 15) from 10 to 1
8m33s       Normal    RolloutPaused            rollout/istio-canary                 Rollout is paused (CanaryPauseStep)
```

2. Next rollout is updated with the stable manifest. This is detected as a "rollback to stable". By looking at the Kubernetes events, we can see that we do things in the correct order:

* scale up stable
* shift traffic away from the previous desired RS (80 -> 20 -> 0)
* switch service selector of canary back to stable RS
* scaled down the old rs (previous desired)

```
8s          Normal    RolloutUpdated           rollout/istio-canary                 Rollout updated to revision 17
8s          Normal    SkipSteps                rollout/istio-canary                 Rollback to stable ReplicaSets
8s          Normal    RolloutResumed           rollout/istio-canary                 Rollout is resumed
8s          Normal    ScalingReplicaSet        rollout/istio-canary                 Scaled up ReplicaSet istio-canary-f7d4dcd68 (revision 17) from 1 to 10
0s          Normal    Updated VirtualService   rollout/istio-canary                 VirtualService `istio-canary` set to desiredWeight '80'
0s          Normal    TrafficWeightUpdated     rollout/istio-canary                 Traffic weight updated from 90 to 80
0s          Normal    Updated VirtualService   rollout/istio-canary                 VirtualService `istio-canary` set to desiredWeight '20'
0s          Normal    TrafficWeightUpdated     rollout/istio-canary                 Traffic weight updated from 80 to 20
0s          Normal    SwitchService            rollout/istio-canary                 Switched selector for service 'istio-canary-canary' from '7b8bcb8869' to 'f7d4dcd68'
0s          Normal    Updated VirtualService   rollout/istio-canary                 VirtualService `istio-canary` set to desiredWeight '0'
0s          Normal    TrafficWeightUpdated     rollout/istio-canary                 Traffic weight updated from 20 to 0
0s          Normal    ScalingReplicaSet        rollout/istio-canary                 Scaled down ReplicaSet istio-canary-7b8bcb8869 (revision 16) from 9 to 0
```

jessesuen:
This is ready for review (remaining work is to add unit + e2e tests)

rollout/canary.go
```diff
@@ -248,21 +242,81 @@ func (c *rolloutContext) scaleDownOldReplicaSetsForCanary(oldRSs []*appsv1.Repli
 	return totalScaledDown, nil
 }
 
-func (c *rolloutContext) replicaSetReferencedByCanaryTraffic(rs *appsv1.ReplicaSet) bool {
+// isReplicaSetReferenced returns if the given ReplicaSet is still being referenced by any of
+// the current, stable, blue-green services. Used to determine if the ReplicaSet can
+// safely be scaled to zero, or deleted.
+func (c *rolloutContext) isReplicaSetReferenced(rs *appsv1.ReplicaSet) bool {
```
zachaller, line 248: Would it make sense to move this to the rollout/context.go since it handles both bluegreen and canary?

jessesuen: Yes, that would make more sense. Will adjust.

jessesuen: I ended up moving it to replicaset.go since it's called `isReplicaSetReferenced`

zachaller (approved):

Gabryel8818:
Hello guys, I have same problem in 1.6.6
image

PR #3081, opened 2023-10-06
chore: refactor reconcileEphemeralMetadata to be more concise

zhaque44:
**Changes made:**

The refactored function behaves exactly the same, but it's more concise. The error handling is slightly better and the switch statement makes the code more idiomatic in Go. There is a unit test there to test the logic now

Checklist:


zhaque44:
@zachaller question about this PR, it’s saying coverage is 0.03 less from 81.74 to 81.71. I ran the tests in epehemeral tests and they all passed

rollout/ephemeralmetadata.go
```diff
@@ -14,43 +14,46 @@ import (
...
-		}
-		// sync stable metadata to the stable rs
-		err = c.syncEphemeralMetadata(ctx, c.stableRS, stableMetadata)
-		if err != nil {
-			return err
+		// If the rollout is not fully completed, we are in an upgrading state where newRS is a canary.
+		// Sync the new metadata to the newRS.
+		err = c.syncEphemeralMetadata(ctx, c.newRS, newMetadata)
+		// If there was no error syncing the new metadata, sync the stable metadata to the stableRS.
+		if err == nil {
```
meeech, line 41: I think this is less clear. I would keep the early err returns thoughtout the ifs...

meeech: Truth is, if you looking to tighten it up/ lose some indentation, you could keep the return in the if, and then lose the wrapping else entirely

zhaque44:
@meeech @zachaller can you guys kick off the build again. Appreciate it

rollout/ephemeralmetadata.go
```diff
@@ -27,21 +27,19 @@ func (c *rolloutContext) reconcileEphemeralMetadata() error {
...
-		if err != nil {
-			return err
-		}
-		// sync stable metadata to the stable rs
-		err = c.syncEphemeralMetadata(ctx, c.stableRS, stableMetadata)
-		if err != nil {
-			return err
-		}
+		return c.syncEphemeralMetadata(ctx, c.newRS, stableMetadata)
+	}
```
meeech, line 31: you have changed the behaviour here with this early return. the for loop never happens down below.

zhaque44: Thanks so much for your help, that was the issue I resolved it.

rollout/ephemeralmetadata.go
```diff
@@ -27,26 +27,24 @@ func (c *rolloutContext) reconcileEphemeralMetadata() error {
...
-		err := c.syncEphemeralMetadata(ctx, c.newRS, newMetadata)
-		if err != nil {
-			return err
-		}
-		// sync stable metadata to the stable rs
-		err = c.syncEphemeralMetadata(ctx, c.stableRS, stableMetadata)
-		if err != nil {
-			return err
-		}
+		return c.syncEphemeralMetadata(ctx, c.newRS, stableMetadata)
```
meeech, line 30: The mistake here is you are short circuiting, missing the `for` loop on line -48/+46 which would execute under the old code.

Issue #3125, opened 2023-10-25
Istio Subset-level traffic splitting removes other DestinationRoutes


dnguy078:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**
- Argo Rollouts removes any subset level DestinationRoutes in a VirtualService that it not modifying during a Canary Rollout.
- I have a VirtualService where I want to route a certain percentage to a newer system while maintaining a canary release for a legacy service. 
- In the following case below, Argo Rollouts will remove the DestinationRoutes with host `rollout-alpha-example` following a rollout. Ideally, I'd also want the flexibility to define a maxWeight in which Rollouts would setWeight up to the maxWeight for a HTTP DestinationRoute. 

```
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: istio-gateway
  labels:
    app: istio-gateway
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - "*"
---
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: rollout-vsvc
spec:
  gateways:
  - istio-rollout-gateway
  http:
  - match:
    - uri:
        prefix: "/somepath"
  - name: primary
    route:
    - destination:
        host: rollout-example
        subset: stable
      weight: 100
    - destination:
        host: rollout-example
        subset: canary
      weight: 0
    - destination:
        host: rollout-alpha-example
        subset: stable1
      weight: 20
---
apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: rollout-destrule
spec:
  host: rollout-example
  subsets:
  - name: canary
    labels:
      app: rollout-example
  - name: stable
    labels:
      app: rollout-example
---
apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: rollout-alpha-destrule
spec:
  host: rollout-alpha-example
  subsets:
  - name: stable1
    labels:
      app: rollout-alpha-example
```
Rollout.yaml
```
---
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: rollout-example
  labels:
    app: rollout-example
spec:
  strategy:
    canary:
      trafficRouting:
        istio:
          virtualService:
            name: rollout-vsvc  
            routes:
            - primary  
          destinationRule:
[...]

PR #3126, opened 2023-10-25
fix: istio destionationrule subsets enforcement

dnguy078:
Checklist:
Relates to issue in https://github.com/argoproj/argo-rollouts/issues/3125 


rollout/trafficrouting/istio/istio.go
```diff
@@ -255,6 +255,23 @@ func (r *Reconciler) reconcileVirtualService(obj *unstructured.Unstructured, vsv
 		if err = ValidateHTTPRoutes(r.rollout, vsvcRouteNames, httpRoutes); err != nil {
 			return nil, false, err
 		}
+
+		host, err := r.getDestinationRuleHost()
+		if err != nil {
+			return nil, false, err
+		}
+
+		if host != "" {
```
agrawroh, line 264: Is it possible to have an empty host? We should check if Istio allows it.

agrawroh (approved):
Looks good to me overall. @zachaller Could you please take a look as well?

rollout/trafficrouting/istio/istio.go
dnguy078: looks like the host is required for Istio [DestinationRule](https://istio.io/latest/docs/reference/config/networking/destination-rule/#DestinationRule). We can check do this check on the DestinationRule if included in `.spec.strategy.canary.trafficRouting.istio.destinationRule`

dnguy078:
Thank you for the review @agrawroh !

rollout/trafficrouting/istio/istio.go
dnguy078: ah that check already exist :) https://github.com/argoproj/argo-rollouts/blob/423be7b9ee37fade38a652b4fc6a43da8c842b5d/rollout/trafficrouting/istio/istio.go#L795

zachaller (approved):

zachaller:
@agrawroh thanks for the review, I also picked this back to v1.6.1 which I am releasing now

zachaller:
@dnguy078 This might have caused this: https://cloud-native.slack.com/archives/C01U781DW2E/p1698875164861259 you mind taking a peak?

zachaller:
Here is the issue: https://github.com/argoproj/argo-rollouts/issues/3141

dnguy078:
I'll take a look @zachaller

PR #2828, opened 2023-06-10
chore(deps): upgrade k8s version and client-go

fengshunli:
Checklist:

* [ ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [ ] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [ ] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)
* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ ] My builds are green. Try syncing with master if they are not. 
* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

PR #3211, opened 2023-12-04
chore(deps): Bump k8s dependencies to v1.26.0.

dlorenc:
This addresses a few CVEs, but was slightly challenging because of some Kubernetes API changes.

In particular, the Ingress type signatures changed. This required some surgery on a function that wrapped the extensionsv1beta1 Ingress and v1 Ingress types, attempting to provide some compatibility.

The API changes were kind of tough to deal with, but fortunately the only field required by anything else in this project was the Hostname field, so I changed the function to just normalize the hostnames into a slice.

Checklist:


dlorenc:
This was attempted back here, but it ran into the same k8s api type changes: https://github.com/argoproj/argo-rollouts/pull/2828

zachaller:
This is great, I did make a few changes to finish bumping the replace deps which fixed a few more tests but now it looks like manifest generation has some incompatibilities as well.

dlorenc:
> This is great, I did make a few changes to finish bumping the replace deps which fixed a few more tests but now it looks like manifest generation has some incompatibilities as well.

Thanks! I'm a bit out of my depth now, not sure how to debug these failures.

dlorenc:
Amazing work!!!

zachaller:
Thanks for contributing, was really helpful to have you go through and figure out all the logic required for this upgrade.

PR #3237, opened 2023-12-08
fix: make sure we use the updated rs when we write back to informer

zachaller:
We do this in addScaleDownDelay as well this was just missing and so we are writing the same value we pass into the function instead of the updated rs.


leoluz (approved):
LGTM

Issue #3256, opened 2023-12-15
the rollout will stuck in "progressing", if the new spec is applied with ("replicas" increasement, and pod spec is changed) when the rollout is in "paused".


andyliuliming:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

**To Reproduce**
1. deploy one "Rollout" with:
replicas: 3
  strategy:
    canary:
      dynamicStableScale: true
      minPodsPerReplicaSet: 1
      maxSurge: 1
      maxUnavailable: 1
      steps:
      - setCanaryScale: {weight: 60}
      - setWeight: 60
      - pause: {duration: 1000m}
 
 and then deploy change the replias of the Rollout.

then it will be stuck there. the promote also not work.

I think it's because the FindNewReplicaSet will return nil in this case.

**Expected behavior**

**Screenshots**

**Version**

latest in master branch.

```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

zachaller:
Are you able to reproduce the 100% of the time, I am suspicious that something is going on but am unable to reproduce. If you can get this to trigger easily can you maybe explain a bit more your exact steps.

zachaller:
I have been able to replicate something similar if I update both the replica count and the pod template at the same time, was this what you where doing?

andyliuliming:
> I have been able to replicate something similar if I update both the replica count and the pod template at the same time, was this what you where doing?

yes, forgot to mention that the pod spec is changed together with the replias.

andyliuliming:
I think this might help fix that:
https://github.com/argoproj/argo-rollouts/pull/3257

PR #3257, opened 2023-12-17
fix: fix the rollout stuck issue when pod/replicas changed together for canary rollout. 

andyliuliming:
when the rollout is in the "Pause" status,
and replica/poc spec changed together, then the rollout will stuck there forever.
this is because the newRS will be nil,
and there's one short circut return in reconcileNewReplicaSet().

issue reported in:
https://github.com/argoproj/argo-rollouts/issues/3256

the change will follow the "BlueGreen" style to handle this.
pick the replica set which the canary svc targeted to. and reconcile it.

Checklist:

### DIFF
--- a/rollout/service.go
+++ b/rollout/service.go
@@ -286,7 +286,7 @@ func (c *rolloutContext) reconcileStableAndCanaryService() error {
 }
 
 func (c *rolloutContext) getCanaryReplicaSet() (*appsv1.ReplicaSet, error) {
-	if c.rollout.Spec.Strategy.Canary == nil {
+	if c.rollout.Spec.Strategy.Canary == nil || c.rollout.Spec.Strategy.Canary.CanaryService == "" {
 		return nil, nil
 	}
 	svc, err := c.servicesLister.Services(c.rollout.Namespace).Get(c.rollout.Spec.Strategy.Canary.CanaryService)
--- a/rollout/service.go
+++ b/rollout/service.go
@@ -17,6 +17,7 @@ import (
 	serviceutil "github.com/argoproj/argo-rollouts/utils/service"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	patchtypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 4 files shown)


andyliuliming:
will add ut later.

rollout/service.go
```diff
@@ -284,6 +285,27 @@ func (c *rolloutContext) reconcileStableAndCanaryService() error {
 	return nil
 }
 
+func (c *rolloutContext) getCanaryReplicaSet() (*appsv1.ReplicaSet, error) {
+	if c.rollout.Spec.Strategy.Canary == nil {
+		return nil, nil
+	}
+	svc, err := c.servicesLister.Services(c.rollout.Namespace).Get(c.rollout.Spec.Strategy.Canary.CanaryService)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
carolwu-1206, line 294: If there is no canaryService, the rollout will be stuck too?

andyliuliming: yes, but that's a negative case.

but this is really a goot catch! i should check whether the canary service is an empty string first.

zachaller:
Linking to the BlueGreen issue for reference: https://github.com/argoproj/argo-rollouts/pull/3060

zachaller:
@andyliuliming I could still be doing something wrong so still digging but this PR does not seem to fix the issue for me.

zachaller:
I also think the issue is around this code:

```
func (c *rolloutContext) rolloutCanary() error {
	var err error
	if replicasetutil.PodTemplateOrStepsChanged(c.rollout, c.newRS) {
		c.newRS, err = c.getAllReplicaSetsAndSyncRevision(false)
		if err != nil {
			return err
		}
		return c.syncRolloutStatusCanary()
	}

	c.newRS, err = c.getAllReplicaSetsAndSyncRevision(true)
	if err != nil {
		return err
	}
```

zachaller:
This also happens in Rollouts 1.5

andyliuliming:
@zachaller 

> @andyliuliming I could still be doing something wrong so still digging but this PR does not seem to fix the issue for me.

this only works when there's canaryservice defined in rollout.
the fix works as expected in our rollout.

zachaller:
You are right I don't know how I missed that in my test rollout. Still going to spend some time on it though because I feel there is a more underlying issue, I want to investigate that a bit more but this does seem to fix the issue as long as you have canary service defined.

zachaller:
I think this would also fix the issue but also work for rollouts without a canary service: https://github.com/argoproj/argo-rollouts/pull/3272/files

andyliuliming:
> I think this would also fix the issue but also work for rollouts without a canary service: https://github.com/argoproj/argo-rollouts/pull/3272/files

tested. it works!

PR #3339, opened 2024-01-30
Add exception to `requireCanaryStableServices` to disable validation when using the `hashicorp/consul` plugin

wilkermichael:
Checklist:


david-yu:
Hi @zachaller friendly nudge, if you're able to review this PR for Consul Service Mesh.

pkg/apis/rollouts/validation/validation_test.go
```diff
@@ -129,6 +129,80 @@ func TestValidateRolloutStrategyBlueGreen(t *testing.T) {
 	assert.Equal(t, ScaleDownLimitLargerThanRevisionLimit, allErrs[1].Detail)
 }
 
+func TestValidateRolloutStrategyCanaryMissingServiceNames(t *testing.T) {
```
zachaller, line 132: Can we add a few test that use two traffic routers one where both are required to have canary service and one where it one requires it and one does not.

thisisnotashwin: Added some test cases with multiple routers where one requires the canary services and one or more do not to ensure they error as expected.

zachaller (approved):

thisisnotashwin:
Thanks so much for helping with this PR @zachaller

PR #3350, opened 2024-02-04
feat: allow analysis run to use separate kubeconfig for jobs

### LINKED ISSUE #3337: Option to specify a different KUBECONFIG where Analysis vs. Jobs would run
# Summary

Currently, when using the Job metric, the Job object will be created in the same cluster where the AnalysisRun resides. This proposal is to introduce a configuration option (environment variable) where a JOB_KUBECONFIG could be pointed to a **_different_** cluster where Jobs would be created (different from the AnalysisRuns).

# Use Cases

Having this option would enable some multi-cluster use cases, namely the ability for AnalysisTemplates, AnalysisRuns to be defined at a central location, without having to deploy/distribute the CRDs or AnalysisTemplates/Runs to workload clusters.

This should of course be an optional setting -- if the option is unspecified, the controller would maintain the current behavior. 

## Other notes

In addition to having a remote setting, the jobs should also have the option to be created using in-cluster credentials (instead of static credentials).

---

gdsoumya:
Fixes #3337 

This PR :
1. Enables setting `ARGO_ROLLOUTS_ANALYSIS_JOB_KUBECONFIG` to specify custom k8s config for running analysis jobs.
2. Also provides `ARGO_ROLLOUTS_ANALYSIS_JOB_NAMESPACE` to specify a different namespace from the analysis run to execute the job.

NOTE:

Setting custom `ARGO_ROLLOUTS_ANALYSIS_JOB_NAMESPACE` and `ARGO_ROLLOUTS_ANALYSIS_JOB_KUBECONFIG` will prevent automatic GC of jobs when analysis runs are deleted.

Checklist:


jessesuen (approved):
I did a careful review of this one and could not find anything. This LGTM!

jessesuen:
@zachaller did you want a chance to review this before merging? FWIW, we have also done manual, end-to-end testing of the feature.

zachaller:
I will take deeper look tomorrow, however one of my first thoughts is consistency with ENV var naming. Due to past [conventions](https://github.com/argoproj/argo-rollouts/pull/1956#discussion_r855856441) both plugins and various metric providers use `ARGO_ROLLOUTS_`  prefixes for env vars. I lean towards keeping consistent with that even though this somewhat piggy backs of of kubeclients KUBECONFIG env var. I tend to always prefer consistency though. Thoughts?

gdsoumya:
> I will take deeper look tomorrow, however one of my first thoughts is consistency with ENV var naming. Due to past [conventions](https://github.com/argoproj/argo-rollouts/pull/1956#discussion_r855856441) both plugins and various metric providers use `ARGO_ROLLOUTS_` prefixes for env vars. I lean towards keeping consistent with that even though this somewhat piggy backs off of kubeclients KUBECONFIG env var. I tend to always prefer consistency though. Thoughts?

Would you prefer `ARGO_ROLLOUT_ANALYSIS_JOB_KUBECONFIG` and `ARGO_ROLLOUTS_ANALYSIS_JOB_NAMESPACE` ? I dropped the prefix as the env was getting a bit long and also because we recently added the only analysis mode which makes it a standalone feature so thought just having `ANALYSIS_` prefix might work. Though I don't have a strong opinion on this and if you and @jessesuen feel a bigger env name isn't a issue I'll add it.

jessesuen:
Thanks @zachaller, that's a great call out. I agree the ARGO_ROLLOUTS_ prefix is an important convention to keep. So despite the long name, let's use that.

gdsoumya:
Updated env names

zachaller (approved):
LGTM

PR #3371, opened 2024-02-14
feat: ping pong support for istio

zachaller:
Checklist:


test/e2e/istio_test.go
```diff
@@ -492,3 +492,124 @@ func (s *IstioSuite) TestIstioSubsetSplitExperimentStep() {
...
+		Then().
+		//Assert(assertWeights(s, "ping-service", "pong-service", 100, 0))
+		Assert(func(t *fixtures.Then) {
+			vsvc := t.GetVirtualService()
+			assert.Equal(s.T(), int64(100), vsvc.Spec.HTTP[0].Route[0].Weight)
+			assert.Equal(s.T(), int64(0), vsvc.Spec.HTTP[0].Route[1].Weight)
+		})
+}
+
+func (s *IstioSuite) TestIstioSubsetSplitPingPong() {
```
mayzhang2000, line 547: I am not sure we need this test case since splinting by subset does not need to provide stable or canary service, nor ping pong service.

pkg/apis/rollouts/validation/validation_test.go
```diff
@@ -284,6 +284,10 @@ func TestValidateRolloutStrategyCanary(t *testing.T) {
 		validRo.Spec.Strategy.Canary.Steps[0].SetWeight = pointer.Int32(10)
 		validRo.Spec.Strategy.Canary.CanaryService = ""
 		validRo.Spec.Strategy.Canary.StableService = ""
+		validRo.Spec.Strategy.Canary.PingPong = &v1alpha1.PingPongSpec{
```
leoluz, line 287: Shouldn't we add a new testcase for istio?

rollout/trafficrouting/istio/istio_test.go
```diff
@@ -263,6 +298,72 @@ spec:
         host: canary
       weight: 0`
 
+const regularMixedVsvcPingPong = `apiVersion: networking.istio.io/v1alpha3
```
leoluz, line 301: I prefer using `go:embed` directive to accomplish the same goal of having a large file instantiating its contents in a variable. This approach is cleaner/easier to maintain and doesn't require any additional tooling. 

See the following PR as an example:
https://github.com/argoproj/argo-cd/pull/17404/files

leoluz (commented):
Please check my minor comments

rollout/trafficrouting/istio/istio_test.go
zachaller: Yea, I kept it the way it is because it is consistent with all the other tests for istio

pkg/apis/rollouts/validation/validation_test.go
zachaller: added new test instead of piggy backing off one

leoluz (approved):
LGTM

rollout/trafficrouting/istio/istio_test.go
leoluz: sure.. but we have to start somewhere ¯\_(ツ)_/¯

PR #3377, opened 2024-02-16
docs: Update shell autocompletion instructions

jopit:
The docs currently describe how to enable shell auto completion for the rollouts CLI when run as a standalone binary, but the don't describe how to get auto completion to work when used from `kubectl`.

This PR updates the docs with instructions on how this can be done, based on the information here: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2379-kubectl-plugins#shell-completion-support

This works with `kubectl` version 1.26 and newer.

Checklist:

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -81,7 +81,21 @@ kubectl argo rollouts version
 
 ## Shell auto completion
 
-The CLI can export shell completion code for several shells.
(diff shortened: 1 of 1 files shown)


zachaller (approved):

kzap:
@jopit how does this work just by putting a file in `/usr/local/bin/` ? Is that a known path or must one also setup `kubectl argo rollouts completion bash` ?

Issue #3361, opened 2024-02-12
A default valueFrom not accepted on AnalysisTemplate


rolandkool:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

An `AnalysisTemplate` with a parameter that has a default value using `valueFrom` does not work. The `Rollout` will fail with a message that the parameter needs to be supplied.

**To Reproduce**

The `AnalysisTemplate` contains a parameter like this:

```
spec:
  args:
  - name: namespace
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
```

However, when the analysis is called without providing the `namespace` parameter, the rollout fails. If the `namespace` parameter is provided without a value, the value used by the analysis is empty.

According to the docs, this seems to be possible if it were a secret reference:
https://argo-rollouts.readthedocs.io/en/stable/features/analysis/#analysis-template-arguments

**Expected behavior**

It is expected that the `AnalysisTemplate` can be called without specifying the `namespace` parameter.

**Version**

v1.6.5

**Logs**

```
Degraded - RolloutAborted: Rollout aborted update to revision 2: Unable to resolve metric arguments: argument "namespace" was not supplied
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

todaywasawesome:
@kostis-codefresh can you take a look?

newtondev:
I have added a potential way of solving this with the ability to extend for any future fields that might want to be referenced. Any suggestions for change is also welcome. Thanks

PR #3436, opened 2024-03-10
fix: allow namespace default using valueFrom as arg for AnalysisTempl…

newtondev:
…ate (#3361)

Checklist:

### DIFF
--- a/analysis/analysis.go
+++ b/analysis/analysis.go
@@ -56,7 +56,7 @@ func (c *Controller) reconcileAnalysisRun(origRun *v1alpha1.AnalysisRun) *v1alph
 		run.Status.MetricResults = make([]v1alpha1.MetricResult, 0)
 	}
 
-	resolvedMetrics, err := getResolvedMetricsWithoutSecrets(run.Spec.Metrics, run.Spec.Args)
+	resolvedMetrics, err := getResolvedMetricsWithoutSecrets(run)
 	if err != nil {
 		message := fmt.Sprintf("Unable to resolve metric arguments: %v", err)
 		logger.Warn(message)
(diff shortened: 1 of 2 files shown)


analysis/analysis.go
```diff
@@ -260,6 +267,21 @@ func parseMetricInterval(logCtx log.Entry, metricDurationString v1alpha1.Duratio
 	return metricInterval, nil
 }
 
+// resolveArgsFromValueFromFieldRef resolves args when using ValueFrom.FieldRef.FieldPath
+// returns argument with default value from lookup
+func resolveArgsFromValueFromFieldRef(arg v1alpha1.Argument, namespace string) v1alpha1.Argument {
+	if arg.ValueFrom.FieldRef != nil && arg.ValueFrom.FieldRef.FieldPath != "" {
+		switch arg.ValueFrom.FieldRef.FieldPath {
+		case "metadata.namespace":
```
zachaller, line 275: Does this mean that we will only support metadata.namespace for fieldPath lookups?

zachaller, line 275: If I am following this correctly does this mean that only metadata.namespace works?

zachaller: Wondering why this does not seem be called form this path: https://github.com/argoproj/argo-rollouts/blob/c6ba82cca216c7c9713c7cd69dae1ebe25ba33a8/utils/analysis/factory.go#L34

zachaller:
How are you creating the AnalysisRun, is the argo rollouts controller creating it or are you creating it say via kubectl or something similar outside of rollouts controller?

newtondev:
So it seems like the problem here is mainly that when wanting to not have to pass in the namespace from the Rollout to the AnalysisTemplate. If the arg is not specified but is referenced in the AnalysisTemplate like namespace, it should default to the namespace from the Rolllout to be used in metrics provider like Prometheus.

Thanks @zachaller , I will take a deeper look into the ExtractFieldPathAsString to see if there is a good way to handle this kind of default.

PR #3467, opened 2024-03-23
fix(controller): prevent negative vsvc weights on a replica scaledown following a canary abort for istio trafficrouting

shubhajyoti-bagchi-groww:
fixes: https://github.com/argoproj/argo-rollouts/issues/3470

Checklist:


shubhajyoti-bagchi-groww:
Hi @OpenGuidou, thanks for the response

Right, this happens in cases whenever stable.AvailableReplicas > rollout.spec.Replicas

HPA is one way where stable.AvailableReplicas > rollout.spec.Replicas without any manual intervention
However, in the issue in which we faced, we had a need to reduce the number of replicas of the rollout which was in aborted state and came across the issue.

The e2e test here focuses on breaking the existing flow using one of the scenarios, without this fix. It kind of reproduces exactly what we faced.

OpenGuidou (approved):
LGTM, thanks !

Eslam-mohammed-anwar:
we have a similar issue https://github.com/argoproj/argo-rollouts/issues/3470

zachaller (approved):

Eslam-mohammed-anwar:
@zachaller when should it be released ?

PR #3531, opened 2024-04-18
docs: provide recommendation for strategies

kostis-codefresh:
Checklist:

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -60,7 +60,7 @@ of blue/green deployments but only the basic capabilities of canaries.
 * Blue/Green also works with services that use queues and databases (workers that fetch tasks). Argo Rollouts doesn't control traffic flow for
 connections it doesn't understand (i.e. binary/queue channels).
 
-Here is a summary table for the two approaches.
+Here is a summary table for the possible approaches.
 
 |                           |         Blue/Green        |       Basic Canary         | Canary with Traffic manager    |
 |--------------------------:|:-------------------------:|:--------------------------:| :-----------------------------:|
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -54,13 +54,15 @@ In general Blue/Green is the easier strategy to start with, but also the more li
 
 You also need to examine if your application can handle canaries or not.
 
-* Blue/Green always works because only one application is active at a time. Not all applications can have different versions running in parallel at the same time (which is what canaries are doing)
-* Blue/Green is simpler because it works WITHOUT a traffic manager. Not all people like service meshes, and not all people have an ingress that is supported by Argo Rollouts
-* Blue/Green also works with services that use queues and databases (workers that fetch tasks)
+* Blue/Green always works because only one application is active at a time. Not all applications can have different versions running in parallel at the same time (which is what canaries are doing). This can be a showstopper for adopting canary deployments especially for legacy applications.
+* Blue/Green is simpler because you can get their full value WITHOUT a traffic manager. While canaries can also work without a traffic manager, most of their advanced features assume a fine-grained way to control traffic. If you don't have a traffic manager, then you can easily get the full value
+of blue/green deployments but only the basic capabilities of canaries.
+* Blue/Green also works with services that use queues and databases (workers that fetch tasks). Argo Rollouts doesn't control traffic flow for
+connections it doesn't understand (i.e. binary/queue channels).
 
 Here is a summary table for the two approaches.
 
-|                           |         Blue/Green         | Canary         |
+|                           |         Blue/Green         | Canary with Traffic manager         |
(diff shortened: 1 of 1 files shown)


docs/concepts.md
```diff
@@ -46,4 +46,30 @@ A Canary deployment exposes a subset of users to the new version of the applicat
...
+and percentages of traffic according to your use case.
+
+## Which strategy to choose
+
+In general Blue/Green is the easier strategy to start with, but also the more limited. We recommend you start with Blue/Green deployments first and as you gain confidence for your metrics and applications switch to Canaries.
+
+You also need to examine if your application can handle canaries or not.
+
+* Blue/Green always works because only one application is active at a time. Not all applications can have different versions running in parallel at the same time (which is what canaries are doing)
+* Blue/Green is simpler because it works WITHOUT a traffic manager. Not all people like service meshes, and not all people have an ingress that is supported by Argo Rollouts
```
meeech, line 58: my only issue here is it muddies the fact that progressive canary also has a no traffic manager option.

docs/concepts.md
```diff
@@ -46,4 +46,30 @@ A Canary deployment exposes a subset of users to the new version of the applicat
...
+* Blue/Green also works with services that use queues and databases (workers that fetch tasks)
+
+Here is a summary table for the two approaches.
+
+|                           |         Blue/Green         | Canary         |
+|--------------------------:|:-------------------------:|:--------------------------:|
+|                Ease of adoption |           Low             |     High                   | 
+|               Flexibility | Low        |  High                        | 
+|               Needs traffic provider | No        |  Yes                        |
```
meeech, line 67: I don't know that I agree with the conclusions in this chart.

docs/concepts.md
meeech: ( i just saw the note at the bottom, but comment still stands)

docs/concepts.md
```diff
@@ -46,4 +46,30 @@ A Canary deployment exposes a subset of users to the new version of the applicat
...
+
+You also need to examine if your application can handle canaries or not.
+
+* Blue/Green always works because only one application is active at a time. Not all applications can have different versions running in parallel at the same time (which is what canaries are doing)
+* Blue/Green is simpler because it works WITHOUT a traffic manager. Not all people like service meshes, and not all people have an ingress that is supported by Argo Rollouts
+* Blue/Green also works with services that use queues and databases (workers that fetch tasks)
+
+Here is a summary table for the two approaches.
+
+|                           |         Blue/Green         | Canary         |
```
meeech, line 63:
```suggestion
|                           |         Blue/Green         | Canary  w/Service Mesh       |
```

kostis-codefresh: True. But the big blocker is that Canary needs your application to have 2 instances that are both receiving live traffic at the same time. And not all  apps can do this. The traffic manager is important but completely unrelated to that requirement.

docs/concepts.md
kostis-codefresh: ok. I will update to make it clear.

docs/concepts.md
```diff
@@ -46,4 +46,30 @@ A Canary deployment exposes a subset of users to the new version of the applicat
...
+
+* Blue/Green always works because only one application is active at a time. Not all applications can have different versions running in parallel at the same time (which is what canaries are doing). This can be a showstopper for adopting canary deployments especially for legacy applications.
+* Blue/Green is simpler because you can get their full value WITHOUT a traffic manager. While canaries can also work without a traffic manager, most of their advanced features assume a fine-grained way to control traffic. If you don't have a traffic manager, then you can easily get the full value
+of blue/green deployments but only the basic capabilities of canaries.
+* Blue/Green also works with services that use queues and databases (workers that fetch tasks). Argo Rollouts doesn't control traffic flow for
+connections it doesn't understand (i.e. binary/queue channels).
+
+Here is a summary table for the two approaches.
+
+|                           |         Blue/Green         | Canary with Traffic manager         |
```
zachaller, line 65: Do you think we should also call out basic canary, aka canary without traffic router

kostis-codefresh: @zachaller I added a third row to the table to make it clear.

docs/concepts.md
meeech: Fair. I guess we rank that a bit different (in terms of ease). but i guess there's a big bag of gotchas/tradeoffs, and I get this is meant as an intro guide, so I'll try to avoid getting bogged in details.

docs/concepts.md
meeech: thansk. I guess my issue was the big WITHOUT traffic manager, implying its a necessity for canary.

zachaller (approved):

PR #3602, opened 2024-05-28
fix(istio): Fix downtime on initial deployment using Istio DestinationRule Subsets. Fixes #2507

### LINKED ISSUE #2507: Downtime On initial deployment using Istio destination rule
Checklist:

**Describe the bug**
When we initialized rollout for a deployment with Istio traffic routing, we found rollout controller created a new RS from the current RS template with a new label key named 'rollouts-pod-template-hash', and the controller update Istio DR with the label immediately without considering whether the new RS is available or not. This leads 503 http error, no healthy upstreams.

**To Reproduce**
1. create virtualservice & destinationrule resources as [this](https://github.com/argoproj/rollouts-demo/tree/master/examples/istio-subset).
2. create rollout resource following the migration [doc](https://argoproj.github.io/argo-rollouts/migrating/) 
3. While creating the rollout resource, the rollout controller will create a RS for it ,and immediately update the Istio Destination Rule and Virtual Service resource, which leading `no healthy upstream` error because the the RS is not ready for traffic.

**Expected behavior**
The rollout controller shift the traffic to rollout's RS while the RS is available.

**Version**
argo-rollouts: v1.2.2

Could we add some determine statements to ensure the new RS is available before updating the Istio DR, which is the same like [ensureSVCTarget](https://github.com/argoproj/argo-rollouts/blob/master/rollout/service.go#L272) does?

---

wmuizelaar:
Checklist:


newtondev (approved):
Looks good. I sat with Wietse on a call and we went through it together.

newtondev:
LGTM

wmuizelaar:
@zachaller are you able to take a look at this? It would be much appreciated! 🙏

zachaller:
Yea, at first glance I think I would like to try and not change the traffic router interface, this will cause all plugins to have to update etc etc. I think an ok alternative is to pass the replicasetInformer/Lister/k8s client or just add roCtx.allRSs into the istio reconciler context and use that to get the replicasets. If this where to be a plugin they would have to do something similar as well.

wmuizelaar:
Thanks for your remark, I agree that it would be really nice if the interface could stay the same. Let me experiment if I can get it passed to the Istio Reconciler in a different way.

wmuizelaar:
@zachaller can you take a look again please? This feels indeed a lot simpler and easier way of implementation.

Bennett-Lynch:
@zachaller Gentle bump. Looking to see if this also resolves https://github.com/argoproj/argo-rollouts/issues/3681.

zachaller:
Sorry, it's been a while I will take a look at this soon.

rollout/trafficrouting/istio/istio.go
```diff
@@ -315,6 +320,17 @@ func (r *Reconciler) reconcileVirtualService(obj *unstructured.Unstructured, vsv
 }
 
 func (r *Reconciler) UpdateHash(canaryHash, stableHash string, additionalDestinations ...v1alpha1.WeightDestination) error {
+	// We need to check if the replicasets are ready here as well if we didn't define any services in the rollout
+	// See: https://github.com/argoproj/argo-rollouts/issues/2507
+	if r.rollout.Spec.Strategy.Canary.CanaryService == "" && r.rollout.Spec.Strategy.Canary.StableService == "" {
```
Bennett-Lynch, line 325: Do we know for sure this is being properly checked when services are defined? In the example in https://github.com/argoproj/argo-rollouts/issues/3681 I am defining both a canary service and stable service but still experiencing a brief downtime/outage after the rollout ends.

wmuizelaar: Aaah, if you explicitly define the services, the existing code should actually be working - the fix I built here is explicitly for when you are using the Istio DestionationRule-method to update your traffic (https://argo-rollouts.readthedocs.io/en/stable/features/traffic-management/istio/#subset-level-traffic-splitting to be precise).

If you define the services (I did not see those in your example, so that's why I thought it might be related to this), there are other codepaths that actually check service-readiness. There might be a big/issue in there, but then your issue is at least definitely NOT related to mine.

zachaller (approved):

PR #3664, opened 2024-06-24
fix(controller): use the stableRS from the rollout context rather tha…

### LINKED ISSUE #3663: A scaling event during a post promotion analysis can leave rollout in irreconcilable state, as the stableRS is not updated.
Checklist:

**Describe the bug**
A scaling event which occurs during a post-promotion analysis template run causes only the active replicaset to scaled and not the stable replicaset. This in-turn causes the rollout to be in an irreconcilable state as it waits for the stable replicaset to have the same number of replicas as the rollout requires.

**To Reproduce**

1. Deploy a new version of a bluegreen rollout, with a post promotion analysis template
2. Wait for the rollout to automatically promote the new replicaset and run it's post promotion analysis template
3. While it runs the analysis template the new replica set is considered _active_ but not _stable_
4. Run a scaling event during this period (while the new replicaset is considered active but not stable)
5. This causes the active (new replicaset) to be scaled to whatever replica count you request but the stable replicaset is untouched
6. Wait for the analysis template to finish
7. Observe that while the (old replicaset) _stable_ is not equal to the replica count desired by the rollout the rollout is stuck waiting to for minimum availability.

**Expected behavior**
During the period of post promotion analysis, a scaling event should affect both the _stable_ and the _active_ replicaset replicas. This should ensure that in the case there is a genuine traffic spike and error concurrently we rollback to a stable replicaset with enough replicas and also to ensure that we don't get stuck waiting for a rollout that is irreconcilable.

**Screenshots**

**Version**
v1.7.0 

**Logs**
infinite loop of:
```
time="2024-06-21T10:59:40Z" level=info msg="Syncing replicas only due to scaling event" namespace=staging rollout=my-app
time="2024-06-21T10:59:40Z" level=info msg="Reconciling stable ReplicaSet 'my-app-6b6649b84d'" namespace=staging rollout=my-app
time="2024-06-21T10:59:40Z" level=info msg="No status changes. Skipping patch" generation=1324 namespace=staging resourceVersion=159634477 rollout=my-app
time="2024-06-21T10:59:40Z" level=info msg="Queueing up Rollout for a progress check now" namespace=staging rollout=my-app
time="2024-06-21T10:59:40Z" level=info msg="Reconciliation completed" generation=1324 namespace=staging resourceVersion=159634477 rollout=my-app time_ms=29.734882
time="2024-06-21T10:59:40Z" level=info msg="Started syncing rollout" generation=1324 namespace=staging resourceVersion=159634477 rollout=my-app
time="2024-06-21T10:59:40Z" level=info msg="Syncing replicas only due to scaling event" namespace=staging rollout=my-app
```
---

Impacted by this bug? Give it a :+1:. We prioritize the issues with the most :+1:.

benminter-treatwell:
…n inferring it from the active selector, to deal with the edge case where the stableRS != activeRS during analysis templates. Fixes #3663

Checklist:

### DIFF
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -71,18 +71,14 @@ func (c *rolloutContext) rolloutBlueGreen() error {
 	return c.syncRolloutStatusBlueGreen(previewSvc, activeSvc)
 }
 
-func (c *rolloutContext) reconcileBlueGreenStableReplicaSet(activeSvc *corev1.Service) error {
-	if _, ok := activeSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey]; !ok {
-		return nil
-	}
-	activeRS, _ := replicasetutil.GetReplicaSetByTemplateHash(c.allRSs, activeSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey])
-	if activeRS == nil {
-		c.log.Warn("There shouldn't be a nil active replicaset if the active Service selector is set")
+func (c *rolloutContext) reconcileBlueGreenStableReplicaSet() error {
+	if c.stableRS == nil {
+		c.log.Info("stableRS is nil on first deployment, we don't need to scale it in this case")
 		return nil
 	}
 
-	c.log.Infof("Reconciling stable ReplicaSet '%s'", activeRS.Name)
-	_, _, err := c.scaleReplicaSetAndRecordEvent(activeRS, defaults.GetReplicasOrDefault(c.rollout.Spec.Replicas))
+	c.log.Infof("Reconciling stable ReplicaSet '%s'", c.stableRS.Name)
+	_, _, err := c.scaleReplicaSetAndRecordEvent(c.stableRS, defaults.GetReplicasOrDefault(c.rollout.Spec.Replicas))
 	if err != nil {
 		return fmt.Errorf("failed to scaleReplicaSetAndRecordEvent in reconcileBlueGreenStableReplicaSet: %w", err)
 	}
@@ -94,7 +90,7 @@ func (c *rolloutContext) reconcileBlueGreenReplicaSets(activeSvc *corev1.Service
 	if err != nil {
 		return err
 	}
-	err = c.reconcileBlueGreenStableReplicaSet(activeSvc)
+	err = c.reconcileBlueGreenStableReplicaSet()
 	if err != nil {
 		return err
 	}
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -73,7 +73,7 @@ func (c *rolloutContext) rolloutBlueGreen() error {
 
 func (c *rolloutContext) reconcileBlueGreenStableReplicaSet() error {
 	if c.stableRS == nil {
-		c.log.Info("stableRS is nil on first deployment, we don't need to scale it in this case")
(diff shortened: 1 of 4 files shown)


rollout/ephemeralmetadata_test.go
```diff
@@ -169,7 +169,7 @@ func TestSyncBlueGreenEphemeralMetadataSecondRevision(t *testing.T) {
 	f := newFixture(t)
 	defer f.Close()
 
-	r1 := newBlueGreenRollout("foo", 1, nil, "active", "preview")
+	r1 := newBlueGreenRollout("foo", 3, nil, "active", "preview")
```
benminter-treatwell, line 172: The number of replicas defined in the rollout were not equal to number in the replicasets. My change correctly caused an extra scaling event where reconcile the fact the replicas were out of sync.

rollout/bluegreen.go
```diff
@@ -71,18 +71,14 @@ func (c *rolloutContext) rolloutBlueGreen() error {
...
-func (c *rolloutContext) reconcileBlueGreenStableReplicaSet(activeSvc *corev1.Service) error {
-	if _, ok := activeSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey]; !ok {
-		return nil
-	}
-	activeRS, _ := replicasetutil.GetReplicaSetByTemplateHash(c.allRSs, activeSvc.Spec.Selector[v1alpha1.DefaultRolloutUniqueLabelKey])
-	if activeRS == nil {
-		c.log.Warn("There shouldn't be a nil active replicaset if the active Service selector is set")
+func (c *rolloutContext) reconcileBlueGreenStableReplicaSet() error {
+	if c.stableRS == nil {
+		c.log.Info("stableRS is nil on first deployment, we don't need to scale it in this case")
```
agrawroh, line 76: This log message talks about `scaling` which seems unintuitive as the method is simply trying to reconcile (scaling is done as part of the reconciliation)

You can just remove the log or change it to say, _"Stable ReplicaSet doesn't exist and hence no reconciliation is required."_

benminter-treatwell: Thanks for the input, I've updated the comment to what you suggested.

agrawroh: Thanks! Could you please also add a test for this scenario?

benminter-treatwell: Hmm, not sure what exactly needs testing here. Other tests which initialize a deployment are inherently testing this path, but there is nothing measurable to test in this case? Scaling the rollout during deployment wouldn't help either as the other methods during the reconciliation process would cause the replicaset to scale, meaning we cannot expect _nothing_ to happen

Any help/specifics would be awesome for this one.

agrawroh: Could you add an explicit test case which asserts on whether the log line: _"Stable ReplicaSet doesn't exist and hence no reconciliation is required."_ is emitted?

benminter-treatwell: Right, I've attempted that - hopefully it looks okay.

agrawroh (approved):
Looks good. @zachaller Could you do a final pass?

benminter-treatwell:
Doing a rebase to keep this up to date and ready to go

benminter-treatwell:
@zachaller dropping a nudge here, would be awesome to get this merged!

Issue #3702, opened 2024-07-04
Support prometheus range queries


mclarke47:
# Summary/Use Cases

Currently the prometheus provider uses the `Query` method which corresponds to the ["instant query" `/query`](https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries) endpoint in prometheus. This endpoint returns the latest value for a time series and so works well when the prometheus query provided boils down the query result to a single value which can then be checked in the `successCondition`/`failureCondition`. 

However it would great to also support [range queries](https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries) which would return multiple results for a time series and thus might be more intuitive for users to understand and write queries for as it's more graph like, than query which is a single point in time.

Likely changes:

- Update the prometheus provider to add configuration of a time range required by the [QueryRange](https://github.com/prometheus/client_golang/blob/1a5f005c914eee4a9f0adf5b0fb19007683a6d99/api/prometheus/v1/api.go#L484) method in the prometheus client. Not 100% clear on this API design, interested in input from the maintainers.
- Update the run method of [the provider](https://github.com/argoproj/argo-rollouts/blob/590206b9bca395a3a71244fd0384c6e1d74585b8/metricproviders/prometheus/prometheus.go#L69) to use `QueryRange` when the time range above is set in the prometheus provider 
- Update docs, we'll need to highlight that since we are returning multiple results notation similar to `all(result, # > {{args.threshold}})` will be necessary

Am willing to submit a PR with guidance 🙏 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #3704, opened 2024-07-04
feat(metricprovider): add prometheus range query support

mclarke47:
fixes: https://github.com/argoproj/argo-rollouts/issues/3702

Checklist:


mclarke47:
It would be nice if we could parameterize `lookBackDuration` through args also, but not 100% sure how to do this

mclarke47:
Query ranges are syntactic sugar for the query endpoint as it says in that blog post. However I think there are 2 main benefits to adding query ranges: 

* They allow writing queries similar to how they are viewed in graphs and so IMO makes it more intuitive, and you can re-use queries from other sources without needing to first convert them to queries that would work with the query endpoint
* Query ranges support the step argument which allows you to specify the windows in which the queries over the range are executed so you get a little more control over how the data it returned from prometheus

zachaller:
That's fair, I do see it being a bit easier to reason about, also do you think it make sense to expose the start and end parameters vs hiding that behind `lookBackDuration`?

mclarke47:
I think that's would be good, although it should probably be relative to `time.Now()` otherwise we would be hard coding a datetime which seems not very useful.

@zachaller is there a way to access args of the analysis template inside the provider so that we can parameterize things?

zachaller:
> I think that's would be good, although it should probably be relative to `time.Now()` otherwise we would be hard coding a datetime which seems not very useful.
> 
> @zachaller is there a way to access args of the analysis template inside the provider so that we can parameterize things?

Pretty sure this works today because we do things like:
```
provider:
        prometheus:
          address: "http://prometheus:{{args.prometheus-port}}"
          query: |-
            (quantile(0.5, quantile_over_time(0.5, namespace_pod_cpu_utilization{namespace="{{args.namespace}}", pod=~".*-{{args.canary-hash}}-.*"}[11m])))
```

Might be worth checking if it still works with your code as well just incase the template system explicity looks for those fields etc.

IvanManz (approved):

mclarke47:
Will double check if this already works

mclarke47:
Confirming:

```
...
  - name: lookback_duration
    value: "5m"
  - name: lookback_step
    value: "1m"
...    
    provider:
      prometheus:
        rangeQuery:
          lookBackDuration: "{{args.lookback_duration}}"
          step: "{{args.lookback_step}}"
...          
```
does work

mclarke47:
There are time functions in [expr](https://expr-lang.org/docs/language-definition#date-functions), I'm going to change the API to be start/end/step, with start's default being `now()`. I'll test this out then update the docs.

mclarke47:
hmmm, it seems this expression logic is only evaluated in success/failure conditions not in the actual fields of the provider, only args substitution works there.

I could:

* Stick with a lookback duration as is currently implemented
* Put in some logic to Evaluate these fields with `util/evaluate`

WDYT @zachaller?


Screenshot 2024-07-08 at 2 15 10 PM

mclarke47:
Hey @zachaller, I pushed that change to treat the start/end fields as expr expressions. 🙏

utils/evaluate/evaluate.go
```diff
@@ -56,6 +57,41 @@ func EvaluateResult(result any, metric v1alpha1.Metric, logCtx logrus.Entry) (v1
 	return v1alpha1.AnalysisPhaseSuccessful, nil
 }
 
+func EvalTime(expression string) (time.Time, error) {
+	var err error
+
+	env := map[string]any{
+		"asInt":   asInt,
+		"asFloat": asFloat,
```
zachaller, line 65: Let's remove asInt, asFloat and just use the built in funcs https://expr-lang.org/docs/language-definition#int

utils/evaluate/evaluate.go
```diff
@@ -56,6 +57,41 @@ func EvaluateResult(result any, metric v1alpha1.Metric, logCtx logrus.Entry) (v1
...
+	output, err := expr.Run(program, env)
+	if err != nil {
+		return time.Time{}, unwrapFileErr(err)
+	}
+
+	switch val := output.(type) {
+	case time.Time:
+		return val, nil
+	default:
+		return time.Time{}, fmt.Errorf("expected string, but got %T", val)
```
zachaller, line 91: expected time.Time

metricproviders/prometheus/prometheus.go
```diff
@@ -112,29 +136,51 @@ func (p *Provider) GarbageCollect(run *v1alpha1.AnalysisRun, metric v1alpha1.Met
...
 
+func sampleValuesToFloatSlice(SampleValues []model.SampleValue) []float64 {
+	results := make([]float64, 0, len(SampleValues))
+	for _, s := range SampleValues {
+		results = append(results, float64(s))
+	}
+	return results
+}
+
+func sampleValuesToResultStr(SampleValues []model.SampleValue) string {
```
zachaller, line 147: Can we lowercase SampleValues

zachaller:
Just a few small changes then this LGTM

metricproviders/prometheus/prometheus.go
mclarke47: fixed!

metricproviders/prometheus/prometheus.go
```diff
@@ -112,29 +136,51 @@ func (p *Provider) GarbageCollect(run *v1alpha1.AnalysisRun, metric v1alpha1.Met
 	return nil
 }
 
+func sampleValuesToFloatSlice(SampleValues []model.SampleValue) []float64 {
```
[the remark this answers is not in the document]
mclarke47: fixed!

utils/evaluate/evaluate.go
mclarke47: fixed!

utils/evaluate/evaluate.go
mclarke47: fixed!

mclarke47:
@zachaller have address your comments 🙏

zachaller:
Thank you for the contribution!

PR #3727, opened 2024-07-17
fix(trafficrouting): Fix setHeaderRoute behavior to affect only related ingress rules

ArenSH:
This PR fixes `ALB::SetHeaderRoute` so it only affects Ingress rules that contain `stableService` as backend.

For example, if we have some ingress:
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: some-ingress
  annotations:
    alb.ingress.kubernetes.io/actions.redirect-to-app: 'some action'
    # ...
spec:
  defaultBackend:
    service:
      name: svc-stable
      port:
        name: some-port
  rules:
  - host: host1.example.local
    http:
      paths:
      - backend:
          service:
            name: redirect-to-app
            port:
              name: use-annotation
        path: /
        pathType: ImplementationSpecific
  - host: host2.example.local
    http:
      paths:
      - backend:
          service:
            name: redirect-to-app
            port:
              name: use-annotation
        path: /
        pathType: ImplementationSpecific
  - http:
      paths:
      - backend:
          service:
            name: svc-stable
            port:
              name: use-annotation
        path: /*
        pathType: ImplementationSpecific
```

And rollout with the following step:
```yaml
#...
      - setHeaderRoute:
          match:
          - headerName: x-user-agent
            headerValue:
              exact: test-framework
          name: route-test-requests
```

Then this `route-test-requests` path will be added to _every_ rule in Ingress. Documentation leaves impression that only rules related to rollout services will be affected.

I'm not sure this fix is consistent with desired argo-rollouts behavior, sou would be happy to receive feedback and adjust the code.

Checklist:


zachaller:
With this PR does this mean that with set heady it will not send 100% of traffic with the configured header to the canary?

ArenSH:
> With this PR does this mean that with set heady it will not send 100% of traffic with the configured header to the canary?

Sorry for delayed response. I'm not sure I understand the case correctly. The `getTrafficForwardActionString` is not affected and it sets 100% weight to canary for matching header. So the annotation works as expected.
 
The problem this fix addresses is that Ingress might contain additional rules (like redirects in example) and they are also affected (the generated `Path` is prepended there too) by `setHeaderRoute`.

With this fix, the generated `Path` is only prepended to rules that target `stable` service. And, if I'm not mistaken, Ingress should not contain explicit rules that route traffic to `canary` - that is argo-rollouts job. So any other rules remain untouched.

Issue #3707, opened 2024-07-05
Broken Datadog apiVersion v2 analyses not specifying aggregator in v1.7.1


aguiraorodriguez:
Checklist:

* [ ] I've included steps to reproduce the bug.
* [ ] I've included the version of argo rollouts.

**Describe the bug**

Datadog analysis using `apiVersion: v2` that don't specify an `aggregator` fail. The error message is:

`"Error Message: received non 2xx response code: 400 {"errors":["Value of parameter 'aggregator' should be any of these [None, 'avg', 'min', 'max', 'sum', 'last', 'mean', 'area', 'l2norm', 'percentile', 'stddev']"]}"`

Probably this is a side effect of https://github.com/argoproj/argo-rollouts/pull/3643 when trying to fix the analyses of `v1`.

**To Reproduce**

Use a Datadog analysis with `apiVersion: v2` and don't specify ` aggregator`:

```yaml
    - name: latency-slo-v2
      interval: 60s
      count: 3
      successCondition: result >= 0.95
      failureLimit: 2
      provider:
        datadog:
          apiVersion: v2
          interval: 60s
          queries:
            a: |
              sum:http_request_duration_seconds.bucket{
                upper_bound:0.5,
                short_image:sre-podinfo,
                kube_namespace:canary-analysis,
                env:{{args.environment}},
                kube_cluster_name:{{args.clusterName}},
                kube_service:{{args.service-name}}
              }.as_rate()
            b: |
              sum:http_request_duration_seconds.count{
                short_image:sre-podinfo,
                kube_namespace:canary-analysis,
                env:{{args.environment}},
                kube_cluster_name:{{args.clusterName}},
                kube_service:{{args.service-name}}
              }.as_rate()
          formula: "a / b"
```

**Expected behavior**

**Screenshots**

**Version**

v1.7.1

**Logs**

```
time="2024-07-05T11:25:18Z" level=warning msg="Measurement had error: received non 2xx response code: 400 {\"errors\":[\"Value of parameter 'aggregator' should be any of these [None, 'avg', 'min', 'max', 'sum', 'last', 'mean', 'area', 'l2norm', 'percentile', 'stddev']\"]}" analysisrun=podinfo-78544d7dd7-285-4 metric=latency-slo-v2 namespace=canary-analysis
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

rvega-clearroute:
[...]

PR #3730, opened 2024-07-18
fix(analysis): explicitly set datadog aggregator to last only on v2

alexef:
the field is required by the v2 API and we currently leave it empty string.

@zachaller you removed it in https://github.com/argoproj/argo-rollouts/pull/3643 and that is fine. this PR will not add it back to the schema and not break the v1 API. 

without this change and without aggregator set on a v2 metric, one gets:

```
received non 2xx response code: 400 {"errors":["Value of parameter
            'aggregator' should be any of these [None, 'avg', 'min', 'max',
            'sum', 'last', 'mean', 'area', 'l2norm', 'percentile', 'stddev']"]}
```

If we set it to None, according to datadog docs it will default to `avg`. According to rollouts docs, we should default to `last` (which makes sense for analysis)

Screenshot 2024-07-18 at 11 46 40

https://docs.datadoghq.com/api/latest/metrics/#query-scalar-data-across-multiple-products

Checklist:

### DIFF
--- a/metricproviders/datadog/datadog.go
+++ b/metricproviders/datadog/datadog.go
@@ -224,6 +224,10 @@ func (p *Provider) createRequestV2(queries map[string]string, formula string, no
 			"formula": formula,
 		}}
 	}
+	// we cannot leave aggregator empty as it will be passed as such to datadog API and fail
+	if aggregator == "" {
+		aggregator = "last"
+	}
 
 	attribs := datadogQueryAttributes{
 		// Datadog requires milliseconds for v2 api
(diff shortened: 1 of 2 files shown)


zachaller (approved):

mickeyrouash:
@alexef  I also experience it , is there a way to workaround it and still using Datadog V2 ?

alexef:
@mickeyrouash yes, the workaround is adding `aggregator: last` to your analysis template (this will now do it implicitely)

aguiraorodriguez:
Is this PR addressing https://github.com/argoproj/argo-rollouts/issues/3707 ?

Issue #3741, opened 2024-07-23
Add timeout attribute for NewRelic metrics provider


orlando-valdez-ck:
# Summary

Add an optional `timeout` attribute to the NewRelic provider spec in the `AnalysisTemplate` so we can send that to the NerdGraph Graphql API. This can look something like this,

```yaml
apiVersion: argoproj.io/v1alpha1
kind: AnalysisTemplate
metadata:
  name: cpu-usage
  namespace: default
spec:
  metrics:
    - name: cpu-usage
      interval: 2m
      failureCondition: "result['percentile[95]'] >= 70"
      failureLimit: 4
      provider:
        newRelic:
          timeout: 10 # this timeout will be sent to and applied by the NerdGraph API to the query
          profile: newrelic
          query: "FROM K8sContainerSample SELECT percentile(`cpuCoresUtilization`, 95)['95'] SINCE 1 WEEK AGO"
```
When not configured, the default of `5` seconds can be sent to the API.

# Use Cases

This can be used when using queries that take longer than the default of `5` seconds that the NerdGraph API uses when no timeout value is sent. This will avoid an AnalysisRun to error out with `NRDB query duration exceeded the set timeout` errors returned by the GraphQL API.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #3742, opened 2024-07-24
feat(analysis): Adds timeout property to NewRelic metrics provider. Resolves: #3741

orlando-valdez-ck:
Resolves: #3741 

Adds a timeout property to the New Relic metrics provider spec to be sent to the NerdGraph API for queries that take longer than the default 5 seconds. I bumped up the newrelic-go-client library while working on this.

Checklist:


metricproviders/newrelic/newrelic.go
```diff
@@ -41,13 +66,30 @@ type NewRelicClient struct {
 }
 
 // Query executes a NRQL query against the given New Relic account
-func (n *NewRelicClient) Query(query string) ([]nrdb.NRDBResult, error) {
-	results, err := n.Nrdb.Query(n.AccountID, nrdb.NRQL(query))
```
orlando-valdez-ck, line 45: There's an equivalent [function](https://github.com/newrelic/newrelic-client-go/blob/36d7acf42b8bd408e77dd1898ab148ee43922a15/pkg/nrdb/nrdb_api.go#L107) that supports passing timeout in the `nrdb` package but it [pulls way more data](https://github.com/newrelic/newrelic-client-go/blob/36d7acf42b8bd408e77dd1898ab148ee43922a15/pkg/nrdb/nrdb_api.go#L213) than the current function which by itself can impact latency.
That's why I switched to the NerdGraph package to send the timeout in the query without incurring in additional data over the wire. I actually slimmed down the query compared to the [used](https://github.com/newrelic/newrelic-client-go/blob/36d7acf42b8bd408e77dd1898ab148ee43922a15/pkg/nrdb/nrdb_api.go#L40) in the `Query` function to just what we need for this metric provider so the GQL API has to do less work to respond (and even less data transmitted back)

zachaller:
Can I have you update with master and fix conflicts, that should re-run the tests now that I have them fixed.

orlando-valdez-ck:
Noticed something during my last round of local testing. Changed to draft while I figure this out.

orlando-valdez-ck:
Ready for review now

zachaller (approved):

PR #3748, opened 2024-07-25
chore: Update k8s version to 1.30

nikoshet:
Checklist:

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -86,6 +86,8 @@ jobs:
           - version: 1.28
             latest: false
           - version: 1.29
+            latest: false
+          - version: '1.30'
             latest: true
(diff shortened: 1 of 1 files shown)


zachaller:
We should remove on on the tail end like 1.26.

mansing2:
Any plan in near future to support K8S v1.30. We want to upgrade the k8s and we are blocked.

zachaller:
@mansing2 I would not let what clients Rollouts uses cause you to not upgrade your k8s cluster. As long as you do not use any of the new pod spec fields if there are any it will be ok we use very little k8s feature set. The only big one being the embedded pod spec into the rollout resource.

blkperl:
This was resolved in https://github.com/argoproj/argo-rollouts/pull/3858

Issue #3757, opened 2024-07-30
feat: enable Profiling support to detect cpu/mem bottlenecks


kostis-codefresh:
# Summary

Idea came from this discussion. https://github.com/argoproj/argo-rollouts/discussions/3752

We need a way to allow people to get profiling data from Argo Rollouts (ideally in a production env) so that we can detect and fix bottlenecks.

# Use Cases

Large installations of Argo Rollouts with many resources on a single cluster would benefit from this
It will also allow us to analyze the performance of new/suggested features in advance.

Ideally this should be a command line argument/flag which is off by default and enabled on demand (to avoid any performance hit when profiling is not used).

Argo Workflows already has this https://argo-workflows.readthedocs.io/en/latest/running-locally/#troubleshooting

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

johnmwood:
The profiling PR has been merged so this issue can be closed.

PR #3769, opened 2024-08-05
feat(controller): enable pprof profiling support

johnmwood:
This PR enables pprof profiling on the Argo Rollouts controller via the `--enable-pprof` flag.  

Issue: [#3757](https://github.com/argoproj/argo-rollouts/issues/3757)

Checklist:

### DIFF
--- a/cmd/rollouts-controller/main.go
+++ b/cmd/rollouts-controller/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"strings"
 	"time"
@@ -79,6 +80,8 @@ func newCommand() *cobra.Command {
 		printVersion                   bool
 		selfServiceNotificationEnabled bool
 		controllersEnabled             []string
+		pprofEnabled                   bool
+		pprofPort                      int
 	)
 	electOpts := controller.NewLeaderElectionOptions()
 	var command = cobra.Command{
@@ -204,6 +207,11 @@ func newCommand() *cobra.Command {
 			ingressWrapper, err := ingressutil.NewIngressWrapper(mode, kubeClient, kubeInformerFactory)
 			checkError(err)
 
+			if pprofEnabled {
+				mux := controller.NewPProfServer()
+				go func() { log.Println(http.ListenAndServe(fmt.Sprintf("localhost:%d", pprofPort), mux)) }()
+			}
+
 			var cm *controller.Manager
 
 			enabledControllers, err := getEnabledControllers(controllersEnabled)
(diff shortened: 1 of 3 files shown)


cmd/rollouts-controller/main.go
```diff
@@ -283,6 +291,7 @@ func newCommand() *cobra.Command {
 	command.Flags().IntVar(&klogLevel, "kloglevel", 0, "Set the klog logging level")
 	command.Flags().IntVar(&metricsPort, "metricsport", controller.DefaultMetricsPort, "Set the port the metrics endpoint should be exposed over")
 	command.Flags().IntVar(&healthzPort, "healthzPort", controller.DefaultHealthzPort, "Set the port the healthz endpoint should be exposed over")
+	command.Flags().IntVar(&pprofPort, "pprof-port", controller.DefaultPProfPort, "Set the port the pprof endpoint should be exposed over")
```
johnmwood, line 294: In my mind as a user, I would like to be able to overwrite the port in case I'm using the default debug port `6060` elsewhere. I will defer on a maintainer's judgement for what pattern makes the most sense in the project given this is creating two flags for one feature.

controller/controller.go
```diff
@@ -69,6 +69,9 @@ const (
 	// DefaultMetricsPort is the default port to expose the metrics endpoint
 	DefaultMetricsPort = 8090
 
+	// DefaultPProfPort is the default port to expose the pprof endpoint
+	DefaultPProfPort = 6060
```
johnmwood, line 73: I recognize that `DefaultMetricsPort` and `DefaultHealthzPort` are being used in a more complex setup for those servers in their respective files. I wanted to mimic that pattern as close as possible without clogging up `main.go`. Open to suggestions here.

controller/profiling.go
```diff
@@ -0,0 +1,24 @@
+package controller
```
johnmwood, line 1: I will caveat that I'm not a pprof expert. This server config worked in my testing on a local image but I'm certainly open to feedback.

I also didn't include any tests as the functionality here is much simpler than the metrics server for example. However, maybe I could add something similar to the healthz server setup to test if the server returns responses?

cmd/rollouts-controller/main.go
zachaller: I am kind of a fan of passing in a single flag of such that we can do `0.0.0.0:6060` or `localhost:6062` . The main reason for this is not all pods can you get into and it might be nice to be a be able to remotely pull pprof files if configured.

controller/profiling.go
zachaller: I am ok without testing for this.

cmd/rollouts-controller/main.go
johnmwood: In terms of wording, a boolean is much easier to convey `--enable-pprof=true` vs. `--pprof-port` being the boolean toggle _and_ the config for the address. Do you have any ideas on how we can convey that? 

I see two options with tradeoffs:

1. Having `--pprof-port` resolve to an address doesn't make it clear that this is disabled by default (maybe the help text is what we need to clarify here).
2. Setting `--enable-pprof` equal to an address instead of a boolean feels strange to me. Maybe something like `--enable-pprof-server-address` is a (verbose) happy medium?

zachaller: Yea, thats hard. We could stick with the status quo with the code base and be a little less secure which uses [0.0.0.0](https://github.com/argoproj/argo-rollouts/blob/master/controller/controller.go#L100) by default and then just have the enable on port flag.

I don't like the above due to the security concerns around the data available on the pprof endpoints. So with that said I think let's go verbos with `--enable-pprof-server-address` or `--enable-pprof-address`  I think I am impartial on weather `server` is there or not, thoughts?

johnmwood: I pushed up a commit to consolidate the two config vars. My remaining question is what you want to do about the default value? We obviously don't want to enable profiling for every user so I don't know know a way around `""` as the default unless you have suggestions?

zachaller: What you did for default is perfect, this LGTM thanks for adding this.

zachaller (approved):

PR #3787, opened 2024-08-13
feat: support multi account Datadog metrics provider

ariadnarouco:
### Summary
This pull adds ability to use namespaced credentials for Datadog metrics provider.  

### Background
Our use case involves an Argo Rollout that operates across multiple namespaces, each managed by tenants with distinct Datadog accounts. As a result, multiple credentials are needed to integrate seamlessly with Datadog.

### Implementation
This setup will allow users to use different credentials than the one setup as default, which is a secret named "datadog" deployed in the same namespaces as the rollouts controller [[refer to doc](https://argo-rollouts.readthedocs.io/en/stable/analysis/datadog/)]. 

This way the users can manage their datadog credentials with the name they want and in their own namespace.  

The process suggested within this PR for retrieving Datadog credentials is as follows:
    1. **If a `secretRef` is defined in the `AnalysisTemplate`:** Argo Rollouts will search for the secret with the specified name in the namespace where the template resides.
    2. **If the secret is not found in the specified namespace:** Argo Rollouts will then check the environment variables.
    3. **If the credentials are not found in environment variables:** Argo Rollouts will look for a secret named "Datadog" in the namespace where Argo Rollouts itself is deployed.

_Notes_:
Based on the code, to have more than one set of credentials is not possible. I also created a [question](https://github.com/argoproj/argo-rollouts/discussions/3726) for this which remains unanswered

Checklist:


docs/analysis/datadog.md
```diff
@@ -41,6 +41,48 @@ stringData:
...
+        interval: 5m
+        successCondition: result <= 0.01
+        failureLimit: 3
+        provider:
+          datadog:
+            apiVersion: v2
+            interval: 5m
+            secretRef:
+              name: "mysecret"
+              namespaced: true
```
zachaller, line 79: I knid of think we just drop this field `namespaced`, i think the presence of a secretRef just means look up secret in the namespace of the AnalysisTemplate. At a later date we can then add a `namespace`  field to secretRef if we want to support cross namespace lookups. Thoughts?

tjamet: Hi @zachaller 
Thanks for the review and the input.

When implementing this feature we discussed the different use-cases and we thought about the following ones:

1. As an argo rollout owner, you also own the integration with a single DataDog account.
2. As an argo rollout owner, you also own the integration with several DataDog accounts
3. As an argo rollout owner, you only own argo rollout and your users own the DataDog ones.

The first case is the currently supported case.

The second case corresponds to `namespaced: false`, and let the "rollouts administrator" to manage different DataDog integrations by providing well-known secret names to the namespace users.

The third one corresponds to `namespaced: true`, where namespace users were completely autonomous to manage the DataDog account integration.

We did consciously exclude the option to have a configurable `namespace` string to maximise security.
Indeed, we want the owners of, say, `namespace-a` `AnalysisTemplate`s and DataDog integrations to have confidence that they will control who can access their metrics.

Having a configurable free text `namespace` would allow users of any other namespace to indirectly access DataDog metrics, which would then break the namespace principles.

WDYT?

zachaller (approved):

Issue #1070, opened 2021-04-07
Infinite Ingress recreation when using rancher managed cluster


stfnbstl:
## Summary 

What happened/what you expected to happen?

I created a canary rollout using nginx for traffic routing. Argo is then creating a second ingress for the canary traffic. Rancher is by default setting some annotations to ingress ressources after creation. As soon as rancher sets the annotations to the canary ingress, argo is removing them because they don't match the desired state that argo knows about.
This results in an infinite loop between rancher setting annotations and argo removing them.

This is the annotation rancher sets with sensitive information removed:
`annotations:
    field.cattle.io/publicEndpoints: >-
      [{"addresses":["xxx"],"port":80,"protocol":"HTTP","serviceName":"rollout-demo-dev:demo-canary","ingressName":"rollout-demo-dev:demo-demo-ing-canary","hostname":"xxx","path":"/demo-dev(/|$)(.*)","allNodes":true}]`

## Diagnostics

What version of Argo Rollouts are you running?
0.10.2
```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

`time="2021-04-07T13:02:34Z" level=info msg="Event(v1.ObjectReference{Kind:\"Rollout\", Namespace:\"rollout-demo-dev\", Name:\"demo\", UID:\"99b773f9-cade-4762-92a9-302c96fcceaf\", APIVersion:\"argoproj.io/v1alpha1\", ResourceVersion:\"54013056\", FieldPath:\"\"}): type: 'Normal' reason: 'PatchingCanaryIngress' Updating Ingress `demo-demo-ing-canary` to desiredWeight '0'"
time="2021-04-07T13:02:34Z" level=info msg="Reconciling StableRS" namespace=rollout-demo-dev rollout=demo
time="2021-04-07T13:02:34Z" level=info msg="No StableRS exists to reconcile or matches newRS" namespace=rollout-demo-dev rollout=demo
time="2021-04-07T13:02:34Z" level=info msg="Reconciling new ReplicaSet 'demo-677464578c'" namespace=rollout-demo-dev rollout=demo
time="2021-04-07T13:02:34Z" level=info msg="Reconciling old replica sets" namespace=rollout-demo-dev rollout=demo
time="2021-04-07T13:02:34Z" level=info msg="No Steps remain in the canary steps" namespace=rollout-demo-dev rollout=demo
[...]

PR #3806, opened 2024-08-26
fix: retain non-nginx canary annotations. Fixes: #1070

jahvon:
The current patch applied for nginx canaries does not play nicely with other controller that may be modifying Ingress annotations. In our case, the Rancher `field.cattle.io/publicEndpoints` is repeatedly added/removed, resulting in a significant amount of updates to Ingress resources in the cluster. 

With this fix, before building and applying the canary ingress patch, all annotations from the existing canary ingress will be copied over to the desired Ingress. 

fixes: https://github.com/argoproj/argo-rollouts/issues/1070

Checklist:

### DIFF
--- a/rollout/trafficrouting/nginx/nginx.go
+++ b/rollout/trafficrouting/nginx/nginx.go
@@ -325,6 +325,7 @@ func (r *Reconciler) SetWeightPerIngress(desiredWeight int32, ingresses []string
 		}
 
 		// Make patches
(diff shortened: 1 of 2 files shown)


jahvon:
I've also tested this in a RKE cluster. Here is a screenshot of the `PATCH|POST|PUT|APPLY` API Server request for `Ingresses` (before and after)

Screenshot 2024-08-26 at 4 19 40 PM

zachaller (approved):

Issue #1841, opened 2022-02-03
abortScaleDownDelaySeconds is not working as expected


mksha:
## Summary 

What happened/what you expected to happen?
when we set abortScaleDownDelaySeconds:0 then it should not delete the canary pod but its deleting it immediately.
## Diagnostics

What version of Argo Rollouts are you running?
1.1.1
```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME>
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

jessesuen:
@mksha this option is ignored if you are doing a basic canary (not using a traffic router). Can you confirm you are using a traffic router? Which one are you using?

mksha:
@jessesuen I am using the canary with traffic routing.
I am using istio with destination rule

huikang:
I can reproduce the issue and will work on a fix. Thanks, @mksha

huikang:
@mksha , after second thought, I think this maybe an expected behavior if you don't use `setCanaryScale` in a step, e.g.,

```
      - setCanaryScale:
          replicas: 3
```

Otherwise, the code will set the new replica to nil:

https://github.com/argoproj/argo-rollouts/blob/a78640a3f2fed23c9855b2ac1d45e3035248e3e9/utils/replicaset/canary.go#L481-L503

jessesuen:
@mksha could you please provide more details including rollout spec, and kubernetes rollout events when this happens? I think we don't have enough information to go off of currently.

mksha:
@jessesuen I already provided the rollout spec above and other details.
I would request you to reproduce using the above instructions and you will see the issue straightforward as @huikang  saw.

mksha:
I am using the same spec shared at https://github.com/argoproj/argo-rollouts/issues/1838.

mksha:
@jessesuen

mksha:
@huikang 
1. You are correct, I tried that and it worked, so one that we need to do is, make sure document is updated so we know how to use it with what.
2. Still there is a weird behaviour, if we use the 
steps:
        - setCanaryScale:
            weight: 100
        - some-analysis-template
        - setCanaryScale:
            matchTrafficWeight: true
[...]

PR #3835, opened 2024-09-17
docs: Improve Documentation for Scaledown on Aborted Rollout

bishalthapa-t:
Improves understanding of how scaling down behaves depending on different configurations:

 • Canary with traffic routing + setCanaryScale does not scales down immediately  when `abortScaleDownDelaySeconds = 0`

 • Canary with traffic routing + setCanaryScale scales down after some  delay of N second `abortScaleDownDelaySeconds = N`
 
Addresses the following issue: https://github.com/argoproj/argo-rollouts/issues/1841

Checklist:

### DIFF
--- a/docs/features/scaledown-aborted-rs.md
+++ b/docs/features/scaledown-aborted-rs.md
@@ -6,13 +6,15 @@ The following table summarizes the behavior under combinations of rollout strate
 `abortScaleDownDelaySeconds = nil` is the default, which means in v1.1 across all rollout strategies, the new replicaset
 is scaled down in 30 seconds on abort by default.
 
-|                                    strategy |         v1.0 behavior         | abortScaleDownDelaySeconds |         v1.1 behavior         |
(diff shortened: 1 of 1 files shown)


kostis-codefresh (changes requested):
I don't understand how this PR makes documentation more clear. In fact it confuses me even more. The original issue is about the value of `abortScaleDownDelaySeconds` when it it **zero**

But the PR diff actually also changes the docs where abortScaleDownDelaySeconds is **N** https://github.com/argoproj/argo-rollouts/pull/3835/files#diff-182a7dd59b6ae0b915e2693efb96ae762eabfe54947eb3d6c43c127b9d82c130R17

Or was this a copy/paste mistake for the table formatting?

Personally, I think that this whole page needs to be rewritten as it talks about Argo Rollouts 1.0 and 1.1, and we are at 1.9. 
I don't think any user cares today about what 1.0 used to do.

Issue #3372, opened 2024-02-14
when dynamicStableScale is enabled, the 2nd canary deployment (when the first one didn't finish) shifts the traffic to the stable 100% without waiting for it to scale up 


MarwanTukhta:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**
When using canary strategy and adding `dynamicStableScale` if a newer release were pushed (v3) while an unstable release (v2) is still being deployed, argo rollouts controller will shift all the traffic to the stable release (v1) without waiting for it to scale up.

**To Reproduce**
Just deploy a simple rollout with traffic routing and  `dynamicStableScale` enabled, then change the image version to `1.25.2` and apply the changes, now the v1 replicaset will have 1 replica while v2 will have 3, change the image to `1.25.3` and the issue will happen, it will immediately shift all traffic to v1 without waiting for it pods to start, since we are using nginx as an example it will be very short, but on production, this caused an outage

```
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: test-argo-rollout
  labels:
    app: test-argo-rollout
spec:
  replicas: 4
  strategy:
    canary:
      canaryService: test-argo-rollout-canary
      stableService: test-argo-rollout
      dynamicStableScale: true
      abortScaleDownDelaySeconds: 300
      maxUnavailable: 0
      trafficRouting:
        nginx:
          stableIngress: test-argo-rollout-ingress
      steps:
        - setWeight: 90
        - pause: {}
  selector:
    matchLabels:
      app: test-argo-rollout
  template:
    metadata:
      labels:
        app: test-argo-rollout
    spec:
      containers:
        - name: nginx
          image: nginx:1.25.1
          imagePullPolicy: IfNotPresent
          ports:
            - containerPort: 80
              name: http
              protocol: TCP
          resources:
            requests:
              cpu: 100m
              memory: 100Mi

---

apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-argo-rollout-ingress
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  ingressClassName: nginx
  rules:
    - host:<HOST-HERE>
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: test-argo-rollout
                port:
                  number: 80

---

apiVersion: v1
kind: Service
metadata:
  name: test-argo-rollout
  labels:
[...]

PR #3862, opened 2024-09-30
fix: update during setWeight: 100 could cause 503 errors

jessesuen:
Resolves https://github.com/argoproj/argo-rollouts/issues/3372

Testing a possible fix for https://github.com/argoproj/argo-rollouts/issues/3372. 

E2E test would be needed

Checklist:


zachaller:
https://github.com/argoproj/argo-rollouts/pull/3878

Abhish2702:
> This seems to almost be the same fix #3878 should we consolidate to one? Also instead of 100 we should use `weightutil.MaxTrafficWeight(c.rollout)`

@zachaller thanks for the suggestion, actually I have raised #3878 along with unit test for the same fix. This PR was raised by @jessesuen for testing purpose only upon my request when I approached him offline with this suggestion. So, yes we can consolidate these PRs

PR #3878, opened 2024-10-08
fix: guardrail to not overload stable replicaset

Abhish2702:
This PR introduces a guardrail in the rollouts controller to help prevent unexpected downtime during rollouts. The guardrail adds an additional check to ensure that the number of replicas in the stable ReplicaSet is sufficient before any traffic switch occurs.

For example, if the desired weight for canary replicas is set to 60%, and there are 10 replicas in total, the code will verify that, before diverting 60% of the traffic to the canary replicas, at least 40% of the replicas (i.e., 4 in this case) are available in the stable ReplicaSet.

This check is crucial to prevent potential downtime. A common scenario where downtime can occur is when a rollout is already in progress, and a new deployment is triggered. In such cases, if the stable replicas are insufficient, it could lead to service disruption.

Resolves https://github.com/argoproj/argo-rollouts/issues/3372
Checklist:

### DIFF
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -27,6 +27,7 @@ import (
 	replicasetutil "github.com/argoproj/argo-rollouts/utils/replicaset"
 	rolloututil "github.com/argoproj/argo-rollouts/utils/rollout"
 	"github.com/argoproj/argo-rollouts/utils/weightutil"
+	appsv1 "k8s.io/api/apps/v1"
 )
 
 // NewTrafficRoutingReconciler identifies return the TrafficRouting Plugin that the rollout wants to modify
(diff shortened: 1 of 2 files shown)


Abhish2702:
> Very similar logic is already defined via
> 
> https://github.com/argoproj/argo-rollouts/blob/5f59344adfcae42cdb0ac941a876311a9ee49045/rollout/service.go#L292
> 
> I do not think it makes sense to add another check

@zachaller but this check is not preventing downtime if we trigger rollout in between the deployment.

KarimTarek:
Just wanna double down that this is super critical for us to ensure that if a rollout occurred in the middle of an already ongoing one, we must ensure that the traffic won't get shifted to the stable replica set before it being fully scaled back to the desired amount specially with dynamicStableScale (talking about late stages in the canary process where stable might be at 25% and suddenly gets all 100% traffic if you triggered a new one). Hope it makes 1.8 to unblock us going into prod.

heshamelsherif97:
@zachaller Is this planned for 1.8?

zachaller:
This would be a bug fix so we could put it into 1.8 but I need time to go through the PR and give it a good review.

zachaller (approved):

Issue #3617, opened 2024-06-06
intermediate replicaset scaledown during ALB traffic shaping causes 503 responses


schlags:
Checklist:

* [ ] I've included steps to reproduce the bug.
* [ ] I've included the version of argo rollouts.

**Describe the bug**

We are using the ArgoCD Image Updater to set image parameters for our application. 

We are using the ALB ingress class, with the `pingPong` strategy

We are running Rollouts v1.6.6

We have readiness-gate-injection enabled on the pods in our namespace where this takes place

We have observed, very frequently, that when a new image is available and set for the rollout, it causes issues if the rollout is still progressing on a previous image.

It typically goes like the following. 

Assume `image A` is the stable, `image B` is next, and `image C` is set while `image B` is still progressing.

1. image A is stable and directed to 100% traffic
2. image B is spun up
3. 25% of traffic is defined to go to image B
	- (Note: we have a pause step of 15 seconds here before running Analysis because we have observed it takes about this long for traffic to finally show up).
5. image C gets discovered by the image updater and set for the rollout
6. Log `scaling down intermediate RS` is written
	- There is a comment here `// It is safe to scale the intermediate RS down, since no traffic is directed to it.` but the following disproves that a bit
7. 25% of our responses from the ALB are showing 503 (all responses from the Target Group).
8. After some time (less than a minute), traffic returns to a normal state of no 503 errors.

I am wondering if we are missing something in our traffic shaping configuration to address this downtime on preemption in rollouts. 

**To Reproduce**

1. Set up a ping pong canary flow where traffic shapes to a certain percentage of traffic
2. Set an image 
3. Once it is on the stage where traffic is directed to the new pods in some percentage, set a newer image
4. Observe that traffic previously being served by the canary is now responding with 503
5. Observe that this is only the case for less than a minute before traffic error rate returns to normal.

**Expected behavior**

All ingress rules should be confirmed to be reset to the state at the beginning of the canary before a new replicaset has gets set for the service selector. (Because otherwise, the ingress would send that percentage of traffic to a service/target group with no assigned IP's - hence the 503 response).

Towards the end of the log output below, you can see that we observe the selector swapping on the service to the new replicaset before restoring the desiredWeight to 0. I believe this is a bug for this reason, but open to ideas on how I'm approaching this incorrectly if this is the intended behavior.

```
[...]

PR #3898, opened 2024-10-16
fix: remove condition where header routes can stay directed at empty service in preemption

schlags:
Attempts to address some concerns with behavior in:
- https://github.com/argoproj/argo-rollouts/issues/3617

It seems like we are forgetting to remove the managed routes when the following occurs:
1. Rollout is stable on V1
2. Rollout is updated with V2, and begins to progress
3. Managed routes are defined (e.g. to send traffic with specific header values to the canary service)
4. Rollout is updated with V3, and must start back at the beginning again where V1 is still considered stable (ditching any ongoing progress with V2)
5. Canary service selector is swapped to V3 replicaset
6. Traffic weights are set back to 0 

It does not seem like any managed routes are considered between steps 4-6. I believe this code that I've made an addition to is where this would need to occur, but would love some feedback.

NOTE: I still don't think this fixes another situation where the service selector is swapped to the new replicaset before the traffic can be reconciled back at the 0% weight, but this at least allows the managed routes to fall back to the stable V1 version, instead of the behavior in the graph below:

image
> `c8e` is `V1` `5ba` is `V2` and `aa9` is `V3`. Note how for a client meeting the requirements for the managed route, it never actually falls back to `V1` but simply 503's until the replicaset for `V3` can finally mark ready (it is never taken away from the canary service)

Checklist:

### DIFF
--- a/rollout/trafficrouting_test.go
+++ b/rollout/trafficrouting_test.go
@@ -1348,7 +1348,7 @@ func TestDontWeightToZeroWhenDynamicallyRollingBackToStable(t *testing.T) {
 
 // TestDontWeightOrHaveManagedRoutesDuringInterruptedUpdate builds off of TestCanaryDontScaleDownOldRsDuringInterruptedUpdate
 // in canary_test when we scale down an intermediate V2 ReplicaSet when applying a V3 spec in the middle of updating.
-// We want to make sure that traffic routing is cleared in both weight AND managed routes when the V2 rs is 
+// We want to make sure that traffic routing is cleared in both weight AND managed routes when the V2 rs is
 // nil or has 0 available replicas.
 func TestDontWeightOrHaveManagedRoutesDuringInterruptedUpdate(t *testing.T) {
 	f := newFixture(t)
@@ -1381,7 +1381,7 @@ func TestDontWeightOrHaveManagedRoutesDuringInterruptedUpdate(t *testing.T) {
 			Ingress: "test-ingress",
 		},
 		ManagedRoutes: []v1alpha1.MangedRoutes{
-			{ Name: "test-header" },
+			{Name: "test-header"},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


maschwenk:
We had some downtime related to this so would be sweet to get a look!

zachaller:
Can we get a unit test for this?

schlags:
> Can we get a unit test for this?
> @zachaller 

Added one just now, was able to verify that this test fails without the presence of that function call!

rollout/trafficrouting_test.go
```diff
@@ -1343,3 +1345,89 @@ func TestDontWeightToZeroWhenDynamicallyRollingBackToStable(t *testing.T) {
...
+	}
+
+	f.objects = append(f.objects, r3)
+	f.kubeobjects = append(f.kubeobjects, rs1, rs2, rs3, canarySvc, stableSvc, ingress)
+	f.replicaSetLister = append(f.replicaSetLister, rs1, rs2, rs3)
+	f.serviceLister = append(f.serviceLister, canarySvc, stableSvc)
+	f.ingressLister = append(f.ingressLister, ingressutil.NewLegacyIngress(ingress))
+
+	f.expectPatchRolloutAction(r3)
+	f.run(getKey(r3, t))
```
schlags, line 1416: This first reconciliation shows that it skips scaling down the old RS since it is still referenced.

rollout/trafficrouting_test.go
```diff
@@ -1343,3 +1345,89 @@ func TestDontWeightToZeroWhenDynamicallyRollingBackToStable(t *testing.T) {
...
+
+	r3.Status.Canary.Weights = &v1alpha1.TrafficWeights{
+		Canary: v1alpha1.WeightDestination{
+			PodTemplateHash: replicasetutil.GetPodTemplateHash(rs3),
+		},
+	}
+
+	f.expectUpdateReplicaSetAction(rs3)
+	f.run(getKey(r3, t))
```
schlags, line 1425: We swap that reference here, and reconcile again to observe the `scaling down intermediate RS` log - where the `else if c.newRS == nil || c.newRS.Status.AvailableReplicas == 0` condition is met while reconciling the traffic routing. 

Adding the removal of the managed traffic routes here allows us to ensure that no traffic will be directed to that intermediate RS.

zachaller (approved):

PR #3899, opened 2024-10-16
fix(controller): fix waiting for stable rs to be fully scaled before canary scale down

y-rabie:
Checklist:

As described in the code comment, when used on a large scale with a cluster autoscaler that can disrupt nodes and evict pods, the canary RS stays scaled-up for a while until the stable RS is fully scaled. This makes sense if the controller scaled down the stable RS during the rollout (using `dynamicStableScale`), but it doesn't make sense if it didn't.


zachaller:
Can we get a test for this? It will also probably help me understand the exact issue a bit better as well, it sounds like the controller will just take a really long time to complete the rollout due to scaling events etc?

y-rabie:
@zachaller 

Yep, exactly. As a principle, if we had _nothing_ to do with scaling down the stable rs, why are we waiting for its full availability? To me that self-evidently _shouldn't be_ the case. Especially that we are forcing a cluster autoscaler to consolidate on a current state (stable and canary pods), just so that when it finishes, we terminate the canary pods (effectively changing that state and causing a new process of consolidation to begin). A sort of a wasted effort. 

When we do have something to do with it however, I don't think there is a way around it. I thought of maybe instead of full availability, wait for stable RS PDB (if exists) non-violation? But I don't think it'd be a correct/expected behavior. So I left that part out of the PR, if you have any thoughts on it

I'm not sure what kind of test to write for this bit though.

zachaller:
@y-rabie Want to give this PR a good review, will get this into a 1.8 release still

nebojsa-prodana:
We have lots of large services with a high number of beefy pods affected by this (15 CPU, 30 gigs of ram, 150+ pods). I am happy to test this version manually if need be.

Is there anything that can be done to expedite getting this one reviewed @zachaller ?

zachaller:
Having a  unit or e2e test to lock the behavior in would be appreciated.

nebojsa-prodana:
Sample unit tests here: https://github.com/argoproj/argo-rollouts/pull/4190. I basically wanted to cover the canary strategy since the existing replicaset scale tests were covering B/G without DynamicStableScale.

Let me know if these are going in the right direction given that this is my first contribution before I continue to add more.

y-rabie:
@zachaller I closed this one in favor of #4239, appreciate you taking a look when you can.

jessesuen:
Removed cherry-pick label since this was closed

Issue #3708, opened 2024-07-05
There should be an error message when URL is invalid and the plugin isn't downloaded


Alegrowin:
When deploying the gatewayAPI plugin using the helm chart like this
``` yaml
  trafficRouterPlugins: 
    trafficRouterPlugins: |-
      - name: "argoproj-labs/gatewayAPI"
        location: "https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/releases/download/0.3.0/gateway-api-plugin-linux-amd64"
# This is the right url, there was a missing v in the version v.0.3.0
# location: "https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/releases/download/v0.3.0/gateway-api-plugin-linux-amd64"
```

The logs show that the plugin downloaded successfully, it also shows a successful reconciliation and shows some misleading errors about stdout and being unable to get the plugin client:
```
level=info msg="Argo Rollouts starting" version=v1.7.1+6a99ea9
# ...
level=info msg="Downloading plugin argoproj-labs/gatewayAPI from: https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/releases/download/0.3.0/gateway-api-plugin-linux-amd64"
level=info msg="Download complete, it took 216.038207ms"
# ...
2024-07-05T20:23:30.787Z [DEBUG] plugin: starting plugin: path=/home/argo-rollouts/plugin-bin/argoproj-labs/gatewayAPI args=[/home/argo-rollouts/plugin-bin/argoproj-labs/gatewayAPI]
level=error msg="roCtx.reconcile err failed to get traffic router plugin argoproj-labs/gatewayAPI: unable to start plugin system: unable to get plugin client (argoproj-labs/gatewayAPI): fork/exec /home/argo-rollouts/plugin-bin/argoproj-labs/gatewayAPI: exec format error" generation=6 namespace=test resourceVersion=4560026783 rollout=test
level=info msg="Reconciliation completed" generation=6 namespace=test resourceVersion=4560026783 rollout=test time_ms=4.15409
# ...
level=error msg="rollout syncHandler error: failed to get traffic router plugin argoproj-labs/gatewayAPI: unable to start plugin system: unable to get plugin client (argoproj-labs/gatewayAPI): fork/exec /home/argo-rollouts/plugin-bin/argoproj-labs/gatewayAPI: exec format error" namespace=test rollout=test
[...]

PR #3905, opened 2024-10-18
feat(metricprovider): credentials to download plugin

ariadnarouco:
In our company, we have a use case where we need to create a new plugin as a metric provider. For security reasons the plugin needs to be private and cannot be uploaded to a public storage. 
Therefore, we have the need to use credentials for downloading the plugin securely. 

We have followed k8s secretRef pattern to specify how the credentials are passed. 

Additionally, we had an issue with 404 status code. The downloader says that the plugin is downloaded correctly but it’s not true. 

Fixes: #3708

Checklist:

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -238,7 +238,7 @@ func NewAnalysisManager(
 		log.Fatalf("Failed to init config: %v", err)
 	}
 
-	err = plugin.DownloadPlugins(plugin.FileDownloaderImpl{})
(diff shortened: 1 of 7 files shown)


zachaller (approved):

cmalladi7:
Can we create a release with this change in please? This is useful for us to make our plugin private as well. Would also love to have a release made in the [argo-helm](https://github.com/argoproj/argo-helm/tree/main/charts) project for the new version.

Thank you!

Issue #3969, opened 2024-12-01
Add ConsecutiveSuccessLimit feature to Analysis


y-rabie:
# Summary

Adding a `ConsecutiveSuccessLimit` feature to analyses. An option where the analysis only succeeds if there was `n` successful measurements in a row.

This is somewhat similar somewhat to K8s container probes, with `successThreshold`.

This can be coupled and/or detached from `FailureLimit`. That is, adding a way to disable `FailureLimit` for an analysis.

* With `ConsecutiveSuccessLimit` disabled, and `FailureLimit` used

    This is currently the normal and only behvaiour.

* With `ConsecutiveSuccessLimit` set, and `FailureLimit` disabled
    
    The analysis will keep running, reading however many failed measurements, possibly mixed with successful measurements, until `n` successful measurements in a row. It then transitions to a `AnalysisPhaseSuccessful` state.

* With `ConsecutiveSuccessLimit` set, and `FailureLimit` used

    The analysis will keep running, reading failed measurements (limited by `FailureLimit`, which when violated, it transitions to `AnalysisPhaseFailed`), possibly mixed with successful measurements, until `n` successful measurements in a row. It then transitions to an `AnalysisPhaseSuccessful` state.

# Use Cases

This provides the *inverse* of what an analysis currently does. The main use case for this would be to allow users to wait on a condition (in a step analysis especially) to hold, before proceeding to the next step.

This can be also really useful for orchestration with external systems through metrics. Where the rollout can basically very easily become event-driven.

An example of things this allows is orchestration between different rollouts, where a rollout can wait on another (through `rollout.phase` metric) to complete successfully before proceeding to 100% scale.

I've opened a [PR](https://github.com/argoproj/argo-rollouts/pull/3970) with the changes and further explanations of all possible scenarios. The changes there should be backward-compatible as well.

---
**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #3970, opened 2024-12-01
feat(analysis): Add ConsecutiveSuccessLimit feature to Analysis

y-rabie:
Checklist:

---

This regards the [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/3969), see there for a brief description first.

First, regarding the fields and their possible values:

  * The value used for `FailureLimit` to disable it is `-1`. The default is left to be `0` as is currently the case. This preserves backward compatibility. The rationale for adding a value to disable `FailureLimit` is the convenience for a user who wants to only use `ConsecutiveSuccessLimit` (waiting on a condition to hold/event to happen), instead of forcing them to type an arbitrarily big value for the `FailureLimit`.
  
  * The default value for `ConsecutiveSuccessLimit` is `0`, which makes it effectively disabled by default. This is explicitly an opt-in feature to use.
  
  * Both of them can be used side-by-side (`ConsecutiveSuccessLimit` > 0 and `FailureLimit` >= 0). Notes regarding behavior follow here shortly after.
  
  * They cannot be both disabled. Validation throws an error for that.

---

Second, a lot of the following resembles the way metric measurements themselves are evaluated in [utils/evaluate/evaluate.go](https://github.com/argoproj/argo-rollouts/blob/5f59344adfcae42cdb0ac941a876311a9ee49045/utils/evaluate/evaluate.go#L18), with regards to `FailureCondition` and `SuccessCondition`. That is:

* When `FailureCondition` is satisfied, it fails (even if `SuccessCondition` is specified separately and is satisifed).
* When neither is satisfied, the measurement is inconclusive.

---

All possible scenarios should be obvious from reading the added unit tests, but the particular ones I think deserve mention are:

* On terminating/cancelling analyses ***prematurely***, that is, either any indefinite analysis, or limited analysis with `count` not yet reached

    They are always terminated with `AnalysisPhaseSuccessful`, UNLESS, it happens that `FailureLimit` is enabled and violated, then they terminate with `AnalysisPhaseFailed`.

    This is already the case, but my emphasis here is that `ConsecutiveSuccessLimit` changes *nothing* from that. Even if it was not yet reached, the analysis is terminated successfully.

* For limited analysis (with `count` specified)

    When count is reached, the note above regarding similaritly to metric measurement evaluation comes into play.
    If both `FailureLimit` and `ConsecutiveSuccessLimit` are enabled, and

    * `FailureLimit` is violated and `ConsecutiveSuccessLimit` is not satisfied     => `AnalysisPhaseFailed`
    * `FailureLimit` is violated and `ConsecutiveSuccessLimit` is satisfied         => `AnalysisPhaseFailed`
    * `FailureLimit` is not violated and `ConsecutiveSuccessLimit` is satisfied     => `AnalysisPhaseSuccessful`
    * `FailureLimit` is not violated and `ConsecutiveSuccessLimit` is not satisfied => `AnalysisPhaseInconclusive`

---

I've written unit tests for the added logic, and ran and tested the controller myself. The minor changes in the UI dashboard have not been tested (building and accessing it always cause the tab to load indefinitely for some reason, and I doubt that is related to my changes).


zachaller (approved):

zachaller:
@y-rabie Probably something on this section of the docs. https://argo-rollouts.readthedocs.io/en/stable/features/analysis/

y-rabie:
@zachaller Sure, will do and open a PR with them

Issue #4003, opened 2024-12-13
Prometheus Metrics Provider Should Evaluate Prometheus Query Before Sending to Prometheus


liesenfeldn:
# Summary

Currently the prometheus metrics provider simply hands of the `Query` to prometheus, which will fail if for example you are using expression language to handle multiple cases of queries. It would be great if the prometheus provider could first evaluate the `Query` field, parse out which part of it is actually the query and then only pass that along to prometheus. 

# Use Cases

We currently have the use case that a metric needs to be swapped out for our template users. The problem however is 2-fold. On the one hand there will be a timeframe where our template users will have either the old metric or the new one, so they will be split for some time, which yes we could solve by just creating a new template, however that means pushing multiple PRs to our users repositories to migrate them, which can cause disruption. On the other hand we also would be able to reduce the amount of template that we currently have by allowing for this behavior. 

# Example
In the current state of argo rollouts, a query such as the following fails, because prometheus doesn't know what to do with `"some_arg" == "not_some_arg"`
```
apiVersion: argoproj.io/v1alpha1
kind: ClusterAnalysisTemplate
metadata:
  name: some-template
spec:
  metrics:
  - name: some-metric
    provider:
      prometheus:
        address: prometheus-provider-url
        query: '"some_arg" == "not_some_arg" ? "query1" : ( "some_arg" == "some_arg" ? 'sum(rate(some_metric{filter1="filter1_value",filter2="filter2_value",filter3=~"filter3_value",filter4=~"filter4_value",filter5!~"filter5_value"}[5m])) by(some_value))' : "query2")'
...
```
However if the prometheus metrics provider were to evaluate `Query` first, then in this case it would pass this query to prometheus, which would be a valid prometheus query:
```
`sum(rate(some_metric{filter1="filter1_value",filter2="filter2_value",filter3=~"filter3_value",filter4=~"filter4_value",filter5!~"filter5_value"}[5m])) by(some_value))`
```

# Expected Changes
- Make `Query` evaluate the expression first, then run the query (by adding a EvalQuery function)
- Adapt the `ResolvedPrometheusQuery` field to only show the actually run query
- Add tests

# Alternative Designs of this feature
1. Adapt the prometheus metrics provider to handle expression language in `Query`
2. Add a new field to the struct, that is a boolean that tells the prometheus metrics provider to evaluate the expression first, then run the query
[...]

PR #4004, opened 2024-12-13
feat: Prometheus Metrics Provider Should Evaluate Prometheus Query Before Sending to Prometheus

liesenfeldn:
This relates to [this issue](https://github.com/argoproj/argo-rollouts/issues/4003). 

I also set up a local kind cluster and tested this. I checked that this works with the templates we are attempting to create, but also i tested some of our old templates to make sure it is backwards compatible.

I mentioned some alternative implementations of this feature in my issue. Specifically i wanted to call out that we could also add a new field called `QueryExpression` or something similar, that way the `Query` behavior stays exactly as is, and this features get added in a new field in the struct.

Checklist:


liesenfeldn:
I've checked off the `My builds are green. Try syncing with master if they are not.`  part, because it says all checks have passed, but i'm not sure that was correct, since above it does link to [this](https://github.com/argoproj/argo-rollouts/runs/34464011675) which indicated e2e test failures. Can someone help me understand this? Thanks!

kostis-codefresh (changes requested):
This seems like a simple and harmless PR but it is not. It completely changes the way we send queries to Prometheus

In essence, it introduces 2 levels of evaluation, one for Argo Rollouts and one in Prometheus. And if this PR is merged, it is enabled by default with no way to opt out.

"I checked that this works with the templates we are attempting to create, but also i tested some of our old templates to make sure it is backwards compatible" is a very small sample size.  I would expect some rigorous testing first to prove that it is not opening a can of worms.

Maybe there are specific templates that either have parsing/evaluation problems, or they send to Prometheus a malformed query, or we might have crashes/errors from the 2-level evaluation process.

Since comprehensive testing might be difficult, I would prefer if the code of  PR is disabled by default (same as today) and users need to explicitly enable this behavior (and accept the risks). It is actually mentioned as an "alternative design" in https://github.com/argoproj/argo-rollouts/issues/4003

PR #4013, opened 2024-12-18
fix: Use correct type union in CRD spec for Quantity fields

heckler1:
This fix sets the expected types based on the current K8S upstream OpenAPI v3 types: https://github.com/kubernetes/kubernetes/blob/release-1.32/api/openapi-spec/v3/api__v1_openapi.json#L8637


heckler1:
Hmm, this doesn't seem to pass e2e, I'll iterate locally, sorry for the noise!

PR #4022, opened 2024-12-22
fix(experiments): move recorder event to after experiment reconciliation, fixes #4021

### LINKED ISSUE #4021: Experiment Notification using old rollout/pre-trigger state - related to #3535?
Checklist:

**Describe the bug**

I've been trying to use the `on-experiment-` triggers to send slack notifications due to an issue that seems similar to #3535  - the  experiment object in the trigger has the pre reconciliation state of the experiment, which means I cannot use it to send the error message, or analysis results.

**To Reproduce**

Create an experiment, and use one of the `on-experiment-xxx` triggers. In the trigger template, trying to use something like `.rollout.status.phase` (also not clear why there is only a `.rollout` object even though when examined you get an Experiment object).
For example, when an experiment fails, we get status `Pending`.

**Expected behavior**

I expected to be able to use the experiment object to send the slack notification to users, such as the failure message, failing analysis.
Ideally I'd also have wanted that the context would include the queries and measurements from the analysis runs.

**Version**

Tested in 1.7.2 in production cluster, but reproduced in 1.8.0-rc1 and master branch locally.

Hellspam:
fixes: https://github.com/argoproj/argo-rollouts/issues/4021

Checklist:


Hellspam:
I have not seen any tests for notifications so I did not add any tests.
If needed, I can add testing with some pointers on how to test this.

Locally I've checked that this works and sends the up-to-date Experiment object and not the one with the previous status.

zachaller (approved):

PR #4035, opened 2025-01-09
fix(trafficrouting): Fix rollback behavior for canary with trafficrouting and .DynamicStableScale=true

ArenSH:
Currently, in case of rollback controller always scales stable up to 100%, while dynamically scaling canary down. If rollback or abort occurs on later steps of rollout, it can cause various issues. For example, our deployment has several hundreds of pods, and rollback can cause surge hundreds of new pods.

This fix ensures that when a rollout is aborted and `DynamicStableScale` is enabled, the StableRS dynamically scales up to 100% as NewRS scales down based on steps in reverse order.
<br/><br/>
By spec, `SetCanaryScale` used to diverge from scaling canary according to traffic weights https://argo-rollouts.readthedocs.io/en/stable/features/canary/#dynamic-canary-scale-with-traffic-routing. Therefore, I expect it to be used only as an intermediate step in rollout that can be ignored.
Attempting to guess and handle all the combinations of setCanaryScale with matchWeights/weights/replicas is complex and I don't have a robust solution at hand (especially without actual use-cases).
<br/><br/>
Since I don't have comprehensive understanding of the argo-rollouts architecture, this fix might be somewhat clunky or suboptimal. Please feel free to suggest a better implementation.

Checklist:

* [X] I've signed my commits with [DCO](https://github.com/argoproj/argoproj/blob/main/community/CONTRIBUTING.md#legal)

### DIFF
--- a/rollout/canary.go
+++ b/rollout/canary.go
@@ -113,7 +113,7 @@ func (c *rolloutContext) reconcileCanaryStableReplicaSet() (bool, error) {
 		// causing us to flap and scale up the stable 100 temporarily (before scaling down to 0 later).
 		// Therefore, we send c.rollout.Status.Canary.Weights so that the stable scaling happens in
 		// a *susbsequent*, follow-up reconciliation, lagging behind the setWeight and service switch.
-		_, desiredStableRSReplicaCount = replicasetutil.CalculateReplicaCountsForTrafficRoutedCanary(c.rollout, c.rollout.Status.Canary.Weights)
+		_, desiredStableRSReplicaCount = replicasetutil.CalculateReplicaCountsForTrafficRoutedCanary(c.rollout, c.newRS, c.stableRS, c.rollout.Status.Canary.Weights)
 	}
 	scaled, _, err := c.scaleReplicaSetAndRecordEvent(c.stableRS, desiredStableRSReplicaCount)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


ssantichaivekin-plaid:
Hi team, I'm planning to upgrading to `v1.9.0` once it's released to pick up fixes related to ephemeral metadata syncs.

However, when I looked at this change, I have some concerns that I want to flag about rollback speed when we have many `setWeight` steps. With this new behavior, stable pod startup becomes serialized since each batch waits on available replicas before proceeding.

For example, with steps (1→2→5→10→20→40→80→100), aborting at 100% step, and ~2 minute pod readiness:

- **Old behavior**: requests all 80 stable replicas immediately → pods start in parallel → rollback takes ~2 minutes.

- **New behavior**: scales stable in 8 batches → rollback takes ~16 minutes.

This slow of a rollback seems undesirable, e.g. it significantly increases mean-time-to-repair during incidents. This also goes against argo-rollouts's own documentation. For example, the documentation on [Rollback Window](https://argo-rollouts.readthedocs.io/en/stable/features/rollback/) feature says:

> It is often undesirable to re-run analysis _and steps_ for a rollout, when the desired behavior is to rollback as soon as possible. _(emphasis added)_

This implies that the correct behavior is for rollback to be immediate (or at least as fast as possible).

I may be misreading the code, so I’d appreciate confirmation from the team on whether this is the intended behavior. It seems like this could be a breaking change for users who rely on faster aborts, especially for users with large number of rollout-steps. Would love to hear your thoughts.

ArenSH:
@ssantichaivekin-plaid Hi,

This rollback behavior is applied if `dynamicStableScale` is set to true, to dynamically scale stable up and down and avoid surge of new pods in both cases.

So yes, if you relied on dynamically scaling stable down with new release, but to immediately scale stable up in case of rollback - that behavior will not work.

PR #4062, opened 2025-01-18
add new proposal for external resource plugin

aburan28:
Checklist:

### DIFF
--- a/docs/proposals/external-resource-plugin.md
+++ b/docs/proposals/external-resource-plugin.md
@@ -1,0 +1,103 @@
+---
+title: Statefulset support
+authors:
+  - '@aburan28'
+sponsors:
+  - '@zaller'
+creation-date: 2025-01-18
+---
+
+# Summary
+In order to support other workloads in Rollouts besides Deployments a Rollout resource plugin would be ideal. 
+
+
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Rollout
+metadata:
+  name: statefulset
+  labels:
+    blah: blah
+  annotations:
+    blah: balh
+spec:
+  resourcePlugin:
+    name: statefulset-plugin
+
+  template:
+    metadata:
+    spec:
+      
+
+
+```
+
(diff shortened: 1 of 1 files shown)


docs/proposals/external-resource-plugin.md
```diff
@@ -0,0 +1,274 @@
...
+  conditions:
+    - 
+
+```
+
+#### Examples
+Below are several high-level overviews of how the `RolloutsPlugin` would handle the rollout of each the following workloads. 
+
+##### StatefulSet 
+
```
patrickant001, line 79: Hi, excited to see this proposal!
Would you mind explain how the plugin handles the rollout of STS in the example?
Want to see if it solves my problem

patrickant001:
```
We love the analysisruns, and want Argo Rollout to support StatefulSet as the scaling unit, i.e. with rollout.spec.replicas=10 , I want 10 StatefulSets created. The statefulsets can have random suffix in their name, such as rollout-124567890-12345  (name looks like a pod).  (And of course, each statefulset has its set of pods (like rollout-124567890-12345-0 , rollout-124567890-12345-1 , rollout-124567890-12345-2 etc). Rolling should work similarly with pods.
We currently have a workaround to use control_pod as a mini-orchestrator: Rollout -> ReplicaSet -> control_pod -> StatefulSet -> worker_pods. The control_pod in here does not do any real work (except orchestration and readiness prob proxying), but adds scheduling latency and extra failure domain. A simpler architecture like Rollout -> StsReplicaSet -> StatefulSet -> worker_pods would be ideal.
I have two questions.
Are there any existing solution in the Argo ecosystem to support this use case?
If I write a StsReplicaSet CRD, and adds StsTemplate to the Rollout spec, would the community be interested?
If I do above on a fork, based on the nature of the change, how hard to maintain? I.e. how modular this change is?
```
This is the question I asked. 
Any help would be appreciated!

Issue #4009, opened 2024-12-18
Notifications not firing for Analysis Run fails when Analysis Run is part of an Expriment


meeech:
I am starting this ticket to capture the information as I investigate and try to resolve this issue. 
If anyone has any pointers or thoughts, please add them. 

When an analysis run fail happens and that analysis run is part of an inline experiment step, we don't get the on-analysis-run-error or on-analysis-run-fail notification. 

**Analysis Run Error** 
✅ Background Analysis Run: event: AnalysisRunError  object: rollout/basic-rollout
❌ Inline Step Analysis Run:  event: AnalysisRunError  object: experiment/basic-rollout-exp-steps-b66774df5-3-0

We get the RolloutAborted notification for both, because the event that fires belongs to the rollout/* object in both cases

**Analysis Run Fail**
✅ Background Analysis Run: event: AnalysisRunFailed  object: rollout/basic-rollout
❌ Inline Step Analysis Run:  event: AnalysisRunFailed  object: experiment/basic-rollout-exp-steps-bd7bdfcc8-4-0

We get the RolloutAborted notification for both, because the event that fires belongs to the rollout/* object in both cases

So this has me thinking theres a few possible options: 

* with the notif engine, we don't give it access to the experiment object events? 
* Or is this a case where the events are being fired off the wrong object - where we use the experiment EventRecorder, when we should find the parent(?) rollout object and use _its_ EventRecorder?

I'll keep digging. Unsure what the ideal would be:
would we like something like `on-experiment-analysis-run-failed`, `on-experiment-analysis-run-error`... or would things be better served with them using the already existing triggers? I think when its a step it would make sense to use the existing triggers, and have the rollout object available for the templates, but what about stand alone experiments? 

**Version**

1.7.2 (but this has existed as a problem as long as I've been using experiment step, so at least 1.5/1.6

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

meeech:
More rough notes from a conversation with @zachaller:

Rollouts uses the notification engine in 2 ways
  - As a library (handles on-event triggers)
  - As a controller (handles when triggers)

Path to explore

Create a new notification controller / deployment  - this is similar to what Argo CD does.
[...]

PR #4124, opened 2025-02-09
fix(experiments): fire rollout event on experiment step

meeech:
This fix (feat?) resolves the issue raised in #4009. 

I think this meets what users would expect from the use of an Experiment as a Step in a Rollout. 

As a user, if I setup notifications on AnalysisRun events, I would expect an AnalysisRun that fails should generate a notification. To be clear, this is ONLY when the Analysis Run belongs to an Experiment, which itself belongs to the Rollout (As a Step in that Rollout).  https://argo-rollouts.readthedocs.io/en/stable/features/experiment/#integration-with-rollouts

Please let me know if you would like to discuss this further, but this feels like a 'safe' change. It should even be good to backport to 1.7x (which is what I developed it against).  

I could use some guidance around testing this - I didn't see any prior integration test for this code path. But maybe there's some e2e and I wasn't having success finding it.  

Checklist:

### DIFF
--- a/experiments/experiment.go
+++ b/experiments/experiment.go
@@ -419,14 +419,8 @@ func (ec *experimentContext) reconcileAnalysisRun(analysis v1alpha1.ExperimentAn
 			// Handle the case where the Analysis Run belongs to an Experiment, and the Experiment is a Step in the Rollout
 			// This makes sure the rollout gets the Analysis Run events, which will then trigger any subscribed notifications
 			// #4009
-			if experimentutil.BelongsToRollout(ec.ex) {
-				var roRef metav1.OwnerReference
-				for _, owner := range ec.ex.OwnerReferences {
-					if owner.Kind == "Rollout" {
-						roRef = owner
-					}
-				}
-
+			roRef := experimentutil.GetRolloutOwnerRef(ec.ex)
+			if roRef != nil {
 				rollout, err := ec.argoProjClientset.ArgoprojV1alpha1().Rollouts(ec.ex.Namespace).Get(context.TODO(), roRef.Name, metav1.GetOptions{})
 				if err != nil {
 					ec.log.Warnf("Failed to get parent Rollout of the Experiment '%s': %v", roRef.Name, err)
--- a/experiments/experiment.go
+++ b/experiments/experiment.go
@@ -420,7 +420,13 @@ func (ec *experimentContext) reconcileAnalysisRun(analysis v1alpha1.ExperimentAn
 			// This makes sure the rollout gets the Analysis Run events, which will then trigger any subscribed notifications
 			// #4009
 			if experimentutil.BelongsToRollout(ec.ex) {
-				roRef := ec.ex.OwnerReferences[0]
+				var roRef metav1.OwnerReference
+				for _, owner := range ec.ex.OwnerReferences {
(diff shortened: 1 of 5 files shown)


utils/experiment/experiment.go
```diff
@@ -18,6 +18,14 @@ import (
 
 var terminateExperimentPatch = []byte(`{"spec":{"terminate":true}}`)
 
+func BelongsToRollout(experiment *v1alpha1.Experiment) bool {
+	if len(experiment.OwnerReferences) == 0 {
+		return false
+	}
+
+	return experiment.OwnerReferences[0].Kind == "Rollout"
```
zachaller, line 26: Even though I do not know of any use cases where an Experiment has multiple parents the k8s spec does allow it, can we loop through them here instead of using '[0]'

experiments/experiment.go
```diff
@@ -416,6 +415,19 @@ func (ec *experimentContext) reconcileAnalysisRun(analysis v1alpha1.ExperimentAn
 				eventType = corev1.EventTypeWarning
 			}
 			ec.recorder.Eventf(ec.ex, record.EventOptions{EventType: eventType, EventReason: "AnalysisRun" + string(newStatus.Phase)}, msg)
```
zachaller, line 417: I go back on forth on if we should still emit this event when it's emitted on the Rollout resource. It's pretty low volume event but the duplication seems funny.

Trying to think think of what makes the most logical sense.

zachaller: It would be breaking to change this so let's keep it

meeech: Yeah, and I think its fine, since in case someone has rolled their own notif handlers, like you say, we don't break it. we are already duping it (there is one emitted that belongsa to the analysis run itself).

utils/experiment/experiment.go
meeech: updated. lmk if this looks good to you

meeech:
@zachaller tightened things up a bit.

experiments/experiment.go
```diff
@@ -416,6 +415,19 @@ func (ec *experimentContext) reconcileAnalysisRun(analysis v1alpha1.ExperimentAn
...
+
+			// Handle the case where the Analysis Run belongs to an Experiment, and the Experiment is a Step in the Rollout
+			// This makes sure the rollout gets the Analysis Run events, which will then trigger any subscribed notifications
+			// #4009
+			roRef := experimentutil.GetRolloutOwnerRef(ec.ex)
+			if roRef != nil {
+				rollout, err := ec.argoProjClientset.ArgoprojV1alpha1().Rollouts(ec.ex.Namespace).Get(context.TODO(), roRef.Name, metav1.GetOptions{})
```
zachaller, line 424: I am ok leaving this as a client call because the rest of the experiment code does this but it would be nice if we had access to the rollouts informer in experiment controller

zachaller (approved):

experiments/experiment.go
meeech: i'd be open to making that change (maybe later iteration - i think i will be back in this code later around fixing the complation/delay bug I reported). I wasn't sure about the implications around a changelike that, so I took a light touch.

Issue #3284, opened 2024-01-02
rollout doesn't honor `canary.maxSurge` and `maxUnavailable` settings when `steps` filed is defined 


shaoye:
Checklist:

* [ ] I've included steps to reproduce the bug.
* [ ] I've included the version of argo rollouts.

**Describe the bug**

According to the doc, argo-rollout only honors `maxSurge` and `maxUnavailable` fields if only `steps` field is omitted.

If this is intentional by design, we'd like to report a production outage we just had with the following settings.

```
spec:
  strategy:
    canary:
      steps:
        - setWeight: 50
        - setHeaderRoute:
        - pause:
            duration: 1h
      dynamicStableScale: true
      maxSurge: 10%
      maxUnavailable: 0%
```

We set the first step to have a weight of 50% so we can start evaluating the canary version in step 2.

We assumed argo will create new pods gradually to Weight 50 with 10% maxSurge. But it turns out argo created 50% more pods at once and we had a outage due to excessive connections to the DB.

After pausing for 1 hour, the replica set was bumped to 100% capacity again at once because not following the 10% maxSurge strategy, once again triggered an outage. 

We are eager to understand if there is any disadvantages by following both `maxSurge` and `steps` fields.

Otherwise we have to write the steps like this

```
      steps:
        - setWeight: 10
        - setWeight: 20
        - setWeight: 30
        - setWeight: 40
        - setWeight: 50
        - setHeaderRoute:
        - pause:
            duration: 1h
        - setWeight: 60
        - setWeight: 70
        - setWeight: 80
        - setWeight: 90
        - setWeight: 100
```

**To Reproduce**

**Expected behavior**

**Screenshots**

**Version**

1.5.0

**Logs**

```
# Paste the logs from the rollout controller

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

agaudreault:
Can you share the section of the [documentation](https://argo-rollouts.readthedocs.io/en/stable/features/canary/#dynamic-canary-scale-with-traffic-routing) saying that `maxSurge` and `maxUnavailable` are used for TrafficRouting canary? 

[...]

PR #4239, opened 2025-04-15
feat(controller): adds toleratedUnavailable for flexible availability checking

y-rabie:
This addresses the closed  #3899 and the open #4190.

I ported from the previous PR the case of not checking stable RS full availability when adding scale-down annotation when we had nothing with scaling down the stable replicaset, i.e: bluegreen rollouts, and traffic-routed canary ones that don't use `dynamicStableScale`.

This adds another field under canary `toleratedUnavailable` which solves for the rigid availability checking problem in _multiple_ places, currently exactly 3. The main motivation is that we should not be that rigid especially with dynamic environments, where pods are evicted continuously. This is doing something similar to what a PDB does, but I don't think it'd be desirable for rollouts controller to handle PDBs itself as well.

I know it might not intuitively seem reasonable to use a field in 3 different contexts, but they all are pretty much doing the same thing. The description of the field in the docs covers the current 3 cases, but going again over them here:
* When fully promoting the canary replicaset to be the stable one, we shouldn't wait for full _exact_ availability.
* When switching traffic to the stable replicaset, we shouldn't wait for full exact availability (equal to a desired replicas number corresponding to the desired traffic weight) of the stable replicaset.
* When adding the scaledown delay annotation.

On naming, if the field name is not clear enough, I'm open for suggestions.

Unit tests for everything have been added/adjusted.

Checklist:


docs/features/specification.md
```diff
@@ -219,6 +219,31 @@ spec:
...
+      #    it will be with respect to the desired number of replicas for the
+      #    stable replicaset (which is corresponding to the desired traffic
+      #    weight).
+
+      # This is used to solve the problem in dynamic environments with
+      # cluster autoscalers, where pod evictions can happen for various
+      # reasons: consolidation, rotation, etc. Defaults to 0, which is a
+      # rigid requirement for full availability. Set to a higher value
+      # for more flexibility. Rounded down if a percentage. +optional
+      toleratedUnavailable: '20%'
```
nebojsa-prodana, line 245: Isn't a part of the problem here that maxUnavailable is not working properly for traffic shifted canaries? https://github.com/argoproj/argo-rollouts/issues/3284

To me this borderline sounds like something that could be covered by maxUnavailable, although I guess that would be a breaking change in a way.

nebojsa-prodana: >I don't think it makes sense to respect maxSurge and maxUnavailable for traffic-routed canary
If you go directly from the previous traffic weight to the next traffic weight, then indeed it doesn't, but if you can gradually lerp between the two based on maxUnavailable then it can make sense.

> To me toleratedUnavailable does something similar to what a PDB (per ReplicaSet) does, not maxUnavailable, and I didn't want to overload the meaning of the field. The name could be enhanced however, but this is the best I could come up with 😄

We have a similar thing in our legacy in-house CD system that we call `minimumHealthyPercent` and we have it capped to 80% to help deal with PDB evictions and spot terminations. It basically enforces that at least `minimumHealthyPercent` of pods per RS have to be available.  PDBs themselves are also capped at 80% minAvailable so that this is sort of kept in sync.

Possibly a terrible idea, but I wonder if a rollout object should optionally have a reference to PDB and try to respect it instead of having to define similar behaviour in multiple places 🤔

sherifabdlnaby:
Is this feature still in plan ? This will pretty much a problem any app that is big and dynamic faces.

danp60:
It looks like this recent API change adding `replicaProgressThreshold` is very similar to this feature (https://github.com/argoproj/argo-rollouts/pull/4341). Wondering if it makes sense to have both fields?

Issue #3573, opened 2024-05-13
[Rollouts dashboard] Load time improvements by caching 5mb javascript bundle


CarlosLanderas:
# Summary

What change needs making?

Add cache strategy for argo rollouts main javascript bundle.

Right now, users have to download a **5mb** javascript file everytime they access the dashboard UI.

The javascript is being bundled using a hash, that will change on new versions, so it would make a lot of sense to cache this file.

We are mantaining our own forked version and have added this change to the dashboard server and the results are a decrease of 700ms average on UI load time.

We are using `cache-control: max-age=31536000, immutable`

![image](https://github.com/argoproj/argo-rollouts/assets/16475649/9f162074-9d94-4f35-968c-219be2badd7a)

![image](https://github.com/argoproj/argo-rollouts/assets/16475649/dff25218-c9fb-46c5-b4c1-fce493d67c4a)

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

pablotp:
@CarlosLanderas thanks for creating this issue and sharing how you are working around this. I've create a PR to allow customizing the `Cache-Control` header: https://github.com/argoproj/argo-rollouts/pull/4248

PR #4248, opened 2025-04-23
feat(dashboard): Enable the in-browser cache for static files

pablotp:
Checklist:

This is a fix for https://github.com/argoproj/argo-rollouts/issues/3573

### DIFF
--- a/server/server_static.go
+++ b/server/server_static.go
@@ -6,6 +6,7 @@ import (
 	"io/fs"
 	"mime"
 	"net/http"
+	"os"
 	"path"
 	"regexp"
 	"strconv"
@@ -24,6 +25,7 @@ var (
 const (
 	ContentType   = "Content-Type"
 	ContentLength = "Content-Length"
+	CacheControl  = "Cache-Control"
 )
 
 func (s *ArgoRolloutsServer) staticFileHttpHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


server/server_static.go
```diff
@@ -67,6 +69,12 @@ func (s *ArgoRolloutsServer) staticFileHttpHandler(w http.ResponseWriter, r *htt
 
 	w.Header().Set(ContentType, determineMimeType(embedPath))
 	w.Header().Set(ContentLength, strconv.Itoa(len(fileBytes)))
+	// Set the Cache-Control header if specified in the environment variable
+	// This can be used to enable in-browser cache of static files
+	cacheControl := os.Getenv("CACHE_CONTROL_HEADER")
```
kostis-codefresh, line 74: I don't understand why we need an environment variable to activate this. Is there a scenario where users will NOT want the cache control header?

kostis-codefresh (changes requested):
This a very good feature to implement, but I think it makes sense to examine how Argo CD already solves the same problem and follow/match all the architectural choices from there https://github.com/argoproj/argo-cd/blob/14b9d68079eec5e484f9f38a7cc9ed227efa442f/server/server.go#L1481

server/server_static.go
pablotp: I am closing this PR as it is a very old item

Issue #4249, opened 2025-04-23
[Dashboard] Cache the kubectl list responses


pablotp:
# Summary

Implement a cache to reduce the kubectl calls needed to show the list of rollouts.

# Use Cases

When there are many deployments in the same namespace, the kubectl list commands can take a long time to complete. A cache will improve the user experience when loading the list of rollouts, and it will reduce the load in the control plane.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

pablotp:
PR with a suggestion on how to implement this: https://github.com/argoproj/argo-rollouts/pull/4250

PR #4250, opened 2025-04-23
feat(dashboard): Allow setting an in-memory cache for kubectl calls

pablotp:
Checklist:

Related issue: https://github.com/argoproj/argo-rollouts/issues/4249


zachaller:
What's the motivation behind this I believe that this should all be backed by the view controller which is already using informers, so should be cached there.

pablotp:
> What's the motivation behind this I believe that this should should instead use the viewControllers informer as a cache or setup a new informer as a cache vs rolling our own?

The two motivators are:
- Reduce the number of calls made to the control plane.
- Reduce the loading time of the dashboard. With the amount of services I have running, it may take ~30 to 50 seconds to load the list of rollouts.

I don't have much experience with this source code, so I would appreciate any guidance you could provide.
Where are the viewControllers you mention defined?

zachaller:
After a quick peak I am not sure the viewControllers informers are usable in this exact case, however I would still use an informer vs rolling our own cache. We can just add an informer to the server to cache the resources.

pablotp:
Wouldn't the informer need to load all the resources first? I understand that the benefit of using informers comes not from the initial call but from subsequent ones. Please let me know if you are suggesting something different.

The goal of this change is to avoid loading or listing all the resources on the first call, limiting it to no more than once every 5 minutes (for example). This limitation applies to each pod separately.

zachaller:
it would load it at startup of the dashboard then it sets up a watch to keep updated with any adds/deletes/updates and any get to that then comes from it's cache.

PR #4251, opened 2025-04-23
fix(controller): Fix a bug where pod Metadata is not updated

heshamelsherif97:
Fixes https://github.com/argoproj/argo-rollouts/issues/4256

This PR tackles a critical bug that leads to inaccurate pod metadata (labels/annotations) in setups with high Pod Counts. Updates currently fail silently when pods are modified concurrently (triggering resourceVersion mismatches for the Update API) or when pods are deleted unexpectedly. These failures cascade, preventing subsequent pods from being updated and compromising data used for analysis.

The solution significantly enhances the update process's resilience. It now fetches the latest pod information right before attempting an update, gracefully skipping deleted pods and accounting for any changes in the object that happened after the List call. Furthermore, it isolates update failures, allowing the process to continue with other pods even if one encounters an error.

The reason for opening the PR against `1.8` because I feel this needs to be released as a patch in the current stable version. and it will be difficult to cherry-pick given that there are changes in that area that is already merged related to parallelization.

I will draft another one against master, keeping in mind the changes related to parallelizing the Update calls that have already been merged.

### DIFF
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -832,6 +832,12 @@ func (f *fixture) expectListPodAction(namespace string) int {
 	return len
 }
 
+func (f *fixture) expectGetPodAction(p *corev1.Pod) int {
+	len := len(f.kubeactions)
+	f.kubeactions = append(f.kubeactions, core.NewGetAction(schema.GroupVersionResource{Resource: "pods"}, p.Namespace, p.Name))
+	return len
+}
+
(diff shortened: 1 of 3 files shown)


zachaller:
Can you close this pr and open one against master

heshamelsherif97:
> Can you close this pr and open one against master

@zachaller https://github.com/argoproj/argo-rollouts/pull/4258

Issue #4256, opened 2025-04-25
Inconsistencies with pod metadata during progressive rollouts


heshamelsherif97:
**Describe the bug**

There's a bug causing metadata inconsistencies during rollouts due to fragile pod update logic. When updating ephemeral metadata, if any pod update fails (due to deletion or concurrent modification causing conflicts), the entire process halts for the remaining pods. This leaves many pods with outdated metadata, directly impacting metric accuracy. This problem is reproducible in high-replica scenarios by deleting pods while the system attempts to apply stable metadata. The goal is for all running pods to consistently reflect the stable metadata post-rollout.

**To Reproduce**

1-Create a Rollout resource configured with a high number of replicas.
2- Update the Rollout's specification to trigger a progressive update.
3- Wait until the canary ReplicaSet is fully scaled (100%) and its definition includes the stable metadata.
4- Observe as the system starts applying the stable metadata to the individual pods managed by the canary ReplicaSet.
5- While this metadata update is in progress, manually delete some pods that still belong to the canary ReplicaSet but haven't yet received the stable metadata.
6- Notice that a single failure (attempting to update a deleted pod) disrupts the process, causing many other running pods to remain with outdated (canary) metadata instead of being updated to stable.

**Expected behavior**

Expect all running pods to have the stable metadata

**Version**

`1.8.2`

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

indeed-kelvin:
+1. We've encounter this bug on our end and it's preventing us from being able to mark a rollout as completed which causes our automations to rollback a release after some timeout.

vvvegero97:
+1 faced this issue during testing argo-rollouts. canaryMetadata stays on pods even after finishing the rollout process. Makes metrics inconsistant and unreliable.

PR #4258, opened 2025-04-25
fix(controller): Fix a bug where failure to update pod Metadata cascades to all other pods

heshamelsherif97:
Fixes https://github.com/argoproj/argo-rollouts/issues/4256

This PR tackles a critical bug that leads to inaccurate pod metadata (labels/annotations) in setups with high Pod Counts. Updates currently fail silently when pods are modified concurrently (triggering resourceVersion mismatches for the Update API) or when pods are deleted unexpectedly. These failures cascade, preventing subsequent pods from being updated and compromising data used for analysis.

The solution significantly enhances the update process's resilience. It now fetches the latest pod information right before attempting an update, gracefully skipping deleted pods and accounting for any changes in the object that happened after the List call. Furthermore, it isolates update failures, allowing the process to continue with other pods even if one encounters an error.

### DIFF
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -833,6 +833,12 @@ func (f *fixture) expectListPodAction(namespace string) int {
 	return len
 }
 
+func (f *fixture) expectGetPodAction(p *corev1.Pod) int {
+	len := len(f.kubeactions)
+	f.kubeactions = append(f.kubeactions, core.NewGetAction(schema.GroupVersionResource{Resource: "pods"}, p.Namespace, p.Name))
+	return len
+}
+
 func (f *fixture) expectGetRolloutAction(rollout *v1alpha1.Rollout) int { //nolint:unused
 	len := len(f.actions)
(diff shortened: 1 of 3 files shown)


heshamelsherif97:
@zachaller Not sure what the plan here is. Is it feasible to integrate this specific change into 1.8 and issue a patch release? Or do we need to wait for 1.9

zachaller:
Thanks for working on the unit tests!

indeed-kelvin:
@heshamelsherif97 @zachaller - anything I can help with here to get it across the line for a 1.8 patch?

heshamelsherif97:
I struggled with my attempts to write unit tests for this change probably it's trivial but I am not very familiar with the codebase.

Any guidance on how to proceed with this PR would be appreciated.

@zachaller @indeed-kelvin

indeed-kelvin:
Hello @heshamelsherif97 - I've added a unit test and fixed some issues I saw. Submitted the changes as a pr in your fork, please review: https://github.com/heshamelsherif97/argo-rollouts/pull/1

PR #4259, opened 2025-04-25
docs: blue green w/ ALB not supported without downtime

zachaller:
Blue green with ALB ingress can not guarantee 100% uptime during a rollout.


nickolanick:
Hi @zachaller,
This doc note is still missing from the stable version. Can it be backported?
More importantly, is there an ETA for a fix for the underlying bug? The lack of a 100% traffic guarantee with ALBs seems like a Major issue.

vl-kp:
This is the major blocker to use argo rollout in AWS environment

Issue #4241, opened 2025-04-17
Canary deployment while no traffic management - no canary and stable weights


dgmt:
Checklist:

* [Y] I've included steps to reproduce the bug.
* [Y] I've included the version of argo rollouts.

**Describe the bug**

When using Argo Rollouts without traffic management, the rollout.Status.Canary.Weights field is nil. According to the documentation, the weights should still be approximated and usable even when traffic management is not configured. However, this does not seem to be the case.

**To Reproduce**

1. Create a rollout with a canary strategy but without traffic management. Example rollout spec:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: example-rollout
  namespace: default
spec:
  replicas: 3
  strategy:
    canary:
      steps:
        - setWeight: 20
        - pause: {}
        - setWeight: 50
        - pause: {}
        - setWeight: 100
```
2. Deploy the rollout using `kubectl apply -f rollout.yaml`
3. Retrieve the rollout object using the Argo Rollouts plugin:
`kubectl get rollout sample-service -o yaml`
4. Observe that the rollout.Status.Canary.Weights is nil.

**Expected behavior**

The rollout.Status.Canary.Weights field should be populated with approximated weights based on the canary strategy, even when traffic management is not configured.

**Version**

Tried v1.7.2, then updated to v1.8.2 and the problem persisted.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #4296, opened 2025-05-27
feat(deploy): support setting Canary and Stable weights without traffic routing

dgmt:
### Description

This PR enables the specification of separate weights for Canary and Stable replicas even when traffic routing is disabled. This enhancement allows teams to simulate canary rollout behavior more flexibly, supporting internal testing and rollout strategies without requiring traffic routing.

We originally opened an issue to discuss this feature:
https://github.com/argoproj/argo-rollouts/issues/4241

As there has been no response, we're submitting this PR as a proposed improvement to the upstream project. It enhances usability by allowing calculated weight distribution in non-traffic-routed canary deployments, aligning with real-world deployment needs.

### Tests

We have intentionally kept the tests unchanged at this stage simply for the fact that there is many of them that are affectd by the change,  and it appears that rollout.Status.Canary.Weights is currently used internally as an indicator for traffic routing. Given this, we wanted to open the discussion before making broader test updates.

If the maintainers consider this change to break intended invariants or internal assumptions, we’re happy to close the PR and revisit our approach.

Checklist:

* [ X] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.

### DIFF
--- a/pkg/apis/rollouts/v1alpha1/types.go
+++ b/pkg/apis/rollouts/v1alpha1/types.go
@@ -997,7 +997,8 @@ type CanaryStatus struct {
 	CurrentBackgroundAnalysisRunStatus *RolloutAnalysisRunStatus `json:"currentBackgroundAnalysisRunStatus,omitempty" protobuf:"bytes,2,opt,name=currentBackgroundAnalysisRunStatus"`
 	// CurrentExperiment indicates the running experiment
 	CurrentExperiment string `json:"currentExperiment,omitempty" protobuf:"bytes,3,opt,name=currentExperiment"`
-	// Weights records the weights which have been set on traffic provider. Only valid when using traffic routing
+	// Weights records the weights which have been set for the canary and stable ReplicaSets.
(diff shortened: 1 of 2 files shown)


zachaller:
At a high level I don't think this change is in the correct spot, I think that if the only issue is the status not updating correctly for non traffic routed canaries we should be able to add that as a feature to basic canaries or find the bug there for the reason it is not updating status weights.

dgmt:
> At a high level I don't think this change is in the correct spot, I think that if the only issue is the status not updating correctly for non traffic routed canaries we should be able to add that as a feature to basic canaries or find the bug there for the reason it is not updating status weights.

If i understand you correctly, this behaviour is a bug? Because over the code, there are several tests for whether the canary is handled with traffic routing by checking whether the rollout.Status.Canary.Weights != nil. This threw me off as some internal decision by the maintainer team.

PR #4310, opened 2025-06-06
fix: experiment services deletion before reconciling traffic routing in rollouts

Abhish2702:
**Problem Statement**
This PR addresses an intermittent issue observed in weighted experiments where Service objects associated with the experiments are prematurely deleted before their corresponding weight updates are fully propagated to the HTTPProxy resource. This race condition can result in temporary downtime or traffic misrouting.

**Root Cause**
The issue stems from the parallel reconciliation of HTTPProxy and experiment Service resources by the Rollouts Controller and Experiments Controller, respectively. Specifically, it was observed that the Service objects are sometimes deleted before the HTTPProxy has been updated with the correct traffic weights. This lack of synchronization between the two controllers can cause a window where traffic is routed to non-existent backend services.

**Proposed Solution**
To enforce proper ordering and avoid premature deletion, this PR leverages the existing spec.terminate flag in the Experiment spec. This flag is only set by the Rollouts Controller when it is safe to begin termination of an experiment. By adding a conditional check for spec.terminate == true before deleting experiment Service objects, we ensure that services are only removed after the rollout has finalized HTTPProxy updates and it is safe to do so. We have introduced a new flag to ensure that only experiments created by rollouts are affected.

This solution provides a lightweight and non-invasive mechanism to ensure the correct sequence of events, thereby preventing downtime during traffic shifting in weighted experiments.

**Summary**
1. Ensures experiment services are deleted only when spec.terminate is set by the Rollouts Controller.
2. Aligns service deletion with HTTPProxy updates to maintain traffic continuity.
3. Fixes a race condition without introducing additional coordination overhead between controllers.

Checklist:

### DIFF
--- a/experiments/experiment.go
+++ b/experiments/experiment.go
@@ -137,6 +137,12 @@ func (ec *experimentContext) reconcileTemplate(template v1alpha1.TemplateSpec) {
 			// Create service for template if service field is set
 			if desiredReplicaCount != 0 {
 				ec.createTemplateService(&template, templateStatus, rs)
+			} else {
+				if rs.Status.AvailableReplicas == 0 {
+					// Check if service should be deleted when ReplicaSet has scaled down to 0 available replicas
+					svc := ec.templateServices[template.Name]
+					ec.deleteTemplateService(svc, templateStatus, template.Name)
+				}
 			}
 
 		} else {
@@ -159,11 +165,6 @@ func (ec *experimentContext) reconcileTemplate(template v1alpha1.TemplateSpec) {
 			templateStatus.LastTransitionTime = &now
 		}
 
-		// Check if service should be deleted when ReplicaSet has scaled down to 0 available replicas
-		if desiredReplicaCount == 0 && template.Service != nil && rs.Status.AvailableReplicas == 0 {
-			svc := ec.templateServices[template.Name]
-			ec.deleteTemplateService(svc, templateStatus, template.Name)
-		}
 	}
 
 	if rs == nil {
--- a/experiments/experiment_test.go
+++ b/experiments/experiment_test.go
@@ -175,13 +175,10 @@ func TestScaleDownRSAfterFinish(t *testing.T) {
 	inThePast := timeutil.Now().Add(-10 * time.Second).UTC().Format(time.RFC3339)
 	rs1.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey] = inThePast
 	rs2.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey] = inThePast
-	rs1.Status.AvailableReplicas = 0
-	rs2.Status.AvailableReplicas = 0
 	f := newFixture(t, e, rs1, rs2, s1)
 	defer f.Close()
 
 	updateRs1Index := f.expectUpdateReplicaSetAction(rs1)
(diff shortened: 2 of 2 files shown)


experiments/experiment.go
```diff
@@ -275,7 +275,7 @@ func (ec *experimentContext) scaleTemplateRS(rs *appsv1.ReplicaSet, template v1a
 	} else {
 		if desiredReplicaCount == 0 && template.Service != nil {
 			svc := ec.templateServices[template.Name]
-			ec.deleteTemplateService(svc, templateStatus, template.Name)
+			ec.deleteTemplateService(svc, templateStatus, template.Name, false)
 		}
```
jessesuen, line 279: This is where I think the issue might be. Currently we are deleting the service as soon as the desired replicas is zero, and are not without consideration of available replicas of the `rs`.

```golang
if desiredReplicaCount == 0 && template.Service != nil {
	svc := ec.templateServices[template.Name]
	ec.deleteTemplateService(svc, templateStatus, template.Name, false)
}
```

Since there can be a delay between the desired replicas and available replicas, we might be prematurely deleting the service, where higher-level routing (e.g. rollout traffic routing) might still be pointing to it.

To fix this, we could delay the deletion of Service until the available replicas reaches `0`. So the fix would be adjusting the check to:

```golang
if desiredReplicaCount == 0 && template.Service != nil && rs.Status.AvailableReplicas == 0 {
```

When available replicas reaches zero, the service pointing to it should be safe to delete, since the Pod Endpoints list would anyways be empty.

jessesuen (changes requested):
I don't think this is the right approach. It was intended that Experiments (as well as AnalysisRuns) not have understanding of a Rollout, so adding a flag to indicate as such would break some of the original design principles.

That said, I think I can see one possible bug with the implementation along with a potential fix. Will comment in the code.

jessesuen:
@Abhish2702 do you have logs of the controller when the problem occurs? I would like to see the evidence of the controller decisioning leading up to the outage.

experiments/experiment.go
Abhish2702: Yes, Jesse thanks for the suggestion, we will look into it more.

Abhish2702:
@jessesuen could you please revisit the PR ? We agree with your solution that svc deletion should only check replicas availability.

jessesuen:
The e2e tests (e.g. `TestExperimentSuite/TestExperimentWithServiceNameAndScaleDownDelay`) are failing because they expect the previous behavior where services are deleted immediately:

```
Experiment expectation 'experiment services count == 0' failed
```

The test will need to be adjusted so that it waits for available replicas reaching zero before checking.

Abhish2702:
@jessesuen I have fixed the approach and e2e

experiments/experiment.go
```diff
@@ -213,6 +213,12 @@ func (ec *experimentContext) reconcileTemplate(template v1alpha1.TemplateSpec) {
...
 	}
 
+	// Check if service should be deleted when ReplicaSet has scaled down to 0 available replicas
+	if desiredReplicaCount == 0 && template.Service != nil && rs != nil && rs.Status.AvailableReplicas == 0 {
+		svc := ec.templateServices[template.Name]
+		ec.deleteTemplateService(svc, templateStatus, template.Name)
+	}
```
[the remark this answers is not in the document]
Abhish2702: thanks jesse, I have incorporated your suggestion.

PR #4324, opened 2025-06-17
fix: issue in rollout stuck during scaling

Abhish2702:
Checklist:

### DIFF
--- a/rollout/analysis_test.go
+++ b/rollout/analysis_test.go
@@ -244,7 +244,7 @@ func TestCreateBackgroundAnalysisRun(t *testing.T) {
 	f.objects = append(f.objects, r2, at)
 
 	createdIndex := f.expectCreateAnalysisRunAction(ar)
-	f.expectUpdateReplicaSetAction(rs2)
+	f.expectPatchReplicaSetAction(rs2)
 	index := f.expectPatchRolloutAction(r1)
 
 	f.run(getKey(r2, t))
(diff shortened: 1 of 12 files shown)


UsmanQ-devops (approved):

Abhish2702:
@jessesuen could you please provide review over the design like if you are satisfied with the approach of patching ?
If yes, I can work upon fixing the e2e 
Please let me know if you envisage any challenges involved with this approach

jessesuen (changes requested):
Unfortunately, switching from `Update()` to `Patch` is very risky, and ultimately is just sweeping the underlying root cause of the conflict error under the rug. The `Update()` call serves an important purpose: it prevents misguided controller updates to the `ReplicaSet` when the controller might be operating a stale version of that ReplicaSet. That includes conflicting updates that the controller itself may have just made (not just external parties).

While annoying, I believe that allowing the controller to hit the conflict error, then re-reconcile, is safe, and the error is benign. The original ReplicaSet handling implementation was copied directly from the k8s Deployment controller, which uses `Update()` calls against the `ReplicaSet`.

> This would cause the rollout to get stuck and not progress, as the controller would not be able to successfully update the ReplicaSet’s replica count

In this scenario, I expect the controller to re-reconcile due to the conflict error encountered in the ReplicaSet update. But I do not expect it to get stuck. If this is indeed happening, I feel we should treat this as a separate problem and understand why the re-reconcile upon error either doesn't happen or help.

Issue #4178, opened 2025-03-06
Add support of specifying Lease name for Leader Election


WZHGAHO:
# Summary

Please add support to specify default leader election `Lease` lock name or generate the name to something unique which includes instance-id.

# Use Cases

I use Argo Rollouts in a distributed setup where multiple Argo Rollouts Controllers with `--instance-id` specified are connecting to a central cluster where a central Argo Rollouts Controller is running. The distributed controllers are using kubeconfig of the central cluster, so all of them are are detecting the same Lease (which is currently specified in the code as [constant value](https://github.com/argoproj/argo-rollouts/blob/master/controller/controller.go#L100) `argo-rollouts-controller-lock` and therefore only one is elected, even though they are differentiated because of --instance-id flag.

According to [Kubernetes documentation](https://kubernetes.io/docs/concepts/architecture/leases/#custom-workload), best practice in case multiple instances are deployed of a component is like using a hashed value or a prefix to avoid collisions for the Leases:

> If a cluster operator or another end user could deploy multiple instances of a component, select a name prefix and pick a mechanism (such as hash of the name of the Deployment) to avoid name collisions for the Leases.
> You can use another approach so long as it achieves the same outcome: different software products do not conflict with one another.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

janlonsetteig:
Any hope of getting this prioritized and assigned to a milestone? 🙏

ksawerykarwacki:
Any chance for someone to look at the opened PR?

PR #4325, opened 2025-06-17
feat(controller): add leader-election-lease-lock-name option

antoinedeschenes:
Implements enhancement proposed in https://github.com/argoproj/argo-rollouts/issues/4178

Checklist:


controller/controller.go
```diff
@@ -500,7 +507,7 @@ func (c *Manager) Run(ctx context.Context, rolloutThreadiness, serviceThreadines
 		log.Infof("Leaderelection get id %s", id)
 		leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
 			Lock: &resourcelock.LeaseLock{
-				LeaseMeta: metav1.ObjectMeta{Name: defaultLeaderElectionLeaseLockName, Namespace: electOpts.LeaderElectionNamespace}, Client: c.kubeClientSet.CoordinationV1(),
+				LeaseMeta: metav1.ObjectMeta{Name: electOpts.LeaderElectionLeaseLockName, Namespace: electOpts.LeaderElectionNamespace}, Client: c.kubeClientSet.CoordinationV1(),
```
kostis-codefresh, line 510: It would be great if we can match the Argo Workflows behavior where it automatically sets the lock name according to instance id. This will make operation much simpler instead of forcing people to manually pass the lock name as a parameter. See https://github.com/argoproj/argo-workflows/blob/main/cmd/workflow-controller/main.go#L153

kostis-codefresh (changes requested):
Can we have some unit tests as well?

kostis-codefresh:
Closed as part of https://github.com/argoproj/argo-rollouts/pull/4768

PR #4341, opened 2025-07-01
feat: added available replica pods threshold

### LINKED ISSUE #3991: Add an option to proceed to the next rollout step based on a percentage of available replicas
# Summary

Add an option to allow Argo Rollouts to proceed to the next step in a rollout even if the desired replica count has not been fully achieved. Instead, allow progression based on a predefined percentage of the new ReplicaSet’s replicas being available.

# Use Cases

In certain deployment scenarios, strict adherence to achieving 100% of the desired replica count can cause rollouts to stall indefinitely. For example:

- Dynamic Workloads: Some applications dynamically scale or remove pods as part of their operation. In these cases, the number of AvailableReplicas may fluctuate during the rollout process, even though the application is healthy and able to progress.
- Non-Critical Tolerance: Applications that do not require every single replica to be ready before moving to the next step might benefit from progressing with a threshold percentage instead of waiting indefinitely.

Currently, the rollout process performs a strict check via:

```
func allDesiredAreAvailable(rs *appsv1.ReplicaSet, desired int32) bool {
    return rs != nil && desired == *rs.Spec.Replicas && desired == rs.Status.AvailableReplicas
}

```
This check ensures that the AvailableReplicas exactly matches the desired count. However, in the above scenarios, the rollout can get stuck indefinitely if all replicas are not consistently ready.

# Proposed Solution

Introduce an optional parameter (e.g., progressThreshold) that defines the percentage of the desired replicas that must be available before proceeding to the next step. For example:
```
strategy:
  canary:
    steps:
      - setWeight: 50
    progressThreshold: 90
```
In this case, the rollout would proceed if at least 90% of the desired replicas are available, offering flexibility for applications that tolerate slight deviations.

# Benefits
- Flexibility: Allows rollouts to progress in dynamic or non-critical environments where exact replica availability cannot always be achieved.
- Avoid Stalling: Prevents rollouts from getting stuck indefinitely in scenarios where waiting for 100% of replicas is impractical or unnecessary.
- Backward Compatible: Existing behavior remains unchanged unless the progressThreshold parameter is explicitly defined.

---

emoral435:
Resolves #3991
## Context
As of the date of writing this PR, whenever you want to promote your canary for each and every step, Argo Rollouts waits for the number of available pods grow and equal the *exact* number of desired pods that the step calls for. For example...
1. Argo Rollouts sees that the canary is passing all metric checks, analysis runs, etc.
2. Argo Rollouts 'promotes' the canary by setting increasing the weight of traffic to send to the canary image. For this example, imagine if we have 10 pods, and Argo sets the canary to desire 7 / 10 of those pods.
3. Argo Rollouts will wait until all 7 (and **EXACTLY** 7) of those pods are ready and available to send traffic too.

This current flow presents several problems for several users of Argo Rollouts...
1. If your deployments use a lot of pods, then your spinning up a lot of pods for the canary deployment as well, leading to a huge horizontal increase in pods that some users may want to limit (for cost reasons, etc.)
2. During times of traffic, polling for exact numbers of pods may lead to slower promotions. If you have an HPA configured, and you have spun up 3 extra pods to help with traffic, then the strict check of `desiredPods == availablePods` can lead to staggered or slower promotions.

With this PR, the rollout.spec changes to allow users to customize a number of percentage of available pods they are content with being up and ready in order to promote a canary release to the next step or be fully promoted.

Checklist:


docs/features/specification.md
```diff
@@ -291,6 +291,17 @@ spec:
...
 
+        # The percentage or number of replica pods within the applications ReplicaSet
+        # that are available and ready when a rollout is ready to be promoted. Useful if your application
+        # configured an HPA to help handle different loads of traffic, but you still want quick promotions.
+        # Value can be an absolute number (ex: 5 replica pods) or a percentage of total pods
+        # during the promotion of a rollout (ex: 90%). Defaults to 100%. 
+        # Current percentage is calculated by the following:
+        # PERCENTAGE = available replicas / desired replicas for the current step
+        # +optional
+        - replicaProgressThreshold: '90%'
```
emoral435, line 302: replacing to be more similar to how the HPA configuration is down: see Sat's most recent HPA PR for more info here https://github.com/Nextdoor/infra-cdk-k8s/pull/1347

utils/replicaset/canary.go
```diff
@@ -35,7 +66,7 @@ func AtDesiredReplicaCountsForCanary(ro *v1alpha1.Rollout, newRS, stableRS *apps
 	} else {
 		desiredNewRSReplicaCount, desiredStableRSReplicaCount = CalculateReplicaCountsForTrafficRoutedCanary(ro, weights)
 	}
-	if !allDesiredAreAvailable(newRS, desiredNewRSReplicaCount) {
+	if !allDesiredAreAvailable(ro.Spec.Strategy.Canary.ReplicaProgressThreshold, newRS, desiredNewRSReplicaCount) {
```
emoral435, line 69: If `replicaThreshold != nil`, my funciton, else, allDesired, else return false

docs/features/specification.md
schahal: e.g., https://github.com/kubernetes/api/blob/bfa25161ab1437d38582d78a4580aaab97a0ee3f/autoscaling/v2/types.go#L238-L250

schahal: e.g., something like:

```
replicaProgressThreshold:
  type: Percent
  value: 100
```

or

```
replicaProgressThreshold:
  type: Pods
  value: 7
```

emoral435: Added this feature! Great callout!

docs/features/specification.md
```diff
@@ -291,6 +291,19 @@ spec:
...
+        # The percentage or number of replica pods within the applications ReplicaSet
+        # that are available and ready when a rollout is ready to be promoted. Useful if your application
+        # configured an HPA to help handle different loads of traffic, but you still want quick promotions.
+        #  Defaults to 100% if replicaProgressThreshold is not specified.
+        # thresholdType should be either "Percent" | "Pod"
+        # Current percentage that is checked against the input percent value is calculated by the following:
+        # CURRENT PERCENTAGE = available replicas / desired replicas for the current step
+        # +optional
+        - replicaProgressThreshold:
+            thresholdType: "Percent"
```
schahal, line 303: Based on last commit, this should probably be `type: Percent` (instead of ` thresholdType: "Percent"`)... so key name + removal of quotes to stay with style of doc

schahal (changes requested):

schahal:
Looking at other PR approvals, adding @kostis-codefresh and @zachaller to potentially review as well?

I think this will also solve a significant use case for me that I tried describing in (1) [comment on the root GH issue](https://github.com/argoproj/argo-rollouts/issues/3991#issuecomment-3059503572), and (2) [Slack](https://cloud-native.slack.com/archives/C01U781DW2E/p1752590657709979)

zachaller (approved):

PR #4410, opened 2025-08-20
feat(controller): propagate rollout revision annotation to pod templates. Fixes #3800

### LINKED ISSUE #3800: Set rollout revision number in the replicaSet.spec.template.metadata annotations
# Summary
In order to support seeing metrics and logs for different rollout revisions, we would like to see the rollout revision number being added to the ReplicaSet.spec.template.metadata annotation, so that they are available at the pods level. Currently rollout revision is only associated with ReplicaSet. Since revision number is how user refers to a given deployment in the dashboard, we want to provide that revision number in Datadog Trace, our metrics dashboard, and Splunk log.

With the ephemeral-metadata feature, it only adds the 'canary' and 'stable' labels/annotations to the pods. When user searches Splunk log, those labels mean different revisions depending on the search time frame. That makes it very confusing in troubleshooting, especially, if there were multiple deployments triggered in a short time frame.

With the actual revision number at the pods level, it allows user to see what version of the service was handling the traffic from the metrics dashboard. In Splunk, user can then search log messages by revision number as well.

We tried modifying the rollout-pod-template to supply that revision number to the pod level, but it somehow triggered the re-calculation logic of the revision number when we rollback, causing two revision numbers to be created, even though one of them did not have any pod. So, if that can be supported by Argo Rollouts to add that revision number to ReplicaSet.spec.template.metadata annotations, this can be done properly without the side effect of double revision numbers on rollback.

What change needs making?
Add rolllout_revision to ReplicaSet.spec.template.metadata annotation.

# Use Cases
1. In Splunk, this will allow user to search log by revision number for troubleshooting, esp. when multiple deployments were triggered in a short time frame.
2. In metrics dashboard, metrics can now be tied to a given revision number, so user can see which revisions are handling traffic at what time. 
3. Datadog traces can also have revision number associated with them to help with troubleshooting.

When would you use this?
We would use that in Splunk and on our metrics dashboard to help with troubleshooting. This also helps us showcase the traffic shifting feature of Argo Rollouts by showing the metrics on how different revisions of a service is handling traffic. 

---

shmargum:
Fixes #3800

This change enables rollout revision tracking at the pod level by propagating the rollout.argoproj.io/revision annotation from ReplicaSets to their pod templates, allowing users to access revision information via Kubernetes Downward API.

Changes:
- Modified SetNewReplicaSetAnnotations to always set revision annotation on newRS.Spec.Template.Annotations
- Added comprehensive tests to verify pod template annotation propagation
- Maintains backward compatibility with existing functionality

Use cases:
- Search logs by specific revision number
- Track which service revisions handle traffic
- Associate traces with precise deployment versions
- Utilize increasing revision number as a "priority" within new pods for things like rabbitmq consumer priority

Checklist:

### DIFF
--- a/rollout/canary_test.go
+++ b/rollout/canary_test.go
@@ -928,6 +928,11 @@ func TestRollBackToStable(t *testing.T) {
 	expectedRS1 := rs1.DeepCopy()
 	expectedRS1.Annotations[annotations.RevisionAnnotation] = "3"
 	expectedRS1.Annotations[annotations.RevisionHistoryAnnotation] = "1"
+	// Expect revision annotation on pod template for downward API access
+	if expectedRS1.Spec.Template.Annotations == nil {
+		expectedRS1.Spec.Template.Annotations = make(map[string]string)
+	}
+	expectedRS1.Spec.Template.Annotations[annotations.RevisionAnnotation] = "3"
 	firstUpdatedRS1 := f.getUpdatedReplicaSet(updatedRSIndex)
 	assert.Equal(t, expectedRS1, firstUpdatedRS1)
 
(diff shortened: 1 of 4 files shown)


shmargum:
one note about this

when rolling **forward** to a previous stable that is **still active** (so rolling back i guess) the revision labels will be updated in the replicaset's template.metadata.labels, but since the pods are running their labels will NOT be updated...
not sure if they should be (kind of like what the argocd ephemeral metadata does)

PR #4477, opened 2025-10-01
fix(controller): Always reconcile pod metadata

### LINKED ISSUE #4256: Inconsistencies with pod metadata during progressive rollouts

JRBANCEL:
Fixes #4256 

This addresses an issue we observe often: pod metadata is not consistent.

## Example

Rollout is done.

```
kubectl argo rollouts list rollouts -A
NAMESPACE      NAME                                         STRATEGY   STATUS        STEP   SET-WEIGHT  READY  DESIRED  UP-TO-DATE  AVAILABLE
platform       portal-app-rollout                           Canary     Healthy       10/10  100         2/2    2        2           2
```

A single ReplicaSet has replicas:
```
kubectl get replicaset -n platform | grep portal
...
portal-app-rollout-58ff85fcb                                          2         2         2       17h
```

With the correct spec:
```
spec:
  replicas: 2
  template:
    metadata:
      labels:
        ...
        role: stable
```

Yet, some pods don't have the expected labels:
```
kubectl get pods -n platform -l app=portal --show-labels         
NAME                                 READY   STATUS    RESTARTS   AGE   LABELS
portal-app-rollout-58ff85fcb-bk7kr   2/2     Running   0          17h   ...role=canary...
portal-app-rollout-58ff85fcb-tqd98   2/2     Running   0          17h   ...role=stable...
```

Reconciliation happens and does nothing to fix this:
```
time="2025-10-01T18:41:16Z" level=info msg="Started syncing rollout" generation=314 namespace=platform resourceVersion=398118957 rollout=portal-app-rollout
time="2025-10-01T18:41:16Z" level=info msg="Found 1 TrafficRouting Reconcilers" namespace=platform rollout=portal-app-rollout
time="2025-10-01T18:41:16Z" level=info msg="Reconciling TrafficRouting with type 'Istio'" namespace=platform rollout=portal-app-rollout
time="2025-10-01T18:41:16Z" level=info msg="No StableRS exists to reconcile or matches newRS" namespace=platform rollout=portal-app-rollout
time="2025-10-01T18:41:16Z" level=info msg="No Steps remain in the canary steps" namespace=platform rollout=portal-app-rollout
time="2025-10-01T18:41:16Z" level=info msg="No status changes. Skipping patch" generation=314 namespace=platform resourceVersion=398118957 rollout=portal-app-rollout
time="2025-10-01T18:41:16Z" level=info msg="Reconciliation completed" generation=314 namespace=platform resourceVersion=398118957 rollout=portal-app-rollout time_ms=61.302378
```

## Why does it matter?
We have PDBs on `role=stable` and the mismtach between the RS replica count and the Pods actually matching the label leads to PDB preventing any change. 

## Root Cause

This is because the code bails if the ReplicaSet metadata matches what it should be. So if for any reason: pod update failure, crash, OOM, etc... one of the pod is not updated after the ReplicaSet is updated the first time, the controller will never try to rectify this and the inconsistent state will stay.

Also, the race condition is still present. Between the ReplicaSet update and the Pod listing. It is entirely possible that the ReplicaSet controller is reconciling on the old spec and creates a Pod with the old spec, after the Pod listing.

This ensure the logic to update pod metadata always runs, which is mostly fine efficiency-wise because the current code already doesn't patch pod in the correct state, so it would be a Pod list call at most in the standard case.

## Testing
Asserting on reads doesn't seem to be a pattern and has little value. It would require 80 tests to be updated, makes more sense to ignore it.

The increased reconciliation time reveals a racy behavior in the e2e tests, especially around `promote`.
I added logic to make sure the controller has done its job before promote is applied.

---
Checklist:
* [X] I've signed my commits with [DCO](https://github.com/argoproj/argoproj/blob/main/community/CONTRIBUTING.md#legal)
  * https://github.com/argoproj/argo-rollouts/pull/4476

### DIFF
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -799,7 +799,8 @@ func filterInformerActions(actions []core.Action) []core.Action {
 			action.Matches("list", "services") ||
 			action.Matches("watch", "services") ||
 			action.Matches("list", "ingresses") ||
-			action.Matches("watch", "ingresses") {
+			action.Matches("watch", "ingresses") ||
+			action.Matches("list", "pods") {
 			continue
 		}
 		ret = append(ret, action)
@@ -846,18 +847,6 @@ func (f *fixture) expectPatchReplicaSetAction(rs *appsv1.ReplicaSet) int {
 func (f *fixture) expectUpdatePodAction(p *corev1.Pod) int {
 	len := len(f.kubeactions)
 	f.kubeactions = append(f.kubeactions, core.NewUpdateAction(schema.GroupVersionResource{Resource: "pods"}, p.Namespace, p))
-	return len
-}
-
-func (f *fixture) expectListPodAction(namespace string) int {
-	len := len(f.kubeactions)
-	f.kubeactions = append(f.kubeactions, core.NewListAction(schema.GroupVersionResource{Resource: "pods"}, schema.GroupVersionKind{Kind: "Pod", Version: "v1"}, namespace, metav1.ListOptions{}))
-	return len
-}
-
-func (f *fixture) expectGetRolloutAction(rollout *v1alpha1.Rollout) int { //nolint:unused
-	len := len(f.actions)
-	f.kubeactions = append(f.actions, core.NewGetAction(schema.GroupVersionResource{Resource: "rollouts"}, rollout.Namespace, rollout.Name))
 	return len
 }
 
--- a/rollout/ephemeralmetadata.go
+++ b/rollout/ephemeralmetadata.go
@@ -74,25 +74,26 @@ func (c *rolloutContext) syncEphemeralMetadata(ctx context.Context, rs *appsv1.R
 		return nil
 	}
 	modifiedRS, modified := replicasetutil.SyncReplicaSetEphemeralPodMetadata(rs, podMetadata)
-	if !modified {
-		return nil
-	}
 
 	// Used to access old ephemeral data when updating pods below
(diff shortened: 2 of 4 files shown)


zachaller (approved):

JRBANCEL:
@zachaller :eyes: :pray:

zachaller (approved):

PR #4464, opened 2025-09-20
feat: set start and end time for analysis in Kayenta metric provider

melnikovi:
Checklist:

* [ ] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [ ] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [ ] I've signed my commits with [DCO](https://github.com/argoproj/argoproj/blob/main/community/CONTRIBUTING.md#legal)
* [ ] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ ] My builds are green. Try syncing with master if they are not. 
* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

PR #4482, opened 2025-10-05
fix: proto codegen was not properly working

jessesuen:
Upon reviewing https://github.com/argoproj/argo-rollouts/pull/4464, it was discovered that our code generation stopped working properly.

I believe the root cause is due to a recent K8s library upgrade, our codegen stopped working properly. K8s upstream has moved off of gogoproto, and so all of the extra gogoproto annotations should no longer be generated in the `generated.proto` file. I'm not sure why it suddenly stopped working properly after upgrade, but this should fix it going forward.

This is a one-line change to the Makefile to `rm -f generated.proto`, which ensures that it will always be generated.


jessesuen:
I tested that the "Verify Codegen" check will now fail if there is a diff in the `generated.proto` file. See:

https://github.com/argoproj/argo-rollouts/actions/runs/18274722080/job/52024093609?pr=4482

Issue #2864, opened 2023-07-03
Recommend Ping Pong strategy as default for Canary with Traffic Routing. 


rajeshetty87:
Checklist:

* [X] I've included steps to reproduce the bug.
* [X] I've included the version of argo rollouts.

**Describe the bug**

Canary w/ Traffic Routing with AWS ALB Ingress causes availability issues at the end of rollout.
The rollout progresses successfully through all steps and at the end, service selector labels are switched by Rollouts. 
This triggered two actions initiated by the ALB Ingress Controller
1. Registers the targets from the Canary Target Groups to the Stable Target Groups. 
2. Updates the weights on the ALB listener from 0 --> 100% for Stable and 100 -> 0% for Canary

The above two actions don't necessarily happen in sync and in scenarios when listener updates (Action #2) before (Action #1) , ALB will send the traffic to an empty target group causing 503 errors and thus causing availability drop. 
The issues is mostly observed in swim-lanes with high TPS and lots of targets (>300) whereas low traffic swim-lanes don't 
show similar issue. 

This is a known issue with Canary with Traffic routing and has been widely discussed in these threads([#2061](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2061), [#1283 ](https://github.com/argoproj/argo-rollouts/issues/1283), #[1453](https://github.com/argoproj/argo-rollouts/issues/1453)  ). The solution to resolve this issue was the [Ping Pong feature](https://argoproj.github.io/argo-rollouts/features/traffic-management/alb/#zero-downtime-updates-with-ping-pong-feature) and details are available here. 

The request with this bugs is to make the Ping Pong strategy as a default when using  Canary w/ Traffic Routing + AWS ALB 
over Simple Canary. This will save time for folks performing the switch and help them evaluate the correct solution(Ping Pong) instead of the Canary w/ Traffic Routing that does not provide Zero Downtime deployments. 

**To Reproduce**

Steps to reproduce

- Create a namespace with ingress on AWS ALB 
- Create rollout with Canary w/ traffic routing stragey.
- Send atleast 70K TPS over the ingress or >=300 Pods 
- Do a rollout with Canary Steps as  0-25-50-75 (weightage doesn't matter)
- Once all the steps are completed looks for 503' in the ALB logs 

**Screenshots**

[...]

PR #4510, opened 2025-10-24
fix(trafficrouting): Verify ALB TG health before final canary step. Fixes #4433

### LINKED ISSUE #4433: Prevent 503s during final canary step with AWS ALB
# Summary

## Problem definition

We have a setup with Argo Rollouts, AWS ALB controller and canary deployment with trafficRouting enabled. During releases we observe 503 errors generated by AWS load balancers, which indicate that target group does not have any healthy targets. This problem is essentially the same as the one documented for blue green deployments: https://github.com/argoproj/argo-rollouts/pull/4259/files

It happens after canary steps are finished and `stable` service selector is switched to new replica set.

We are aware that ping pong deployment strategy might fix this issue, but this is not an optimal option for us since we want to have a fixed preview URL.

Attaching Argo Rollouts logs for a release where 503s occurred. 

[canary503.txt](https://github.com/user-attachments/files/22114442/canary503.txt)

The points of interest are: 

```
Switching stable service selector - "Aug 29, 2025 @ 06:41:45.605","Switched selector for service 'example-service-active' from '56b5dc5bc' to '7fdb8fff6b'"
Updating target weights to set 100% to stable - "Aug 29, 2025 @ 06:41:45.628","New weights: &TrafficWeights{Canary:WeightDestination{Weight:0,ServiceName:example-service-preview,PodTemplateHash:7fdb8fff6b,},Stable:WeightDestination{Weight:100,ServiceName:example-service-active,PodTemplateHash:7fdb8fff6b,},Additional:[]WeightDestination{},Verified:nil,}"
Argo confirms that targets are registered in stable target group - "Aug 29, 2025 @ 06:41:55.903","Service example-service-active (TargetGroup active_target_group(replaced_real_arn)) verified: 2 endpoints registered"
```
503 errors where observed between 06:41:45 and 06:41:55

## Proposal

On top of switching service selector (which causes target registrations in AWS), canary deployment also switches all traffic back to `stable` service after canary deployment steps are complete. The issue here is that traffic switch happens faster than targets registration.

Argo Rollouts [already implements](https://github.com/argoproj/argo-rollouts/blob/6cf05d9eecf7cbf393bc2f68559480e1283f2850/utils/aws/aws.go#L353) logic that is capable of verifying that pods from new RS are registered in AWS target group. 

The proposal is simple - use that functionality to verify that `stable` target group is healthy before switching back all traffic to it. So the general flow would look like this:

1. Canary steps are finished (100% of traffic is on `canary` service)
2. Service selector for `stable` switched to new RS
3. Argo Verifies that pods from new RS are registered in `stable` target group
4. Argo switches 100% traffic to `stable` selector

Would it be technically possible to achieve such behaviour?

# Use Cases

We would like to use this functionality to achieve real zero downtime deployment while still having a dedicated `preview/canary` endpoint for manual testing during release.

---

krapie:
Verify ALB target group health before applying the final weight patch to prevent a race condition where new pods are not yet healthy after a selector change, which can cause the ALB to return 503 errors when routing 100% of traffic.

Fixes: https://github.com/argoproj/argo-rollouts/issues/4433

Checklist:

### DIFF
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -53,7 +53,7 @@ func (c *rolloutContext) rolloutBlueGreen() error {
 		return err
 	}
 
-	err = c.awsVerifyTargetGroups(activeSvc)
+	err = c.awsVerifyTargetGroups(activeSvc, false)
 	if err != nil {
 		return err
 	}
--- a/rollout/canary.go
+++ b/rollout/canary.go
@@ -316,7 +316,7 @@ func (c *rolloutContext) canProceedWithScaleDownAnnotation(oldRSs []*appsv1.Repl
 	if err != nil {
 		return false, err
 	}
-	err = c.awsVerifyTargetGroups(stableSvc)
+	err = c.awsVerifyTargetGroups(stableSvc, false)
 	if err != nil {
 		return false, err
 	}
--- a/rollout/service.go
+++ b/rollout/service.go
@@ -130,7 +130,7 @@ func (c *rolloutContext) areTargetsVerified() bool {
 // of the Service's Endpoint IPs and ports registered. Only valid for services which are reachable
 // by an ALB Ingress, which can be determined if there exists a TargetGroupBinding object in the
 // namespace that references the given service
-func (c *rolloutContext) awsVerifyTargetGroups(svc *corev1.Service) error {
+func (c *rolloutContext) awsVerifyTargetGroups(svc *corev1.Service, checkHealth bool) error {
 	if !shouldVerifyTargetGroup(c.rollout, c.newRS, svc) {
 		return nil
 	}
@@ -162,7 +162,7 @@ func (c *rolloutContext) awsVerifyTargetGroups(svc *corev1.Service) error {
 	}
 
 	for _, tgb := range tgBindings {
-		verifyRes, err := aws.VerifyTargetGroupBinding(ctx, c.log, awsClnt, tgb, endpoints, svc)
+		verifyRes, err := aws.VerifyTargetGroupBinding(ctx, c.log, awsClnt, tgb, endpoints, svc, checkHealth)
 		if err != nil {
 			c.recorder.Warnf(c.rollout, record.EventOptions{EventReason: conditions.TargetGroupVerifyErrorReason}, conditions.TargetGroupVerifyErrorMessage, svc.Name, tgb.Spec.TargetGroupARN, err)
 			return err
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -284,6 +284,30 @@ func (c *rolloutContext) reconcileTrafficRouting() error {
 			}
 		}
 
+		// Verify ALB target group health before final weight patch to prevent a race condition where
+		// ALB serves 503s. This can occur when a service's selector is changed, causing new pods to be
+		// registered as targets, but those targets have not yet passed ALB's health checks. If we
+		// update the ALB listener rules to route 100% traffic before targets are healthy, ALB will
+		// return 503 errors since there are no healthy targets available in the target group.
+		if reconciler.Type() == alb.Type && desiredWeight == weightutil.MaxTrafficWeight(c.rollout) {
+			if c.rollout.Spec.Strategy.Canary != nil && c.rollout.Spec.Strategy.Canary.StableService != "" {
+				stableService, err := c.servicesLister.Services(c.rollout.Namespace).Get(c.rollout.Spec.Strategy.Canary.StableService)
+				if err != nil {
+					return err
+				}
+
+				err = c.awsVerifyTargetGroups(stableService, true)
+				if err != nil {
+					return err
+				}
+
+				if !c.areTargetsVerified() {
+					c.log.Infof("Waiting for ALB target group health checks to pass before applying final weight")
+					return nil
+				}
+			}
+		}
+
 		err = reconciler.UpdateHash(canaryHash, stableHash, weightDestinations...)
 		if err != nil {
 			return err
--- a/utils/aws/aws.go
+++ b/utils/aws/aws.go
@@ -342,15 +342,14 @@ type TargetGroupVerifyResult struct {
 	Verified            bool
 	EndpointsRegistered int
 	EndpointsTotal      int
+	EndpointsHealthy    int // New field to track healthy endpoints
 }
 
 // VerifyTargetGroupBinding verifies if the underlying AWS TargetGroup has all Pod IPs and ports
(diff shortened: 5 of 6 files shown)


zachaller:
Does this only happen without Ping Pong, I am generally under the impression that ALBs should always use Rollouts ping pong feature.

krapie:
This issue seems to occur only with ALB + ping-pong strategy. Using the ping-pong strategy prevents the problem, but as noted in issue #4433, there seems to be cases where users wants fixed endpoint during and after the canary rollout.

JyothsnaGuduri:
Hi! I built an image out of this PR and deployed it our env but seeing following error and pod is getting into CrashLoopBackOff. 
I tried with Argo-rollouts chart version`2.40.0`  and `2.40.9`

```
Error: unknown flag: --metricsport
Usage:
  argo-rollouts [flags]
Flags:
      --alb-ingress-classes stringArray               Defines all the ingress class annotations that the alb ingress controller operates on. Defaults to alb (default [alb])
      --alb-tag-key-resource-id string                Set the default AWS LoadBalancer tag key for resource ID that controller uses when verifying target group weights. (default "ingress.k8s.aws/resource")
      --ambassador-api-version string                 Set the Ambassador apiVersion that controller should look when manipulating Ambassador Mappings. (default "getambassador.io/v2")
      --analysis-threads int                          Set the number of worker threads for the Experiment controller (default 30)
      --appmesh-crd-version string                    Set the default AppMesh CRD Version that controller uses when manipulating resources. (default "v1beta2")
      --as string                                     Username to impersonate for the operation
      --as-group stringArray                          Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --as-uid string                                 UID to impersonate for the operation
      --aws-target-group-binding-api-version string   Set the default AWS TargetGroupBinding apiVersion that controller uses when verifying target group weights. (default "elbv2.k8s.aws/v1beta1")
      --aws-verify-target-group                       Verify ALB target group before progressing through steps (requires AWS privileges)
      --burst int                                     Maximum burst for throttle. (default 80)
      --certificate-authority string                  Path to a cert file for the certificate authority
      --client-certificate string                     Path to a client certificate file for TLS
      --client-key string                             Path to a client key file for TLS
      --cluster string                                The name of the kubeconfig cluster to use
      --context string                                The name of the kubeconfig context to use
      --controllers strings                           Explicitly specify the list of controllers to run, currently only supports 'analysis', eg. --controller=analysis. Default: all controllers are enabled
      --disable-compression                           If true, opt-out of response compression for all requests to the server
      --enable-pprof-address string                   Enable pprof profiling on controller by providing a server address.
      --ephemeral-metadata-threads int                Set the number of worker threads for the Ephemeral Metadata reconciler (default 10)
      --experiment-threads int                        Set the number of worker threads for the Experiment controller (default 10)
      --healthzPort int                               Set the port the healthz endpoint should be exposed over (default 8080)
  -h, --help                                          help for argo-rollouts
      --ingress-api-version string                    Set the Ingress apiVersion that the controller should use.
      --ingress-threads int                           Set the number of worker threads for the Ingress controller (default 10)
      --insecure-skip-tls-verify                      If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --instance-id string                            Indicates which argo rollout objects the controller should operate on
      --istio-api-version string                      Set the default Istio apiVersion that controller should look when manipulating VirtualServices. (default "v1alpha3")
      --kloglevel int                                 Set the klog logging level
      --kubeconfig string                             Path to a kube config. Only required if out-of-cluster
      --leader-elect                                  If true, controller will perform leader election between instances to ensure no more than one instance of controller operates at a time (default true)
      --leader-election-lease-duration duration       The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled. (default 15s)
      --leader-election-renew-deadline duration       The interval between attempts by the acting master to renew a leadership slot before it stops leading. This must be less than or equal to the lease duration. This is only applicable if leader election is enabled. (default 10s)
      --leader-election-retry-period duration         The duration the clients should wait between attempting acquisition and renewal of a leadership. This is only applicable if leader election is enabled. (default 2s)
      --logformat string                              Set the logging format. One of: text|json
      --loglevel string                               Set the logging level. One of: debug|info|warn|error (default "info")
      --metricsPort int                               Set the port the metrics endpoint should be exposed over (default 8090)
  -n, --namespace string                              If present, the namespace scope for this CLI request
      --namespaced                                    runs controller in namespaced mode (does not require cluster RBAC)
      --nginx-ingress-classes stringArray             Defines all the ingress class annotations that the nginx ingress controller operates on. Defaults to nginx (default [nginx])
      --password string                               Password for basic authentication to the API server
      --proxy-url string                              If provided, this URL will be used to connect via proxy
      --qps float32                                   Maximum QPS (queries per second) to the K8s API server (default 40)
      --request-timeout string                        The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
      --rollout-resync int                            Time period in seconds for rollouts resync. (default 900)
      --rollout-threads int                           Set the number of worker threads for the Rollout controller (default 10)
      --self-service-notification-enabled             Allows rollouts controller to pull notification config from the namespace that the rollout resource is in. This is useful for self-service notification.
      --server string                                 The address and port of the Kubernetes API server
      --service-threads int                           Set the number of worker threads for the Service controller (default 10)
      --tls-server-name string                        If provided, this name will be used to validate server certificate. If this is not provided, hostname used to contact the server is used.
      --token string                                  Bearer token for authentication to the API server
      --traefik-api-group string                      Set the default Traefik apiGroup that controller uses. (default "traefik.containo.us")
      --traefik-api-version string                    Set the default Traefik apiVersion that controller uses. (default "traefik.containo.us/v1alpha1")
      --traffic-split-api-version string              Set the default TrafficSplit apiVersion that controller uses when creating TrafficSplits. (default "v1alpha1")
      --user string                                   The name of the kubeconfig user to use
      --username string                               Username for basic authentication to the API server
      --version                                       Print version
unknown flag: --metricsport
```

JyothsnaGuduri:
Hi again! I was able to resolve the conflicts in my local and tested the fix in our environment. We didn't see any improvement with this fix. 
We are planning to test ping-pong strategy as mentioned here https://github.com/argoproj/argo-rollouts/issues/2864 to see if it helps

krapie:
Hello @JyothsnaGuduri, thanks for the update. As I'm not working on this PR anymore, I will close the PR.

PR #4545, opened 2025-11-19
feat: Add configurable annotations to skip. Fixes #4546

### LINKED ISSUE #4546: Add configurable annotations to skip
# Summary

Background: Currently annotations are copied from the Rollout to the underlying Replicaset ([code](https://github.com/argoproj/argo-rollouts/blob/e9888ac860759f3657676fee85546256bc0208fb/utils/annotations/annotations.go#L235)) unless it's part of a fixed [list](https://github.com/argoproj/argo-rollouts/blob/e9888ac860759f3657676fee85546256bc0208fb/utils/annotations/annotations.go#L219-L225) 

- Recently our core engineering added a kyverno clusterpolicy to strip certain annotations from replicasets
- This led to an infinite loop where: the rollout controller added the annotation -> clusterpolicy removed it -> repeat
- The loop spammed the k8s api until the max annotation was hit for revision-history
- Note that in our k8s infra, these annotations were needed on the Rollout but not the underlying ReplicaSet (hence core k8s infra added the clusterpolicy to reduce load on etcd)

We would like to be able to supply additional annotations to skip via config to prevent the continuous looping issue

Submitted PR https://github.com/argoproj/argo-rollouts/pull/4545 for this issue

# Use Cases

- To be able to not copy over additional configurable annotations from the Rollout to the underlying ReplicaSet
- Prevents infinite loop situation

---

zelin-l:
Background: Currently annotations are copied from the Rollout to the underlying Replicaset ([code](https://github.com/argoproj/argo-rollouts/blob/e9888ac860759f3657676fee85546256bc0208fb/utils/annotations/annotations.go#L235)) unless it's part of a fixed [list](https://github.com/argoproj/argo-rollouts/blob/e9888ac860759f3657676fee85546256bc0208fb/utils/annotations/annotations.go#L219-L225) 

This PR would allow additional configurable annotations to skip

Use case:
- Recently our core engineering added a kyverno clusterpolicy to strip certain annotations from replicasets 
- This led to an infinite loop where: the rollout controller added the annotation -> clusterpolicy removed it -> repeat
- The loop spammed the k8s api until the max annotation was hit for revision-history
- Note that in our k8s infra, these annotations were needed on the Rollout but not the underlying ReplicaSet (hence core k8s infra added the clusterpolicy to reduce load on etcd)
- By allowing additional configurable annotations to be skipped, the loop/spam issue can be avoided

This PR will:
- Add support for configuring additional annotations to skip via ConfigMap
- Support both YAML array and comma-separated string formats
- Maintain backward compatibility with existing hardcoded annotations
- Add test coverage for changes
- Include documentation and example configuration

The default annotations are still preserved:
- kubectl.kubernetes.io/last-applied-configuration
- rollout.argoproj.io/revision
- rollout.argoproj.io/revision-history
- rollout.argoproj.io/desired-replicas
- notified.notifications.argoproj.io

Users can now add additional annotations via ConfigMap:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argo-rollouts-config
data:
  annotationsToSkip: |
    - "example.com/build-id"
    - "cert-manager.io/cluster-issuer"
```

Issue link: https://github.com/argoproj/argo-rollouts/issues/4546

Checklist:


zachaller:
How does the built in k8s deployment controller deal with a situation like this? Wondering if we should not match what they are doing because Rollouts does try to emulate the Deployment objects behavior.

zelin-l:
Thanks for the review!

Looking at the upstream k8s code it does seems like the default deployment is also using a static list:
https://github.com/kubernetes/kubernetes/blob/dd4d9fad24828f54ad48b5c6c413a6d073b19e90/pkg/controller/deployment/sync.go#L156C5-L156C22
https://github.com/kubernetes/kubernetes/blob/dd4d9fad24828f54ad48b5c6c413a6d073b19e90/pkg/controller/deployment/util/deployment_util.go#L295

It would mean that this would be a feature enhancement over the default deployment. Would be up to the rollout maintainers on if this additional configurability should be added 🙏 

From our end we could explore using a patch downstream as well to inject custom annotations. Please let me know what would be the preferred method - thanks!

PR #4551, opened 2025-11-25
chore: switch to *bool for clusterScope

zachaller:
In kubernetes bool should be a *bool so that tools like argocd do not go out of sync when set to false in git.


rollout/controller.go
```diff
@@ -882,7 +882,7 @@ func (c *rolloutContext) getReferencedAnalysisTemplatesFromRef(templateRefs *[]v
 	templates := make([]*v1alpha1.AnalysisTemplate, 0)
 	clusterTemplates := make([]*v1alpha1.ClusterAnalysisTemplate, 0)
 	for _, templateRef := range *templateRefs {
-		if templateRef.ClusterScope {
+		if templateRef.ClusterScope != nil && *templateRef.ClusterScope {
```
agaudreault: How do you feel about adding methods on the type to do `templateRef.IsClusterScope()` ?

PR #4564, opened 2025-12-17
fix(trafficrouting): set weight before hash for new canary

andrewjamesbrown:
Occasionally during an "interrupted canary" (C1 stable, C2 in progress, Rollout updated with C3) we see errors from istio with HTTP 503 UH (no healthy upstream).  Note that we are using subset-level traffic splitting with Istio.
We have the following Rollouts logs, when we received UH from Istio at `10:11:27.598`:
```
10:11:25.678 - Created ReplicaSet myservice-5c88b7f59c
10:11:25.912 - DestinationRule myservice subset updated (canary: 5c88b7f59c, stable: cff675f55)
10:11:25.921 - VirtualService `myservice` set to desiredWeight '0'
10:11:25.939 - Traffic weight updated from 5 to 0
10:11:26.074 - Scaled up ReplicaSet myservice-5c88b7f59c (revision 425) from 0 to 2
```
 
What this looks like is Rollouts first updating the DestinationRule to point to the new canary (which is not yet available) at a weight of 5%, and 9ms later, setting the VS to a weight of 0.  Given there is also propagation delay in the Istio control plane, there is a window of time where Istio may be configured to send to a new canary RS that isn't yet available, due to it having an old weight set as part of the previous in-progress rollout.

This fix addresses this, by setting the desiredWeight to `0` before calling `UpdateHash` (but only for the case where a new canary is triggered - existing cases maintain the same behaviour)

Checklist:

### DIFF
--- a/rollout/trafficrouting_test.go
+++ b/rollout/trafficrouting_test.go
@@ -1710,6 +1710,11 @@ func TestDynamicScalingAbortWithZeroReplicas(t *testing.T) {
 //  3. Scale up new canary pods
 //  4. SetWeight to desired canary weight
 func TestNewRolloutResetsWeightBeforeUpdatingHash(t *testing.T) {
+	const (
+		canaryService = "myservice-canary"
+		stableService = "myservice-stable"
+	)
+
 	f := newFixture(t)
 	defer f.Close()
 
@@ -1726,8 +1731,8 @@ func TestNewRolloutResetsWeightBeforeUpdatingHash(t *testing.T) {
 		// Use a generic traffic routing config (SMI) that uses the mock reconciler
 		SMI: &v1alpha1.SMITrafficRouting{},
 	}
-	r1.Spec.Strategy.Canary.CanaryService = "myservice-canary"
-	r1.Spec.Strategy.Canary.StableService = "myservice-stable"
+	r1.Spec.Strategy.Canary.CanaryService = canaryService
+	r1.Spec.Strategy.Canary.StableService = stableService
 
 	// r2 is the "previous canary" - rollout in progress at 5% weight
 	r2 := bumpVersion(r1)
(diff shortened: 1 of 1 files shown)


andrewjamesbrown:
@zachaller I'm not sure the release/review process for this repo - any thoughts on how to get it closer to being merged?  We've been using this on an internal build for the past ~month and it has resolved our issue, so I'm keen to get this into a release so we can start using upstream again instead of our fork.  Thanks!

zachaller:
@andrewjamesbrown  Can you resolve the conflict I will take a look this week.

ssantichaivekin-plaid:
Hi team, I was looking over the changes from version 1.8.2 to the current version 1.9.0 to prepare for a future upgrade, and this PR stood out to me as a potential breaking change.

Specifically, it might break the guardrail implemented in https://github.com/argoproj/argo-rollouts/pull/3878
```go
if !c.checkReplicasAvailable(c.stableRS, weightutil.MaxTrafficWeight(c.rollout)-desiredWeight) {
	return nil
}
```

It seems to me that this guardrail is supposed to always be called _before_ all weight and route management, e.g `reconciler.UpdateHash` `reconciler.SetWeight`, etc. This commit broke this invariant by calling `reconciler.SetWeight` before this check.

Here's the scenario I'm thinking about. During deployments with `dynamicStableScale` enabled, let's say we are rolling out from A to B and we are at 10% A 90% B. At this point, if we "supersede" this deployment by changing the desired pod-hash from B to C, this new code will call `SetWeight(...)` immediately, causing 100% of requests to go to 10% of A pods.

100% of requests will continue to go to 10% of A pods for the amount of time it takes for A to spin up 90% of pods. If A takes 2 minutes to scale-up to 100%, the system immediately experiences 2 minutes of downtime. If A takes 16 minutes to scale up to 100% (due to this other recent change that slows-down `dynamicStableScale` immediate scale-up https://github.com/argoproj/argo-rollouts/pull/4035#issuecomment-3976629863) we will experience 16 minutes of downtime.

This issue seems to only affect services with `dynamicStableScale: true`

Note that the scenario described is a thought experiment and I haven't tried to reproduce this issue in realistic systems yet, so it is very possible that my analysis is incorrect. If anyone has tried out this commit, please let me know.

PR #4580, opened 2026-01-06
feat: Rollout plugin support

Hariharasuthan99:
Implementation of the merged proposal https://github.com/argoproj/argo-rollouts/pull/4387 
Currently getting piloted within our company

Scope:
RolloutPlugin CRD + Controller (Optionally enabled by users)
Basic Canary Support with RolloutPlugin (set weight, pause, inline and background analysis run)
Built In Support for Statefulset
Support external plugins for other k8s resource kinds
Pause,Resume,Abort,Restart,PromoteFull actions
Metrics and Notifications with RolloutPlugin

Checklist:


cmd/rollouts-controller/main.go
```diff
@@ -288,9 +333,168 @@ func newCommand() *cobra.Command {
 					ephemeralMetadataThreads,
 					ephemeralMetadataPodRetries)
 			}
-			if err = cm.Run(ctx, rolloutThreads, serviceThreads, ingressThreads, experimentThreads, analysisThreads, electOpts); err != nil {
-				log.Fatalf("Error running controller: %s", err.Error())
+
+			// Setup RolloutPlugin Controller if enabled (uses controller-runtime)
+			setupRolloutPlugin := enabledControllers[controllerRolloutPlugin]
+
+			if setupRolloutPlugin {
```
tperdue321: this can be one line just like 			`if enabledControllers[controllerAnalysis] ` above to keep style consistent

cmd/rollouts-controller/main.go
```diff
@@ -47,10 +63,28 @@ const (
...
+)
+
+func init() {
+	// Set controller-runtime logger to a null logger to suppress the warning
+	// We use logrus for our own logging
+	ctrl.SetLogger(logr.New(ctrllog.NullLogSink{}))
+
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(v1alpha1.AddToScheme(scheme))
+}
```
tperdue321: why is this init function necessary? I'm always concerned about defining an init function on the main package, especially if it's specific to a plugin that may never be used by most users of this project.

tperdue321: Could this go in the `newCommand` or in some way that fits more naturally to the Cobra style of initialization?

cmd/rollouts-controller/main.go
```diff
@@ -225,11 +259,24 @@ func newCommand() *cobra.Command {
...
 			}
 
-			var cm *controller.Manager
-
 			enabledControllers, err := getEnabledControllers(controllersEnabled)
 			errors.CheckError(err)
 
+			// Create shared MetricsServer that will be used by all controllers
+			log.Info("Creating shared metrics server")
+			metricsServer := metrics.NewMetricsServer(metrics.ServerConfig{
```
tperdue321, line 278: I'm not very familiar with the analysis controller manager. However, previously it explicitly set `RolloutLister` and `ExperimentLister` as `nil`. Now these are being set with `				RolloutLister:                 tolerantinformer.NewTolerantRolloutInformer(dynamicInformerFactory).Lister(),` and `				ExperimentLister:              tolerantinformer.NewTolerantExperimentInformer(dynamicInformerFactory).Lister(),`. Why is globalizing of the metrics server necessary for adding plugins?

To be clear, I'm not arguing for or against this change in a hypothetical vacuum. However, with how large this PR is, I think we should keep it as tightly scoped to the actual feature logic of the new Plugin to make it easier to catch any potential issues specific to the plugin logic without polluting the PR with drive by changes. 

Please correct me if I'm wrong, but this looks like an unrelated driveby change

tperdue321: If my assumption of this being a drive by is correct, please self-review and try to cleanup any other drivebys to make it easier to keep my (and any other) PR reviews more focused to the core values this PR adds.

cmd/rollouts-controller/main.go
```diff
@@ -288,9 +333,168 @@ func newCommand() *cobra.Command {
...
+					electOpts.LeaderElectionNamespace = leaderElectionNamespace
+				}
+
+				mgrOpts := ctrl.Options{
+					Scheme: scheme,
+					Metrics: metricsserver.Options{
+						BindAddress: "0", // Disable metrics server, use standard controller's metrics
+					},
+					HealthProbeBindAddress:  "0",   // Disable health probe, use standard controller's healthz
+					LeaderElection:          false, // Disable leader election, use standard controller's leader election
```
zachaller: not sure that we can turn this off, it should also follow any flags that enable/dissable on the main one.

zachaller: Setting `LeaderElectionID` while `LeaderElection: false` has no effect. This needs to either (a) enable controller‑runtime leader election with the same lock, or (b) actually start `mgr` only after the standard manager wins leadership. As written, the comment's intent ("use standard controller's leader election") is not realized.

pkg/apis/rollouts/v1alpha1/rolloutplugin_types.go
```diff
@@ -0,0 +1,271 @@
...
+// Only canary strategy is supported for RolloutPlugin
+type RolloutPluginStrategy struct {
+	// Canary strategy configuration. Uses the shared CanaryStrategy type, but only the
+	// following fields are observed by the RolloutPlugin controller:
+	//   - steps (setWeight, pause, analysis)
+	//   - analysis (background analysis)
+	// All other CanaryStrategy fields (traffic routing, service names, ReplicaSet scaling, etc.)
+	// are ignored and have no effect on RolloutPlugin behaviour.
+	// +optional
+	Canary *CanaryStrategy `json:"canary,omitempty" protobuf:"bytes,2,opt,name=canary"`
```
zachaller: Using the shared CanaryStrategy type here seems too broad for what this controller actually supports. Even with the comment, the API still accepts lots of valid-looking fields like traffic routing, service settings, anti-affinity, scaling knobs, etc. that are effectively ignored. That creates a pretty misleading UX because users can write configs that look supported but silently do nothing. I think this would be safer as a narrower RolloutPlugin-specific canary schema that only exposes the fields we actually honor today, with unsupported fields rejected rather than accepted-and-ignored.

pkg/apis/rollouts/v1alpha1/rolloutplugin_types.go
```diff
@@ -0,0 +1,271 @@
...
+	// Name of the referenced resource
+	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
+
+	// Namespace of the referenced resource (optional, defaults to the RolloutPlugin's namespace)
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,4,opt,name=namespace"`
+}
+
+// PluginConfig contains configuration for the resource plugin
+type PluginConfig struct {
```
zachaller: I think the unsupported fields need to be supported here, especially now that we are using this config to enable build-in controllers

zachaller: Do we need a new PluginConfig type here at all? We already have an existing plugin config model in utils/plugin/types.PluginItem that is used via the controller ConfigMap for plugin resolution/installation (name, location, sha256, args, etc.). Defining a second plugin config shape in the CRD seems likely to create overlap and confusion about which layer owns what. My expectation would be that the ConfigMap continues to own plugin installation/resolution, while the CRD only references the plugin by name and carries resource-specific runtime inputs if needed.

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+			// Record pause event
+			if r.Recorder != nil {
+				r.Recorder.Eventf(rolloutPlugin, record.EventOptions{
+					EventReason: conditions.RolloutPluginPausedReason,
+				}, conditions.RolloutPluginPausedMessage)
+			}
+
+			newStatus.Message = "manually paused"
+		}
+		return ctrl.Result{}, r.updateStatus(ctx, rolloutPlugin, newStatus, logCtx)
```
zachaller, line 255: Doesn't this early return skip `checkPausedConditions`? If someone pauses longer than timeoutSeconds and resumes, won't `RolloutPluginTimedOut` immediately degrade the rollout (and abort it with timeoutAbort)? The comment at L299 says pause time doesn't count toward the timeout, but that only seems to hold for step pauses.

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+	} else if newStatus.Phase == v1alpha1.RolloutPluginPhasePaused && !rolloutPlugin.Spec.Paused && len(newStatus.PauseConditions) == 0 {
+		logCtx.Info("Manual resume detected, clearing pause state")
+		newStatus.Message = "Rollout resumed"
+	}
+
+	if newStatus.Abort && !newStatus.Aborted {
```
zachaller, line 261: Since the paused branch above returns early, does `status.abort` just get ignored while paused? In the Rollout controller abort wins over pause — is the difference intentional here? If so we should probably surface something in status so the user knows their abort didn't take.

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+		predicate predicate.Predicate
+	}{
+		{id: "statefulset", object: &appsv1.StatefulSet{}, predicate: statefulSetPredicate},
+	}
+
+	b := ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.RolloutPlugin{}, builder.WithPredicates(rolloutPluginPredicate)).
+		Owns(&v1alpha1.AnalysisRun{}, builder.WithPredicates(analysisRunPredicate))
+
+	for _, w := range workloadWatches {
```
zachaller, line 1222: What happens when someone runs with only an external RPC plugin? Looks like no workload watch gets registered at all, so once the RolloutPlugin converges an image bump never triggers a reconcile. Is the RPC e2e only passing because the config also enables builtin://statefulset for the same GVK?

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+	// List all RolloutPlugin resources in the same namespace as the workload
+	var rolloutPlugins v1alpha1.RolloutPluginList
+	if err := r.Client.List(ctx, &rolloutPlugins, client.InNamespace(obj.GetNamespace())); err != nil {
+		logCtx.WithError(err).Error("Failed to list RolloutPlugin resources")
+		return []reconcile.Request{}
+	}
+
+	var requests []reconcile.Request
+	for _, rp := range rolloutPlugins.Items {
+		if rp.Spec.WorkloadRef.Kind == workloadKind &&
```
zachaller, line 1271: Shouldn't this also compare `rp.Spec.WorkloadRef.Namespace`? As is, a cross-namespace workloadRef never gets events (we list in the workload's namespace), and a same-name workload in the RolloutPlugin's namespace would spuriously match.

rolloutplugin/analysis.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+
+	// Do not create or keep a background run if the rollout is fully promoted
+	if rp.Status.CurrentRevision == rp.Status.UpdatedRevision {
+		logCtx.Info("Rollout is fully promoted, not running background analysis")
+		return nil, nil
+	}
+
+	// Do not create background analysis before the configured starting step
+	if rp.Spec.Strategy.Canary.Analysis.StartingStep != nil && rp.Status.CurrentStepIndex != nil {
```
zachaller, line 282: Isn't CurrentStepIndex still nil when this runs on a fresh revision (it gets set to 0 later in processCanaryRollout)? If so this guard falls through and we create the background AR before step 0, then next reconcile returns nil and that AR is orphaned — never cancelled, never in status. Am I reading the ordering right?

rolloutplugin/analysis.go
```diff
@@ -0,0 +1,513 @@
...
+	return currentAr, nil
+}
+
+// convertAnalysisRunArgsToArguments converts AnalysisRunArgument to Argument
+func convertAnalysisRunArgsToArguments(args []v1alpha1.AnalysisRunArgument) []v1alpha1.Argument {
+	result := make([]v1alpha1.Argument, len(args))
+	for i, arg := range args {
+		result[i] = v1alpha1.Argument{
+			Name:  arg.Name,
+			Value: &arg.Value,
```
zachaller, line 324: Should Value be left nil when ValueFrom is set? MergeArgs prefers non-nil Value over ValueFrom, so a fieldRef arg ends up substituting `""` into the metric query. The Rollout path resolves fieldRef via BuildArgumentsForRolloutAnalysisRun first — do we need the same here?

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,483 @@
...
+
+	replicas := int32(1)
+	if sts.Spec.Replicas != nil {
+		replicas = *sts.Spec.Replicas
+	}
+
+	// Calculate partition based on weight
+	// Example: 10 replicas, 40% weight -> partition = 10 - (10 * 40 / 100) = 6
+	// This means pods 0-3 (4 pods) will be updated, pods 4-9 stay at old version
+	partition := replicas - (replicas * weight / 100)
```
zachaller, line 209: Have you checked small weights here? replicas=3 with setWeight 33 floors to partition=3, so zero pods update — and VerifyWeight does the same math, so the step verifies without the new version taking any traffic. Should this ceil, or fail verification when weight > 0 resolves to 0 pods?

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,483 @@
...
+	p.logCtx.WithFields(log.Fields{
+		"oldPartition": oldPartition,
+		"podsToDelete": podsToDelete,
+	}).Info("Deleting updated pods one-by-one to force graceful rollback")
+
+	deletedCount := int32(0)
+	failedDeletes := []string{}
+
+	// Delete in reverse order (replicas-1 down to oldPartition) for graceful rollback which matches StatefulSet's natural ordering
+	for i := replicas - 1; i >= oldPartition; i-- {
```
zachaller, line 354: This runs synchronously inside Reconcile, up to 600s per pod — a 10-replica abort could pin a worker for an hour, and the RolloutPlugin can't process a resume or promote-full in the meantime. Could this be requeue-driven instead? The sleeps also don't check ctx.

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,483 @@
...
+		}
+
+		deletedCount++
+		p.logCtx.WithFields(log.Fields{
+			"pod": podName,
+		}).Info("Deleted pod for rollback, waiting for replacement to be Ready")
+
+		// Wait for the replacement pod to be Ready before deleting the next one
+		// This ensures service availability during rollback
+		err = p.waitForPodReady(ctx, workloadRef.Namespace, podName, 600) // TODOH make it configurable 600 second timeout
```
zachaller, line 383: Right after the delete, doesn't the terminating pod still report Ready=True (and the cached client can hand back the pre-delete object)? If the wait returns immediately, the one-by-one rollback turns into deleting everything back-to-back. Should it check DeletionTimestamp or wait for a new UID first?

rolloutplugin/plugin/client/client.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		typedPlugin, ok := pluginInstance.(types.RpcResourcePlugin)
+		if !ok {
+			return nil, fmt.Errorf("unexpected type from plugin")
+		}
+		r.instances[pluginName] = typedPlugin
+
+		resp := r.instances[pluginName].InitPlugin(namespace)
+		if resp.HasError() {
```
zachaller, line 106: If InitPlugin fails, don't we leave processClient/rpcConnClient/instances populated? Then every later call takes the already-initialized branch, Ping passes, and InitPlugin never gets retried until the process dies. Should this call `cleanupPlugin` like the Ping failure path below?

zachaller (changes requested):

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+
+		// Clear pause state from previous rollout
+		newStatus.PauseConditions = nil
+		newStatus.ControllerPause = false
+		pCtx.ClearPauseConditions()
+	}
+
+	// Check if we need to start a rollout
+	if resourceStatus.CurrentRevision != resourceStatus.UpdatedRevision {
+		if !conditions.IsRolloutPluginProgressing(newStatus) {
```
zachaller, line 417: I think controller-initiated pauses break here. Once a pause step sets pauseConditions, the next reconcile's checkPausedConditions flips Progressing to Unknown — so IsRolloutPluginProgressing is false, this block treats it as a brand-new rollout, resets CurrentStepIndex to nil (L455), sets Progressing back to True and re-emits the revision event. Every reconcile. The pause duration is never checked, the step index keeps regressing, and after a manual resume the earlier steps re-run (including re-creating AnalysisRuns). Traced it over two reconcile cycles — as far as I can tell pause steps can't ever complete. This needs a guard for the Paused reason, and honestly a unit test that walks a pause step across two reconciles.

cmd/rollouts-controller/main.go
```diff
@@ -303,9 +348,202 @@ func newCommand() *cobra.Command {
...
+						log.Info("RolloutPlugin controller stopped")
+					}()
+
+					pluginWg.Wait()
+				}()
+
+				log.Info("All controllers started successfully")
+
+				// Wait for context cancellation
+				<-ctx.Done()
```
zachaller, line 530: What happens on lost leader election here? cm.Run returns nil after RunOrDie exits, both goroutines finish, and then we block on the signal context forever — with healthz already shut down. The plugin-disabled path below returns and the process exits to re-contend, which is the old behavior. As written, a pod that loses the lease just sits dead-but-Running until a liveness probe happens to kill it.

Makefile
```diff
@@ -265,18 +267,25 @@ test-kustomize: ## run kustomize tests
...
+	@kubectl patch --context='${E2E_K8S_CONTEXT}' configmap argo-rollouts-config -n argo-rollouts --type merge --patch-file test/e2e/rolloutplugin/argo-rollouts-config.yaml
 	@rm -rf plugin-bin
 	@go build -gcflags="all=-N -l" -o plugin-bin/e2e-step-plugin test/cmd/step-plugin-e2e/main.go
+	@go build -gcflags="all=-N -l" -o plugin-bin/e2e-resource-plugin test/cmd/resource-plugin-e2e/main.go
 
 .PHONY: start-e2e
 start-e2e: ## start e2e test environment
 	mkdir -p coverage-output-e2e
-	GOCOVERDIR=coverage-output-e2e go run -cover ./cmd/rollouts-controller/main.go --instance-id ${E2E_INSTANCE_ID} --loglevel debug --kloglevel 6
+	GOCOVERDIR=coverage-output-e2e go run -cover ./cmd/rollouts-controller/main.go --instance-id ${E2E_INSTANCE_ID} --loglevel debug --kloglevel 6 --controllers=rolloutplugin
```
zachaller, line 280: Two problems with forcing --controllers=rolloutplugin here: (1) rollout-plugin-crd.yaml was never added to manifests/crds/kustomization.yaml, and CI installs CRDs from that kustomization without running setup-e2e — so the manager fails cache sync on the missing CRD and os.Exit(1)s ~2 min in, taking the shared e2e controller down for every suite. (2) Even once that's fixed, the plugin-disabled startup path is no longer exercised by e2e at all. Also, doesn't this belong only in a plugin-specific target rather than the shared start-e2e?

pkg/apis/rollouts/v1alpha1/rolloutplugin_types.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Kind of the referenced resource (e.g., StatefulSet, DaemonSet)
+	Kind string `json:"kind" protobuf:"bytes,2,opt,name=kind"`
+
+	// Name of the referenced resource
+	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
+
+	// Namespace of the referenced resource (optional, defaults to the RolloutPlugin's namespace)
+	// +optional
+	Namespace string `json:"namespace,omitempty" protobuf:"bytes,4,opt,name=namespace"`
```
zachaller, line 67: Is cross-namespace workloadRef intentional? Nothing validates it, so a RolloutPlugin in a low-privilege namespace can drive partition patches and pod deletions in any other namespace through the controller's cluster RBAC. Rollout deliberately doesn't have a namespace field on workloadRef for exactly this reason. If there's no strong use case I'd drop the field; if there is, it needs validation and an explicit opt-in. (Note the workload watch mapper also doesn't handle it — cross-namespace refs get no events today.)

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+		conditions.RemoveRolloutPluginCondition(newStatus, v1alpha1.RolloutPluginConditionInvalidSpec)
+	}
+
+	if refErr != nil {
+		return ctrl.Result{}, r.updateStatus(ctx, rolloutPlugin, newStatus, logCtx)
+	}
+
+	// Get the plugin from the singleton PluginManager
+	// Plugins are registered at startup and shared across all reconcilers
+	plugin, err := r.PluginManager.GetPlugin(rolloutPlugin.Spec.Plugin.Name)
```
zachaller: For external plugins GetPlugin also fails transiently (process crashed, ping failed) — but every error lands here as InvalidSpec with ctrl.Result{} and no requeue. The condition-only status write gets filtered by shouldReconcileRolloutPluginUpdate, and external-only configs have no workload watch, so nothing ever retries: a crashed plugin leaves the RolloutPlugin stuck Degraded until someone edits it. Shouldn't transient plugin errors return an error (for backoff) and keep InvalidSpec for actual config problems?

rolloutplugin/plugin/client/client.go
```diff
@@ -0,0 +1,149 @@
...
+		rpcClient, err := r.processClient[pluginName].Client()
+		if err != nil {
+			return nil, fmt.Errorf("unable to get plugin client (%s): %w", pluginName, err)
+		}
+
+		// Cache the RPC client to avoid calling Client() again
+		r.rpcConnClient[pluginName] = rpcClient
+
+		// Request the plugin
+		pluginInstance, err := rpcClient.Dispense("RpcResourcePlugin")
```
zachaller: Same cleanup problem as the InitPlugin path, but nastier: if the process died and we re-exec, a failure here (or at Client()/the type assert) leaves rpcConnClient pointing at the new process while instances still holds the stale instance bound to the dead connection. Every later call pings the new conn (fine) and returns the stale instance — permanently wedged until the new process happens to die. Shouldn't every error return between exec and the Ping call cleanupPlugin? The client tests currently assert the no-cleanup behavior, so they'd need updating too.

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,483 @@
...
+			"spec": map[string]any{
+				"updateStrategy": map[string]any{
+					"rollingUpdate": map[string]any{
+						"partition": partition,
+					},
+				},
+			},
+		},
+	}
+	return p.client.Patch(ctx, stsPatch, client.Apply, client.ForceOwnership, client.FieldOwner(FieldManager))
```
zachaller, line 59: What happens with updateStrategy: OnDelete? apps/v1 rejects rollingUpdate config for OnDelete, so this SSA patch 422s and every operation fails — while GetResourceStatus (partition defaults to 0) reports the workload as fully updated. Feels like validation should reject non-RollingUpdate StatefulSets up front.

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,483 @@
...
+		"oldPartition": oldPartition,
+		"podsToDelete": podsToDelete,
+	}).Info("Deleting updated pods one-by-one to force graceful rollback")
+
+	deletedCount := int32(0)
+	failedDeletes := []string{}
+
+	// Delete in reverse order (replicas-1 down to oldPartition) for graceful rollback which matches StatefulSet's natural ordering
+	for i := replicas - 1; i >= oldPartition; i-- {
+		podName := fmt.Sprintf("%s-%d", sts.Name, i)
```
zachaller, line 355: This assumes ordinals run 0..replicas-1. With spec.ordinals.start set (say 100), the partition math breaks everywhere — any weight updates all pods at once (every real ordinal >= partition), VerifyWeight still passes, and this loop deletes pod names that don't exist, counts the NotFounds as deleted, then waits 600s each for pods that will never appear. Either offset by sts.Spec.Ordinals.Start throughout or reject such StatefulSets in validation.

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,483 @@
...
+		replicas = *sts.Spec.Replicas
+	}
+
+	partition := int32(0)
+	if sts.Spec.UpdateStrategy.RollingUpdate != nil && sts.Spec.UpdateStrategy.RollingUpdate.Partition != nil {
+		partition = *sts.Spec.UpdateStrategy.RollingUpdate.Partition
+	}
+
+	// In StatefulSets, pods with ordinal >= partition are at new version
+	updatedReplicas := replicas - partition
```
zachaller, line 159: Why derive this from the partition instead of using sts.Status.UpdatedReplicas? The moment we patch the partition this reports the update as done, before any pod has rolled — and it goes negative if the STS is scaled down while partition > replicas. Same theme below: availableReplicas is fed from ReadyReplicas (ignores minReadySeconds), and Ready never checks status.observedGeneration == generation, so decisions can be made on stale STS status.

rolloutplugin/analysis.go
```diff
@@ -0,0 +1,513 @@
...
+	limitFailedArs := defaults.DefaultAnalysisRunUnsuccessfulHistoryLimit
+	if rp.Spec.Analysis != nil {
+		if rp.Spec.Analysis.SuccessfulRunHistoryLimit != nil {
+			limitSucceedArs = *rp.Spec.Analysis.SuccessfulRunHistoryLimit
+		}
+		if rp.Spec.Analysis.UnsuccessfulRunHistoryLimit != nil {
+			limitFailedArs = *rp.Spec.Analysis.UnsuccessfulRunHistoryLimit
+		}
+	}
+	arsToDelete := analysisutil.FilterAnalysisRunsToDelete(otherArs, nil, limitSucceedArs, limitFailedArs)
```
zachaller, line 203: Passing nil ReplicaSets here breaks the GC: FilterAnalysisRunsToDelete keeps runs whose pod-template-hash matches a live RS, but our ARs never get that label — so every non-current AR is deleted immediately and successful/unsuccessfulRunHistoryLimit is never honored. A failed run that explains an abort is gone by the next reconcile. This filter needs a RolloutPlugin-aware variant.

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+			// the abort block below sets Aborted=true, and the next reconcile takes the
+			// `Restart && Aborted` path into processRestart.
+			logCtx.Info("Restart requested while abort is in-flight; deferring until abort completes")
+		} else {
+			logCtx.Warn("Restart rejected: rollout has not been aborted")
+			newStatus.Restart = false
+		}
+	}
+
+	if newStatus.PromoteFull {
```
zachaller, line 225: PromoteFull only gets cleared inside processCanaryRollout — if someone issues promote-full while the rollout is idle/fully promoted, the flag persists in status and silently full-promotes the NEXT revision, skipping every step and analysis gate. Should this clear the flag (or no-op with an event) when there's nothing to promote?

cmd/rollouts-controller/main.go
```diff
@@ -303,9 +348,202 @@ func newCommand() *cobra.Command {
...
+
+				// Set up the RolloutPlugin controller
+				if err = (&rolloutplugin.RolloutPluginReconciler{
+					Client:            mgr.GetClient(),
+					Scheme:            mgr.GetScheme(),
+					KubeClientset:     kubeClient,
+					ArgoProjClientset: argoprojClient,
+					DynamicClientset:  dynamicClient,
+					PluginManager:     pluginManager,
+					InstanceID:        instanceID,
```
zachaller, line 457: InstanceID is stored on the reconciler but never checked, and the manager cache has no instance-id filter — so two instance-id-partitioned installations (each with its own lease) will both reconcile every RolloutPlugin in the cluster. Every legacy controller filters via instanceIDTweakListFunc; this one needs the equivalent (cache label selector or a predicate).

cmd/rollouts-controller/main.go
```diff
@@ -303,9 +348,202 @@ func newCommand() *cobra.Command {
...
+					kubeClient,
+					metrics.MetricRolloutEventsTotal,
+					metrics.MetricNotificationFailedTotal,
+					metrics.MetricNotificationSuccessTotal,
+					metrics.MetricNotificationSend,
+					rolloutPluginApiFactory,
+				)
+
+				// Create NotificationController for RolloutPlugin for custom triggers
+				rolloutPluginNotificationsController := notificationcontroller.NewControllerWithNamespaceSupport(
```
zachaller, line 431: Two questions on the notifications wiring: (1) this always uses NewControllerWithNamespaceSupport, ignoring --self-service-notification-enabled — the standard path (controller.go) switches on that flag specifically because the namespace variant logs spurious 'trigger not configured' errors without self-service. (2) It's fed by the instance-id-filtered dynamic informer while the reconciler uses the unfiltered manager cache — with an instance-id set, objects get reconciled but their notifications are silently dropped. Intentional?

cmd/rollouts-controller/main.go
```diff
@@ -236,11 +270,24 @@ func newCommand() *cobra.Command {
...
+			// Create shared MetricsServer that will be used by all controllers
+			log.Info("Creating shared metrics server")
+			metricsServer := metrics.NewMetricsServer(metrics.ServerConfig{
+				Addr:                          fmt.Sprintf(listenAddr, metricsPort),
+				RolloutLister:                 tolerantinformer.NewTolerantRolloutInformer(dynamicInformerFactory).Lister(),
+				AnalysisRunLister:             tolerantinformer.NewTolerantAnalysisRunInformer(dynamicInformerFactory).Lister(),
+				AnalysisTemplateLister:        tolerantinformer.NewTolerantAnalysisTemplateInformer(dynamicInformerFactory).Lister(),
+				ClusterAnalysisTemplateLister: tolerantinformer.NewTolerantClusterAnalysisTemplateInformer(clusterDynamicInformerFactory).Lister(),
+				ExperimentLister:              tolerantinformer.NewTolerantExperimentInformer(dynamicInformerFactory).Lister(),
+				RolloutPluginLister:           tolerantinformer.NewTolerantRolloutPluginInformer(dynamicInformerFactory).Lister(),
```
zachaller, line 285: This registers the RolloutPlugin informer on the shared dynamic factory unconditionally — so a deployment that never enabled the feature (and doesn't have the CRD) gets an endlessly-erroring reflector once it becomes leader. Can the lister/collector wiring be gated on the rolloutplugin controller actually being enabled, so the feature is fully inert when off? (Also worth checking: analysis-only mode now gets Rollout/Experiment/RolloutPlugin listers it previously didn't, which starts watches and duplicate metric series from that instance.)

utils/conditions/rolloutplugin.go
```diff
@@ -0,0 +1,176 @@
...
+// RolloutPluginTimedOut checks if the RolloutPlugin has timed out based on the plugin config timeoutSeconds.
+func RolloutPluginTimedOut(rolloutPlugin *v1alpha1.RolloutPlugin, newStatus *v1alpha1.RolloutPluginStatus) bool {
+	condition := GetRolloutPluginCondition(*newStatus, v1alpha1.RolloutPluginConditionProgressing)
+	if condition == nil || condition.Reason == RolloutPluginAbortedReason || condition.Reason == RolloutPluginPausedReason {
+		return false
+	}
+
+	// Already timed out - if observedGeneration changed (spec was edited), allow re-evaluation.
+	if condition.Reason == RolloutPluginTimedOutReason {
+		if newStatus.ObservedGeneration != rolloutPlugin.Generation {
```
zachaller, line 145: This escape hatch is dead code — the controller sets newStatus.ObservedGeneration = Generation at the top of reconcile before calling this, so the comparison never fires and a TimedOut/Degraded rollout can't be recovered by editing the spec (raising timeoutSeconds does nothing; only a new revision escapes). Separately: LastUpdateTime is only refreshed at rollout start and pause/resume, never on step advance — so timeoutSeconds is a whole-rollout wall clock, not the 'no progress for N seconds' deadline Rollout users expect. 15 slow-but-progressing steps will trip it.

pkg/apis/rollouts/v1alpha1/rolloutplugin_types.go
```diff
@@ -0,0 +1,259 @@
...
+	RestartCount int32 `json:"restartCount,omitempty" protobuf:"varint,22,opt,name=restartCount"`
+
+	// RestartedAt indicates when the last restart occurred
+	// +optional
+	RestartedAt *metav1.Time `json:"restartedAt,omitempty" protobuf:"bytes,23,opt,name=restartedAt"`
+
+	// Restart when set to true will restart the rollout from step 0.
+	// Similar to status.Abort and status.PromoteFull, this is a one-shot trigger field.
+	// +optional
+	Restart bool `json:"restart,omitempty" protobuf:"varint,24,opt,name=restart"`
```
zachaller: Some API-design questions on using status as the command channel: the controller rewrites status wholesale each reconcile (merge patch, no optimistic lock), so a user's status.abort/restart patch can race a concurrent controller write and be silently dropped. And restart diverges from Rollout's spec.restartAt — as a status field it can't be expressed declaratively via GitOps at all. Also worth documenting: restart is only honored when aborted=true, which the comment doesn't mention.

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1300 @@
...
+
+	// Skip the patch when nothing changed at all. Without this, every reconcile writes an
+	// identical status, bumping the resourceVersion and re-triggering the self-watch — a
+	// self-sustaining reconcile loop. A genuine change (including a condition timestamp
+	// transition) makes DeepEqual false, so it is still patched.
+	if apiequality.Semantic.DeepEqual(&rolloutPlugin.Status, newStatus) {
+		return nil
+	}
+
+	patch := client.MergeFrom(rolloutPlugin.DeepCopy())
```
zachaller, line 1075: MergeFrom without optimistic locking on a status that doubles as the command channel: a reconcile working from a stale read can resurrect pauseConditions the user just cleared, swallowing their promote. MergeFromWithOptimisticLock (and retry on conflict) seems safer here.

test/e2e/rolloutplugin_test.go
```diff
@@ -0,0 +1,1721 @@
...
+		Assert(func(t *fixtures.Then) {
+			rp := t.GetRolloutPlugin()
+			assert.Equal(s.T(), rov1.RolloutPluginPhaseHealthy, rp.Status.Phase)
+		})
+}
+
+// Abort Tests
+
+// TestRolloutPluginAbort tests aborting a rollout mid-canary.
+func (s *RolloutPluginSuite) TestRolloutPluginAbort() {
```
zachaller, line 401: All the abort tests (this one, AbortBeforePause, AbortIdempotent, AbortEvent) assert only RolloutPlugin status fields — none check the StatefulSet itself (partition reset, pods back on the old revision). A rollback that updates status and touches nothing would pass every one of these. Can we assert workload state on at least the main abort path? Same gap for the intermediate setWeight steps — partition is never checked at 25/75.

Makefile
```diff
@@ -265,18 +267,25 @@ test-kustomize: ## run kustomize tests
...
 
 .PHONY: test-e2e
 test-e2e: install-devtools-local
-	${DIST_DIR}/gotestsum --rerun-fails-report=rerunreport.txt --junitfile=junit-e2e-test.xml --format=testname --packages="./test/e2e" --rerun-fails=5 -- -timeout 60m -count 1 --tags e2e -p ${E2E_PARALLEL} -parallel ${E2E_PARALLEL} -v --short ./test/e2e ${E2E_TEST_OPTIONS}
+	${DIST_DIR}/gotestsum --rerun-fails-report=rerunreport.txt --junitfile=junit-e2e-test.xml --format=testname --packages="./test/e2e" --rerun-fails=5 -- -timeout 120m -count 1 --tags e2e -p ${E2E_PARALLEL} -parallel ${E2E_PARALLEL} -v --short ./test/e2e ${E2E_TEST_OPTIONS}
+
+.PHONY: test-e2e-rolloutplugin
+test-e2e-rolloutplugin: install-devtools-local ## run RolloutPlugin E2E tests only
+	${DIST_DIR}/gotestsum --rerun-fails-report=rerunreport.txt --junitfile=junit-e2e-rolloutplugin-test.xml --format=testname --packages="./test/e2e" --rerun-fails=5 -- -timeout 60m -count 1 --tags e2e -p ${E2E_PARALLEL} -parallel ${E2E_PARALLEL} -v --short ./test/e2e -run 'TestRolloutPluginSuite' $${E2E_TEST_OPTIONS}
```
zachaller, line 288: Two issues: the -run 'TestRolloutPluginSuite' regex doesn't match TestRolloutPluginRpcSuite, and that suite silently skips when the plugin isn't in the configmap — so the external/RPC path never runs in any automated context. And $${E2E_TEST_OPTIONS} expands the shell variable, not the make variable, so `make test-e2e-rolloutplugin E2E_TEST_OPTIONS=...` silently drops the options.

utils/log/log.go
```diff
@@ -22,6 +22,8 @@ const (
 	ExperimentKey = "experiment"
 	// AnalysisRunKey defines the key for the analysisrun field
 	AnalysisRunKey = "analysisrun"
+	// RolloutPluginKey defines the key for the rolloutplugin field
+	RolloutPluginKey = "rolloutplugin"
```
zachaller, line 26: The constant got added but WithObject's kind-fallback switch and KindNamespaceName below weren't extended for RolloutPlugin. Objects from the manager cache have empty TypeMeta, so every RolloutPlugin event log gets an empty-string field key, and the kind=="RolloutPlugin" branch in record.go never fires — which also means rolloutplugin events are counted under rollout_events_total (and the new rolloutplugin_events_total metric is registered but never incremented).

zachaller (changes requested):

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1371 @@
...
+// These conditions are needed so that we won't accidentally report lack of progress for resumed rollouts
+// that were paused for longer than the plugin config timeoutSeconds.
+func checkPausedConditions(rolloutPlugin *v1alpha1.RolloutPlugin, newStatus *v1alpha1.RolloutPluginStatus, logCtx *log.Entry) error {
+	progCond := conditions.GetRolloutPluginCondition(*newStatus, v1alpha1.RolloutPluginConditionProgressing)
+	progCondPaused := progCond != nil && progCond.Reason == conditions.RolloutPluginPausedReason
+
+	// Paused state is determined by spec.Paused (manual) or PauseConditions (controller-set step pauses)
+	isPaused := rolloutPlugin.Spec.Paused || len(newStatus.PauseConditions) > 0
+	abortCondExists := progCond != nil && progCond.Reason == conditions.RolloutPluginAbortedReason
+	completedCondExists := conditions.GetRolloutPluginCondition(*newStatus, v1alpha1.RolloutPluginConditionCompleted) != nil
```
zachaller: This checks that the `Completed` condition *exists*, not that it's `True`. Every RolloutPlugin gets `Completed=True` on its first reconcile, and `calculateFinalRolloutPluginConditions` turns it into `Completed=False` (never removes it) once a new revision starts, so this is effectively always true and the L606 block never flips `Progressing` to Paused.

Consequences I reproduced with the fake client:
- **`spec.paused` doesn't pause anything.** Setting it mid-rollout, the controller still called `SetWeight(50)` and `SetWeight(75)` and ran through all steps while reporting `phase: Paused`.
- **Paused time counts toward `timeoutSeconds`.** Sitting at a `pause: {}` step for 11m → `Degraded / ProgressDeadlineExceeded` (and an abort with `timeoutAbort: true`).

`TestRolloutPluginManualPause` passes because it only asserts the phase, never that steps stopped advancing.

Heads up that the one-line fix (`&& Status == True`) isn't enough on its own — see the next comment.

rolloutplugin/controller.go
```diff
@@ -0,0 +1,1371 @@
...
+		newStatus.PromoteFull = false
+		newStatus.Message = "No rollout in progress to promote"
+		if r.Recorder != nil {
+			r.Recorder.Eventf(rolloutPlugin, record.EventOptions{
+				EventReason: "PromoteFullNoOp",
+			}, newStatus.Message)
+		}
+	}
+
+	if conditions.IsRolloutPluginProgressing(newStatus) {
```
zachaller: Using the `Progressing` condition as the gate for whether `processCanaryRollout` runs at all means that once pausing *does* flip it to Unknown, nothing evaluates the pause any more. I tried the one-line fix above: manual pause and the timeout were fixed, but a `pause: {duration: 5s}` step never completed — still at step 1 after 10s, `RequeueAfter: 0`, because `CompletedCanaryPauseStep` and the timed requeue only live inside `processCanaryRollout`, and the L494 Paused guard returns before it.

The Rollout controller keeps running the step/pause reconciliation while paused and only uses the condition for reporting. Could we do the same here, i.e. always run the step logic, and have it no-op on steps that mutate the workload when `spec.paused` / pause conditions are set, while still evaluating pause durations and requeueing? I'd also like to see unit tests that walk (a) a timed pause, (b) an indefinite pause + promote, and (c) `spec.paused` mid-rollout across several reconciles, asserting the step index and the plugin calls, not just the phase.

rolloutplugin/analysis.go
```diff
@@ -0,0 +1,547 @@
...
+	if currentAr == nil {
+		return true
+	}
+
+	// If the controller has already paused for an inconclusive result, we need to create a new AR to retry.
+	if rp.Status.ControllerPause &&
+		getRolloutPluginPauseCondition(rp, v1alpha1.PauseReasonCanaryPauseStep) == nil {
+		return currentAr.Status.Phase == v1alpha1.AnalysisPhaseInconclusive
+	}
+	return rp.Status.AbortedAt != nil
```
zachaller: `markAborted` sets `AbortedAt` (controller.go L160), but nothing clears it: not `processRestart` (L972), not the new-revision paths (L451 / L509). In the Rollout controller `CalculateAbortStatus` nils it as soon as the abort is lifted. Here it sticks for the life of the object, so this returns `true` for every current AR and a replacement is created each reconcile.

Repro (abort rev `b`, push rev `c`, analysis step with every AR completing):

| scenario | result |
|---|---|
| prior abort, every AR **Successful** | 11 ARs (`rp-c-1` … `rp-c-1.10`), stuck on the analysis step |
| prior abort, every AR **Failed** | 7 ARs, **never aborts** — each failure is replaced before it's evaluated |
| no prior abort (control) | 1 AR, all steps complete |

So after a single abort, every future analysis gate on that RolloutPlugin can never pass *and* silently ignores failures. Background analysis is affected the same way. Should `AbortedAt` be cleared alongside `Aborted` (restart + new revision), and could we add a test for abort → new revision → step analysis?

rolloutplugin/plugins/statefulset/plugin.go
```diff
@@ -0,0 +1,532 @@
...
+	p.logCtx.WithFields(log.Fields{
+		"expectedPartition": expectedPartition,
+		"actualPartition":   partition,
+		"expectedUpdated":   expectedUpdated,
+		"actualUpdated":     actualUpdated,
+		"readyReplicas":     sts.Status.ReadyReplicas,
+		"totalReplicas":     replicas,
+	}).Info("Weight verification")
+
+	verified := partition == expectedPartition && actualUpdated >= expectedUpdated
```
zachaller: `status.updatedReplicas` counts pods *created* on the update revision, regardless of readiness. With 4 replicas at `setWeight: 25` and `sts-3` in CrashLoopBackOff (3/4 ready), this returns `true`, even though `GetResourceStatus` on the same object reports `Ready: false`. The following pause timer / step analysis then starts against a canary that isn't serving.

The Rollout canary path waits on the canary RS's `availableReplicas` (respecting `minReadySeconds`) before completing a setWeight step. Could this count pods on `UpdateRevision` that are Available (the plugin already caches pods), and also require `observedGeneration == generation` like `GetResourceStatus` now does?

rolloutplugin/controller.go
zachaller: Following up after re-testing at `76159a71`: the Paused-reason guard (now L494) is effectively unreachable today, because `Progressing` never actually flips to Paused (see the new comment on L600: `completedCondExists` is always true). Step pauses only appear to work because `processCanaryRollout` handles them independently. If `completedCondExists` is fixed, this guard becomes reachable, and timed pauses then hang forever (see the comment on L551). Leaving this thread open until pausing is reworked and has multi-reconcile unit tests.

utils/plugin/types/types.go
```diff
@@ -25,6 +27,59 @@ func (e RpcError) HasError() bool {
...
+	Kind    string
+}
+
+// RpcResourcePlugin is the RPC interface for resource plugins used in RolloutPlugins.
+// It's nearly identical to the controller's ResourcePlugin interface, but:
+// - Returns RpcError instead of error (for RPC serialization)
+// - No context.Context (can't be serialized over RPC)
+// Built-in plugins implement the controller interface directly.
+// External RPC plugins implement this interface.
+type RpcResourcePlugin interface {
```
zachaller: Right now there are two places plugin config could live, and neither quite works:

- **ConfigMap `args`:** the `rolloutPlugins` entries in `argo-rollouts-config` pass `args` to the plugin process at startup. That config is controller-wide, shared by every RolloutPlugin.
- **`spec.plugin.config`:** this is on the CRD, but it never reaches the plugin. Every method only gets `(namespace, workloadRef)`, and the only code that reads `config` is the controller's own `timeoutSeconds` / `timeoutAbort` parsing (`defaults.go` L401, `validation_rolloutplugin.go` L43).

What was the intent for per-RolloutPlugin config?

1. **If plugins should get config per RolloutPlugin**, which is how traffic router, metric and step plugins all work, could the methods take `*v1alpha1.RolloutPlugin` instead of `(namespace, workloadRef)`? That's the same as traffic routers taking the whole `Rollout`. It carries namespace, workloadRef and config, and later spec fields won't need interface changes.
2. **If ConfigMap `args` are meant to be enough for now**, could we drop `spec.plugin.config` until there's a use for it, so users don't set values that are silently ignored?

Either way, I think the timeouts should move out of `plugin.config` into first-class spec fields, e.g. `progressDeadlineSeconds` / `progressDeadlineAbort` like Rollout. The controller shouldn't be reading its own settings out of the plugin's config blob, and today a malformed `config` is silently ignored in `defaults.go`.

Related: `status.pluginStatus` is declared but never read or written. Should it be wired up, or removed for now?

I'd like this settled before merge, because changing the Go interface later is a compile-time break for external plugins.

Hariharasuthan99: Hello @zachaller The plugin config is intended to be config values that a plugin might make use of per RolloutPlugin object and not a global config for all RolloutPlugin objects. The current implementation of the statefulset plugin does not require any config values so only the timeoutSeconds was added. This timeoutSeconds was originally was not part of the plugin config and it was spec field called progressDeadlineSeconds for RolloutPlugin CRD itself but in our previous discussions we decided to rename it(since progressDeadlineSeconds applies only to Deployment k8s object) and move it to stateful plugin.
I agree controller using values from plugin config is not the correct approach, I will move timeoutSeconds to CRD.
Plugin config validation is done as part of validation_rolloutplugin.go lines 43-51 and adds a RolloutPluginConditionInvalidSpec condition.

status.pluginStatus is not used by the current statefulset plugin but it is something I thought could be good to have if in future other plugins want to persist some details.

zachaller: I think both config spots are probably valid, but we need to make sure we get the RPC interface correct since changing those later are breaking changes for all plugins.

Issue #4128, opened 2025-02-12
Rollout Stuck Indefinitely When New Pods Fail to Come Up – Abort and Progress Deadline Not Triggering


revandarth:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

When deploying a Canary Rollout (using workloadRef) in Argo Rollouts v1.8.0, if the new version's pods fail to come up (e.g. due to ImagePullBackOff), the rollout remains stuck in a "Progressing" state indefinitely. Even with `progressDeadlineSeconds: 600` and `progressDeadlineAbort: true` configured, the rollout never marks itself as failed and manual abort commands (`kubectl argo rollouts abort <rollout-name>`) do not take effect unless the failing ReplicaSet is manually deleted. This blocks any subsequent updates, as the rollout remains stuck with the failing new ReplicaSet.

**To Reproduce**

1. Create a Canary Rollout using `workloadRef` to an existing Deployment.
2. Update the rollout to use a new image that is misconfigured or non-existent so that pods fail (e.g., trigger ImagePullBackOff).
3. Confirm that the rollout status remains "Progressing" with a message like:

```
Name:            xapproxy
Namespace:       red
Status:          ◌ Progressing
Message:         more replicas need to be updated
Strategy:        Canary
  Step:          0/4
  SetWeight:     10
  ActualWeight:  0
Images:         dockerhub.io/istio/proxyv2:1.17.1 (canary, stable)
                dockerhub.io/myproxy:25.2.1.3-745.76dfaae (canary)
                dockerhub.io/myproxy:25.2.1.3-746.76dfaae (stable)
```

4. Run the command:
```bash
kubectl argo rollouts abort <rollout-name> -n <namespace>
```

and observe that the rollout does not transition to a failed/aborted state. 

5. As a workaround, executing argo undo/abort:
`kubectl argo rollouts undo <rollout-name> -n <namespace>` or `kubectl argo rollouts abort <rollout-name> -n <namespace>`

followed by manually deleting the failing ReplicaSet (e.g., `kubectl delete rs <failing-replicaset> -n <namespace>`) is required before any new version updates are applied.

**Expected behavior**

- The rollout controller should detect that the new ReplicaSet’s pods are not becoming Ready within the defined progress deadline and automatically abort or mark the rollout as failed.
- The abort command should work regardless of the pod state, allowing the rollout to revert to the stable ReplicaSet and enabling new updates.
[...]

Issue #4390, opened 2025-08-04
If stable is not ready a new canary can cause traffic to be still on old RS while its being scaled down


n1koo:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

When you are in the middle of a release with `Canary` (eg 70% stable, 30% canary) but have another release overwriting that `argo-rollouts` tries to start over with the new version (eg 95% stable, 5% canary)

During this flip to the new version we check the health of `stable` in [UpdateHash](https://github.com/argoproj/argo-rollouts/blob/master/rollout/trafficrouting.go#L287) but it has a specific case [where it doesn't return an error](https://github.com/argoproj/argo-rollouts/blob/master/rollout/trafficrouting/istio/istio.go#L333-L334) but also didn't do what we thought it did.

This causes us to continue in `trafficrouting` and starting the scale down with `reconcileOtherReplicaSets()`. 

This causes us to end up in a situation where:

- DestinationRule is still pointing to old-canary
- old-canary scale down is triggered
- new-canary is healthy but not getting traffic
- after scale down is done, 503/UH due to no endpoints being left

This will get eventually fixed, eg as stable RS recovers.

**To Reproduce**

- Trigger a deployment with `Istio` and `Canary` mode. Use `subset` for traffic routing
- Let it run to a phase where roll-out is mid way (eg. 70/30 split)
- Continuously cause stable RS to be not fully up by eg deleting pods randomly in a loop
- Trigger a new deployment

This should show logs like
- `New weights: &TrafficWeights{Canary:WeightDestination{Weight:5,ServiceName:,PodTemplateHash:<new-canary-rs>,},Stable:WeightDestination{Weight:95,ServiceName:,PodTemplateHash:<stable-rs,},Additional:[]WeightDestination{},Verified:nil,}`
- `Previous weights: &TrafficWeights{Canary:WeightDestination{Weight:70,ServiceName:,PodTemplateHash:<old-canary-rs>,},Stable:WeightDestination{Weight:30,ServiceName:,PodTemplateHash:<stable-rs>,},Additional:[]WeightDestination{},Verified:nil,}`
- `delaying destination rule switch: ReplicaSet <stable-rs> not fully available`

And key here is that we do not see:

`DestinationRule <rule> subset updated (canary: <new-rs>, stable: <stable-rs>)`

**Expected behavior**

Theres two routes we could take

[...]

PR #4612, opened 2026-02-11
fix(trafficrouting): ensure DestinationRule is updated before SetWeight on rollback

andrewjamesbrown:
We have experienced several incidents where we have a canary in-progress, and we want to do a rollback to a previous stable release that is within the rollback window (i.e. fast rollback).  What we experience is that Argo Rollouts will delay the DestinationRule switch (since the target is not ready), but still set the weight to the final weight in the `steps` list, resulting in a canary getting 70% of traffic in our case, while it is only scaled to handle 5%.  This creates an overload condition on the canary, spiking latency and impact to any clients calling the service.

The fix:
  - Return error from UpdateHash when delaying DR switch so SetWeight is not called, preventing VS weight update before DR subset switch (fixes rollback traffic hitting wrong subsets).
  - When shifting traffic to stable only (abort or dynamic rollback), only require stable RS availability so abort still completes when canary is failed (no regression of #4128).

Related PRs: 
https://github.com/argoproj/argo-rollouts/pull/4299
https://github.com/argoproj/argo-rollouts/pull/3602

Checklist:


andrewjamesbrown:
Possibly related to https://github.com/argoproj/argo-rollouts/issues/4390 ?

kostis-codefresh:
I am ok with this change on a technical level. 

Regarding the code organization, however, we overload the updateHash method with different types of outcomes. Somebody would expect that updateHash either does what it says on the tin OR fails because of an actual error. But now we overload it also to say "I am not ready". And we need to look at the log line to understand what actually happened.

Do we envision this change to be Istio only? If no, then maybe we could change the traffic routing interface to have a new "areweready" method or similar. If the problem is only evident with destination rules, then the code is ok as it is.

andrewjamesbrown:
> I am ok with this change on a technical level.
> 
> Regarding the code organization, however, we overload the updateHash method with different types of outcomes. Somebody would expect that updateHash either does what it says on the tin OR fails because of an actual error. But now we overload it also to say "I am not ready". And we need to look at the log line to understand what actually happened.
> 
> Do we envision this change to be Istio only? If no, then maybe we could change the traffic routing interface to have a new "areweready" method or similar. If the problem is only evident with destination rules, then the code is ok as it is.

@kostis-codefresh at this point I would say it's Istio only, but my organization only uses Istio so it's where we've discovered this defect.  I did a preliminary refactoring to introduce a "areweready" method and it seems straight-forward - it has the added benefit of doing a `return nil` in the reconciler versus `return err` and a re-queue.  Happy to switch to that if you think it's warranted here, but I also want to see this shipped as well.  We've been running this fix in production for a couple of weeks (against 1.8.3, not 1.9) and we haven't seen any stalls etc. or related HTTP 503 errors returned by Istio.

zachaller (approved):

Issue #4073, opened 2025-01-24
Rollout stuck - persistent update conflict


risentveber:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

Rollout is stuck, due to persistent `the object has been modified error` during reconciliation.

There is race condition between rollouts informer watch and `writeBackToInformer`
https://github.com/argoproj/argo-rollouts/blob/v1.7.2/rollout/controller.go#L454.

**To Reproduce**

1. controller updates rollout - get version 1
2. somebody updates rollout - get version 2
3. informer watches version 2
4. writeBackToInformer rewrites version 1

Bingo - we have stale data in the informer.

Maybe related to https://github.com/argoproj/argo-rollouts/issues/3316

**Expected behavior**

not stuck

**Screenshots**

<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/64c0c05b-f971-4849-b344-c0983fef1a33" />

**Version**

1.7.2

**Logs**

only on screenshot

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

zachaller:
This matches the root cause addressed in [#4786](https://github.com/argoproj/argo-rollouts/pull/4786): `writeBackToInformer` could overwrite a fresher informer entry (from a concurrent user or watch update) with state derived from a stale reconcile, leaving the Rollout informer cache permanently wrong.

The fix removes direct informer store writes and requeues when the cache has not yet observed the controller's last write. [#4784](https://github.com/argoproj/argo-rollouts/pull/4784) complements this by exiting reconciliation early after mid-pass Rollout writes.

Both are targeted for **v1.10**. Closing as fixed; if you still see this on **v1.10**, please re-open with version details and logs.

PR #4624, opened 2026-02-23
fix(controller): Skip write-back to informer when informer resourceVersion is newer

jonathan-innis:
Checklist:

### DIFF
--- a/rollout/bluegreen_test.go
+++ b/rollout/bluegreen_test.go
@@ -955,7 +955,7 @@ func TestBlueGreenRolloutStatusHPAStatusFieldsNoActiveSelector(t *testing.T) {
 	f.replicaSetLister = append(f.replicaSetLister, rs)
 	f.serviceLister = append(f.serviceLister, activeSvc)
 
-	ctrl, _, _ := f.newController(noResyncPeriodFunc)
+	ctrl, _, _, _ := f.newController(noResyncPeriodFunc)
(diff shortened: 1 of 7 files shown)


kostis-codefresh:
Have you run this change in your organization internally? Do you find that it solves https://github.com/argoproj/argo-rollouts/issues/4073 and https://github.com/argoproj/argo-rollouts/issues/3316 ?

Also might be related to https://github.com/argoproj/argo-rollouts/pull/4324#pullrequestreview-2999282524

jonathan-innis:
> Have you run this change in your organization internally

We haven't run this internally yet, but I could intuit that this would help. The issue that we _have_ seen in production is hanging on reconciliation because there was effectively a missed update due to the write-back. 

I can intuit that if we skip updating and trigger a requeue due to an error, that would do a lot towards making it less likely that things just skip updates and hang in prod

kostis-codefresh:
@jonathan-innis Did you stop working on this? Why was it closed?

jonathan-innis:
> @jonathan-innis Did you stop working on this? Why was it closed?

@kostis-codefresh I closed this because this solution ended up still having a race condition in it (the informer gets updated after the check is performed). I created another PR (#4674) which solves for the race condition that this one didn't. We're running this fix in production now and haven't seen the issue occur since it was shipped out.

PR #4629, opened 2026-02-24
fix(dashboard): Show page size selector in table view

gheine:
## Summary
- Enable the page size selector dropdown in the rollouts table view by adding `showSizeChanger: true` to the Ant Design Table pagination config
- Users can now choose between 10, 20, 50, and 100 items per page

## Test plan
- [ ] Open the Argo Rollouts dashboard in table view
- [ ] Verify the page size dropdown is visible in the pagination controls
- [ ] Change the page size and verify the table updates accordingly


kostis-codefresh:
Local testing results

Before:

before-table-view

After this change

after-table-view

Drop-down is also available

after-dropdown-open

kostis-codefresh (approved):
Many thanks for this!

zachaller (approved):

PR #4635, opened 2026-02-28
feat(analysis): add caPath support for Prometheus TLS with custom CA

### LINKED ISSUE #2298: Prometheus metrics should support self-signed certs.
Many metrics providers that are locally hosted on the same Kubernetes cluster use internal, self-signed certs. This can throw errors when trying to use the available https endpoints. We should add support for self-signed certs. 

https://github.com/argoproj/argo-rollouts/blob/4d8fbd151e10bb7c08fe6a573f65857cb91ec330/metricproviders/prometheus/prometheus.go#L146-L173

mark-liu:
## Summary

Add a `caPath` field to `PrometheusMetric` that allows specifying a PEM-encoded CA certificate file for TLS verification when querying Prometheus. This enables connecting to Prometheus instances that use self-signed or private CA certificates — common in Kubernetes clusters using cert-manager or internal PKI.

Fixes #2298

## Changes

- Add `CAPath` field to `PrometheusMetric` struct (protobuf field 8)
- `buildRoundTripper()` reads the CA cert and configures a custom root CA pool
- `Insecure` takes precedence over `caPath` — if both are set, TLS verification is skipped entirely
- Documentation in `docs/analysis/prometheus.md` with example and precedence note
- 5 new test cases covering: default behaviour, insecure skip, custom CA, missing cert, invalid cert
- All generated files updated (`generated.pb.go`, `generated.proto`, `openapi_generated.go`, CRD YAMLs, JSON schema)

## Design decisions

**File path vs inline cert**: This PR uses a file path (`caPath`) rather than inline cert content. A file-based approach is consistent with how the Web metric provider handles TLS, and avoids putting certificate data in CRD specs. An inline `caCert` field or Secret reference could be added in a follow-up if there is demand — file path covers the primary use case (volume-mounted CA certs from cert-manager or ConfigMaps).

**Protobuf field numbering**: `CAPath` uses field number 8 (next available). Existing fields `Headers` (6) and `RangeQuery` (7) are unchanged to preserve wire-format compatibility.

## Test plan

- `TestBuildRoundTripperDefaults` — no TLS config returns shared secure transport
- `TestNewPrometheusAPITLSInsecure` — `insecure: true` connects to self-signed server
- `TestNewPrometheusAPITLSSelfSignedFails` — default config rejects self-signed cert
- `TestNewPrometheusAPITLSWithCAPath` — custom CA cert allows connection
- `TestNewPrometheusAPITLSWithMissingCAPath` — missing cert file returns error
- `TestNewPrometheusAPITLSWithInvalidCAPath` — invalid PEM returns error


mark-liu:
Friendly ping — adds `caPath` support for Prometheus TLS analysis so custom CAs work without disabling verification. CI is all green. Happy to address any feedback.

mark-liu:
Friendly ping — this adds `caPath` support for custom CA bundles in Prometheus analysis providers. Happy to rebase or address feedback.

mark-liu:
Hi @kostis-codefresh — saw the `needs-follow-up` label go on a few weeks back. CI is green and there's no review feedback in the thread, so I'm not sure what the follow-up is. Could you point me at what you'd like changed? Happy to address.

kostis-codefresh:
@mark-liu Have you tested this PR locally on your workstation with a real Prometheus instance?

PR #4645, opened 2026-03-12
fix(controller): honor scaleDownDelaySeconds before scaling down the intermediate RS

andrewjamesbrown:
We are experiencing a problem with our use of Istio and Argo Rollouts.  During an "interrupted canary", we see:
- Set weight to 0 for all Istio VirtualServices
- Update DestinationRule to point to new canary
- Scale down intermediate replicaset
- Set weight to 0 for all Istio VirtualServices (this is done twice in 1.9 to avoid setting the weight prior to the DR - see https://github.com/argoproj/argo-rollouts/pull/4639 and https://github.com/argoproj/argo-rollouts/pull/4564

The issue occurs because the Istio control plane is async - it takes time for the configuration to sync to all the proxies, and during this time, we've started to scale down the intermediate replicaset.  In this case, it's desirable to delay the scale-down of the intermediate replicaset until the Istio data plane is in sync.

We see this on a 1.9-equivalent release that we build internally.

Checklist:


ajithshah (approved):

Hariharasuthan99 (approved):
LGTM!

Hariharasuthan99:
Hello @andrewjamesbrown, I see after you merged master, one of the unit test has failed, could you check and see if you could fix it?

kostis-codefresh (changes requested):
I have no major comments regarding the code just a question for backwards compatibiility.

The original purpose of `dynamicStableScale`  was for cost reasons where users
don't want to keep too many pods active.

But if this PR is merged, then canary users that have enabled `dynamicStableScale` will see
out of the blue the intermediate RS still running for 30 seconds (default value for `scaleDownDelaySeconds`) in interrupted canaries even though they have explicitly defined that they don't want any extra pods. So their workflow will change simply by upgrading to a new version of Rollouts.

Not only that, but if they try to add `scaleDownDelaySeconds` to 0, it turns out that[ existing validation doesn't allow](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/validation/validation.go#L301) both `scaleDownDelaySeconds` and `dynamicStableScale` to be active. 

Do I get that correctly?

Also it turns out that there is a similar PR at https://github.com/argoproj/argo-rollouts/pull/4925 It has a quick fix that drops the validation and allows people to define **both** `scaleDownDelaySeconds` and `dynamicStableScale`. But I am not sure if this is the correct solution either. It will open a can of worms as it doesn't make sense to have both of them enabled at the same time (for the happy path scenario).

I think the bigger question to ask here is if `scaleDownDelaySeconds` needs to be overloaded with this responsibility. Maybe we need a brand new field here?

I see three ways forward with this PR

1) Merge it as is and note in the release notes the behavior change. Accept the risk of breaking existing users

2) Drop the validation of both fields together. Merge and then accept the future risk of opening a can of worms. Users will set `scaleDownDelaySeconds` expecting it to apply even in the happy path scenario and it won't.

3) Create a brand new variable (needs design and discussion) specifically about intermediate/interrupted canaries.

@zachaller any thoughts?

PR #4648, opened 2026-03-14
feat: Add optional namespace support for secretKeyRef in Analysis argument

### LINKED ISSUE #4611: Allow a central secret to be referenced by (Cluster)AnalysisTemplates
# Summary

Consider the following analysis template:

```yaml
kind: ClusterAnalysisTemplate
metadata:
  name: secret-arg-example
spec:
  args:
  - name: api-token
    valueFrom:
      secretKeyRef:
        name: token-secret
        key: apiToken
  metrics:
  - name: webmetric
    provider:
      web:
        url: https://example.com/
        headers:
          - key: Authorization
```

Although the `ClusterAnalysisTemplate` is a centrally managed resource, the secret it references, `token-secret`, must currently be distributed to all namespaces that invoke that template. This defeats some of the central management value of the ClusterAnalysisTemplate and still leaves an operational burden of distributing the secret across all namespaces for the operator.

To make centralization more useful, `secretKeyRef` could be improved to support referencing a namespace preconfigured and populated by a rollout admin with a shared secret. e.g.:

```yaml
    valueFrom:
      secretKeyRef:
        name: token-secret
        namespace: shared-namespace
        key: apiToken
```

Requirements:
* This should be opt-in, disabled by default
* The permitted namespace(s) should be explicitly specified by the rollout admin. e.g. we would not want to default to `argo-rollouts` namespace, but rather one they specified.

# Use Cases

Currently, to use a secret in ClusterAnalysisTemplate, you must distribute that secret to all the namespaces that need it. Distribution is not only an operational burden, but also a security concern. e.g. you may wish to allow for the secret to be used by your end users, but not seen by them (who may have full access to their own namespace).

# Open Issues
* For job-based ClusterAnalysisTemplates, there is no getting around having the secret exist locally in the namespace, since the secret needs to be mounted by the Pod. I consider this out of scope for this request.

---

Mujib-Ahasan:
## Description
This PR adds support for referencing secrets from explicitly allowed namespaces when resolving `secretKeyRef` in Analysis arguments.
so currently, secrets referenced by `ClusterAnalysisTemplate` must be copied into every namespace where the template is used. This PR allows referencing a shared secret stored in an admin-controlled namespace instead. This PR also takes care of backward compatibility.

### changes
- CRDs have been generated
- Added an optional `namespace` field to `SecretKeyRe`
-  Implemented cross-namespace secret resolution in resolveArgs()
- Added controller flags:  
      - `--enable-cross-namespace-secret-refs`
      - `--allowed-secret-ref-namespaces`

- Added validation to ensure cross-namespace secret access is: disabled by default (backward compatibilty)
- restricted to explicitly allowed namespace only

fixes: #4611 

Checklist:


Mujib-Ahasan:
I will write the unit tests shortly.

Mujib-Ahasan:
test cases have been added! Ready for the review.

zachaller:
I am not sure I love the security implications of this, the usage also makes it easy to say pull secrets from any listed namespace in the allowed list so in say multi team setups if you allow namespace a and b but you want to not let b see a you can't currently in this PR.

I wonder if a middle ground would be to just do something under the secret ref like controllerNamespace: true that then pulls secrets from the namespace the controller is in, this just make it seem more limited to a single namespace?

I do see the issue with ClusterAnalysis needing secrets though, just not sure which way is the best option yet.

kostis-codefresh:
The PR is also missing any kind of documentation. Not only it needs documentation, but also we should place some big red warnings about the security implications that Zach already highlighted.

Since this is disabled by default and admins need to enable it, I think it is understandable that they choose the risk. But they should at least know about the risk from the docs.

Also the messages "cross-namespace secret references are disabled" and "cross-namespace secret references to namespace %q are not allowed" are not very descriptive. It would be better to explain to the user where the error comes from (the analysis) and what causes it (the namespace property there).

I don't see any e2e tests. Maybe we need some? Or do the unit tests cover everything?

Mujib-Ahasan:
> The PR is also missing any kind of documentation. Not only it needs documentation, but also we should place some big red warnings about the security implications that Zach already highlighted.
>
>Since this is disabled by default and admins need to enable it, I think it is understandable that they choose the risk. But they should at least know about the risk from the docs.

I totally understand the risk. and I am gonna add them properly.

> Also the messages "cross-namespace secret references are disabled" and "cross-namespace secret references to namespace %q are not allowed" are not very descriptive. It would be better to explain to the user where the error comes from (the analysis) and what causes it (the namespace property there).

Sure, I will update these shortly.🙂

> I don't see any e2e tests. Maybe we need some? Or do the unit tests cover everything?

probably, I need to take a deeper look.

Mujib-Ahasan:
I have updated the docs and also error messages. also IMO these unit test are sufficient for this changes because the behavior is focused in analysis args resolution than in a broader rollout flowing. Still, let me know if I need to work on it.

cc: @kostis-codefresh, @zachaller

zachaller:
I still think I want to scope it down to a single namespace not a list because it better aligns the k8s rbac with the behavior vs documenting it, because at that point if admins want someone to be able to see the secrests at the k8s rbac level they have to grant them access to the single namespace and they are not misslead in say team a can put secrets in namespace a and team b can use namespace b, if they just miss or ignore the fact that team b can pull teams a secrete via analysis.

Would a single namespace meet your requimrnets?

Mujib-Ahasan:
Makes sense, I am ok with either way. I think for single namespace we still have the copy secrets to a single namespace. Although the risk is less here but manual work is there. for multiple namespace, high risk and no manual work. I am confused😅. Even though the risks are documented, admins have to choose their security approaches. 
(should we ping the issue creator? ). I need a solid green signal.

zachaller:
> Makes sense, I am ok with either way. I think for single namespace we still have the copy secrets to a single namespace. Although the risk is less here but manual work is there. for multiple namespace, high risk and no manual work. I am confused😅. Even though the risks are documented, admins have to choose their security approaches. (should we ping the issue creator? ). I need a solid green signal.

Yea to me it's just more of an alignment with k8s rbac, I do get that we could document the use and put warning that all teams could pull secrets from any of the listed namespaces etc but that just becomes more apparent without reading docs when it is only allowed as a single namespace. It's also much easier to add multi namespace support later as a non-breaking change than it is to take it away becoming a breaking change. So, I feel its a good place to start single namespace expand to multi later if absolutely required and if it is maybe we allow some two way configuration saying x namespaces are allowed to pull from x secret namespace which is a much bigger coding task so lets kick that down the road and just do single namespace today.

Mujib-Ahasan:
just a gentle ping @zachaller, whenever you get time, would appreciate a review.

kostis-codefresh:
@Mujib-Ahasan Can you fix the failed checks please?

Mujib-Ahasan:
Hey @kostis-codefresh, I have updated the CRDs, still not sure why getting Go/Verify Codegen error.

Mujib-Ahasan:
@zachaller `Go / Verify Codegen` is failing. Is there any way I can fix this.??

kostis-codefresh (commented):
@Mujib-Ahasan Could you please look at the conflicts?

Mujib-Ahasan:
@kostis-codefresh I updated this PR. Please have a look!

PR #4674, opened 2026-03-27
fix(controller): Handle requeue when getting stale resource version

jonathan-innis:
Checklist:

## Description

We were experiencing issues, similar to the issues that were documented in https://github.com/argoproj/argo-rollouts/issues/3316#issuecomment-3802000786 where we would have a change be applied to the Rollout but the rollout informer would not react to it because the Rollout informer became out-of-date due to the `writeBackToInformer` behavior. In a nutshell, here's what we're fairly confident is happening:

1. A change is made to the Rollout to do something like unpause the rollout with the CLI, at the same time a separate Rollout reconciliation loop has already started
2. The rollout informer gets a watch event from the unpause done in the first step so the informer's view of the Rollout is updated
3. The Rollout controller makes some other change to the rollout and doesn't see the unpause that was just committed. The rollout controller then directly writes that change to the informer, which means that the change performed earlier was just missed.
4. Now, what we have is that the cache informer and the actual state of the world don't match and never will match until another event comes in to trigger the informer to update and the controller to run. The actual Rollout stored in etcd has the unpaused state that was committed by the user, but the cache informer had that change forcefully overridden by the controller. 

As was mentioned in the issue that I linked, the process of directly overwriting the informer is a pretty bad anti-pattern for the reasons that are shown across that issue -- you get inconsistent views of what the world looks like, not just delayed views.

The whole intention of this `writeBackToInformer` logic is to make sure that we see updates that we make to the apiserver (since there is no write-through caching that is done). A much safer alternative to achieve this is to mark down _that_ you made an update and then wait around for the cache to catch up to that update. You see this done with the [deployment-controller](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L273) for example.

This change branches off of that a bit because the amount of work that would be required to see what specific changes need to be done to make sure you can either: 1) Handle the eventual consistent state or 2) store the state that you need to store so that you can wait until you see the resource with that changes would be pretty substantial from what I can see. This change basically uses the resourceVersion and the fact that etcd generally monotonically increases its resourceVersions to wait until you see a newer resourceVersion in your cache before allowing reconciliation to continue.

This addresses the issue of not seeing the changes you have made while also making sure that we are making destructive changes to the write informer that can cause other issues.


Hariharasuthan99:
@jonathan-innis , could you add some description to the PR about the reasoning behind the change?

kostis-codefresh:
@jonathan-innis Have you been running this change in production over the last month?

jonathan-innis:
> @jonathan-innis Have you been running this change in production over the last month? 

Yep! No issue since we deployed it. I don't think it's necessarily perfect since you aren't guaranteed monotonically increasing resource versions, but the fallback for that is just you'll wait to requeue a bit longer

jonathan-innis:
@Hariharasuthan99 I wrote a pretty lengthy description -- let me know if the description doesn't make sense or you feel like I should make changes to address this a slightly different well. FWIW, this has completely mitigated our issue after we deployed this patch on our side

jonathan-innis:
I'd love to get traction on this if we could get this into the newest versions of argo

zachaller:
@jonathan-innis thank you so much for contributing and pointing this out I did kind of redo a lot of it in a different PR but this is a great addition.

PR #4678, opened 2026-03-29
docs: update installation doc to be arch/vendor neutral and remove MacOS quarantine bit

mifisignal:
Checklist:

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -53,7 +53,7 @@ brew install argoproj/tap/kubectl-argo-rollouts
 
 ### Using mise
 
-If you are an enthusiast user of [mise](https://github.com/jdx/mise) polyglot tool version manager, you can use a command like this :
+If you are an enthusiast user of [mise](https://github.com/jdx/mise) polyglot tool version manager, you can use a command like this:
 
 ```shell
 mise use -g argo-rollouts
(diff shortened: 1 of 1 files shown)


Hariharasuthan99 (approved):

kostis-codefresh:
Instead of using variable expansions and trying to make universal documentation that works across all OS, I think it is much easier to write sections per OS exactly like Argo CD does https://argo-cd.readthedocs.io/en/stable/cli_installation/

In this PR if I am a linux user I don't care about the MacOS quarantine bit. 

/usr/local/bin/ is also the correct choice for multi-user system. So .local/bin assumes silently that this is a single user installation. I am also not sure if .local/bin is in the Path for all systems.

I would split this page in specific sections for Mac and Linux instead of mixing those two.

docs/installation.md
```diff
@@ -63,25 +63,32 @@ mise use -g argo-rollouts
 
 1. Install [Argo Rollouts Kubectl plugin](https://github.com/argoproj/argo-rollouts/releases) with curl.
     ```shell
-    curl -LO https://github.com/argoproj/argo-rollouts/releases/latest/download/kubectl-argo-rollouts-darwin-amd64
+    os=$(uname -s | tr '[A-Z]' '[a-z]')
```
kostis-codefresh, line 66: Let's follow the Argo CD docs where each OS has a dedicated sections instead of trying to create generic documentation for everything https://argo-cd.readthedocs.io/en/latest/cli_installation/#mac

kostis-codefresh (changes requested):

PR #4679, opened 2026-03-31
fix(analysis): provide actionable error for empty metric results in AnalysisRun

### LINKED ISSUE #2101: Improve the error message for AnalysisRun
## Summary 

Today, when AnalysisRun fails it returns the message with the string "Error" to the user. We should improve Rollouts error handling to provide a better message.

Consider the success condition bellow:
```yaml
successCondition: 'result[0] <= {{args.cpu-utilization-limit-perc}}'
```

Sometimes the error is related to the metric success condition not being met but sometimes it is related to the metric not available causing result[0] == nil with `IndexOutofBound`. 

Rollout should be able to distinguish the 2 conditions and provide a friendly message to the user.

## Diagnostics

argo-rollouts 1.2.1

Kunalbehbud:
## Summary
When an AnalysisRun metric condition indexes into `result` (for example `result[0]`) and the provider returns an empty result set, Argo Rollouts currently surfaces a low-level evaluation error.

This change returns a more actionable message explaining that the metric result is empty or unavailable and that the condition should be guarded with `len(result) > 0`.

It adds unit coverage for this behavior in `utils/evaluate/evaluate_test.go` and `metricproviders/prometheus/prometheus_test.go`.

## Testing
- `make lint`
- `PATH="/home/kunal/Desktop/contribute/Argo/argo-rollouts/dist:$PATH" ./test/kustomize/test.sh`
- `go test ./...`

Fixes #2101

### DIFF
--- a/metricproviders/prometheus/prometheus_test.go
+++ b/metricproviders/prometheus/prometheus_test.go
@@ -614,6 +614,22 @@ func TestProcessInvalidResponse(t *testing.T) {
 
 }
 
+func TestProcessResponseWithEmptyVectorIndexedCondition(t *testing.T) {
+	logCtx := log.WithField("test", "test")
+	p := Provider{
+		logCtx: *logCtx,
+	}
+	metric := v1alpha1.Metric{
+		SuccessCondition: "result[0] <= 10",
+	}
+
+	value, status, err := p.processResponse(metric, model.Vector{})
+	assert.Equal(t, "[]", value)
+	assert.Equal(t, v1alpha1.AnalysisPhaseError, status)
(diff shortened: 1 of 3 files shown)


Kunalbehbud:
Quick follow-up here: after the coverage fix in df6b17108, the full run went green on my side. If the error wording still feels off, I am happy to iterate.

kostis-codefresh (changes requested):
Issue https://github.com/argoproj/argo-rollouts/issues/2101 was already solved by PR https://github.com/argoproj/argo-rollouts/pull/4680

Please reopen if think that your PR addresses something different.

PR #4689, opened 2026-04-12
fix: correct stale desired-replicas annotation after HPA removal

### LINKED ISSUE #4407: HPA removal leaves RS scaled by stale desired-replicas
# Summary

When an HPA that previously scaled a Rollout is removed (or disabled) and the Rollout is set back to a static .spec.replicas, the per-ReplicaSet annotation rollout.argoproj.io/desired-replicas can remain at the old HPA value. If the Rollout is paused (e.g., BlueGreen with autoPromotionEnabled: false) or otherwise not completing a full reconcile, affected ReplicaSets may stay at the stale replica count.

Proposed change (enhancement): When no HPA targets the Rollout, the controller should automatically reconcile each managed RS so its desired-replicas (and spec.replicas) align with the Rollout’s current desired scale—even while paused—respecting strategy fields like previewReplicaCount.

Workarounds

- Manual fix (works immediately): Patch the preview RS:.

# Use Cases

BlueGreen, paused: HPA scaled preview RS to 2; HPA is removed and Rollout set to replicas: 1; preview RS remains at 2 due to stale annotation. Auto-healing avoids manual patching and extra pods.

---

jaspreet-sirswal:
## Summary

Fixes #4407 — after an HPA is removed and the Rollout goes back to a static `spec.replicas`, the `rollout.argoproj.io/desired-replicas` annotation on the ReplicaSet keeps the old HPA-set value. `isScalingEvent()` then reports a scaling event on every sync while the ReplicaSet stays parked at the stale count.

**Root cause:** a ReplicaSet that is already at `spec.replicas` needs no scaling, so the strategy reconcilers leave it untouched — and while progress is halted they do not run at all. `reconcileCanaryReplicaSets()` returns early on `haltProgress()` (paused Rollout, or inconclusive analysis), so `scaleReplicaSetAndRecordEvent()`, which is what normally rewrites the annotation as part of scaling, is never reached. The stale annotation survives every reconcile.

**Fix:** `syncReplicasOnly()` now calls `syncStaleDesiredReplicasAnnotation()`, which realigns the annotation with `spec.replicas` on the newRS/stableRS that are already at the desired size. ReplicaSets whose size does not match are left alone — that is a genuine scaling event, and the scaling path already rewrites the annotation. `isScalingEvent()` stays a read-only predicate. Update failures are returned rather than swallowed, so the Rollout is requeued and the sync retried.

**Scope note:** I could not reproduce the BlueGreen half of the issue. `reconcileBlueGreenReplicaSets()` has no `haltProgress()` gate, so `reconcileBlueGreenStableReplicaSet()` → `scaleReplicaSetAndRecordEvent()` still rewrites the annotation while paused. The reproducible case is the halted canary path.

**Changed files:**
- `rollout/sync.go` — new `syncStaleDesiredReplicasAnnotation()`; `isScalingEvent()` restored to read-only
- `rollout/sync_test.go` — paused fixture and tests

## Test plan
- [x] Paused Rollout and inconclusive-analysis tests both fail on master and pass with the fix
- [x] Update-failure path surfaces the error instead of swallowing it
- [x] No-op when the ReplicaSet size differs, so genuine scaling events are untouched
- [x] `go test ./rollout/...` passes, including the existing scaling tests


jaspreet-sirswal:
@zachaller @leoluz Could one of you review this when you get a chance? All CI checks are green. This fixes #4407 — stale `desired-replicas` annotation left behind after HPA removal causing RS to stay at wrong replica count.

jaspreet-sirswal:
@zachaller Friendly bump — let me know if you'd like any changes.

jaspreet-sirswal:
@zachaller Friendly bump — all CI checks are green. Could you take a look when you have a moment?

jaspreet-sirswal:
@zachaller Friendly bump — all CI checks are green. Could you take a look when you have a moment?

pkg/kubectl-argo-rollouts/cmd/list/list_test.go
```diff
@@ -206,9 +206,10 @@ func TestListWithWatch(t *testing.T) {
 	watcher.Add(bg)
 	watcher.Add(can1copy)
 	watcher.Add(can2)
-	// Don't stop immediately to allow all events to be processed
+	// Don't stop immediately to allow all events to be processed.
+	// Use a longer sleep to avoid flakiness under CI load.
 	go func() {
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(1 * time.Second)
```
kostis-codefresh: Why did we increase the timeout here? Is this change related to the original issue in any way?

rollout/sync.go
```diff
@@ -332,7 +334,18 @@ func (c *rolloutContext) isScalingEvent() (bool, error) {
...
 			continue
 		}
-		if desired != defaults.GetReplicasOrDefault(c.rollout.Spec.Replicas) {
+		if desired != rolloutReplicas {
+			// Fix stale desired-replicas annotation left behind after HPA removal (#4407).
+			// Only correct when RS already has the right replica count but stale annotation.
+			// If the update fails, syncReplicasOnly will retry on next reconciliation.
+			if *(rs.Spec.Replicas) == rolloutReplicas && annotations.SetReplicasAnnotations(rs, rolloutReplicas) {
+				ctx := context.TODO()
+				if _, err := c.updateReplicaSet(ctx, rs); err != nil {
```
kostis-codefresh: Reading just the title of the method I would expect isScalingEvent to be a readonly function. Having side effects and "update" methods inside a function that is called isX is not a good practice IMHO

rollout/sync_test.go
```diff
@@ -875,3 +876,61 @@ func TestIsScalingEventMissMatchedDesiredOldReplicas(t *testing.T) {
 	assert.Equal(t, int32(13), roStatus.Status.ReadyReplicas)
 	assert.Equal(t, int32(0), roStatus.Status.UpdatedReplicas)
 }
+
+// setupStaleHPAAnnotationFixture creates a rollout (5 replicas) with a stable RS
+// whose desired-replicas annotation is stale (20, left from prior HPA scaling).
```
kostis-codefresh: I think the original issue talks about a Rollout that is paused. It this what we test here?

kostis-codefresh (changes requested):

pkg/kubectl-argo-rollouts/cmd/list/list_test.go
jaspreet-sirswal: @kostis-codefresh Not related to #4407. It was a drive-by deflake for `TestListWithWatch` that I pushed onto this branch to get CI green, and it got swept into the PR diff.

Reverted — `list_test.go` is back to master. I'll send it separately.

rollout/sync.go
jaspreet-sirswal: @kostis-codefresh Agreed, fixed. `isScalingEvent` is a predicate again with no writes in it.

The correction now lives in `syncStaleDesiredReplicasAnnotation`, called from `syncReplicasOnly` — already the write path for ReplicaSets.

For context on why the existing scaling path doesn't cover it: a ReplicaSet that is already at `spec.replicas` needs no scaling, so the strategy reconcilers leave it untouched, and `reconcileCanaryReplicaSets` returns early on `haltProgress()` while the rollout is paused, so they don't run at all. Either way the stale annotation survives every sync.

Also changed the update failure to be returned rather than logged and swallowed, so the rollout is requeued and the sync retried.

rollout/sync_test.go
jaspreet-sirswal: @kostis-codefresh You're right, it wasn't — the old fixture was an unpaused canary rollout, so it never exercised the paused case.

Fixed: the fixture now sets `spec.paused: true`, and I added a second test for the other `haltProgress()` branch (inconclusive analysis). Both fail on master and pass with the fix.

One thing worth flagging: I could not reproduce the BlueGreen half of the issue. `reconcileBlueGreenReplicaSets` has no `haltProgress()` gate, so `reconcileBlueGreenStableReplicaSet` → `scaleReplicaSetAndRecordEvent` still rewrites the annotation while paused. I wrote a BlueGreen test for it and dropped it after confirming it passed with the fix reverted. The reproducible case is the halted canary path.

jaspreet-sirswal:
@zachaller @kostis-codefresh  Friendly bump — all CI checks are green. Could you take a look when you have a moment?

jaspreet-sirswal:
@zachaller @kostis-codefresh @Hariharasuthan99 Friendly bump. Please take a look when you have a moment?

PR #4668, opened 2026-03-24
feat(dashboard): add token-based authentication support

matanbaruch:
## Summary

Adds optional authentication to the Argo Rollouts dashboard with **Bearer token** auth, **OIDC SSO** (Okta, Google, Azure AD, Keycloak, etc.), and **per-user Kubernetes RBAC enforcement**. Closes #1323.

### Features

- **`--auth-mode token`** — requires a valid Kubernetes bearer token for API access
- **OIDC SSO** — browser-based login flow via `--oidc-issuer-url`, `--oidc-client-id`, etc.
- **Per-user RBAC** — in token auth mode, every K8s API call uses the user's own token, so Kubernetes RBAC policies are enforced per user (e.g., a read-only user cannot promote/abort rollouts)
- Token validation via Kubernetes `TokenReview` API with caching
- CSRF protection for OIDC flow
- Default behavior (`--auth-mode server`) is unchanged — fully backward compatible

### Usage

```bash
# No auth (default, backward compatible)
kubectl argo rollouts dashboard

# Bearer token auth with per-user RBAC
kubectl argo rollouts dashboard --auth-mode token

# OIDC SSO (e.g., Okta)
kubectl argo rollouts dashboard --auth-mode token \
  --oidc-issuer-url https://your-org.okta.com \
  --oidc-client-id my-client-id \
  --oidc-client-secret my-client-secret \
  --oidc-redirect-url http://localhost:3100/rollouts/auth/callback
```

### How RBAC works

When `--auth-mode=token` is set, the server creates per-request Kubernetes clients using the user's bearer token. This means:
- A user with `get` permissions on rollouts can view but not modify them
- A user with `update` permissions can promote/abort/restart
- Namespace restrictions in RBAC are enforced naturally
- No custom policy engine needed — standard Kubernetes RBAC applies

### Files changed

| Area | Files | What |
|------|-------|------|
| Token auth | `server/auth.go` | HTTP middleware, gRPC interceptors, TokenReview validation, cache |
| OIDC SSO | `server/oidc.go` | OIDC discovery, login redirect, callback token exchange, CSRF |
| Server + RBAC | `server/server.go` | `AuthMode`, `RESTConfig` in `ServerOptions`, per-request client creation via `getClients()`, all handlers updated |
| CLI | `dashboard.go` | `--auth-mode`, `--oidc-*` flags, REST config passthrough |
[...]

PR #4690, opened 2026-04-12
feat(dashboard): add client auth mode for token-based authentication. Fixes #1323

### LINKED ISSUE #1323: Dashboard - Authentication
Sign in dashboard with token.

Refer to: https://github.com/argoproj/argo-workflows/issues/1813

matanbaruch:
Checklist:

## Summary

Adds a `--auth-mode` flag to the dashboard with two modes, following the same pattern as [Argo Workflows' client auth mode](https://argo-workflows.readthedocs.io/en/latest/argo-server-auth-mode/):

- **`server`** (default) — existing behavior, uses the server's own kubeconfig credentials
- **`client`** — requires a Kubernetes bearer token from the user; the server creates per-request Kubernetes clients using the user's token so native Kubernetes RBAC is enforced per-user

This is a scoped-down follow-up to #4668, containing **only the token-passing feature** as discussed in [#1323](https://github.com/argoproj/argo-rollouts/issues/1323#issuecomment-2403294810). No OIDC/SSO, no custom RBAC engine, no Dex integration — just passing the user's K8s token through to the API server.

### How it works

1. User starts the dashboard with `kubectl argo rollouts dashboard --auth-mode client`
2. The UI shows a login page where the user pastes their Kubernetes bearer token
3. The token is sent with all API requests via the `Authorization: Bearer` header
4. The server creates per-request Kubernetes clients using the user's token
5. Kubernetes RBAC is enforced naturally — a user with read-only permissions can view rollouts but cannot promote/abort them

### What changed

| Area | Files | What |
|------|-------|------|
| Server core | `server/server.go` | `AuthMode`/`RESTConfig` in `ServerOptions`, `getClients(ctx)` for per-request clients, HTTP middleware + gRPC interceptors, all handlers updated |
| CLI | `dashboard.go` | `--auth-mode` flag |
| Frontend | `auth.tsx`, `api.tsx`, `App.tsx`, `header.tsx`, `watch.ts` | Auth context, token-aware API client, login page, logout button, EventSource token passing |
| Tests | `server_test.go`, `dashboard_test.go` | Token extraction, middleware, client mode, CLI flags |
| Docs | `kubectl-argo-rollouts_dashboard.md` | Regenerated |

### Default behavior is unchanged

When `--auth-mode` is not specified (or set to `server`), the dashboard behaves exactly as before. No authentication is required and the server uses its own kubeconfig credentials.

## Test plan

- [x] `go build ./...` — project compiles
- [x] `go test ./server/...` — server tests pass (token extraction, middleware, client mode)
- [x] `go test ./pkg/kubectl-argo-rollouts/cmd/dashboard/...` — CLI tests pass (flag parsing, validation)
- [x] `cd ui && yarn build` — UI builds successfully
- [x] `go run ./hack/gen-docs/main.go` — docs regenerated
- [x] Manual: `--auth-mode client` shows login page, valid token grants access
- [x] Manual: invalid token shows error notification
- [x] Manual: read-only RBAC user cannot promote/abort (gets K8s 403)
- [x] Manual: logout returns to login page


matanbaruch:
@kostis-codefresh This is a brand new PR following up on your feedback in #4668. Here's what changed:

**Scope:** Only the token-passing feature - no OIDC/SSO, no custom RBAC, no Dex. This follows the same "client" auth mode pattern as [Argo Workflows](https://argo-workflows.readthedocs.io/en/latest/argo-server-auth-mode/), as suggested by @zachaller in https://github.com/argoproj/argo-rollouts/issues/1323#issuecomment-2403294810.

**How it works:** `--auth-mode client` makes the dashboard require a Kubernetes bearer token. The server creates per-request K8s clients using that token, so native Kubernetes RBAC is enforced per-user, no custom policy engine needed.

**Default behavior is unchanged** - without the flag, the dashboard works exactly as before.

I've followed the PR template checklist, signed commits with DCO, and written unit tests. Happy to address any review feedback.

server/server.go
```diff
@@ -80,6 +100,140 @@ const (
...
+
+// tokenFromHTTPRequest extracts a bearer token from an HTTP request.
+// It checks the Authorization header first, then falls back to the "token" query parameter
+// (needed for EventSource/SSE which cannot set custom headers).
+func tokenFromHTTPRequest(r *http.Request) string {
+	token := extractBearerToken(r.Header.Get("Authorization"))
+	if token != "" {
+		return token
+	}
+	return r.URL.Query().Get("token")
```
kostis-codefresh: I am not an expert on Argo Workflows, but it seems to me that Argo Workflows specifically moved away from using URL based parameters. Mentioned here https://github.com/argoproj/argo-workflows/issues/1949 and fixed here https://github.com/argoproj/argo-workflows/pull/2058

Why doesn't this PR follow the same approach?

server/server.go
```diff
@@ -80,6 +100,140 @@ const (
...
+		return ""
+	}
+	return extractBearerToken(authHeaders[0])
+}
+
+// getClients returns the appropriate Kubernetes clients for the current request.
+// In server mode, it returns the shared server clients.
+// In client mode, it creates per-request clients using the user's bearer token.
+func (s *ArgoRolloutsServer) getClients(ctx context.Context) (*serverClients, error) {
+	if s.Options.AuthMode != AuthModeClient || s.Options.RESTConfig == nil {
```
kostis-codefresh: I dont' understand the check about RestConfig here. Deciding if we are in client or server mode should be based solely on the Authmode option. Whether a restconfig is available or not is a side effect and not a decision.

Why was this conditional added here?

If we have the case where auth is client and restconfig is not available doesn't this result in a misconfigured client mode which becomes less secure?

server/server_test.go
```diff
@@ -83,16 +98,818 @@ func TestNewHTTPServer(t *testing.T) {
...
+		interceptor := s.newAuthStreamInterceptor()
+		md := metadata.Pairs("authorization", "Bearer valid-token")
+		ctx := metadata.NewIncomingContext(context.Background(), md)
+		stream := &mockServerStream{ctx: ctx}
+		err := interceptor(nil, stream, &grpc.StreamServerInfo{}, mockHandler)
+		assert.NoError(t, err)
+	})
+}
+
+func TestVersion(t *testing.T) {
```
kostis-codefresh: I am all for testing, but this seems a bit excessive. Does this test ever fail? The project should have tests that help with regression and not just increase test coverage in an inflated manner...

server/server_test.go
```diff
@@ -83,16 +98,818 @@ func TestNewHTTPServer(t *testing.T) {
...
+			AuthMode:   AuthModeClient,
+			RESTConfig: &rest.Config{Host: "https://localhost:6443"},
+		},
+	}
+	_, err := s.GetNamespace(context.Background(), &empty.Empty{})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "missing bearer token")
+}
+
+func TestRolloutToRolloutInfoClientModeNoToken(t *testing.T) {
```
kostis-codefresh: I am confused about what exactly this test covers? Also it seems to me that RolloutToRolloutInfo is used only in this test and nowhere else. Is my understanding correct?

If there is a function that is not used anywhere, let's just remove it. No need to write tests about it.

ui/src/app/components/login/login.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
```
kostis-codefresh, line 1: There are no e2e tests for the UI. Argo Workflows seems to have tests that verify the token login process.

kostis-codefresh:
Apart from all  the code comments I added this PR is missing

1) E2e tests for the UI
2) User documentation that explains how the feature is used and more specifically the different ways of how a user can obtain a token to enter in the UI.

kostis-codefresh:
I tried to run the code locally for this PR and it simply doesn't work for me.

Entering a correct token just gives me an empty grey page


https://github.com/user-attachments/assets/1f321f85-0355-4dff-adc1-5c1e59a460ad

No error message, no log error of any kind.

Entering an invalid value just reloads the page. Again no error message


https://github.com/user-attachments/assets/d3011702-9203-4629-b231-962b0cafa0bd

Maybe it was a mistake on my end, and this is why I say that user documentation is important.
But even if all code comments are addressed the usability factor of a new feature is equally important.

ui/src/app/shared/context/auth.tsx
```diff
@@ -0,0 +1,66 @@
...
+    setAuthRequired: () => {},
+});
+
+export const AuthProvider = (props: {children: React.ReactNode}) => {
+    const [token, setTokenState] = React.useState<string | null>(localStorage.getItem(AUTH_TOKEN_KEY));
+    const [authRequired, setAuthRequired] = React.useState(false);
+
+    const setToken = (newToken: string | null) => {
+        if (newToken) {
+            localStorage.setItem(AUTH_TOKEN_KEY, newToken);
```
kostis-codefresh: Same question as the backend. Argo Workflows seems to be using cookies and not local storage. Why does this PR use local storage for authentication?

kostis-codefresh (changes requested):

server/server.go
matanbaruch: Good catch - switched to a cookie like argoproj/argo-workflows#2058, so the browser sends `Cookie: authorization=...`, API clients keep using the header, and the token is not read from the query string at all anymore.

server/server.go
matanbaruch: You are right, it was a decision made on a side effect and it degraded to the server's own credentials - client mode is now decided only by `AuthMode`, a missing REST config is an error instead of a fallback, and `dashboard.go` fails at startup rather than serving in that state.

server/server_test.go
matanbaruch: Removed - and the nine near-identical `TestXxxClientModeNoToken` functions are now one table-driven test while the six `TestXxxServerMode` ones assert the error actually maps to `NotFound` instead of just "some error happened", so the file is down from 915 to 877 lines.

server/server_test.go
matanbaruch: Correct, only that test used it - function and test both deleted.

ui/src/app/shared/context/auth.tsx
matanbaruch: Same fix as the backend comment - it is a `SameSite=Strict` session cookie scoped to `--root-path` now, local storage is gone, and that is also what lets EventSource authenticate without a token in the URL.

ui/src/app/components/login/login.tsx
matanbaruch: Added `ui/src/app/App.test.tsx` with jest + React Testing Library (both already dev deps) covering login, rejected token, mid-session 401, logout and server mode - but there is no browser harness in this repo at all, so tell me if you want Cypress/Playwright wired up and I will do it as a separate PR rather than bolt it onto this one.

matanbaruch:
> I tried to run the code locally for this PR and it simply doesn't work for me

Not a mistake on your end - after login the app fetched namespaces through the tokenless `RolloutAPI` singleton instead of the token-aware client, so it always got a 401 and then rendered nothing (the grey page), and the login check hit `Version()` which never talks to Kubernetes, which is why `wrong` was accepted with no error; both are fixed and verified against a real cluster - no token 401, `wrong` 401 with a real message, valid token loads the dashboard, SSE streams on the cookie alone, and a read-only SA gets a 403 from Kubernetes on promote.

matanbaruch:
> 1) E2e tests for the UI 2) User documentation that explains how the feature is used and more specifically the different ways of how a user can obtain a token

Docs are in `docs/dashboard.md` now - `kubectl create token`, the service-account Secret fallback for older clusters, why cert and exec-plugin users (EKS/GKE/OIDC) cannot paste anything and need a service account, and copy-pasteable RBAC for read-only vs promote/abort; the UI tests are answered in your code comment on `login.tsx`.

1-login

2-invalid-token

3-dashboard

4-rollout-detail

5-rbac-denied

6-logout


https://github.com/user-attachments/assets/6ec2df70-ecb6-4c4f-9843-70a1a46baaf4

Issue #2250, opened 2022-09-19
Inconclusive result on Kubernetes Job metric


object88:
# Summary

The current implementation of the [Job](https://argoproj.github.io/argo-rollouts/analysis/job/) metric does not allow for an "inconclusive" result.  This puts it at a technical deficit against other metrics.

# Use Cases

I can think of 2 use cases:
1) The job is unable to start, as with a CLB or Image Pull Error, as mentioned in [Issue 1329](https://github.com/argoproj/argo-rollouts/issues/1329)
2) The job successfully starts, but it itself is unable to determine if the results was certainly a success or a failure.

My personal use-case is the second: we have a Job that starts a GHA workflow.  In some cases, the workflow simply does not run or complete in time (say, because Github is down, or we have a terrible backlog).  I would like to be able to go to an Argo console or CLI and indicate whether the promotion should proceed.  For example, maybe I know that my GHA workflow (some particular time) is going to fail due to `$REASONS`, but I want to ignore that and proceed anyway.

The current implementation only allows me to work in absolutes.

## Notes

The framework for an [inconclusive result](https://argoproj.github.io/argo-rollouts/features/analysis/#inconclusive-runs) is established for other metrics.

I would propose that if a Job returns JSON, the [existing evaluation](https://github.com/argoproj/argo-rollouts/blob/master/utils/evaluate/evaluate.go) could be applied.

I would also propose that the Job be augmented with a new property that describes the anticipated reaction if the Job does not start or has a non-zero exit code.  The default would be consistent with current behavior (`success`) so that existing Jobs are not impacted.  Non-default values would indicate that the Job was a `fail` or `inconclusive`.

This should maybe be two separate proposals?

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

Abhishek-Kaudare:
Is there any updates on this enhancement?

kostis-codefresh:
Related https://github.com/argoproj/argo-rollouts/issues/3562

Issue #2158, opened 2022-07-24
When defining an analysis with multiple items, when one item fails, others are aborted but marked as successful.


marcportabellaclotet-mt:
Checklist:

* [x ] I've included steps to reproduce the bug.
* [x ] I've included the version of argo rollouts.

**Describe the bug**

When defining an analysis with multiple items, when one item fails, others are aborted but marked as successful.
**To Reproduce**
Create an analysis template like this:
```
kind: AnalysisTemplate
apiVersion: argoproj.io/v1alpha1
metadata:
  name: sleep-and-fail
spec:
  metrics:
  - name: sleep-and-fail
    count: 1
    interval: 5s
    failureLimit: 0
    provider:
      job:
        spec:
          ttlSecondsAfterFinished: 60
          template:
            spec:
              containers:
              - name: sleep
                image: alpine:3.8
                command: [sh, -c]
                args: [sleep 10 && exit 1]
              restartPolicy: Never
          backoffLimit: 0
  - name: webmetric
    successCondition: result == "3.3.3.3"
    provider:
      web:
        url: "http://ip.jsontest.com/"
        timeoutSeconds: 20 # defaults to 10 seconds
        headers:
          - key: Authorization
            value: "Bearer lol"
        jsonPath: "{$.ip}"
``` 
create a rollout pointing to the above analysis:
```
  strategy:
    canary:
      canaryService: service-canary
      stableService: service
      canaryMetadata:
        labels:
          deployment: canary
      stableMetadata:
        labels:
          deployment: stable
      steps:
      - setWeight: 0
      - setCanaryScale:
          weight: 100
      - analysis:
          templates:
          - templateName: sleep-and-fail
```

**Expected behavior**

The expected behavior is that the "sleep-and-fail" item in the analysis is not marked as successful.
It could be marked as "aborted" or a similar naming.
It would be great to add a conditional spec to force items to complete, similar to the spec requiredForCompletion in experiments.

**Screenshots**

![image](https://user-images.githubusercontent.com/56038098/180644509-cc8c52a3-c40c-4321-9393-30a2367551a1.png)

**Version**

v1.2.1
**Logs**

```
# Paste the logs from the rollout controller
```
time="2022-07-24T11:19:19Z" level=info msg="Metric Assessment Result - Failed: failed (1) > failureLimit (0)" metric=webmetric
[...]

PR #4692, opened 2026-04-14
fix(analysis): detect infrastructure errors early and mark job analysis as inconclusive

### LINKED ISSUE #3562: Job Analysis Run never fails when the job image can't be pulled (it just succeeds and promotes).
Checklist:

**Describe the bug**

Analysis Run of type job never fails when the job image can't be pulled (it just succeeds and promotes). When Job pods can't pull image it goes into `ErrImagePull` indefinitely and the job itself does not Fail. However, Argo rollouts  waits for some time and considers this as success and then promotes the canary.

**To Reproduce**

1. create a namespace called `test-namespace`
2. run the following kubernetes configs
3. Wait for the pods to come up
4. change the deployment image to `nginx:1.19.0` to trigger a rollout

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment
  namespace: test-namespace
spec:
  replicas: 0
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:1.9.8 # update it to 1.19.0 to trigger a rollout, as the first time will only create the pods.
          ports:
            - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-stable
  namespace: test-namespace
spec:
  selector:
    app: nginx
  ports:
  - port: 80
    targetPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-canary
  namespace: test-namespace
spec:
  selector:
      app: nginx
  ports:
  - port: 80
    targetPort: 8080
---
kind: ClusterAnalysisTemplate
metadata:
  name: acceptance-test
spec:
  metrics:
    - name: acceptance-test
      provider:
        job:
          spec:
            backoffLimit: 2
            completions: 1
            template:
              spec:
                containers:
                  - name: test
                    image: docker.io/library/lol:5 # this image does not exist.
                    command: ["bash"]
                    args: ["-c", "echo 'Hello world! going to exit with 0 (success).' && sleep 10 && exit 0"]
                restartPolicy: Never
---
kind: Rollout
metadata:
  name: awesome-rollout
  namespace: test-namespace
spec:
  replicas: 2
  workloadRef:
    apiVersion: apps/v1
    kind: Deployment
    name: test-deployment
  strategy:
    canary:
      canaryService: nginx-canary
      stableService: nginx-stable
      analysis:
        templates:
          - templateName: acceptance-test
            clusterScope: true
      steps:
        - setWeight: 20
        - pause: {duration: 10s}
        - setWeight: 40
        - pause: {duration: 10s}
        - setWeight: 80
        - pause: {duration: 10s}
```

**Expected behavior**

* It will succeed and promote and won't report a failure even though the job pod didn't start and was in `ErrImagePull`. It will also delete both the job and the rollout from the cluster.

**Screenshots**

Screenshot 2024-05-17 at 4 41 52 PM

Screenshot 2024-05-17 at 4 42 09 PM

**Version**
Tested on v1.6.6 and previous versions. 

MohammedShetaya:
### Description

This PR introduces an early-detection mechanism in the job metric provider for terminal pod errors, such as `ErrImagePull`. Currently, jobs that can't pull their image wait indefinitely without ever setting a `Failed` condition, causing the analysis to silently succeed and promote the canary.

### Changes

* The `JobProvider.Resume` function is now actively called during analysis reconciliation. Previously, it was only invoked on manual resumes.
* `Resume` now inspects pod container statuses (including init containers) for terminal waiting states (`ErrImagePull`, `ImagePullBackOff`, `InvalidImageName`) and marks the measurement as `Inconclusive` immediately.
* The `ResumeAt` interval grows with the job execution time using `√(2t)`. This matches the nature of the targeted errors — they surface early, so frequent checks are needed at job start, but not later.
* In-progress measurements are no longer skipped during termination due to a future `ResumeAt`, ensuring `Terminate()` is always called.
* Unit/E2E tests were added/updated to cover the new cases. The E2E fixture was made deterministic by using an indefinite pause instead of a timed one.

### What is out of scope?

* Long-running jobs are still an issue in the current implementation; they report an inconclusive state without any effect on the rollout lifecycle.
* A change needs to be discussed and implemented on how background analysis should be treated for jobs since they are different in nature than other metric types, which are not expected to run for much time (either immediately return or time out). Job metrics, on the other hand, are expected to run for some time.
* Job metrics are the only ones hitting the `Provider.Terminate` function, so a change can be made to do extra analysis after the metric is terminated, for example, (condition xyz should be met to promote the rollout after termination of the running metrics)

Closes: #3562  
Help with: #2250

#### Checklist:

* Either (a) I've created an enhancement proposal and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* The title of the PR is (a) conventional with a list of types and scopes found here, (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.
* I've signed my commits with DCO
* My builds are green. Try syncing with master if they are not.
* I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* I have run all tests locally (including the flaky ones) and they pass on my workstation
* I have used LLM/AI/Agent tools for this PR but I am responsible for all code of this PR
* I understand what the code does and WHY/HOW it works in several scenarios
* I know if my code is just adding new functionality or changing old functionality for existing users
* My organization is added to USERS.md.


kostis-codefresh:
I think we agreed that a job that cannot start [would be inconclusive](https://github.com/argoproj/argo-rollouts/issues/3562#issuecomment-2238041627). Correct?

Now this is a major breaking change.

Initially we had failed job -> Succesful analysis. But there [are](https://github.com/argoproj/argo-rollouts/issues/2158) [several](https://github.com/argoproj/argo-rollouts/issues/2250) bugs (and I agree with them ) that this doesn't make any sense. 
So with my [PR](https://github.com/argoproj/argo-rollouts/pull/4602) we said it is inconclusive (which matches user expectations)

But with this PR (please correct me if I am missing something), a job that does not start fails the analysis metric.

I think that pausing the Rollout when an inconclusive analysis happens is a good design choice.

If I run a deployment with Argo Rollouts using smoke tests for an analysis (in a Job metric), I think there is a big difference between

1. The smoke tests run, finished, and returned a business error (fail the deployment)
2. The smoke tests had an error and could not run. The main deployment might actually be ok

In the second case I would expect an inconclusive run that allows me to intervene.

test/e2e/analysis_test.go
```diff
@@ -834,20 +834,19 @@ func (s *AnalysisSuite) TestCanaryBackgroundAnalysisInvalidImageJob() {
 		ExpectAnalysisRunCount(1).
 		ExpectBackgroundAnalysisRunPhase("Running").
 		When().
-		PromoteRollout().
-		WaitForRolloutStatus("Healthy").
-		WaitForBackgroundAnalysisRunPhase("Inconclusive").
+		WaitForBackgroundAnalysisRunPhase("Failed").
```
kostis-codefresh: This is an important change. I think that now users will open new issues complaining that when an analysis has failed (as a job) it is not clear if the job itself has a "terminal error" or if the job actually run and it found an error (i.e. the smoke tests did not pass)

MohammedShetaya: Thanks! I agree, it will cause more confusion. I've changed the behavior to be inconclusive. There's still one unhandled case where the job is a long-running job (it still promotes), however, this shouldn't block this PR. I will follow up on it later.

metricproviders/job/job.go
```diff
@@ -207,7 +212,30 @@ func (p *JobProvider) Resume(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric,
...
+		err = processJobPods(podList)
+		if err != nil {
+			measurement.FinishedAt = &now
+			measurement.Message = err.Error()
+			measurement.Phase = v1alpha1.AnalysisPhaseInconclusive
+		}
+	}
+	elapsed := timeutil.MetaNow().Sub(measurement.StartedAt.Time)
+	// growing delay with respect to the elapsed time since the job started
+	delay := time.Duration(math.Sqrt(2*elapsed.Seconds()) * float64(time.Second))
```
kostis-codefresh, line 246: Why do we use math.Sqrt here instead of the standard wait/backoff k8s package?

metricproviders/job/job.go
```diff
@@ -207,7 +212,30 @@ func (p *JobProvider) Resume(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric,
...
+			LabelSelector: labelSelector,
+		})
+		if err != nil {
+			return metricutil.MarkMeasurementError(measurement, err)
+		}
+		err = processJobPods(podList)
+		if err != nil {
+			measurement.FinishedAt = &now
+			measurement.Message = err.Error()
+			measurement.Phase = v1alpha1.AnalysisPhaseInconclusive
```
kostis-codefresh: If we are finished and have an error and have decided already about the measurement being inconclosive don't we need to return straigth away? Do we miss a return statement here?

metricproviders/job/job.go
```diff
@@ -227,6 +255,33 @@ func (p *JobProvider) Terminate(run *v1alpha1.AnalysisRun, metric v1alpha1.Metri
 	return measurement
 }
 
+var terminalWaitingReasons = map[string]bool{
```
kostis-codefresh, line 274: I would add a specific comment here explaining that for these types of errors we clearly decide that the job is inconclusive and we want a human to intervene instead of having the rollout progress.

The "terminal" wording here implies that the rollout itself is/will be terminated and this is not the case.

metricproviders/job/job.go
```diff
@@ -227,6 +255,33 @@ func (p *JobProvider) Terminate(run *v1alpha1.AnalysisRun, metric v1alpha1.Metri
...
+	"ImagePullBackOff": true,
+	"InvalidImageName": true,
+}
+
+// processJobPods checks if the job should be considered as failed
+// this can happen because of any terminal condition that stops the container from starting
+// such as ErrImagePull in one of the pod containers
+func processJobPods(pods *v1.PodList) error {
+	for _, pod := range pods.Items {
+		for _, cs := range pod.Status.InitContainerStatuses {
```
kostis-codefresh, line 285: Have you also tested the code with an init container example? Or just the main container of a job?

metricproviders/job/job_test.go
```diff
@@ -205,6 +205,88 @@ func TestRunCreateCollision(t *testing.T) {
 	assert.Nil(t, measurement.FinishedAt)
 }
 
+func TestResumeRunningJob(t *testing.T) {
```
kostis-codefresh, line 212: what does this test exactly? Happy path scenario? Can we add a comment?

metricproviders/job/job.go
kostis-codefresh: Never mind. Just saw the unit test.

test/e2e/analysis_test.go
```diff
@@ -34,7 +34,6 @@ func (s *AnalysisSuite) SetupSuite() {
 	s.ApplyManifests("@functional/analysistemplate-fail-multiple-job.yaml")
 	s.ApplyManifests("@functional/analysistemplate-long-running-job.yaml")
 	s.ApplyManifests("@functional/analysistemplate-invalid-image-job.yaml")
-	s.ApplyManifests("@functional/analysistemplate-invalid-image-job-deadline.yaml")
```
kostis-codefresh, line 37: Why was this file removed?

test/e2e/functional/rollout-inline-invalid-image-job.yaml
```diff
@@ -1,6 +1,6 @@
-# Rollout with an inline analysis that uses a non-existent docker image
-# The job has a deadline of 10 seconds which is less than the 30 second pause, so the job should be terminated 
-# and the analysis should be failed.
+# Rollout with an inline analysis that uses a non-existent docker image.
+# The job pod stays in ErrImagePull/ImagePullBackOff/InvalidImageName, which causes the
```
kostis-codefresh, line 2: This seems to change the test of another scenario. Please keep the previous test (which was specifically about having a deadline) and add a brand new test for your scenario.

test/e2e/functional/rollout-background-invalid-job.yaml
```diff
@@ -13,14 +13,16 @@ spec:
...
       steps:
       - setWeight: 10
-      - pause: {duration: 10s}      
+      - analysis: # make the rollout busy until analysis run returns Inconclusive
+          templates:
+            - templateName: long-running-job
   selector:
     matchLabels:
-      app: rollout-background-long-running-job
+      app: rollout-background-invalid-job
```
kostis-codefresh, line 21: This seems to change the test of another scenario. Please keep the previous test (which was specifically about having a deadline) and add a brand new test for your scenario.

It is two separate tests

1) a Job the never finished (long-running) but hasn't thrown an error just yet
2) A Job that has an error that we detect during the course of the rollouit

test/e2e/functional/analysistemplate-invalid-image-job-deadline.yaml
```diff
@@ -1,23 +0,0 @@
...
-metadata:
-  name: invalid-image-job-deadline
-spec:
-  metrics:
-  - name: invalid-image-job-deadline-metric
-    count: 1
-    provider:
-      job:
-        spec:
-          activeDeadlineSeconds: 10
```
kostis-codefresh, line 14: We need this test. Why was it removed? It tests the scenario of having a long running job AND defining activeDeadlineSeconds in the rollout analysis

test/e2e/analysis_test.go
```diff
@@ -798,22 +797,22 @@ func (s *AnalysisSuite) TestCanaryInlineAnalysisInvalidImageJob() {
...
 			ar := t.GetRolloutAnalysisRuns().Items[0]
-			assert.Equal(s.T(), v1alpha1.AnalysisPhaseFailed, ar.Status.Phase)
+			assert.Equal(s.T(), v1alpha1.AnalysisPhaseInconclusive, ar.Status.Phase)
 			if len(ar.Status.MetricResults) > 0 {
 				metricResult := ar.Status.MetricResults[0]
-				assert.Equal(s.T(), v1alpha1.AnalysisPhaseFailed, metricResult.Phase)
+				assert.Equal(s.T(), v1alpha1.AnalysisPhaseInconclusive, metricResult.Phase)
 				if len(metricResult.Measurements) > 0 {
 					measurement := metricResult.Measurements[len(metricResult.Measurements)-1]
-					// The job that was terminated due to deadline should be Failed
```
kostis-codefresh, line 815: Here we seem to change a test of a scenario to another scenario. I think this was testing having a deadline AND also having a long running job. I think this test should stay the same and any new behavior should be tested on another test.

metricproviders/job/job.go
```diff
@@ -207,7 +212,30 @@ func (p *JobProvider) Resume(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric,
 	}
 	if measurement.Phase.Completed() {
 		p.logCtx.Infof("job %s/%s completed: %s", job.Namespace, job.Name, measurement.Phase)
+		return measurement
 	}
+
+	labelSelector := metav1.FormatLabelSelector(job.Spec.Selector)
+	if labelSelector != "" && labelSelector != "<none>" {
+		podList, err := p.kubeclientset.CoreV1().Pods(jobName.Namespace).List(context.TODO(), metav1.ListOptions{
```
kostis-codefresh: do we know the performance implications for hitting directly the K8s API every time this code runs? Maybe we need to use a watcher/lister/caching mechanism instead?

kostis-codefresh (changes requested):
I added some specific comments on the code. But I also have two general comments.

The [job documentation page should also be updated](https://argo-rollouts.readthedocs.io/en/latest/analysis/job/) to explain what the code assumes and how it behaves. At the very least we should now mention that job errors are instanly detected and make the analysis inconclosive

I am also really confused about the tests added/update from this PR. I would expect to see only brand new tests and no changes in existing tests. The main behavior regarding the result was already updated by the PR here https://github.com/argoproj/argo-rollouts/pull/4602/files#diff-84b1a3050cb2b346abd90a02e6c89c812bf8917dd0306c18199a0928d57f3d67L225

As a sidenote,  here are all the tests that I run locally and that I would expect to see for this area https://github.com/kostis-codefresh/testing-argo-rollouts/tree/main/job-analysis in respective e2e tests

The tests this PR should add would be *Faster* detection of failed jobs (instead of waiting the timeout or forever). But not changing existing tests. Unless I am missing something here.

metricproviders/job/job.go
MohammedShetaya: The intuition was that it is already scoped down to job pods only, but yeah, I think it makes more sense to have an informer for this. 
Implemented ✅

test/e2e/functional/rollout-background-invalid-job.yaml
MohammedShetaya: For this test case, I did two changes:

* fixed the wrong selector
* needed something to make the RO wait until the inconclusive state pops up, without adding a pause step to make sure the pause state is because of the inconclusive analysis.

For the two cases you mentioned above

* `a Job that never finished (long-running) but hasn't thrown an error just yet` is covered in `TestCanaryBackgroundAnalysisLongRunningJob.` 
* `A Job that has an error that we detect during the course of the rollout` is covered in `TestCanaryBackgroundAnalysisInvalidImageJob`

metricproviders/job/job.go
MohammedShetaya: Yes, this was missing

metricproviders/job/job.go
MohammedShetaya: Wait.Backoff would require persisting its state on `Measurement`

MohammedShetaya:
All the comments related to tests point to deleting the analysis template `rollout-background-invalid-job` and changing the behavior of its test case. Here's why I did these changes:

* invalid-image-job-deadline was deleted because its only purpose was to give the old test a way to eventually terminate an ErrImagePull pod via activeDeadlineSeconds. With the new short-circuit, an invalid image is detected and the analysis is marked Inconclusive long before the deadline can fire, so a deadline-bound invalid-image template no longer has any observable effect that's distinct from any other failing template. In other words, the test it supported was really testing "deadline kills a stuck Job → analysis Failed" ("always" because the deadline was less than the pause time), and the invalid image was just a contrived way to keep the Job stuck.

* Knowing that, I don't think this test case will be different from a test case that just returns a failure. I still added it back as an actual long-running job with a deadline and included it in the `TestCanaryInlineAnalysisLongRunningJobWithDeadline` test case, and used clearer naming for it. 

* TestCanary{Inline,Background}AnalysisInvalidImageJob now exercises the actual new behavior > immediate Inconclusive from an ErrImagePull pod with the rollout transitioning to Paused for InconclusiveAnalysisRun — without needing a deadline as scaffolding.

docs/analysis/job.md
```diff
@@ -27,10 +27,10 @@ metrics:
 
 The possible outcomes of a job metric are:
 
-1. Jobs starts, runs and ends succesffully with exit 0. Analysis has passed
+1. Jobs starts, runs and ends successfully with exit 0. Analysis has passed
 1. Jobs starts, runs and end with exit code non-zero. Analysis has failed
-1. Job cannot start because of infrastructure error (e.g. image not found). Analysis is inconclusive
-1. Job was still running but was terminated because it is not needed anymore.  Analysis is inconclusive
+1. Job cannot start because one of its pods is stuck in a terminal waiting state such as `ErrImagePull`, `ImagePullBackOff`, or `InvalidImageName` The metric short-circuits to **Inconclusive** without waiting for the Job to time out, and the rollout transitions to **Paused**
```
kostis-codefresh, line 32: Yes the comments were exactly what I had in mind. A clear description of all the outcomes of a job so that users are not confused.  Thanks!

metricproviders/job/job.go
```diff
@@ -119,6 +125,14 @@ func newMetricJob(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric, jobNS strin
 
 	ownerRef := []metav1.OwnerReference{*metav1.NewControllerRef(run, analysisRunGVK)}
 
+	podLabels := metric.Provider.Job.Spec.Template.GetLabels()
+	if podLabels == nil {
+		podLabels = make(map[string]string)
+	}
+	// this label is used by the pod informer to cache metric job pods
+	podLabels[AnalysisRunUIDLabelKey] = string(run.UID)
+	metric.Provider.Job.Spec.Template.SetLabels(podLabels)
```
kostis-codefresh, line 134: Can you explain a bit what is happening here? If podLabels is not nil, then we modify an entry and then save back the whole thing? But it was already there anyway.  I think this statement is needed only in some scenarios.

Did you also do any analysis here regarding racing conditions? Or there is no problem here because we assume each rollout has its own analysis/job/pods sets?

MohammedShetaya: > If podLabels is not nil, then we modify an entry and then save back the whole thing

Yes, but we don't save back `the whole thing` since it is just the map hmap. Also, I think this is clearer to read than moving the assignment line in the if condition. I don't mind moving it, though.  

> Did you also do any analysis here regarding race conditions? Or there is no problem here because we assume each rollout has its own analysis/job/pods sets?

Honestly, I'm not thinking of a race condition here since the `newMetricJob` is called once by the `Run` function (subsequent reconciles hit Resume, not Run), which is always called once, + the AnalysisRun is [deep copied](https://github.com/argoproj/argo-rollouts/blob/master/analysis/analysis.go#L53) per reconcile.

kostis-codefresh:
I tested locally the PR with my own tests at https://github.com/kostis-codefresh/testing-argo-rollouts/tree/main/job-analysis

It correctly catches the scenarios, error-job and background-error-job as expected. Every other scenario has the same behavior as before.

kostis-codefresh (approved):
@zachaller This PR has a very simple goal:

Previously if a job analysis had an issue itself (i.e. the job image was not found), the analysis just stayed there doing nothing and would only timeout at the end of the canary

With this PR we actually look at what the pods of the job did and Argo Rollouts learns straight away if the job itself has issues. If yes, then it returns inconclusive right away (same as before but without waiting for the canary timeout)

The code changes are deeper than I originally thought so PTAL.

zachaller (approved):

Issue #4672, opened 2026-03-26
kstatus compatibility


artem-nefedov:
# Summary

Currently, argo-rollouts custom resources (specifically Rollout) are not compatible with [kstatus format](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md), which is de-facto standard for status/health checks used by various utilities. It was a long time problem for flux2 kustomize-controller (see https://github.com/fluxcd/flux2/issues/4632), and now it became a more prominent problem since [helm release v4](https://github.com/helm/helm/releases/tag/v4.0.0), which also adopted kstatus health checks. Any release installed with helm CLI/library that creates Rollouts (and possibly other argo-rollouts resources) will hang until timeout if `--wait=watcher` is used. It also affects flux2 helm-controller since it adopted helm v4 library.

The proposal is to enhance "status" field to work with kstatus. The most important part is to add "Ready" condition. Logically, Rollout should become "Ready" when it has fully progressed.

Another problem is `status.observedGeneration` being the wrong type, resulting in `of type string, expected int64` error.
The algorithm of kstatus includes the comparison of generation with observedGeneration, and [it is defined by convention that this field must be of type int64](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md?plain=1#L450) (here `observedGeneration` is inside a condition, but I believe the same applies when it's a standalone status field as well).

# Use Cases

Ability to use argo-rollouts resources properly with installation methods other than ArgoCD.

---

**Message from the maintainers**:

Need this enhancement? Give it a 👍. We prioritize the issues with the most 👍.

isugimpy:
https://github.com/argoproj/argo-rollouts/issues/3402 is an older issue related to this.

Issue #1339, opened 2021-07-12
docs: Restructure/Review documentation. Also make it beginner friendly.


kostis-codefresh:
# Current documentation status

As more and more people adopt Argo Rollouts we need to ensure that the documentation pages cover all common scenarios and can guide both beginner and advanced users to progressive delivery practices.

We need different kinds of documentation for different personas. The documentation right now is mostly of the reference type and caters more to the advanced user/operator. 

 Some pages are really terse and make too many assumptions for the knowledge of the reader. We should improve also the concepts and architecture sections so nothing is blurry and readers know exactly what the controller is doing behind the scenes.

## Suggested structure and documentation sections

I suggest we follow the [Documentation system](https://documentation.divio.com/introduction/#the-secret) that has 4 distinct areas:

* Tutorials
* How-To guides
* Explanation
* Reference

## Alignment with the other Argo Projects

It would be great to match the docs with the [new ArgoCD structure](https://github.com/argoproj/argo-cd/issues/5635). We should also gather the frequent questions asked in the CNCF slack channel and make sure that they are documented.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

kostis-codefresh:
Still valid

PR #4711, opened 2026-05-01
docs: Added separate developer guide section

Kevinjoeharris:
The idea is to create more clear documentation for developers like one which we have in Argo CD. Starting off with **Code Quality and Security Scanning** and will add others files one by one in the upcoming PRs.

**Screenshot from my local machine:**
roll-1
roll-2

Checklist:

### DIFF
--- a/docs/developer-guide/code-quality-and-security-scanning.md
+++ b/docs/developer-guide/code-quality-and-security-scanning.md
@@ -1,0 +1,11 @@
+# Code Quality and Security Scanning
+
+We use the following code quality and security scanning tools:
+
+* `golangci-lint` and `eslint` for compile time linting
+* [CodeQL](https://codeql.github.com/) - for semantic code analysis
+* [codecov.io](https://codecov.io/gh/argoproj/argo-cd) - for code coverage
+* [snyk.io](https://app.snyk.io/org/argoproj/projects) - for image scanning
+* [sonarcloud.io](https://sonarcloud.io/organizations/argoproj/projects) - for code scans and security alerts
+
+These are at least run daily or on each pull request.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -133,17 +133,19 @@ nav:
     - generated/kubectl-argo-rollouts/kubectl-argo-rollouts_terminate_experiment.md
     - generated/kubectl-argo-rollouts/kubectl-argo-rollouts_undo.md
     - generated/kubectl-argo-rollouts/kubectl-argo-rollouts_version.md
+- Developer Guide:
+  - Code Quality and Security Scanning: developer-guide/code-quality-and-security-scanning.md
+  - Contributing:
+    - Contribution Guide: CONTRIBUTING.md
+    - Environment Setup: getting-started/setup/index.md
+    - Releasing: releasing.md
+    - Plugins: plugins.md
 - Best Practices: best-practices.md
 - Migrating: migrating.md
 - FAQ: FAQ.md
 - Security:
   - Security Policy: security/security.md
   - Verify Release Artifacts: security/signed-release-assets.md
-- Contributing:
-  - Contribution Guide: CONTRIBUTING.md
-  - Environment Setup: getting-started/setup/index.md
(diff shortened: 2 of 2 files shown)


docs/developer-guide/code-quality-and-security-scanning.md
```diff
@@ -0,0 +1,11 @@
+# Code Quality and Security Scanning
```
kostis-codefresh, line 1: I don't see the value of having just a single page with only a list of tools. In fact several of these tools are cloud based so a developer cannot really run them locally.

Security scanning is also notorious for showing different results according to which tool you are using.

kostis-codefresh (commented):
One thing that we are missing right now is actually architecture discussion. If we are to improve documentation we need to expand this page https://argo-rollouts.readthedocs.io/en/stable/architecture/ and also explain high level decisions.

Please ask the WHY behind Argo Rollouts. 

Examples:
* https://github.com/argoproj/argo-rollouts/issues/4672
* https://github.com/argoproj/argo-rollouts/discussions/4608

Having pages that explain the architectural decision is more important than just listing tools IMHO (which anybody can see anyway by looking at GitHub actions)

Kevinjoeharris:
Thanks for the thorough review, @kostis-codefresh  Thanks for the feedback! I wanted to share the context behind the approach here.

The inspiration for this PR came directly from Argo CD's developer guide - specifically pages like their [Toolchain Guide](https://argo-cd.readthedocs.io/en/stable/developer-guide/toolchain-guide/) and [Static Code Analysis](https://argo-cd.readthedocs.io/en/stable/developer-guide/static-code-analysis/) page, which follow the exact same pattern: dedicated docs pages per topic, building up a structured developer section incrementally. Argo Rollouts currently has none of that structure, so this PR is meant to be the first page of what becomes a fuller section - not a standalone listing.

I completely agree that architectural decision docs (the "why") are more valuable in the long run, and I fully intend to contribute those. Issues like [#4672](https://github.com/argoproj/argo-rollouts/issues/4672) and discussion [#4608](https://github.com/argoproj/argo-rollouts/discussions/4608) are exactly the kind of material I would like to capture in subsequent PRs, similar to how Argo CD documents its architectural decisions.

That said, I'm happy to expand this first page to go beyond a bare tool listing - adding context around why each tool was chosen, what gap it fills, and how they relate to the CI pipeline. That would make it a more meaningful entry point even on its own.

Would that direction work for you, or would you prefer I reorder the contributions and tackle an architectural decisions page first?

kostis-codefresh:
I understand the need to mimic what Argo CD does and I agree that both projects should have a uniform documentation style. But the scope of Argo Rollouts is more specific than Argo CD and developing each project if a bit different, so all "contribution" page in Argo Rollouts must be specific to Rollouts instead of cloning Argo CD docs.

Yes,  I think it is more important to have pages like this https://argo-cd.readthedocs.io/en/latest/developer-guide/architecture/components/ in Argo Rollouts than simply a page that lists tools.

One my ideas was also to look at these AI generated pages, verify/review them and add a subset of their content to the official Argo Rollout docs.

* https://deepwiki.com/argoproj/argo-rollouts
* https://codewiki.google/github.com/argoproj/argo-rollouts

But as with everything that is LLM generated, we should validate them and confirm what they say is correct by cross-referecing with the actual code, and not blindly copy paste their contents.

On a related note, even improving the user documentation would be a big win for me. See also https://github.com/argoproj/argo-rollouts/issues/1339

kostis-codefresh:
I forgot to mention that for any PRs that include documentation  you do not need to attach screenshots. We have setup already Readthedocs to create temporary build for each PR. It is one of the checks attached in this PR.

Here is rendering of your changes https://argo-rollouts--4711.org.readthedocs.build/en/4711/developer-guide/code-quality-and-security-scanning/

Kevinjoeharris:
@kostis-codefresh 
I can definitely see the scope for better documentation. Will starting working on it one by one.!

kostis-codefresh (changes requested):

PR #4712, opened 2026-05-01
chore: Migrate to pnpm (#4699)

### LINKED ISSUE #4699: chore: migrate to pnpm
Argo CD has already moved to pnpm.  (See merged PR: https://github.com/argoproj/argo-cd/pull/23937)

Rollouts should as well.

keithchong:
Fixes: https://github.com/argoproj/argo-rollouts/issues/4699 

I've used Blake's [PR](https://github.com/argoproj/argo-cd/pull/23937) as a guide (eg. for consistency), as well as other migration resources.

- Zach, right now, in my local environment, I have pnpm installed globally. So, I did not look into a way to install the script in the dist folder
- I did get the same sass-loader issues that Blake had, so the same change was made to webpack.common.js
-  For the version of pnpm used in the package.json, I selected the same version that Argo CD uses: "packageManager": "pnpm@10.28.1",
- For missing dependencies, I tried to use the same version as listed in the original yarn.lock file.  If there are multiple versions, then I selected the latest version.
- Tested with various make targets
  - make test ran fine for me locally (See CI results too)
  - make image, make controller, plugin, etc. were ok
  - Generated the binary (I ran `make plugin-darwin`) and launched the dashboard and did a side-by-side comparison of a couple of the views: (See Table View tooltip in the top right area of the first screenshot)

Left-original-RIght-pnpm

Left-original-Right-pnpm-details

- pnpm run build - some warnings common to what I saw when building argo-cd.  Perhaps these warnings can be addressed later
- Dockerfile changes - I've locked the dependencies to whatever is in the lock file, so we get/build with what we have set in the file in the repo
  - RUN pnpm install --frozen-lockfile

Checklist:


ui/package.json
```diff
@@ -2,6 +2,7 @@
     "name": "ui",
     "version": "0.1.0",
     "private": true,
+    "packageManager": "pnpm@10.28.1",
```
blakepettersson, line 5: With argoproj/argo-cd#27615 we're looking to upgrade to upgrade to pnpm 11 to get improved SBOM generation goodness

keithchong: Hi @blakepettersson , thanks for reviewing. Is it possible to make that change in a separate PR, like that for Argo CD?  At least this way, we've migrated over to use pnpm.

blakepettersson: Sounds good to me. Would you also be looking into getting pnpm into argo-ui?

keithchong: Yes, I will look at argo-ui too.

keithchong:
@blakepettersson , @zachaller , could we merge this soon? BTW, since the Argo CD change is already merged into master already and I believe it will be in 3.5, which Rollout release will this be in, so it will 'match' in terms of having pnpm as the package manager?

zachaller (approved):

PR #4718, opened 2026-05-06
Fix documentation on rollout specification for autoPromotionEnabled b…

sncariad:
…eing true

As the autoPromotionEnabled is by default true when not specified the example in docs is misleading

Checklist:

### DIFF
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -128,7 +128,7 @@ spec:
       # Indicates if the rollout should automatically promote the new ReplicaSet
       # to the active service or enter a paused state. If not specified, the
       # default value is true. +optional
-      autoPromotionEnabled: false
+      autoPromotionEnabled: true
 
       # Automatically promotes the current ReplicaSet to active after the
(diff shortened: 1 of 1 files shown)


kostis-codefresh (changes requested):
Actually it is the other way around. In most cases in the spec file the default is mentioned in the comments, and the example value shows how you **OVERRIDE** something.

Some quick examples:
* `minPodsPerReplicaSet` default is 1, but the spec shows how to override to 2
* `scaleDownDelayRevisionLimit` default is nil, but the spec shows how to override to 2

So in the case of `autoPromotionEnabled`, the comment says default is true and the example value shows how to **override** with false.

It is true that there are some cases that don't follow this pattern (quick example would be `progressDeadlineSeconds`). But just changing `autoPromotionEnabled` on its own doesn't really clear up anything and it will be confusing to the next reader.

I believe a better PR would be to fix the remaining values in the spec to be uniform with everything else (i.e. mention default in comment and show example of **override**).

PR #4729, opened 2026-05-16
fix(controller): honor scaleUpPreviewCheckPoint reset on blueGreen pod template change. Fixes #4728

### LINKED ISSUE #4728: blueGreen: prePromotionAnalysis skipped on new revision when previous revision's scaleUpPreviewCheckPoint was already true
Checklist:

**Describe the bug**

In blueGreen mode with `previewReplicaCount` < `replicas`, if a new pod template is rolled out while the *previous* revision had already completed `prePromotionAnalysis` (and thus set `status.blueGreen.scaleUpPreviewCheckPoint = true`), Argo Rollouts skips creating a new AnalysisRun for the new revision and proceeds straight to scaling preview to full `replicas` and cutting over the active service.

The `scaleUpPreviewCheckPoint` flag is supposed to be reset on pod template change — `resetRolloutStatus` is called for that purpose and explicitly sets `newStatus.BlueGreen.ScaleUpPreviewCheckPoint = false`. The subsequent call to `calculateScaleUpPreviewCheckPoint` then reads the *pre-reset* value from `c.rollout.Status` rather than the just-mutated `newStatus`, immediately overwriting the reset with `true`.

**To Reproduce**

1. blueGreen Rollout with `previewReplicaCount: 1`, `replicas: 4`, a `prePromotionAnalysis` configured, `autoPromotionEnabled: true`.
2. Update the rollout (revision N). Wait for `prePromotionAnalysis` to complete successfully and `scaleUpPreviewCheckPoint` to become `true`. **Do not let the rollout fully promote** — e.g. it's still in the scale-up / `autoPromotionSeconds` window.
3. Update the rollout again to a new image (revision N+1) while it's in that window.
4. Observe: the new revision's preview RS is scaled to full `replicas` immediately, no AnalysisRun is created for revision N+1, and cutover proceeds without `prePromotionAnalysis` ever running.

Expected: revision N+1 runs its own `prePromotionAnalysis` before active-service cutover, just like revision N did.

**Expected behavior**

A new pod template must invalidate the previous revision's `scaleUpPreviewCheckPoint` so the new revision's `prePromotionAnalysis` is re-evaluated against the new preview ReplicaSet. The function's own docstring states this:

```go
// calculateScaleUpPreviewCheckPoint ...
// It get reset to false when the pod template changes, or the rollout fully
// promotes (stableRS == newRS)
```

**Root cause**

[`rollout/bluegreen.go`](https://github.com/argoproj/argo-rollouts/blob/master/rollout/bluegreen.go):

```go
func (c *rolloutContext) syncRolloutStatusBlueGreen(...) error {
    newStatus := c.calculateBaseStatus()
    newStatus.StableRS = c.rollout.Status.StableRS

    if replicasetutil.CheckPodSpecChange(c.rollout, c.newRS) {
        c.resetRolloutStatus(&newStatus)   // sets newStatus.BlueGreen.ScaleUpPreviewCheckPoint = false
    }
    ...
    if reason := c.shouldFullPromote(newStatus); reason != "" {
        c.promoteStable(&newStatus, reason)
    } else {
        newStatus.BlueGreen.ScaleUpPreviewCheckPoint = c.calculateScaleUpPreviewCheckPoint(newStatus)
        //                                              ^ ignores newStatus, reads c.rollout.Status — gets pre-reset value
    }
}

func (c *rolloutContext) calculateScaleUpPreviewCheckPoint(newStatus v1alpha1.RolloutStatus) bool {
    if c.rollout.Spec.Strategy.BlueGreen.PreviewReplicaCount == nil {
        return false
    }
    prevValue := c.rollout.Status.BlueGreen.ScaleUpPreviewCheckPoint   // <-- bug: reads stale state
    if prevValue {
        return true
    }
    ...
}
```

`resetRolloutStatus` correctly clears the flag on `newStatus`, but `calculateScaleUpPreviewCheckPoint` then reads `prevValue` from `c.rollout.Status` (the pre-reconcile snapshot) and short-circuits with `true`, clobbering the reset.

**Proposed fix**

Read `prevValue` from `newStatus` so the reset applied earlier in the same reconcile is honored:

```diff
- prevValue := c.rollout.Status.BlueGreen.ScaleUpPreviewCheckPoint
+ prevValue := newStatus.BlueGreen.ScaleUpPreviewCheckPoint
```

PR to follow.

**Version**

Reproducible against `master` (HEAD `2ccdae835`). Originally observed on a deployed controller in v1.x — affects every version since `calculateScaleUpPreviewCheckPoint` started reading from `c.rollout.Status`.

**Logs / live-cluster evidence**

Rollout state from an affected production cluster:

```
currentPodHash: 64bdd8c78c        # preview RS for revision 234
prePromotionAnalysisRunStatus: null
scaleUpPreviewCheckPoint: true     # stale, inherited from revision 233
```

All AnalysisRuns owned by the rollout (ordered):

```
230-pre   Successful
231-pre   Successful
232-pre   Failed     (transient flake)
232-pre.1 Failed     (same)
233-pre   Successful  <-- last one. NO 234-pre.
```

Revision 234's preview RS was scaled to the full `replicas` count, `scaleUpPreviewCheckPoint` stayed `true`, and the rollout proceeded toward active-service cutover without ever creating `234-pre`.

abizer:
## What

Read `prevValue` from `newStatus` instead of `c.rollout.Status` in `calculateScaleUpPreviewCheckPoint`. When `syncRolloutStatusBlueGreen` calls `resetRolloutStatus` on pod template change, the reset mutates `newStatus.BlueGreen.ScaleUpPreviewCheckPoint = false`; reading from `c.rollout.Status` immediately afterwards picks up the *pre-reset* value, clobbering the reset.

```diff
- prevValue := c.rollout.Status.BlueGreen.ScaleUpPreviewCheckPoint
+ prevValue := newStatus.BlueGreen.ScaleUpPreviewCheckPoint
```

## Why

Matches the function's documented behavior:

> scaleUpPreviewCheckPoint... gets reset to false when the pod template changes, or the rollout fully promotes

Without this fix, a new revision started while the previous one was in the scale-up / `autoPromotionSeconds` window inherits `scaleUpPreviewCheckPoint: true` and skips `prePromotionAnalysis` entirely — the new preview RS scales straight to `replicas` (bypassing `previewReplicaCount`), no AnalysisRun is created, and cutover proceeds untested.

Full repro + root cause + live-cluster evidence in #4728.

## Tests

* Added unit regression test `TestCalculateScaleUpPreviewCheckPointHonorsReset` in `rollout/bluegreen_test.go`. Verified it fails on the unfixed code (`Should be false`) and passes with the fix.
* Existing `TestPreviewReplicaCountHandleScaleUpPreviewCheckPoint` sub-tests (`TrueAfterMeetingMinAvailable`, `FalseAfterActiveServiceSwitch`, `TrueWhenScalingUpPreview`) continue to pass.
* Full `./rollout/...` package suite passes locally (`make test` equivalent).

## Checklist

* [x] This is a bug fix.
* [x] PR title is conventional (`fix(controller): ...`), states what changed, and suffixes the related issue number (`Fixes #4728`).
* [x] Commits are DCO signed.
* [x] Unit tests written and passing.
* [x] All tests run locally and pass.
* [x] This changes behavior for existing users of `blueGreen` + `previewReplicaCount` + `prePromotionAnalysis`: previously a back-to-back revision would skip the gate; now it correctly runs `prePromotionAnalysis` for the new revision. No API or config change.
* [ ] My organization is added to USERS.md. (Out of scope for this bug fix — separate PR if needed.)

### DIFF
--- a/rollout/bluegreen.go
+++ b/rollout/bluegreen.go
@@ -326,8 +326,11 @@ func (c *rolloutContext) calculateScaleUpPreviewCheckPoint(newStatus v1alpha1.Ro
 	}
 
 	// Once the ScaleUpPreviewCheckPoint is set to true, the rollout should keep that value until
-	// the newRS becomes the new stableRS or there is a template change.
-	prevValue := c.rollout.Status.BlueGreen.ScaleUpPreviewCheckPoint
+	// the newRS becomes the new stableRS or there is a template change. Read from newStatus rather
+	// than c.rollout.Status so that a reset applied earlier in this reconcile (e.g. by
+	// resetRolloutStatus on a pod template change) is honored instead of being clobbered by the
+	// pre-reset value.
+	prevValue := newStatus.BlueGreen.ScaleUpPreviewCheckPoint
 	if prevValue {
 		return true
 	}
--- a/rollout/bluegreen_test.go
+++ b/rollout/bluegreen_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/argoproj/argo-rollouts/utils/annotations"
 	"github.com/argoproj/argo-rollouts/utils/conditions"
 	"github.com/argoproj/argo-rollouts/utils/hash"
+	logutil "github.com/argoproj/argo-rollouts/utils/log"
 	rolloututil "github.com/argoproj/argo-rollouts/utils/rollout"
 	timeutil "github.com/argoproj/argo-rollouts/utils/time"
 )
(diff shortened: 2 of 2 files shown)


abizer:
cc @argoproj/argo-rollouts-approvers

abizer:
@kostis-codefresh would you be able to take a look at this when you have a chance? it is a one line fix

kostis-codefresh (changes requested):
I feel that this unit test doesn't really cover the scenario that was described in https://github.com/argoproj/argo-rollouts/issues/4728

It seems to me that we don't have any e2e tests that use previewReplicaCount anyway, so I want to be very cautious about the effects of this PR.

So can we add some e2e test(s) that showcase the bug and verify the actual fix in a more realistic way please?

PR #4730, opened 2026-05-17
fix: remove revision==1 gate for scaleDown onsuccess. Fixes #4681

### LINKED ISSUE #4681: workloadRef.scaleDown: onsuccess never scales down Deployment for blue-green with postPromotionAnalysis
---
  Checklist:

  - I've included steps to reproduce the bug.
  - I've included the version of argo rollouts.

  Describe the bug

  workloadRef.scaleDown: onsuccess never scales down the referenced Deployment when using a blue-green strategy with postPromotionAnalysis configured. The Rollout reaches
   Healthy phase but the Deployment replicas remain unchanged.

  Two bugs in promoteStable() (rollout/sync.go) interact to cause this:

  Bug 1: Initial deploy bypasses PostPromotionAnalysis, creating a permanent dead state

  shouldFullPromote() returns "Initial deploy" when c.stableRS == nil, bypassing all strategy-specific checks including PostPromotionAnalysis. This creates the following
  sequence:

  1. Cycle 1: stableRS is nil → shouldFullPromote returns "Initial deploy" → promoteStable runs → sets stableRS → the scaleDown check fails because c.rollout.Status.Phase
   reflects the previous reconciliation's status (not yet Healthy)
  2. Cycle 2+: stableRS is now set → enters BlueGreen branch in shouldFullPromote → PostPromotionAnalysis != nil → checks c.currentArs.BlueGreenPostPromotion for a
  successful AnalysisRun → finds nil (no AnalysisRun was ever created because the initial deploy path bypassed it) → returns "" → promoteStable is never called again →
  scaleDown never fires

  The Rollout reaches Healthy, but shouldFullPromote permanently returns "" because it requires a successful PostPromotionAnalysis AnalysisRun that was never created for
  the initial deploy.

  Bug 2: Hardcoded revision == 1 check prevents scaleDown on revision 2+

  // rollout/sync.go - promoteStable()
  if revision == 1 && c.rollout.Status.Phase == v1alpha1.RolloutPhaseHealthy &&
     c.rollout.Spec.WorkloadRef != nil &&
     c.rollout.Spec.WorkloadRef.ScaleDown == v1alpha1.ScaleDownOnSuccess {

  Even if Bug 1 were fixed, scaleDown would never fire on revision 2+ due to the revision == 1 guard. If the Deployment wasn't scaled down on revision 1 (due to Bug 1),
  it can never be corrected on subsequent revisions.

  To Reproduce

  1. Create a Deployment with replicas: 3
  2. Create a Rollout with:
  spec:
    workloadRef:
      apiVersion: apps/v1
      kind: Deployment
      name: my-deployment
      scaleDown: onsuccess
    strategy:
      blueGreen:
        activeService: my-service
        previewService: my-service-preview
        autoPromotionEnabled: true
        postPromotionAnalysis:
          templates:
            - templateName: my-analysis-template
  3. Wait for the Rollout to reach Healthy phase
  4. Check the Deployment: kubectl get deploy my-deployment -o jsonpath='{.spec.replicas}'
  5. Observe replicas remain at 3 — never scaled to 0

  Expected behavior

  After the Rollout reaches Healthy and promotion completes, the referenced Deployment should be scaled to 0 replicas, as documented in the
  https://argo-rollouts.readthedocs.io/en/stable/migrating/#reference-deployment-from-rollout.

  Workaround

  Disable postPromotionAnalysis for the initial migration deploy (e.g. by not configuring analysis metrics on the first deploy). Without PostPromotionAnalysis configured,
   shouldFullPromote returns "Completed blue-green update" on cycle 2, promoteStable is called with phase=Healthy, and scaleDown fires correctly. Re-enable
  PostPromotionAnalysis on subsequent deploys.

  Suggested Fix

  The scaleDown logic should not be gated behind promoteStable/shouldFullPromote. It should be a separate check in the main reconciliation loop that fires whenever:
  - The Rollout phase is Healthy
  - workloadRef.scaleDown == onsuccess
  - The referenced Deployment has replicas > 0

  This decouples scaleDown from the promotion flow and eliminates both the timing issue and the revision == 1 limitation.

  Version

  v1.8.3 (confirmed same code exists in v1.7.2 and v1.9.0)

SimbaKingjoe:
## Summary

Remove the `revision == 1` gate from `promoteStable()` so `workloadRef.scaleDown: onsuccess` triggers on every stable promotion, not just the initial deploy.

## Background

When using BlueGreen strategy with `workloadRef.scaleDown: onsuccess`, the referenced Deployment should be scaled down every time a new ReplicaSet is promoted to stable. However, the check in `promoteStable()` (rollout/sync.go:1040) included `revision == 1`, restricting the scale-down to only the very first deployment.

For revision >= 2, `shouldFullPromote()` correctly runs PostPromotionAnalysis (if configured), returns "Completed blue-green update", and calls `promoteStable()`, but the scale-down is silently skipped because `revision != 1`.

## Modifications

- `rollout/sync.go`: Remove `revision == 1 &&` from the scaleDown condition in `promoteStable()`
- `rollout/sync_test.go`: Add `TestScaleDownDeploymentOnSuccessRevision2` verifying scale-down triggers on revision 2

## Verification

```bash
go test ./rollout/ -run 'TestScaleDownDeploymentOnSuccess' -v
# PASS: TestScaleDownDeploymentOnSuccess (revision 1)
# PASS: TestScaleDownDeploymentOnSuccessRevision2 (revision 2)
```

Checklist:

- [x] This is a bug fix
- [x] The title follows conventional commits: `fix: remove revision==1 gate for scaleDown onsuccess`
- [x] DCO sign-off included
- [x] Unit tests written and passing
- [ ] My organization is added to USERS.md (N/A - bug fix)

### DIFF
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -1099,7 +1099,7 @@ func (c *rolloutContext) promoteStable(newStatus *v1alpha1.RolloutStatus, reason
 			conditions.RolloutCompletedMessage, revision, newStatus.CurrentPodHash, reason)
 	}
 
-	if revision == 1 && c.rollout.Status.Phase == v1alpha1.RolloutPhaseHealthy && c.rollout.Spec.WorkloadRef != nil && c.rollout.Spec.WorkloadRef.ScaleDown == v1alpha1.ScaleDownOnSuccess {
+	if c.rollout.Status.Phase == v1alpha1.RolloutPhaseHealthy && c.rollout.Spec.WorkloadRef != nil && c.rollout.Spec.WorkloadRef.ScaleDown == v1alpha1.ScaleDownOnSuccess {
 		var targetScale int32 = 0
 		err := c.scaleDeployment(&targetScale)
 		if err != nil {
--- a/rollout/sync_test.go
+++ b/rollout/sync_test.go
@@ -882,3 +882,22 @@ func TestIsScalingEventMissMatchedDesiredOldReplicas(t *testing.T) {
 	assert.Equal(t, int32(13), roStatus.Status.ReadyReplicas)
 	assert.Equal(t, int32(0), roStatus.Status.UpdatedReplicas)
 }
+
+// TestScaleDownDeploymentOnSuccessRevision2 verifies that scaleDown: onsuccess
+// triggers scale-down even when revision > 1 (previously gated by revision==1).
+func TestScaleDownDeploymentOnSuccessRevision2(t *testing.T) {
+	ctx := createScaleDownRolloutContext(v1alpha1.ScaleDownOnSuccess, 5, true, nil)
+	// Override the revision annotation to simulate a second rollout revision
+	ctx.rollout.Annotations[annotations.RevisionAnnotation] = "2"
+	newStatus := &v1alpha1.RolloutStatus{
+		CurrentPodHash: "2f646bf702",
+		StableRS:       "15fb5ffc01",
+	}
+	err := ctx.promoteStable(newStatus, "reason")
+
+	assert.Nil(t, err)
+	k8sfakeClient := ctx.kubeclientset.(*k8sfake.Clientset)
+	updatedDeployment, err := k8sfakeClient.AppsV1().Deployments("default").Get(context.TODO(), "workload-test", metav1.GetOptions{})
+	assert.Nil(t, err)
+	assert.Equal(t, int32(0), *updatedDeployment.Spec.Replicas)
+}


Hariharasuthan99:
Hello @SimbaKingjoe , scale down on success is expected to be used only during the initial migration from deployment to rollout and so the reason behind hardcoded value of 1. This change would not only try to trigger scale down even after the initial migration it also does not fix the core issue mentioned in #4681 i.e Initial deploy bypasses PostPromotionAnalysis, creating a permanent dead state. Could you please recheck?

PR #4731, opened 2026-05-18
fix(metricprovider): parse grouped Datadog queries (by {tag}) instead of crashing. Fixes #4276

### LINKED ISSUE #4276: Support evaluating grouped Datadog metrics (`by {tag}`) in AnalysisTemplates
# Summary

Support evaluating grouped Datadog metrics in AnalysisTemplates. Specifically, enable the use of `by {tag}` clauses in Datadog queries.

# Use Cases

Currently, the Datadog provider expects queries to return a single scalar value. This prevents users from writing queries that group metrics by tags, such as:

```
sum:trace.http.request.errors{service:my-service} by {resource_name}.as_count()
```

Such grouping is useful for detecting regressions in a subset of entities (e.g., a specific `resource_name`) that might be hidden when looking only at global aggregates. For example, a single misbehaving deployment pod may cause a spike in errors, but its impact can be diluted in the overall sum or average.

This feature would allow users to:
- Group by a tag in a Datadog query
- Define an aggregation strategy in the AnalysisTemplate (e.g. max) to reduce grouped results to a scalar value
- Fail an analysis based on the aggregated result (e.g. if the max value across groups exceeds a threshold)

---

cgreeno:
Checklist:

  > **(b) bug fix.** Today, any Datadog v2 query that uses `by {tag}` crashes in `parseResponseV2` with `json: cannot unmarshal string into Go struct field .Data.Attributes.Columns.Values of type []*float64` — the v2 group column contains strings while the parser expected `[]*float64`. This PR fixes that crash. Issue #4276 tracks the user-visible symptom. The new `reducer` field is the mechanism the fix uses to collapse the multiple per-group values into a scalar for evaluation.

  > Title: `fix(metricprovider): parse grouped Datadog queries (by {tag}) instead of crashing. Fixes #4276`. Type `fix`, scope `metricprovider` (both allowed in `.github/workflows/pr-title-check.yml`), states what changed, suffixed with `Fixes #4276`. `Validate PR title` check is green.

  > All four commits include `Signed-off-by: Chris Greeno <cgreeno@gmail.com>`. DCO check is green on this PR.

  > Build, Lint Go code, Verify Codegen, Run unit tests, e2e (1.32 / 1.33 / 1.34 / 1.35), CodeQL, Snyk all passed on the latest commit. The earlier e2e "failure" was a cancellation from a follow-up push, not a real test failure.

  > 11 new table cases in `metricproviders/datadog/datadogV2_test.go` covering each reducer, the no-reducer error path, all-null grouped responses, Datadog returning errors, and `failing-groups` metadata on both Failed and Successful phases. Focused unit tests in `metricproviders/datadog/datadog_test.go` for `Numeric`, `Strings`, `findGroupValues`, `identifyFailingGroups`, `reduceValues`, plus a direct `parseResponseV2` body-read-error test. `validateIncomingProps` extended with v1/v2 `reducer` cases.

  > `go test ./...` → 83 packages, 0 failures. `golangci-lint run ./metricproviders/datadog/... ./pkg/apis/rollouts/v1alpha1/...` → 0 issues. Local package coverage for `metricproviders/datadog` moves from 90.0% to 93.2%.

  > Yes — I used an LLM-based coding assistant during development. I have read and own every line: the API design (`reducer` as a new field vs. overloading the existing `Aggregator`), the per-group attribution decision (informational metadata, phase still driven by reduced scalar), and the trade-offs around when to error vs. fall through.

  > Walking through the scenarios:
  > 1. **Ungrouped query, 1 value, no reducer** → `Numeric()` returns one float, `reduceValues` short-circuits on `len==1`, behavior is byte-identical to today.
  > 2. **Ungrouped query, 0 values** → empty-result path, same `nilFloat64` evaluation as before.
  > 3. **Grouped query, no reducer** → `reduceValues` returns an explicit error telling the user to set `reducer`. Better than today's cryptic JSON-unmarshal panic.
  > 4. **Grouped query with reducer** → number column found by `Type=="number"`, group column found by `Type=="group"`, values reduced, condition evaluated per group, failing groups joined into `Metadata["failing-groups"]`.
  > 5. **Mismatched response (more values than groups)** → `identifyFailingGroups` breaks out of the loop rather than indexing past the slice. Tested.
  > 6. **Bad success-condition expression** → per-group eval error is swallowed (continue), so attribution doesn't bubble up unrelated errors. Tested.

  > This is **purely additive** for working users — `reducer` is a new optional field; ungrouped queries are byte-identical in behavior. The one existing test that asserted the JSON-unmarshal error string (`cannot unmarshal string into Go struct field .Data.Attributes.Columns.Values of type []*float64`) needed its expectation updated because the response struct now uses `[]json.RawMessage`. No working AnalysisTemplate in the wild needs to be touched — anything that worked before still works; anything that crashed before now works.

  > Fresha added in `USERS.md` (alphabetical position after Flipkart).

## Changes

Fixes #4276. Today, Datadog v2 scalar queries that use a `by {tag}` clause crash in `parseResponseV2`. The v2 scalar API returns *two* columns for a grouped query — a string "group" column holding the tag values and a "number" column holding the per-group scalars — but the existing parser typed `Values` as `[]*float64`, so the very first JSON unmarshal panics on the string column. This PR makes that parse work and lets the result be evaluated as a single scalar.

### Mechanism: new `reducer` field

Added an optional field `reducer` to `DatadogMetric` (`avg | min | max | sum | last`). It collapses the per-group scalars into a single value for the success/failure condition.

- Required when the query returns more than one value. Without it the analysis errors out with a clear message pointing at the missing field, rather than silently picking the first value.
- Rejected on `apiVersion: v1` (v1 doesn't support grouped scalar queries).
- Validated via `+kubebuilder:validation:Enum` at the CRD level.

### Per-group failure attribution

When the query is grouped, the same success/failure condition is additionally evaluated against each group's individual value, and the failing groups are surfaced on the measurement:

- `measurement.Metadata["failing-groups"]` — comma-separated `name=value` pairs
- `measurement.Message` — short human-readable summary, e.g. `reducer=max applied; 1 failing group(s): POST /v2/checkout=0.087`

The phase is still driven by the reduced scalar; per-group attribution is informational. This means `reducer: avg` can drive the rollout decision while operators still see which sub-entity is misbehaving in `kubectl describe analysisrun`, the dashboard, and notification templates.

### Example

```yaml
metrics:
- name: per-endpoint-error-rate
  interval: 30s
  successCondition: result < 0.05
  provider:
    datadog:
      apiVersion: v2
      interval: 5m
      reducer: max
      query: "sum:trace.http.request.errors{service:my-service} by {resource_name}.as_count()"
```

If one endpoint regresses, the measurement reports:

```
Phase: Failed
Value: 0.087
Message: reducer=max applied; 1 failing group(s): POST /v2/checkout=0.087
Metadata:
  failing-groups: "POST /v2/checkout=0.087"
```

### Backward compatibility

- Ungrouped queries are unaffected.
- Responses without a `type` field on columns (existing test fixtures and any legacy caller) are treated as numeric, preserving the previous "first column wins" behavior.
- Pre-existing grouped queries already crashed, so no working configuration regresses.

### Files

- `metricproviders/datadog/datadog.go` — response struct, reducer + attribution logic, validation.
- `metricproviders/datadog/datadog_test.go`, `metricproviders/datadog/datadogV2_test.go` — tests.
- `pkg/apis/rollouts/v1alpha1/analysis_types.go` — new `Reducer` field.
- `pkg/apis/rollouts/v1alpha1/generated.{proto,pb.go}`, `openapi_generated.go`, `pkg/apiclient/rollout/rollout.swagger.json`, `ui/src/models/rollout/generated/api.ts`, `manifests/crds/*.yaml`, `manifests/install.yaml`, `docs/features/kustomize/rollout_cr_schema.json` — regenerated via `make codegen`.
- `docs/analysis/datadog.md` — documents `reducer`, the no-reducer error, and `failing-groups` metadata.


cgreeno:
@kostis-codefresh - love to get a review if you have the time please. :)

cgreeno:
Why does the entire CI run when you change Users.md? as near as I can tell nothing is done with it?  Github is already almost broken... perhaps add an ignore and help them out a little??

cgreeno:
@meeech - assume you might suggest this as a plugin?

I could also just fix the crash bug when by {tag} - with better error to migrate to plugin if you want to use tags and then just provide all the datadog functionality in a plugin? Which you prefer? 


ref: https://github.com/argoproj/argo-rollouts/pull/3892

metricproviders/datadog/datadog.go
```diff
@@ -335,17 +407,102 @@ func (p *Provider) parseResponseV2(metric v1alpha1.Metric, response *http.Respon
...
+			continue
+		}
+		if phase == v1alpha1.AnalysisPhaseFailed {
+			failing = append(failing, fmt.Sprintf("%s=%s", groupNames[i], strconv.FormatFloat(v, 'f', -1, 64)))
+		}
+	}
+	return failing
+}
+
+func reduceValues(values []float64, reducer string) (float64, error) {
```
kostis-codefresh, line 464: All these operators (and more) are already supported natively by the expr library that Argo Rollouts already uses in Analysis results See https://expr-lang.org/docs/language-definition

kostis-codefresh (changes requested):
First a quick disclaimer that I am not a Datadog expert by any means. I do however understand the problem but I believe it might be solved in another way.

Right now this PR adds a brand new field with a limited set of operators (max, min etc).

Argo Rollouts already uses the Expr library that includes [all these operators and more](https://expr-lang.org/docs/language-definition). I don't see why we need a new "reducer" field and custom golang code to implement this limited set of operators just for Datadog.

Instead, this PR (or a brand new PR) should make the Datadog provider behave like the rest of the providers where query calls a provider specific metric and higher level "reducers" are in the **successCondition** property (example successCondition: max(result) < 0.05)

So this PR right now:

1.  Adds a brand new field needed only for Datadog
2. Contains only a very limited subset of what expr supports

What I suggest (in the same or different PR) will 

1. Get the full power of expr library to Datadog results
2. Bring the Datadog provider to the same level as all the other providers
3. not require a new field

I looked at how other providers deal with the same problem and I think Prometheus/Wavefront code can be used as an example of how the Datadog provider must also handle returned results.

Again, I am not an expert on Datadog, so let me know if I am missing something.

metricproviders/datadog/datadog.go
```diff
@@ -335,17 +407,102 @@ func (p *Provider) parseResponseV2(metric v1alpha1.Metric, response *http.Respon
...
+	if len(values) == 1 {
+		return values[0], nil
+	}
+	if reducer == "" {
+		return 0, errors.New("Datadog query returned multiple values but no reducer was set. Set `reducer` (avg, min, max, sum, or last) when using grouped queries.")
+	}
+
+	switch reducer {
+	case "avg":
```
kostis-codefresh, line 473: avg is available already in expr as [mean](https://expr-lang.org/docs/language-definition#mean) (median also exists).

All others (min, max, sum, last) exist with the exact same name

USERS.md
```diff
@@ -28,6 +28,7 @@ Organizations below are **officially** using Argo Rollouts. Please send a PR wit
 1. [Factorial](https://factorialhr.com)
 1. [Farfetch](https://www.farfetch.com/)
 1. [Flipkart](https://flipkart.com)
+1. [Fresha](https://www.fresha.com/)
```
kostis-codefresh, line 31: You can also add this on a separate PR on its own if you want.

cgreeno:
@kostis-codefresh - Awesome thanks for the feedback and direction as well as filling in the knowledge gaps! I will have another shot on it with another PR. 🙇

PR #4734, opened 2026-05-23
test: use t.Log instead of fmt.Println

bhargavpillati21:
Replace fmt.Println with t.Log in tests for utils/record package.


kostis-codefresh (changes requested):
Already in by https://github.com/argoproj/argo-rollouts/pull/4735

PR #4740, opened 2026-05-26
fix: allow plugins to implement ping-pong (and not just ALB/Istio)

### LINKED ISSUE #4739: RFC: Extend ping-pong support to plugin-based traffic routers (initially Gateway API)
## Summary

The ping-pong feature is currently restricted by validation to ALB and Istio traffic routers. With the Gateway API plugin becoming the standard path for new traffic-routing integrations, and with AWS NLB now supporting weighted target groups via the AWS Load Balancer Controller (kubernetes-sigs/aws-load-balancer-controller#4732), there is a clear need for ping-pong support on plugin-based traffic routers - specifically the Gateway API plugin - to enable true zero-downtime updates for long-lived TCP and gRPC workloads behind NLBs, and on the HTTPRoute side as users migrate away from Ingress.

This RFC describes the gap, shows that the core controller logic is already router-agnostic, and proposes two possible paths forward. Feedback on the preferred approach is requested before opening implementation PRs.

## Motivation

Argo Rollouts users running gRPC or long-lived TCP workloads behind AWS NLBs cannot currently achieve zero-downtime rollouts with the Gateway API plugin. The standard canary flow with a `canaryService`/`stableService` pair performs a service-selector swap at the end of the rollout, which causes a brief window where the AWS Load Balancer Controller deregisters and re-registers targets, dropping in-flight connections. This is exactly the problem the ping-pong feature was introduced to solve for ALB.

With NLB weighted target groups now supported by the AWS Load Balancer Controller, the underlying infrastructure can hold both ping and pong target groups simultaneously and shift weight between them without target deregistration. Combined with `TCPRoute`/`GRPCRoute`/`HTTPRoute` `backendRefs[].weight`, this gives us everything we need to replicate the ALB ping-pong story for Gateway API - except that today the rollout validation rejects this configuration.

This also matters for the broader Ingress-to-Gateway-API migration: as users move HTTP workloads off Ingress to HTTPRoute, ALB ping-pong becomes unavailable to them (since Argo Rollouts ALB integration is Ingress-based), and there is currently no equivalent on the Gateway API side regardless of which load balancer sits in front.

## Current state of the code

A walk through the existing implementation, since most of the work is structurally already in place:

**Status transition is router-agnostic.** `Status.Canary.StablePingPong` is flipped on promotion in `rollout/sync.go` (lines 1027–1033 on master), gated only by `trafficrouting.IsPingPongEnabled(c.rollout)`. There is no router-type check.

**Selector reconciliation already handles ping-pong correctly.** `reconcilePingAndPongService` in `rollout/service.go` (line 247) calls `trafficrouting.GetStableAndCanaryServices(c.rollout, true)` to point the current canary-side service at `newRS`, with no router-type check. `reconcileStableAndCanaryService` calls `ensureSVCTargets`, which short-circuits cleanly when `svcName == ""` (`rollout/service.go` line 296), so a rollout that defines only `pingPong` (without `canaryService`/`stableService`) does not trigger any selector swap on promotion.

**The shared helper is exported.** `trafficrouting.GetStableAndCanaryServices(ro, isPingpongPreferred)` is already a public function (`rollout/trafficrouting/service_helper.go`) and is consumed by ALB, Istio, and the controller-level selector reconciliation. Plugins can import and call it. Its existing semantics also correctly handle the case where both `pingPong` and `canaryService`/`stableService` are configured - see PR #3628, which scoped that combined-configuration behaviour to ALB.

**The single remaining ALB/Istio-only constraint is in validation.** In `pkg/apis/rollouts/validation/validation.go` (line 268):

```go
if canary.PingPong != nil {
    if canary.TrafficRouting != nil && canary.TrafficRouting.ALB == nil && canary.TrafficRouting.Istio == nil {
        allErrs = append(allErrs, field.Invalid(..., PingPongWithRouterOnlyMessage))
    }
    ...
}
```

`requireCanaryStableServices` (line 238) does not include `TrafficRouting.Plugins` in its list, so a plugin-based rollout with only `pingPong` (no `canaryService`/`stableService`) is already valid from that perspective.

**The gap is on the plugin side.** The Gateway API plugin reads `rollout.Spec.Strategy.Canary.CanaryService` and `.StableService` directly (e.g. `pkg/plugin/tcproute.go` lines 17–18, with identical patterns in `httproute.go`, `grpcroute.go`, `tlsroute.go`) rather than going through `GetStableAndCanaryServices`. Even if validation allowed the configuration, the plugin would not pick up the ping/pong service names from `Spec.Strategy.Canary.PingPong`.

## Proposed change (high level)

1. **In argo-rollouts core:** relax the ping-pong validation to permit plugin-based traffic routers, and update `PingPongWithRouterOnlyMessage` accordingly. No controller logic changes required - the existing implementation is already router-agnostic.
2. **In the Gateway API plugin:** replace the direct `Spec.Strategy.Canary.CanaryService`/`StableService` reads with a call to `trafficrouting.GetStableAndCanaryServices(rollout, true)` in all four route handlers (TCP, TLS, HTTP, gRPC). Bump the `github.com/argoproj/argo-rollouts` dependency to a version that exports the helper (already exported in current stable).
3. **Tests** at both layers - validation tests in core, unit tests for each route handler in the plugin.
4. **Docs** - update `docs/features/traffic-management/plugins/` (or wherever appropriate) to note ping-pong support, and add an example to the Gateway API plugin docs.

The change is small in code volume but carries an API-surface decision, hence this RFC.

## Two options for the validation/declaration mechanism

The question is how the core should know that a given plugin actually supports ping-pong. Plugins are external binaries; a misconfigured user could theoretically point at a plugin that doesn't implement ping-pong and get a silently broken rollout.

### Option A - Permissive (minimal change)

Simply allow `PingPong != nil` together with `TrafficRouting.Plugins != nil`. Trust the user to use a plugin that handles it. The Gateway API plugin will be updated to handle it; other plugins are free to do so.

Pros:
- No new API surface.
- Symmetric with how plugins are treated elsewhere (the core doesn't know what capabilities a plugin has).
- Smallest possible diff in core.

Cons:
- A user configuring a plugin that doesn't implement ping-pong sees no validation error - the rollout will proceed with `StablePingPong` flipping correctly, but the plugin will continue routing to the wrong service names, breaking the rollout silently.

### Option B - Declarative (plugin opts in)

Introduce a way for a plugin to declare ping-pong support - e.g. a new optional field in `RolloutTrafficRouting.Plugin` config (something like `supportsPingPong: true`) - and gate the validation on that. Or, alternatively, introduce a new RPC method `SupportsPingPong() bool` on the plugin interface that the controller calls during validation.

Pros:
- Surfaces misconfiguration early.
- Establishes a pattern for declaring other plugin capabilities in the future.

Cons:
- New API surface in the plugin RPC interface (or in the trafficRouting CRD), requires a coordinated change across plugin authors.
- Heavier than the problem strictly demands.

### My read

Option A is the smallest, most consistent change and gets us unblocked. Option B is cleaner long-term but is a larger conversation about plugin-capability declaration in general, which feels out of scope for ping-pong specifically. I'd lean toward A unless there's appetite for a broader capability mechanism - happy to be guided either way.

## Backwards compatibility

The change is strictly additive. Existing ALB and Istio ping-pong rollouts are unaffected. The semantics of `GetStableAndCanaryServices(ro, true)` already correctly handle the case where both `pingPong` and `canaryService`/`stableService` are set (pingpong takes priority), so the Gateway API plugin update remains backwards-compatible for users not on ping-pong.

## Testing plan

- **Unit, core:** Add a validation test in `pkg/apis/rollouts/validation/validation_test.go` covering the new plugin + ping-pong combination.
- **Unit, plugin:** Add tests in `pkg/plugin/tcproute_test.go` (and the three siblings) covering:
  - PingPong-only configuration applies weights to ping/pong backendRefs.
  - When `StablePingPong` flips between revisions, weights are applied to the swapped pair.
  - When both ping-pong and canary/stable services are set, ping-pong takes precedence (matching existing helper semantics).
- **E2E:** Manual verification with a kind cluster + Envoy Gateway for TCPRoute weight propagation, then with AWS NLB + AWS Load Balancer Controller in a real environment to verify zero-downtime promotion.

## Out of scope

- Any change to the controller logic for ping-pong (the existing implementation is already router-agnostic, as shown above).
- A general plugin-capability declaration mechanism (Option B above stays an open question if there's interest, but I'm not proposing to design it as part of this work).
- Changes to other plugins. This RFC concerns the core validation relaxation plus the Gateway API plugin. Other plugins can adopt the same pattern when their authors choose to.

## Asks

1. Is the permissive approach (Option A) acceptable, or is there a preference for plugin-declared capability (Option B) or some other mechanism I'm not seeing?
2. Are there controller-level edge cases I haven't accounted for? Specifically I'd appreciate a sanity-check on `status.canary.weights` reporting for plugin-based routers under ping-pong - I believe it already routes through `GetStableAndCanaryServices`, but want confirmation.
3. Any preference on the wording/structure of the updated `PingPongWithRouterOnlyMessage` (e.g. "ping-pong is supported by ALB, Istio, and plugin-based traffic routers")?

Happy to draft PRs against both repos once there's directional alignment.

## References

- ALB ping-pong feature documentation: https://argo-rollouts.readthedocs.io/en/stable/features/traffic-management/alb/#zero-downtime-updates-with-ping-pong-feature
- PR #3628 (scoped pingpong-vs-canary/stable precedence to ALB)
- AWS LBC NLB weighted target groups fix: kubernetes-sigs/aws-load-balancer-controller#4732
- Gateway API plugin: https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi

ajax-boiko-y:
## Motivation

Closes #4739.

Plugin-based traffic routers (e.g. the [Gateway API plugin](https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi)) need to implement ping-pong support for zero-downtime rollouts with long-lived TCP/gRPC connections (e.g. AWS NLB with weighted target groups via AWS Load Balancer Controller).

Currently the validation in `ValidateRolloutStrategyCanary` rejects any `pingPong` configuration unless `trafficRouting.alb` or `trafficRouting.istio` is set, blocking plugins from using the feature at all.

## What this PR does

Relaxes the validation condition to also permit `trafficRouting.plugins` — matching exactly the pattern used in #4643 for traffic mirroring.

**No controller logic changes are required.** Verified by reading the source:
- `reconcilePingAndPongService` (`rollout/service.go`) has no router-type check.
- `GetStableAndCanaryServices` is already router-agnostic.
- `ensureSVCTargets` short-circuits on empty `svcName`, so no selector swap occurs when only `pingPong` is configured.
- `requireCanaryStableServices` does not include `Plugins` in its required-services list.

## Testing

Added a unit test: rollout with `pingPong` + `trafficRouting.plugins` → validation passes.

Existing test `ping-pong feature without the ALB traffic routing` (Nginx router → rejected) is unchanged and still passes.

## Checklist

- [x] Followed the [contributing guidelines](https://github.com/argoproj/argo-rollouts/blob/master/CONTRIBUTING.md)
- [x] DCO signed (`-s`)
- [x] Unit tests added
- [ ] Added entry in `USERS.md` (organization will be added once the plugin PR is open)

### DIFF
--- a/pkg/apis/rollouts/validation/validation_test.go
+++ b/pkg/apis/rollouts/validation/validation_test.go
@@ -386,6 +386,19 @@ func TestValidateRolloutStrategyCanary(t *testing.T) {
 		assert.Equal(t, PingPongWithRouterOnlyMessage, allErrs[0].Detail)
 	})
 
+	t.Run("ping-pong feature with plugin-based traffic routing is valid", func(t *testing.T) {
+		validRo := ro.DeepCopy()
+		validRo.Spec.Strategy.Canary.Steps[0].SetWeight = ptr.To[int32](10)
+		validRo.Spec.Strategy.Canary.PingPong = &v1alpha1.PingPongSpec{PingService: "ping", PongService: "pong"}
+		validRo.Spec.Strategy.Canary.TrafficRouting = &v1alpha1.RolloutTrafficRouting{
+			Plugins: map[string]json.RawMessage{
+				"my-org/my-plugin": []byte(`{}`),
+			},
+		}
+		allErrs := ValidateRolloutStrategyCanary(validRo, field.NewPath(""))
+		assert.Empty(t, allErrs)
+	})
+
 	t.Run("invalid traffic routing", func(t *testing.T) {
 		invalidRo := ro.DeepCopy()
 		invalidRo.Spec.Strategy.Canary.CanaryService = ""
--- a/docs/features/specification.md
+++ b/docs/features/specification.md
@@ -179,6 +179,17 @@ spec:
       # stable pods. Required for traffic routing.
       stableService: stable-service
 
+      # Ping-pong spec allows zero-downtime rollouts for long-lived TCP/gRPC
+      # connections by avoiding service selector swaps at promotion time.
+      # Instead of swapping selectors between canaryService/stableService,
+      # the rollout alternates which of the two persistent services is
+      # "stable" via Status.Canary.StablePingPong. Supported with ALB,
+      # Istio, and plugin-based traffic routers.
+      # When pingPong is set, canaryService and stableService are not required.
+      pingPong:
+        pingService: ping-service
+        pongService: pong-service
+
       # Metadata which will be attached to the canary pods. This metadata will
       # only exist during an update, since there are no canary pods in a fully
       # promoted rollout.
@@ -451,6 +462,7 @@ spec:
           ingress: ingress # required
           servicePort: 443 # required
           annotationPrefix: custom.alb.ingress.kubernetes.io # optional
+          rootService: root-service # required when ping-pong is enabled
 
         # Service Mesh Interface routing configuration
         smi:
--- a/pkg/apis/rollouts/validation/validation.go
+++ b/pkg/apis/rollouts/validation/validation.go
@@ -81,8 +81,8 @@ const (
 	DuplicatedPingPongServicesMessage = "This rollout uses the same service for the ping and pong services, but two different services are required."
 	// MissedAlbRootServiceMessage indicates that the rollout with ALB TrafficRouting and ping pong feature enabled must have root service provided
 	MissedAlbRootServiceMessage = "Root service field is required for the configuration with ALB and ping-pong feature enabled"
-	// PingPongWithRouterOnlyMessage At this moment ping-pong feature works with the ALB traffic routing only
-	PingPongWithRouterOnlyMessage = "Ping-pong feature works with the ALB and Istio traffic routers only"
+	// PingPongWithRouterOnlyMessage At this moment ping-pong feature works with the ALB, Istio, and plugin-based traffic routers only
+	PingPongWithRouterOnlyMessage = "Ping-pong feature works with the ALB, Istio, and plugin-based traffic routers only"
 	// InvalideStepRouteNameNotFoundInManagedRoutes A step has been configured that requires managedRoutes and the route name
 	// is missing from managedRoutes
 	InvalideStepRouteNameNotFoundInManagedRoutes = "Steps define a route that does not exist in spec.strategy.canary.trafficRouting.managedRoutes"
@@ -265,7 +265,7 @@ func ValidateRolloutStrategyCanary(rollout *v1alpha1.Rollout, fldPath *field.Pat
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("stableService"), canary.StableService, DuplicatedServicesCanaryMessage))
 	}
 	if canary.PingPong != nil {
-		if canary.TrafficRouting != nil && canary.TrafficRouting.ALB == nil && canary.TrafficRouting.Istio == nil {
+		if canary.TrafficRouting != nil && canary.TrafficRouting.ALB == nil && canary.TrafficRouting.Istio == nil && len(canary.TrafficRouting.Plugins) == 0 {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("trafficRouting").Child("alb"), canary.TrafficRouting.ALB, PingPongWithRouterOnlyMessage))
 		}
 		if canary.PingPong.PingService == "" {


pkg/apis/rollouts/validation/validation_test.go
```diff
@@ -386,6 +386,19 @@ func TestValidateRolloutStrategyCanary(t *testing.T) {
 		assert.Equal(t, PingPongWithRouterOnlyMessage, allErrs[0].Detail)
 	})
 
+	t.Run("ping-pong feature with plugin-based traffic routing is valid", func(t *testing.T) {
+		validRo := ro.DeepCopy()
+		validRo.Spec.Strategy.Canary.Steps[0].SetWeight = ptr.To[int32](10)
+		validRo.Spec.Strategy.Canary.PingPong = &v1alpha1.PingPongSpec{PingService: "ping", PongService: "pong"}
```
kostis-codefresh, line 392: It seems to me that ping/pong is not documented at https://argo-rollouts.readthedocs.io/en/stable/features/specification/

Could you also add it there? It can be in a separate PR as well.

kostis-codefresh (approved):
LGTM. 

I added a minor comment that we should also update the spec page with ping/pong. This was an existing omission, not blocking this PR in any way.

zachaller (approved):

Issue #2758, opened 2023-05-05
Dashboard crashes when result from AnalysisRun is [NaN]


FredrikAugust:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

When clicking the "Analysis XX-post" button in the dashboard, the dashboard crashes. The console logs are

```
SyntaxError: Unexpected token 'N', "[NaN]" is not valid JSON
    at JSON.parse (<anonymous>)
    at revision.tsx:209:86
    at Array.map (<anonymous>)
    at c (revision.tsx:201:55)
    at ni (react-dom.production.min.js:157:137)
    at Ii (react-dom.production.min.js:180:154)
    at Ws (react-dom.production.min.js:269:343)
    at Ll (react-dom.production.min.js:250:347)
    at Ml (react-dom.production.min.js:250:278)
    at vl (react-dom.production.min.js:250:138)
```

Which points to this line of code: https://github.com/argoproj/argo-rollouts/blob/f72e565960e074fd885c14d88334168859c44a66/ui/src/app/components/rollout/revision.tsx#L220

**To Reproduce**

I have an analysistemplate which queries a linkerd prometheus instance which returns `0 / 0` which in turn becomes `[NaN]` by the time it hits the analysistemplate.

**Expected behavior**

It should not crash.

**Screenshots**

![Screenshot 2023-05-05 at 13 18 52](https://user-images.githubusercontent.com/7691547/236444296-09aed484-c296-48f8-b08f-9fdc7c6213a8.png)
![Screenshot 2023-05-05 at 13 19 03](https://user-images.githubusercontent.com/7691547/236444330-435075c0-a713-4571-8de8-e3299c058a54.png)

**Version**

`quay.io/argoproj/kubectl-argo-rollouts:v1.4.1`

**Logs**

```
time="2023-05-05T11:17:03Z" level=info msg="Enqueueing analysis after 4m59.970939132s" analysisrun=init-forms-665c967c8d-18-post namespace=init
time="2023-05-05T11:17:03Z" level=info msg="No status changes. Skipping patch" analysisrun=init-forms-665c967c8d-18-post namespace=init
time="2023-05-05T11:17:03Z" level=info msg="Reconciliation completed" analysisrun=init-forms-665c967c8d-18-post namespace=init time_ms=1.374922
[...]

PR #4744, opened 2026-05-27
fix(dashboard): guard null chartValue in metric chart tooltip. Fixes #4743

### LINKED ISSUE #4743: Dashboard crashes on metric chart tooltip hover when AnalysisRun measurement value is [NaN]
Checklist:

* [X] I've included steps to reproduce the bug.
* [X] I've included the version of argo rollouts.

**Describe the bug**

When the analysis modal renders the metric chart for an AnalysisRun whose measurement value is `[NaN]`, hovering the corresponding data point crashes the dashboard to a blank white screen. The table view does not crash for the value.

**To Reproduce**

Minimal AnalysisRun that reproduces the crash whenever Prometheus returns 0 / 0 (no traffic — a common case in canary analysis):

```yaml
kind: AnalysisRun
metadata:
  name: nan-tooltip-repro
  namespace: default
spec:
  metrics:
    - name: 5xx-ratio
      successCondition: len(result) == 0 || isNaN(result[0]) || result[0] <= 0.05
      interval: 30s
      count: 10
      provider:
        prometheus:
          address: prometheus-server.address
          query: |
            (
              sum(increase(nginx_ingress_controller_requests{
                ingress="my-application",
                canary!="",
                status=~"5.."
              }[1m]))
              or vector(0)
            )
            /
            sum(increase(nginx_ingress_controller_requests{
              ingress="my-application"
              canary!="",
            }[1m]))
```

Resulting measurement (truncated):

```
Status:
  Metric Results:
    Measurements:
      Finished At:  2026-05-26T18:03:20Z
      Phase:        Successful
      Started At:   2026-05-26T18:03:20Z
      Value:        [NaN]
```

Steps:

1. Apply the AnalysisRun (or trigger one via a Rollout using the same successCondition against a service with no traffic) — the controller marks the measurement Successful with Value: [NaN] (correct behavior per the success condition).
2. Open the dashboard.
3. Navigate to the Rollout, click the analysis badge to open the analysis modal.
4. Select the metric (e.g. 5xx-ratio) and switch to chart view.
5. Hover the data point for the [NaN] measurement.

The dashboard unmounts to a blank white screen. The console shows:

```
TypeError: Cannot read properties of null (reading '0')
    at metric-chart.tsx:59:158
    at Array.map (<anonymous>)
    at b (metric-chart.tsx:59:41)
    at react-dom.production.min.js:...
```

**Expected behavior**

Hovering an unchartable (NaN) measurement should render the tooltip with the timestamp and status indicator, gracefully omitting the numeric value (the same way the table view handles it). No crash, no console error.

**Screenshots**

https://github.com/user-attachments/assets/fc468c9a-1f28-4753-8749-acc8bf2f9ef5

**Version**

- Controller: `v1.9.0`
- Dashboard: `v1.9.0`
  - Issue is tested and verified that it exists on the current `master` branch too (200f934fd6c94c39ab29f74547f0551af50062ab). The screen recording is from that commit on the `master` branch.

**Logs**

The controller behaves correctly — it reports the measurement as Successful per the `isNaN(result[0])` clause in the success condition. The bug is in the dashboard UI rendering layer, not in the controller.

---

berk-karaal:
Closes #4743.

Checklist:

## Summary

The analysis modal metric chart crashes the dashboard to a blank white screen when the user hovers a data point whose measurement value is `[NaN]` (or any other value that fails `JSON.parse`). The crash is a `TypeError: Cannot read properties of null (reading '0')` at `ui/src/app/components/analysis-modal/metric-chart/metric-chart.tsx:59`.

This is the same crash family as closed issue #2758. PR #3801 fixed it at the time by switching to `JSON5.parse`. PR #3881 reverted to `JSON.parse + try/catch` for other reasons, which re-introduced the consumer-side crash on a new surface (the metric chart tooltip rather than `revision.tsx`).

## Root cause

When `transformMeasurementValue` cannot parse the measurement value, it returns the documented null sentinel:

```ts
return { canChart: true, chartValue: null, tableValue: null };
```

The metric chart tooltip then dereferences it without a null guard:

```ts
const sublabels = conditionKeys.map((cKey) => (conditionKeys.length > 1
    ? `${valueFormatter(data.chartValue[cKey])} (${cKey})`
    : valueFormatter(data.chartValue[cKey])));
```

null['0'] throws, no error boundary catches it, the whole tree unmounts. The table view path does not crash because it already null-guards.

## Fix

Guard the `chartValue` dereference with optional chaining + nullish coalescing, matching the formatter's existing null contract (`valueFormatter` already accepts `number | string | null`).

The tooltip label derivation was extracted out of `TooltipContent` into a small pure helper, `getTooltipLabel` (`ui/src/app/components/analysis-modal/metric-chart/tooltip-label.ts`), so the guard sits at — and can be unit-tested at — the exact code path that crashed.

Why consumer-side instead of patching `transforms.ts`:

1. Preserves the existing transforms contract. `chartValue: null` is the documented sentinel for unchartable data; the existing `transforms measurements correctly with NaN numbers in list` test in `transforms.test.ts` asserts exactly that. A data-layer fix invalidates it.
2. Smaller blast radius — the guard lives at the actual crash site.
3. Generalizes. Defends every null-`chartValue` path — NaN, Infinity, -Infinity, malformed JSON, future provider quirks — not only NaN.

## Tests

Two layers:

**`getTooltipLabel` (`tooltip-label.test.ts`) — validates the fix.** The label logic was extracted from `TooltipContent` into a pure `getTooltipLabel` helper so it can be tested directly. The null-`chartValue` cases throw `TypeError: Cannot read properties of null (reading '0')` *without* the guard and pass *with* it — so the test exercises the exact code path that crashed.

**`transformMeasurementValue` (`transforms.test.ts`) — pins the data-layer contract the guard relies on.** Unchartable values collapse to `{ canChart: true, chartValue: null, tableValue: null }`, covered for: `undefined`, empty string, `[NaN]`, bare `NaN`, `[NaN, 5]`, `[Infinity]`, `[-Infinity]`, bare `Infinity`, and malformed JSON. `transformMeasurementValue` is now exported so it can be tested directly. These complement the fix by documenting the contract every chart consumer depends on; the `getTooltipLabel` test above is what validates the fix itself.

## Fix Demo

https://github.com/user-attachments/assets/16a35326-ff51-4f76-94c7-7f23d023e283

## Reproducer

See #4743 for the full reproducer (AnalysisRun YAML, steps and crash screen recording).

### DIFF
--- a/ui/src/app/components/analysis-modal/metric-chart/metric-chart.tsx
+++ b/ui/src/app/components/analysis-modal/metric-chart/metric-chart.tsx
@@ -56,10 +56,13 @@ const TooltipContent = ({active, conditionKeys, payload, valueFormatter}: Toolti
     if (data.phase === AnalysisStatus.Error) {
         label = data.message ?? 'Measurement error';
     } else if (conditionKeys.length > 0) {
-        const sublabels = conditionKeys.map((cKey) => (conditionKeys.length > 1 ? `${valueFormatter(data.chartValue[cKey])} (${cKey})` : valueFormatter(data.chartValue[cKey])));
+        const sublabels = conditionKeys.map((cKey) => {
+            const v = data.chartValue?.[cKey] ?? null;
+            return conditionKeys.length > 1 ? `${valueFormatter(v)} (${cKey})` : valueFormatter(v);
+        });
         label = sublabels.join(' , ');
     } else {
-        label = valueFormatter(data.chartValue);
+        label = valueFormatter(data.chartValue ?? null);
     }
 
     return (
--- a/ui/src/app/components/analysis-modal/metric-chart/metric-chart.tsx
+++ b/ui/src/app/components/analysis-modal/metric-chart/metric-chart.tsx
@@ -8,6 +8,7 @@ import {Typography} from 'antd';
 import {AnalysisStatus, FunctionalStatus, TransformedMeasurement} from '../types';
 import {ANALYSIS_STATUS_THEME_MAP} from '../constants';
 import {isValidDate} from '../transforms';
+import {getTooltipLabel} from './tooltip-label';
 
 import StatusIndicator from '../status-indicator/status-indicator';
 
@@ -52,18 +53,7 @@ const TooltipContent = ({active, conditionKeys, payload, valueFormatter}: Toolti
     }
 
     const data = payload[0].payload;
-    let label;
-    if (data.phase === AnalysisStatus.Error) {
-        label = data.message ?? 'Measurement error';
-    } else if (conditionKeys.length > 0) {
-        const sublabels = conditionKeys.map((cKey) => {
-            const v = data.chartValue?.[cKey] ?? null;
-            return conditionKeys.length > 1 ? `${valueFormatter(v)} (${cKey})` : valueFormatter(v);
-        });
-        label = sublabels.join(' , ');
-    } else {
-        label = valueFormatter(data.chartValue ?? null);
-    }
+    const label = getTooltipLabel(data, conditionKeys, valueFormatter);
 
     return (
         <div className={cx('metric-chart-tooltip')}>
--- a/ui/src/app/components/analysis-modal/metric-chart/tooltip-label.test.ts
+++ b/ui/src/app/components/analysis-modal/metric-chart/tooltip-label.test.ts
@@ -1,0 +1,20 @@
+import {getTooltipLabel} from './tooltip-label';
+import {AnalysisStatus} from '../types';
+
+const valueFormatter = (value: number | string | null) => (value === null ? '' : value.toString());
+
+describe('getTooltipLabel()', () => {
+    test.each([
+        ['single condition key with a value', {phase: AnalysisStatus.Successful, chartValue: {0: 5}}, ['0'], '5'],
+        ['multiple condition keys with values', {phase: AnalysisStatus.Successful, chartValue: {0: 5, 1: 10}}, ['0', '1'], '5 (0) , 10 (1)'],
+        ['no condition keys with a scalar value', {phase: AnalysisStatus.Successful, chartValue: 5}, [], '5'],
+        ['error phase with a message', {phase: AnalysisStatus.Error, message: 'boom'}, ['0'], 'boom'],
+        ['error phase without a message', {phase: AnalysisStatus.Error}, ['0'], 'Measurement error'],
+        ['null chartValue with a single condition key', {phase: AnalysisStatus.Successful, chartValue: null}, ['0'], ''],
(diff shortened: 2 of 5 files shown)


berk-karaal:
Hi @kostis-codefresh, could you review this PR when you have time? Tests + manual A/B verification details are in the description. Thanks!

kostis-codefresh (changes requested):
Many thanks for the detailed analysis and the video recordings :-)

The Sonar check fails however. I suspect because of duplicated code. Can you make the new tests tabular/parameterized? I think that would solve the duplication.

berk-karaal:
Refactored tests, Sonar check passes now

kostis-codefresh:
I am a bit confused about the tests. The fix was on the UI level for TooltipContent. But the new tests are on transformMeasurementValue which the PR doesn't seem to touch. So do the new tests actually validate the fix? Shouldn't this PR have a new test testing explicitly the way TooltipContent works (as the fix was done there)?

Or the added tests are not validating the fix, and you just added them to strengthen the data layer testing suite?

Disclaimer: I am not an expert on UI code, so I am happy to be corrected here.

berk-karaal:
@kostis-codefresh you're right. The `transformMeasurementValue` tests only validate the data-layer contract, they don't validate the actual fix. I had added them that way because the project doesn't have any React component rendering tests where I could test the fix directly.

But on second thought after your comment, I separated the tooltip label generation logic (the place where the dashboard was crashing at) out of the `TooltipContent` component into its own TypeScript function. That makes it easy to write unit-test without a component rendering requirement. The tests are added in the last commit (7a16dbb), covering the actual bug-fix.

---

About the SonarCloud check: it reports "0.0% Coverage on New Code". Sonar config excludes all TypeScript from coverage (`sonar.coverage.exclusions` includes `**/*.ts*`) and the UI jest suite doesn't run in CI. The added tests are only for local/reviewer validation I guess 🤔. If that's the case, I can take a look at writing CI for running TypeScript unit tests on a separate PR

kostis-codefresh:
> Sonar config excludes all TypeScript from coverage

Actually, the biggest problem for me is that UI tests are currently not run by GitHub actions at all. We should fix that first before dealing with SonarCube coverage.

> If that's the case, I can take a look at writing CI for running TypeScript unit tests on a separate PR

That would be great, thank you. As you did here, please open an associated GitHub issue first with a clear description of what needs to happen, before actually creating a PR.

kostis-codefresh (approved):
LGTM, but to get real value from the added tests, we also need a separate PR for running UI tests in GitHub actions.

zachaller (approved):

Issue #4571, opened 2025-12-30
Possible memory leak in controller - grows from ~140Mi to 8.8GB over 127 days


jacobk-papaya:
### Checklist

- [x] I've included steps to reproduce the bug.
- [x] I've included the version of argo rollouts.

### Describe the bug

The Argo Rollouts controller exhibits a memory leak where memory consumption grows continuously over time. After 127 days of uptime, the leader controller pod consumed 8.8GB of memory. After restarting, the same pod uses only ~140Mi.

### Environment

- **Argo Rollouts version:** v1.8.3
- **Kubernetes version:** v1.33 (EKS)
- **HA mode:** Yes (2 replicas with leader election)

### Observed Behavior

| Pod | Role | Uptime | Memory |
|-----|------|--------|--------|
| argo-rollouts-7ccd66b9d5-g4rvs | Leader | 127 days | 8877Mi |
| argo-rollouts-7ccd66b9d5-78v26 | Standby | 110 days | 34Mi |

After restart:

| Pod | Role | Uptime | Memory |
|-----|------|--------|--------|
| argo-rollouts-6f589bd668-9hkn6 | Leader | ~1.5 hours | 142Mi |
| argo-rollouts-6f589bd668-schtq | Standby | ~1.5 hours | 31Mi |

### Cluster Details

- **Rollouts:** 25
- **AnalysisRuns:** 135 (active)
- **Deployment frequency:** High (continuous deployments throughout the day)

### Controller Configuration

```
--healthzPort=8080
--metricsport=8090
--loglevel=info
--logformat=text
--kloglevel=0
--leader-elect
--aws-verify-target-group
--rollout-threads=20
--analysis-threads=30
```

No resource limits were configured (empty `resources: {}`).

### Steps to Reproduce

1. Deploy Argo Rollouts v1.8.3 in HA mode (2 replicas)
2. Configure multiple rollouts with analysis templates
3. Perform frequent deployments over weeks/months
4. Monitor memory consumption of the leader pod

### Expected Behavior

Memory usage should remain stable over time, not grow unbounded.

### Additional Context

- The standby pod (non-leader) maintains low memory (~31-34Mi) regardless of uptime
- Only the leader pod exhibits the memory growth
- Similar issue reported in #2443 (closed without resolution)
- The dashboard component shows the same leak pattern (15GB after 110 days)

Issue #517, opened 2020-05-21
controller-gen not foud in $PATH


Ivan-Zhu:
make codegen
./hack/update-mocks.sh
./hack/update-codegen.sh
Generating deepcopy funcs
Generating clientset for rollouts:v1alpha1 at github.com/argoproj/argo-rollouts/pkg/client/clientset
Generating listers for rollouts:v1alpha1 at github.com/argoproj/argo-rollouts/pkg/client/listers
Generating informers for rollouts:v1alpha1 at github.com/argoproj/argo-rollouts/pkg/client/informers
./hack/update-openapigen.sh
go run ./hack/gen-crd-spec/main.go
panic: exec: "controller-gen": executable file not found in $PATH

jessesuen:
controller-gen is built in `dist` directory and the makefile add the dist directory in the path

Issue #389, opened 2020-01-29
Spec validation failed in Rollout CR


vgunapati:
Spec validation failed in Rollout CR. This will block all rollout CR's in the cluster 

```
.Rollout: v1alpha1.Rollout.Spec: v1alpha1.RolloutSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Resources: v1.ResourceRequirements.Limits: unmarshalerDecoder: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$', error found in #10 byte of ...|pu":"nill","memory":|..., bigger context ...|outSeconds":1},"resources":{"limits":{"cpu":"nill","memory":"12400M"},"requests":{"cpu":"2","memory"|...
E0129 05:37:15.324090       1 reflector.go:123] github.com/argoproj/argo-rollouts/pkg/client/informers/externalversions/factory.go:117: Failed to list *v1alpha1.Rollout: v1alpha1.RolloutList.Items: []v1alpha1.Rollout: v1alpha1.Rollout.Spec: v1alpha1.RolloutSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Resources: v1.ResourceRequirements.Limits: unmarshalerDecoder: quan# Please edit the object below. Lines beginning with a '#' will be ignored,
```

If rollout CR is not valid, CR should be rejected

jessesuen:
This is a known issue which should be addressed via validation webhook, since Kubernetes OpenAPI validation has gaps/limits, namesly it cannot handle `intstr.StringOrInt` data type.

An alternative to validation webhook, is to use an unstructured informer which updates/saves the rollout as a valid Rollout object which will not fail unmarshalling.

jessesuen:
Fixed via a tolerant informer which does best effort unmarshalling.

PR #4745, opened 2026-05-27
fix(controller): convert unstructured→typed to avoid memory bloat

niels-s:
Replace per-call reflection-based conversion in tolerantinformer List/Get with a cache.SharedIndexInformer.SetTransform that runs once on object insert/update. The cache now stores typed objects and the generated typed listers from pkg/client/listers can be used directly.

Why: heap profiling (issue #4571) showed ~60% of allocations in the controller flow through tolerantinformer.convertObject → runtime.DefaultUnstructuredConverter.FromUnstructured. The path is hit on every Prometheus scrape and every rollout reconcile, churning hundreds of MB/s and inflating RSS over time.

Behavior preserved:
- Malformed-object tolerance (PR #666 / issues #389, #517): the transform still falls back to encoding/json when FromUnstructured fails fast, so partially-decoded objects remain usable.
- Public constructor signatures unchanged; callers in main.go, controller managers, and downstream consumers are not touched.

Implementation notes:
- SetTransform must be installed before SharedIndexInformer.Run; the constructors run during wiring (before factory.Start), and panic if invoked after start.
- Test fixture refactored: newFakeDynamicInformer constructs the tolerant wrappers before calling factory.Start, returns a fakeInformers struct that tests reuse instead of re-constructing.

Net diff: -227 lines (custom listers replaced by generated typed listers).

Checklist:

### DIFF
--- a/utils/tolerantinformer/analysisrun.go
+++ b/utils/tolerantinformer/analysisrun.go
@@ -1,8 +1,6 @@
 package tolerantinformer
 
 import (
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/tools/cache"
@@ -13,9 +11,11 @@ import (
 )
 
 func NewTolerantAnalysisRunInformer(factory dynamicinformer.DynamicSharedInformerFactory) rolloutinformers.AnalysisRunInformer {
-	return &tolerantAnalysisRunInformer{
-		delegate: factory.ForResource(v1alpha1.AnalysisRunGVR),
-	}
+	delegate := factory.ForResource(v1alpha1.AnalysisRunGVR)
+	installTransform(delegate.Informer(),
+		makeTransform(func() *v1alpha1.AnalysisRun { return &v1alpha1.AnalysisRun{} }),
+		"AnalysisRun")
+	return &tolerantAnalysisRunInformer{delegate: delegate}
 }
 
 type tolerantAnalysisRunInformer struct {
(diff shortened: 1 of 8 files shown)


niels-s:
## Allocation impact — before/after benchmark

Added `BenchmarkAnalysisRunListerList` in `utils/tolerantinformer/bench_test.go` (commit 0281726). It exercises the hot path: namespaced AnalysisRun lister `List` at N=10/100/1000 ARs — the same path called on every Prometheus scrape and every Rollout reconcile (the path flagged by the heap profile in #4571).

Compared `master` (200f934fd) vs this branch (028172698) using `go test -bench BenchmarkAnalysisRunListerList -benchmem -benchtime=2s -count=8` on the same machine (Apple M1 Max, Go 1.26.3). `benchstat` output:

### Time per call (sec/op)

| N | master | branch | delta |
|---|-------:|-------:|------:|
| 10   | 114.59 µs | 9.94 µs  | **−91.33%** (p<0.001) |
| 100  | 1.11 ms   | 89.90 µs | **−91.92%** (p<0.001) |
| 1000 | 10.76 ms  | 829.3 µs | **−92.29%** (p<0.001) |
| geomean | 1.11 ms | 90.49 µs | **−91.86%** |

### Allocations per call (allocs/op)

| N | master | branch | delta |
|---|-------:|-------:|------:|
| 10   | 300    | 69    | **−77.00%** |
| 100  | 2,913  | 612   | **−78.99%** |
| 1000 | 29,016 | 6,015 | **−79.27%** |
| geomean | 2,938 | 633 | **−78.44%** |

### Bytes per call (B/op)

| N | master | branch | delta |
|---|-------:|-------:|------:|
| 10   | 27.13 KiB | 23.27 KiB | −14.25% |
| 100  | 269.2 KiB | 230.4 KiB | −14.43% |
| 1000 | 2.62 MiB  | 2.24 MiB  | −14.43% |

### Why the gap widens with N

- Master: every `List` reflection-converts every AR (`runtime.DefaultUnstructuredConverter.FromUnstructured` × N). Cost scales linearly and is reflection-bound — slow and alloc-heavy.
- Branch: `SetTransform` converts each AR once on insert/update; `List` walks the cached typed pointers and `DeepCopy()`s each (generated code, no reflection). Cost still scales linearly but with a much smaller per-AR factor.

### Reproduce

```
go test -run XXX -bench BenchmarkAnalysisRunListerList -benchmem \
  -benchtime=2s -count=8 ./utils/tolerantinformer/
```

Compare branches via [`benchstat`](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat).

### Note

GC pressure scales with **alloc count**, not bytes. The ~80% drop in allocs/op is the load-bearing number for the controller-RSS-balloon symptom reported in #4571; the time and bytes wins are gravy.

niels-s:
It's early days, but we're running this fix + #4666 in our production cluster. We're monitoring as we go and see if things remain stable. Previously, we saw a big memory build-up over time and performance degradation, as if the controller got stuck.
We will report back if this is helping or not

niels-s:
We've been running this PR in production, we see significant performance improvements in both CPU and Memory, see chart below. And our engineers haven't complained anymore if Argo rollouts getting stuck

CleanShot 2026-06-15 at 12 46 20@2x

zachaller:
This is great to see, I want to merge this first: https://github.com/argoproj/argo-rollouts/pull/4674/changes I will do that here probably today and it will affect this PR a bit.

zachaller:
@niels-s are you able to resolve the conflicts

niels-s:
@zachaller conflicts are resolved

zachaller (approved):

Issue #4268, opened 2025-04-30
Fix flaky tests and make builds deterministic


kostis-codefresh:
# Summary

There are several E2E tests that break all the time. This makes contributor work very difficult.

# Use Cases

Examples of commits with failed tests

https://github.com/argoproj/argo-rollouts/actions/runs/14712277237/job/41287384542
https://github.com/argoproj/argo-rollouts/actions/runs/13277626443/job/37069985746
https://github.com/argoproj/argo-rollouts/actions/runs/14755408824/job/41422279926

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

kostis-codefresh:
Related https://github.com/argoproj/argo-rollouts/issues/1200

PR #4746, opened 2026-05-27
fix(metricprovider): parse grouped Datadog queries (by {tag}) instead of crashing. Fixes #4276

### LINKED ISSUE #4276: Support evaluating grouped Datadog metrics (`by {tag}`) in AnalysisTemplates

cgreeno:
## What this fixes

Datadog v2 queries using `by {tag}` crash today. The `/api/v2/query/scalar` endpoint returns a string "group" column alongside the number column, but the response struct types every column's values as `[]*float64`, so the unmarshal fails:

```
Could not parse JSON body: json: cannot unmarshal string into Go struct field
.Data.Attributes.Columns.Values of type []*float64
```

This is a rework of #4731 — @kostis-codefresh pointed out the custom `reducer` field there reinvented what other providers already do with Expr, so this version drops the new API field entirely.

## What changed

- Column values are now unmarshalled as `[]json.RawMessage` and interpreted based on the column `type`. `findColumns` picks out the number and group columns, and `extractValues` pulls the floats (skipping nulls) and pairs each one with its group tag via `groupNameAt`.
- Single-value responses still return a scalar, so existing `successCondition: result < X` configs behave exactly as before. Grouped responses return `[]float64` — same as the Prometheus provider does for vector results — so you can write `max(result) < 0.05`, `all(result, # < 0.05)`, etc.
- For grouped queries the measurement also gets `metadata.groups`, a JSON array of `{"name": ..., "value": ...}` pairs so you can tell which group tripped the condition (JSON rather than CSV because tag values can legally contain `,` and `=`).
- Docs updated with a grouped-query example.

```yaml
metrics:
- name: per-endpoint-error-rate
  interval: 30s
  successCondition: max(result) < 0.05
  failureLimit: 3
  provider:
    datadog:
      apiVersion: v2
      interval: 5m
      query: "sum:trace.http.request.errors{service:my-service} by {resource_name}.as_count()"
```

## Backwards compatibility

- **v1 users:** `parseResponseV1` is untouched. The wrapper passes through with no groups, so metadata is never set for v1.
- **v2 ungrouped queries:** same scalar, same formatting (`strconv.FormatFloat(v, 'f', -1, 64)`), same evaluation, no metadata.
- **Columns without a `type` field** (older fixtures / responses) are treated as numeric, preserving the old "first column wins" behavior.
- Two small deltas, both on paths that already errored or returned no usable value:
  - An empty / all-null result now reports `[]` as the measurement value instead of the marshalled raw column (e.g. `[null]`). Evaluation is unchanged (still evaluated as a nil `*float64`).
  - A non-numeric entry in a number column now errors with `could not parse numeric value ...` instead of the generic JSON unmarshal error. The one existing test asserting the old error string was updated accordingly.

## Tests

New table cases in `datadogV2_test.go`: grouped query with `max(result)`, grouped with `all(result, # < X)`, grouped with a null entry (checks group/value alignment), non-numeric value in a number column, and a populated Datadog `errors` response. Unit tests in `datadog_test.go` for `extractValues`, `groupNameAt`, and a `parseResponseV2` body-read-error case.


cgreeno:
@kostis-codefresh - Thanks again for the previous review and guidance!

Hopefully this looks a bit better? If not feedback more than welcome. :)

cgreeno:
Heya @kostis-codefresh - just seeing if you had a chance to do a review. More than happy to hear your feedback

cgreeno:
CI's `Run unit tests` failed on `rollout.TestRolloutUseDynamicWeightOnPromoteFull` (`rollout/controller_test.go:788`) — appears to be an unrelated flake; this PR only touches `metricproviders/datadog/`, the same test passes on the other recently-opened PRs (#4761–#4767), and both `DynamicStableScale_true`/`DynamicStableScale_false` subtests passed (only the parent's setup-level assertion fired).

Traced it to a cache-sync race in the test fixture, not the test itself:

`rollout/controller_test.go:783` only waits for the ReplicaSet and Rollout informer caches:
```go
assert.True(f.t, cache.WaitForCacheSync(stopCh, c.replicaSetSynced, c.rolloutsSynced))
```
…but the reconciler validates `spec.strategy.canary.stableService` via the Services lister, which reads from the Services informer cache. On a slow CI runner the Services cache hasn't populated yet → validation reports `service "stable" not found` → expected `UpdateReplicaSetAction` never fires. The subtests still log as PASS because `f.t` points at the parent's `*testing.T`, so all the `f.t.Errorf` calls roll up to the parent (which is why only the parent shows red).

A one-line fix would be to extend the `WaitForCacheSync` call to include the Services informer (and probably any other listers the reconciler queries).

@zachaller / @kostis-codefresh — could you re-run the failed job when convenient? Happy to push the cache-sync fix as a separate PR while #4746 is in review (keeping #4746 strictly the Datadog parse fix), or you can track it under #4268 for someone else to pick up — whichever you prefer.

kostis-codefresh:
Can you please cleanup the PR description and mention only/exactly what was implemented please?

 > Response struct now keeps Values []json.RawMessage per column, with Numeric() and Strings() helpers that filter to the typed values caller cares about. Nulls (Datadog's "no point" marker) are dropped.

I don't see any such helpers in the diff.

Also it would be great if you can mention your analysis about backwards compability. We should be certain that nothing is broken for existing datadog users (either the v1 users or v2 users  that do not use tag queries at all).

cgreeno:
@kostis-codefresh - updated PR with backwards compatibility apologies for that the description was written against an earlier iteration of the code. I changed approach partway through (the `Numeric()`/`Strings()` helpers got refactored into `findColumns`/`extractValues`/`groupNameAt`) and didn't go back and update the PR text. Honestly sorry for wasting your time on the mismatch :(  I've now rewritten the description so it matches exactly what's in the diff.

On backwards compatibility, the evidence is mostly pre-existing tests that this PR
doesn't touch:

**v1** — `parseResponseV1` is unchanged, and the whole v1 suite in `datadogV1_test.go` (~13 cases) is not modified by this PR and still passes.

**v2 without grouped queries**  the existing cases in `TestRunSuiteV2` now run through the new parsing code but keep their original expectations, all still green:

- simple scalar query → same value string and phase
- value failing the condition → same failed phase
- `default(result, x)` with data / no data
- empty result (`"values": []`) → same `[]` value and nil-result handling
- 401/403 auth errors and non-2xx errors → same messages

Worth noting: all the pre-existing v2 fixtures omit the `"type"` field on columns, so the legacy "no type = numeric, first column wins" fallback is exercised by every one of those old cases, not just the new unit test for it.

The only pre-existing expectation I had to change is the "error when bad values from dd" case, where the JSON unmarshal error string differs because `Values` is now `[]json.RawMessage` instead of `[]*float64`. That's called out in the updated description, along with one other cosmetic delta (empty/all-null results now report `[]` instead of the marshalled raw column).

cgreeno:
Heya @kostis-codefresh sorry to bug you again - any chance you could take another look at this?

metricproviders/datadog/datadog.go
```diff
@@ -297,56 +379,68 @@ func (p *Provider) parseResponseV1(metric v1alpha1.Metric, response *http.Respon
...
+	// vector results.
+	if len(values) == 1 {
+		status, evalErr := evaluate.EvaluateResult(values[0], metric, p.logCtx)
+		return strconv.FormatFloat(values[0], 'f', -1, 64), status, "", evalErr
+	}
 
-		if jsonErr != nil {
-			return "", v1alpha1.AnalysisPhaseError, fmt.Errorf("Failed to marshall JSON empty Values: %v", jsonErr)
-		}
+	status, evalErr := evaluate.EvaluateResult(values, metric, p.logCtx)
```
kostis-codefresh, line 511: If I understand this correctly, here we look at the response count and if len > 1 we assume it is a grouped query. But is this a hard contract with Datadog or an implicit assumption? Would it be possible to have a single grouped/tag query ever returned? Or this never happens?

kostis-codefresh (commented):
This looks much better than the previous PR and it is contained in datadog code only. Thanks.

The only major comment I have is how the new code actually takes effect. It seems to me that we only look at the response lenght to decide if we have grouped/tagged JSON or not. But is this the correct approach?

I looked at how other providers do it. Wavefront indeed does the same. It just looks at length of response. But prometheus is much smarter. I doesn't care about the lenght of the response (1, 2, or 50 columns). It looks [specifically at "type" of the response](https://github.com/argoproj/argo-rollouts/blob/master/metricproviders/prometheus/prometheus.go#L156)

Could we do the same with Datadog?

Again,  I am not a datadog expert so an acceptable answer here is that just looking at length of query response is enough and tagged groups will only appear for len > 1

kostis-codefresh:
@meeech , @stephen-harris As you are the Datadog experts, this PR might be interesting to you. Any feedback welcome

metricproviders/datadog/datadog.go
```diff
@@ -66,14 +66,90 @@ type datadogResponseV1 struct {
...
+func isJSONNull(r json.RawMessage) bool {
+	return string(bytes.TrimSpace(r)) == "null"
+}
+
+// findColumns locates the first numeric column and the first group column in
+// a v2 scalar response. Legacy responses omit the type field, so an empty
+// type is treated as numeric for backward compatibility with the previous
+// behavior of picking column 0.
+func findColumns(columns []datadogV2Column) (numCol, groupCol *datadogV2Column) {
+	for i, c := range columns {
```
meeech: i don't know the response well enough, but theres no chance of more than one number column? because if the for should find first, then it should exit the loop as soon as its found?

metricproviders/datadog/datadog.go
```diff
@@ -66,14 +66,90 @@ type datadogResponseV1 struct {
...
+	for i, r := range numCol.Values {
+		if isJSONNull(r) {
+			continue
+		}
+		var f float64
+		if err := json.Unmarshal(r, &f); err != nil {
+			return nil, nil, fmt.Errorf("could not parse numeric value %q in column %q: %v", string(r), numCol.Name, err)
+		}
+		values = append(values, f)
+		if name, ok := groupNameAt(groupCol, i); ok {
```
meeech: you mention keeping the 2 return slices aligned. Any chance they become misaligned here? like what if a group name is not ok?

metricproviders/datadog/datadog.go
```diff
@@ -162,13 +238,18 @@ func (p *Provider) Run(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric) v1alph
...
 
-	value, status, err := p.parseResponse(metric, response, dd.ApiVersion)
+	value, status, groups, err := p.parseResponse(metric, response, dd.ApiVersion)
 	if err != nil {
 		return metricutil.MarkMeasurementError(measurement, err)
 	}
 
 	measurement.Value = value
 	measurement.Phase = status
+	if groups != "" {
```
meeech: small risk of a string literal "null" bleeding through here? but that would depend on dd returning multiple values with no group column. is that possible? i dont know the dd responses well enough - if its not possible, ignore this comment

meeech (commented):
overall looks like a great improvement. called out a few small things, feel free to ignore if they are impossible to happen.

meeech:
@kostis-codefresh thanks for the ping.

cgreeno:
> The only major comment I have is how the new code actually takes effect. It seems to me that we only look at the response lenght to decide if we have grouped/tagged JSON or not. But is this the correct approach?
> 
> I looked at how other providers do it. Wavefront indeed does the same. It just looks at length of response. But prometheus is much smarter. I doesn't care about the lenght of the response (1, 2, or 50 columns). It looks [specifically at "type" of the response](https://github.com/argoproj/argo-rollouts/blob/master/metricproviders/prometheus/prometheus.go#L156)

Ya I was not that fond of that as well, felt quite hacky - Let me take another look at the documentation on DD 

Awesome thanks so much for taking a look!

cgreeno:
Thanks again @kostis-codefresh and @meeech for the review! It is much appreciated. I looked at the OpenAPI spec and datadog documentation. 

1. Dispatch on response shape, not value count (@kostis-codefresh main point + inline :428)
`parseResponseV2` now branches on whether a group column is present (grouped flag from extractValues), not on len(values) == 1. This is the Prometheus pattern....  I confirme against Datadog's OpenAPI-generated client: the scalar response is a oneOf of DataScalarColumn{type:"number"} and GroupScalarColumn{type:"group"}.

2. Fixed the single-group bug (@kostis-codefresh  :428 "could a grouped query return one value?")
Yes it can! should of caught this and the old code silently scalar-ized it and dropped metadata.groups. Now a by {tag} query that matches one group still returns [0.087] + group metadata. New end-to-end test locks this in.

3. Fixed the real [][]string wire shape (@meeech  :146 alignment + :248 — and your test-masking concern)
 The group column's values are arrays of tag values per group ([["GET /a"],["GET /b"]]), not flat strings. The old groupNameAt unmarshalled into a string, so against real Datadog it would have silently failed and never populated metadata.groups. The tests passed only because the fixtures used the wrong flat-string shape. I corrected every fixture to the real shape and added multi-tag (by {env, resource_name} → joined prod,GET /a) coverage.

4. Alignment is now better, errors surface loudly (@meeech  :146)
In grouped mode every surviving value gets a paired group label; an unparseable label is now an error (could not parse group label at index N) rather than a dropped group that shifts the rest. New unit + end-to-end tests for it.

5. findColumns breaks early once both columns found (@meeech  :101) and is documented as "first number column wins (one per formula)".

Relevent DD doc links


The oneOf discriminator (ScalarColumn is one of the two variants):
-  [openapi_one_of lists GroupScalarColumn and DataScalarColumn.](https://datadoghq.dev/datadog-api-client-ruby/DatadogAPIClient/V2/ScalarColumn.html)

The two column variants (generated Go client, which is generated from Datadog's published OpenAPI spec):
-[ GroupScalarColumn — type: "group", Values [][]string ("the array of tag values for each group"):](https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/model_group_scalar_column.go)

- [DataScalarColumn — type: "number", Values []*float64:](https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/model_data_scalar_column.go)

I pulled the raw model_group_scalar_column.go and confirmed line-for-line:
// GroupScalarColumn A column containing the tag keys and values in a group.
```
type GroupScalarColumn struct {
      Name   *string                `json:"name,omitempty"`
      Type   *ScalarColumnTypeGroup `json:"type,omitempty"`
      // The array of tag values for each group found for the results of the formulas or queries.
      Values [][]string             `json:"values,omitempty"`
}
```

Thanks agin for the feedback!!

kostis-codefresh:
FYI Another Datadog related PR https://github.com/argoproj/argo-rollouts/pull/4812

metricproviders/datadog/datadog.go
```diff
@@ -66,14 +66,120 @@ type datadogResponseV1 struct {
...
+func isJSONNull(r json.RawMessage) bool {
+	return string(bytes.TrimSpace(r)) == "null"
+}
+
+// findColumns locates the number column and the group column in a v2 scalar
+// response. A response may carry several number columns (one per formula or
+// query); like the original implementation we evaluate the first. Legacy
+// responses omit the type field, so an empty type is treated as numeric for
+// backward compatibility with the previous behavior of picking column 0.
+func findColumns(columns []datadogV2Column) (numCol, groupCol *datadogV2Column) {
```
gferrate: Heads up — I think the multi-tag grouping doesn't do what the docs promise. I ran it against the live v2 API to be sure.

Say you gate on per-host error rate and group by two tags, `by {region,host}`. Two hosts in `us-east`, one of them spikes:

| region | host | result |
|--------|--------|--------|
| us-east | host-a | 0.01 |
| us-east | host-b | 0.09 |
| us-west | host-c | 0.02 |

`max(result) < 0.05` correctly fails, so the gate itself is fine. But `metadata.groups` comes out as:

```json
[{"name":"us-east","value":0.01},{"name":"us-east","value":0.09},{"name":"us-west","value":0.02}]
```

`host` is gone and both `us-east` rows look identical, so you can't tell which host caused the `0.09`. Since pinpointing the offending entity is the whole reason `metadata.groups` exists, it's a real loss for any query grouped by more than one tag.

The cause: `/api/v2/query/scalar` doesn't return one group column with `["prod","GET /a"]` rows. It returns a **separate group column per tag**:

```
columns:
  {name:"region", type:"group",  values:[["us-east"],["us-east"],["us-west"]]}
  {name:"host",   type:"group",  values:[["host-a"], ["host-b"], ["host-c"]]}
  {name:"query",  type:"number", values:[0.01,0.09,0.02]}
```

`findColumns` keeps only the first group column and breaks, so everything after `region` is dropped.

One aside that tripped me up: the multi-element arrays you sometimes see (like `["us-east-1","us.east.1"]`) aren't multi-tag — that's tag aliasing within a single dimension. So the join in `groupNameAt` is still needed, just not for the reason the comment gives.

### Suggested fix

Collect every group column instead of the first, then join across them per row:

```go
// returns the number column and ALL group columns, in response order
func findColumns(columns []datadogV2Column) (numCol *datadogV2Column, groupCols []*datadogV2Column) {
	for i := range columns {
		c := &columns[i]
		if numCol == nil && (c.Type == "" || c.Type == datadogV2ColumnNumber) {
			numCol = c
		}
		if c.Type == datadogV2ColumnGroup {
			groupCols = append(groupCols, c)
		}
	}
	return
}

// joins every group column's tag value(s) at idx into one label
func groupNameAt(cols []*datadogV2Column, idx int) (string, bool) {
	parts := make([]string, 0, len(cols))
	for _, col := range cols {
		if idx >= len(col.Values) {
			return "", false
		}
		var tags []string
		if err := json.Unmarshal(col.Values[idx], &tags); err != nil {
			return "", false
		}
		parts = append(parts, tags...)
	}
	return strings.Join(parts, ","), true
}
```

Then in `extractValues`, dispatch on `grouped = len(groupCols) > 0` and call `groupNameAt(groupCols, i)`. For the table above this yields `us-east,host-a`, `us-east,host-b`, `us-west,host-c` — the `prod,GET /a` shape the docs describe.

(Caveat: a real second tag and an aliased value both get flattened into the same comma list, so the label can't tell them apart. Fine for a display label, but worth a comment.)

### Same root cause, two more spots

- `datadogV2_test.go:380` and `datadog_test.go:75` feed a single group column with `["prod","GET /a"]` rows — a shape the API never sends — so they pass while the real path is broken. Worth switching them to the per-tag-column shape so they'd actually catch this.
- `docs/analysis/datadog.md:238` documents the joined `prod,GET /a` behavior, which won't happen until the above is fixed.

Happy to share the raw responses if useful.

metricproviders/datadog/datadog.go
```diff
@@ -297,56 +409,70 @@ func (p *Provider) parseResponseV1(metric v1alpha1.Metric, response *http.Respon
...
+		return strconv.FormatFloat(values[0], 'f', -1, 64), status, "", evalErr
+	}
+
+	status, evalErr := evaluate.EvaluateResult(values, metric, p.logCtx)
+
+	// For grouped queries, surface the (name, value) pairs as JSON so
+	// operators can map an outlier in `result` back to the entity that
+	// produced it. JSON-encoded rather than CSV because Datadog tag values
+	// can legally contain `,` and `=`.
+	groupsJSON, _ := json.Marshal(groups)
```
gferrate: One design thing worth sorting out before this ships: nothing bounds how many groups land in the measurement.

For a low-cardinality tag like `resource_name` it's fine. But `by {host}` or `by {pod}` on a real fleet gets big. A `by {host}` query I ran came back with ~8,500 groups. Every group gets persisted to the AnalysisRun status twice, once in the `result` slice (`formatValueSlice`) and once in `metadata.groups`. At ~8,500 entries that's a few hundred KB per measurement, and the controller keeps up to 10 measurements per metric (`DefaultMeasurementHistoryLimit`), so a single AnalysisRun can run into the megabytes.

Kubernetes caps objects around 1.5MB in etcd. Past some cardinality the status write starts failing with `request entity too large` and the run gets stuck, which is a nasty failure mode for the thing that's meant to gate the rollout.

A couple of ways to handle it:
- Cap `metadata.groups` at the top N (the worst offenders, say) and mark it when truncated. It's display-only, so trimming it doesn't affect evaluation.
- At minimum, document that this is for low-cardinality tags, not something like `by {host}`.

Not a blocker for the normal case, but right now nothing stops someone pointing it at a high-cardinality tag and quietly bricking the AnalysisRun.

metricproviders/datadog/datadog.go
```diff
@@ -66,14 +66,120 @@ type datadogResponseV1 struct {
...
+// a query can group by more than one tag (`by {env, resource_name}` →
+// ["prod", "GET /a"]); the values are joined into a single label. ok is false
+// only when the entry is missing or malformed, which callers treat as an error
+// rather than silently dropping the group.
+func groupNameAt(col *datadogV2Column, idx int) (string, bool) {
+	if col == nil || idx >= len(col.Values) {
+		return "", false
+	}
+	var tags []string
+	if err := json.Unmarshal(col.Values[idx], &tags); err != nil {
```
gferrate: Small one, but the doc comment right above this says `ok` is false when the entry is "missing or malformed" so callers can error instead of dropping a group. A JSON `null` slips past that. `json.Unmarshal([]byte("null"), &tags)` sets `tags` to nil and returns no error, so this returns `("", true)` and the value gets paired with an empty name. An empty array `[]` does the same.

The live API sends arrays so it won't bite in practice, but you wrote explicit tests for the string and number cases erroring, and `null` is the one that quietly doesn't. If you want the loud behavior the comment describes:

```go
if err := json.Unmarshal(col.Values[idx], &tags); err != nil || len(tags) == 0 {
    return "", false
}
```

gferrate (changes requested):

metricproviders/datadog/datadog.go
cgreeno: I will cap and document - I think it better to mark is as truncate and mark than just document it.

cgreeno:
Thanks @gferrate  this was a great catch, and you were right on all three. Pushed fixes for each:

1. Multi-tag grouping dropping dimensions. Took your suggested approach: findColumns now collects every group column and groupNameAt joins across them per row, so by {region, host} keeps both dimensions instead of dropping host. Added a regression case built from your exact example (region/host, one host spiking) plus a 6-tag case, and switched the old fixtures off the single-column shape that was letting the bug pass green.

One thing worth aligning on for the code comments: I found a Datadog source that describes a different multi-tag shape than the per-tag-columns one. The DBM API guide parses a single group column and comments that its values are "one element per group-by tag":

- https://docs.datadoghq.com/database_monitoring/guide/build_apps_with_dbm_api/

```
group_col = next(c for c in columns if c["type"] == "group")
# group_values is a list with one element per group-by tag
```

That's the opposite of the per-tag-column shape (one region column + one host column). The generated Go client is consistent with either reading — GroupScalarColumn.Values is [][]string ("the array of tag values for each group"):

- https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/model_group_scalar_column.go
- https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadogV2/model_data_scalar_column.go
- https://docs.datadoghq.com/api/latest/metrics/

The current code handles both shapes...  it joins across columns and flattens each column's inner array, so a row comes out us-east,host-a whether the tags arrive as N columns or one column with N-element rows (there's a unit test for each). But since the DBM guide is a DBM query and yours was a live metrics query on the same endpoint, I don't want to assume they're identical. Could you paste the raw JSON from the by {region,host} response you captured? That lets me pin the code comments to what the metrics endpoint actually returns rather than guessing between the two.

2. Unbounded cardinality. metadata.groups is now capped (100 entries) with a groups_truncated flag, and the docs note it's for low-cardinality tags. One tweak on your "keep the top N worst offenders" suggestion: the offender is the highest value for an error-rate/latency gate but the lowest for a success-rate gate, and the cap can't see the condition.... SO it keeps both value extremes (lowest + highest halves) to keep the worst outlier visible either way. The result slice used for evaluation is never truncated.

3. null/[] group label. Fixed to match the doc contract — a null or empty label now errors loudly instead of pairing a value with a blank name, with tests for both.

metricproviders/datadog/datadog.go
cgreeno: Ya that is my bad - I only tested it with a single region

gferrate (approved):

cgreeno:
@kostis-codefresh @meeech - any additional feedback or things that need changing.

cgreeno:
Sorry... gentle nudge @kostis-codefresh @meeech 🙇

cgreeno:
Heya sorry to bug your guys again, but any chance I can get a review please?
 @meeech 
@kostis-codefresh 

Thanks

kostis-codefresh:
Hey. We will get to it. Don't worry. The current AI Pr wave is slowing everything down. But rest assured that all PRs are being monitored.

kostis-codefresh (approved):
Thanks to everyone who helped with this PR review!

Issue #4732, opened 2026-05-20
Rollout Duration Tracking


agaudreault:
## Problem Statement

Argo Rollouts currently lacks visibility into how long rollouts actually take to complete. This creates several operational challenges:

1. **No observability into deployment performance**: Teams cannot measure whether rollouts are getting faster or slower over time, making it impossible to track deployment efficiency improvements.
2. **Cannot distinguish active work from waiting time**: When a rollout takes 30 minutes, there's no way to know if it's because the deployment itself is slow, or because it was waiting 25 minutes for manual approval.
3. **No data for SLIs/SLOs**: Organizations cannot set or track Service Level Indicators for deployment duration, preventing them from establishing deployment performance goals.
4. **Difficult to identify bottlenecks**: Without timing breakdowns, it's unclear whether optimization efforts should focus on deployment automation or approval workflows.
5. **Missing forensic data**: When investigating rollout issues, there's no historical record of "how long did rollout X take?" or "what percentage of time was spent paused?"
6. **No differentiation between outcomes**: Failed, aborted, and successful rollouts are not tracked separately, making it impossible to analyze patterns like "do aborted rollouts typically fail faster than successful ones complete?"

### Functional Requirements

**FR1: Track three separate duration metrics**

- Total wall-clock time from rollout start to completion
- Active progression time (excluding manual pauses)
- Manual pause time (time waiting for human intervention)
- Relationship: `total = progression + manual_pause`

**FR2: Support four rollout outcome statuses**

- `promoted`: Normal successful completion
- `manually-promoted`: User triggered full promotion (skipped remaining steps)
- `aborted`: Rollout was aborted (timeout, failed analysis, manual abort)
- `superseded`: Template/steps changed **mid-rollout** while in-progress (replaced by new attempt)
  - Note: Template changes **after** completion do NOT emit "superseded" - metrics were already emitted as "promoted"

**FR3: Distinguish manual vs automatic pauses**

- Manual pauses (require human action):
  - `spec.paused == true`
  - Canary step with `pause: {}` (no duration)
  - Blue-green with `autoPromotionEnabled: false`
  - Inconclusive analysis/experiment results
- Automatic pauses (do NOT count as manual):
  - Canary step with `pause: {duration: Xm}` (timed, auto-resumes)
  - Blue-green with `autoPromotionSeconds > 0` (timed, auto-resumes)
  - Waiting for analysis/experiment to complete (in-progress)

**FR4: Handle edge cases correctly**

[...]

PR #4755, opened 2026-05-28
fix(controller): Keep pod-template-hash stable across k8s library upgrade by aligning with the upstream Deployment controller.

### LINKED ISSUE #4696: Upgrading controller to v1.9.0 triggers spurious new revisions and Canary pause-for-promotion on all Rollouts due to pod template hash change from k8s library bump (#4497)
Checklist:

**Describe the bug**

Upgrading the Argo Rollouts controller from v1.8.3 to v1.9.0 (Helm chart 2.40.2 → 2.40.9) causes every Rollout in the cluster to be treated as a new revision, even when no change was made to any spec.template. Rollouts configured with infinite pause immediately pause waiting for manual promotion, with both the old and new ReplicaSets having identical pod specs.

**Root cause:** v1.9.0 includes #4497 — bump k8s libraries and go to current supported versions, which updates the k8s.io/api and k8s.io/apimachinery Go modules to a newer minor version. These libraries are used internally by ComputePodTemplateHash (via controller.ComputeHash from k8s.io/kubernetes/pkg/controller). The updated library serialises the PodSpec struct differently — due to new fields added to the struct definition — producing a different hash for an identical pod template.
On the first reconcile after the controller restarts, FindNewReplicaSet looks for a ReplicaSet whose rollouts-pod-template-hash label matches the newly computed hash. No match is found because all existing ReplicaSets were stamped with the hash computed by the v1.8.x controller. The lookup returns nil, the controller logs "Pod template change detected", and treats the situation as a genuine spec change — creating a new ReplicaSet and executing the full BlueGreen or Canary update lifecycle.
This is a recurring pattern. The same issue was reported at:

#88 (2019) — first occurrence when k8s deps were bumped
#1356 (2021) — recurred on v0.10 → v1.0.2 upgrade

**To Reproduce**

1. Install Argo Rollouts v1.8.3 (Helm chart 2.40.2) on any Kubernetes cluster
2. Create one or more Rollouts with 
```
strategy:
  canary:
    steps:
       - setWeight: 10
        - pause: {}
        - setWeight: 30
        - pause:
            duration: 2m
        - setWeight: 50
        - pause:
            duration: 2m
        - setWeight: 100
```      
3. Confirm all Rollouts are in Healthy status with a single active ReplicaSet
4. Upgrade the controller to v1.9.0 (Helm chart 2.40.9):
```
helm upgrade argo-rollouts argo/argo-rollouts \
     --namespace argo-rollouts \
     --version 2.40.9 \
     --reuse-values --wait
```
5. Observe all Rollouts immediately transition to Paused status with a new revision

**Expected behavior**

Upgrading the controller should not create new revisions or trigger the BlueGreen/Canary update lifecycle on Rollouts whose spec.template has not changed. The controller should detect that the existing ReplicaSet's pod spec is semantically identical to the desired spec (via PodTemplateEqualIgnoreHash) and skip creating a new ReplicaSet, re-stamping the updated hash label on the existing RS instead.

**Version**

Argo Rollouts: v1.8.3 → v1.9.0
Helm chart: 2.40.2 → 2.40.9
Kubernetes: 1.35.2

**Logs**

Controller logs after upgrade - 
```
time="2026-04-20T16:04:51Z" level=info msg="ComputePodTemplateHash hash changed (expected: 5b4958d4db, actual: 848c6689c8)" namespace=podinfo rollout=podinfo
time="2026-04-20T16:04:51Z" level=info msg="Pod template change detected (new: 65fdd55585, old: 5b84b55b6f)" namespace=demo rollout=argo-demo
time="2026-04-20T16:04:51Z" level=info msg="Pod template change detected (new: 86658bb664, old: 54f64cbbcb)" namespace=argo-podinfo rollout=argo-podinfo
time="2026-04-20T16:04:51Z" level=info msg="Pod template change detected (new: 65fdd55585, old: 5b84b55b6f)" namespace=demo rollout=argo-demo
time="2026-04-20T16:04:51Z" level=info msg="Pod template change detected (new: 86658bb664, old: 54f64cbbcb)" namespace=argo-podinfo rollout=argo-podinfo
time="2026-04-20T16:04:51Z" level=info msg="ComputePodTemplateHash hash changed (expected: 5b4958d4db, actual: 848c6689c8)" namespace=podinfo rollout=podinfo
time="2026-04-20T16:19:46Z" level=info msg="ComputePodTemplateHash hash changed (expected: 5b4958d4db, actual: 848c6689c8)" namespace=podinfo rollout=podinfo
time="2026-04-20T16:19:47Z" level=info msg="ComputePodTemplateHash hash changed (expected: 5b4958d4db, actual: 848c6689c8)" namespace=podinfo rollout=podinfo
time="2026-04-20T16:19:48Z" level=info msg="ComputePodTemplateHash hash changed (expected: 5b4958d4db, actual: 848c6689c8)" namespace=podinfo rollout=podinfo
time="2026-04-20T16:19:48Z" level=info msg="ComputePodTemplateHash hash changed (expected: 5b4958d4db, actual: 848c6689c8)" namespace=podinfo rollout=podinfo
```

Diff of the old and new replicaset

```
% diff old-rs.yml new-rs.yml
8,10c8,9
<     rollout.argoproj.io/ephemeral-metadata: '{"labels":{"role":"stable"}}'
<     rollout.argoproj.io/revision: "7"
<     rollout.argoproj.io/revision-history: 3,5
---
>     rollout.argoproj.io/revision: "6"
>     rollout.argoproj.io/revision-history: "4"
12,13c11,12
<   creationTimestamp: "2026-04-19T03:00:16Z"
<   generation: 12
---
>   creationTimestamp: "2026-04-20T15:29:06Z"
>   generation: 10
18c17
<     rollouts-pod-template-hash: 5b84b55b6f
---
>     rollouts-pod-template-hash: 65fdd55585
24c23
<   name: argo-demo-5b84b55b6f
---
>   name: argo-demo-65fdd55585
33,34c32,33
<   resourceVersion: "1776703357277215006"
<   uid: 1370c637-bd37-47b7-a169-f38ee2891026
---
>   resourceVersion: "1776703371609983008"
>   uid: 451df1ce-7a70-4902-958d-24886a3279g1
37c36
<   replicas: 1
---
>   replicas: 0
40c39
<       rollouts-pod-template-hash: 5b84b55b6f
---
>       rollouts-pod-template-hash: 65fdd55585
62,63c61
<         role: stable
<         rollouts-pod-template-hash: 5b84b55b6f
---
>         rollouts-pod-template-hash: 65fdd55585
206,210c204,205
<   availableReplicas: 1
<   fullyLabeledReplicas: 1
<   observedGeneration: 12
<   readyReplicas: 1
<   replicas: 1
---
>   observedGeneration: 10
>   replicas: 0
```

---

zachaller:
Fixes #4696

## Summary

Upgrading the controller across a Kubernetes library bump (v1.8.x → v1.9.0, i.e. k8s `v0.29` → `v0.34`) causes **every Rollout** to recompute a different `rollouts-pod-template-hash`. That spuriously creates a new revision which progresses through the canary steps and parks at the first indefinite pause, even though nothing in the user's `spec.template` actually changed.

This PR removes the *correctness* dependency on byte-stable hashing (the hash-label lookup remains as a fast path) and aligns Argo Rollouts with how the **upstream Kubernetes Deployment controller** decides ReplicaSet identity:

1. Compute `pod-template-hash` with the upstream `controller.ComputeHash` (spew/reflect based) instead of `json.Marshal`, so the hash is immune to JSON-serialization changes.
2. Fall back to **semantic pod-template equality** (`PodTemplateEqualIgnoreHash`) when no ReplicaSet carries the computed hash label, exactly like `deploymentutil.FindNewReplicaSet`. The hash-label lookup remains the first (fast) tier: the label was written by this controller from the same desired template, so a match is authoritative even when the live template has since diverged (apiserver defaulting from a newer server, or mutating admission on ReplicaSets).

The net effect: a controller/library upgrade is a no-op for existing Rollouts — the existing ReplicaSet is adopted (keeping its original hash label), so no new revision and no redeploy.

## Background / root cause

`hash.ComputePodTemplateHash` previously computed an FNV-32a sum over `json.Marshal(template)`. The hash function itself didn't change between releases — the Kubernetes libraries did:

| | k8s.io/api | k8s.io/apimachinery | k8s.io/kubernetes |
|---|---|---|---|
| v1.8.3 | v0.29.3 | v0.29.3 | v1.29.3 |
| v1.9.0 | v0.34.x | v0.34.x | v1.34.x |

In apimachinery `v0.34`, the JSON struct tag on `metav1.ObjectMeta.CreationTimestamp` gained `omitzero` ([kubernetes/kubernetes#130989](https://github.com/kubernetes/kubernetes/pull/130989)):

```go
// apimachinery v0.29.3
CreationTimestamp Time `json:"creationTimestamp,omitempty" ...`
// apimachinery v0.34.x
CreationTimestamp Time `json:"creationTimestamp,omitempty,omitzero" ...`
```

A pod template's `creationTimestamp` is always the zero value. Previously it serialized as `"creationTimestamp":null`; with `omitzero` (honored by Go 1.24+) it is omitted entirely. That single byte change flips the JSON-based hash for 100% of Rollouts on upgrade. This is a recurrence of the long-standing class of bug in [#88](https://github.com/argoproj/argo-rollouts/issues/88).

### How the upstream Deployment controller avoids this

The Deployment controller never had this problem because:

- **It does not hash via JSON.** `controller.ComputeHash` feeds the template through `hashutil.DeepHashObject` (the `spew` library, reflect-based). JSON struct tags such as `omitzero` are irrelevant to it.
- **It does not identify the current ReplicaSet by hash label.** `deploymentutil.FindNewReplicaSet` iterates ReplicaSets and compares with `EqualIgnoreHash` (a semantic `apiequality.Semantic.DeepEqual` of the pod templates). Even if the hashing algorithm changes, an existing ReplicaSet is still recognized as current.

This PR brings Argo Rollouts in line with both behaviors, but deliberately keeps the hash-label lookup as a fast first tier. Unlike the in-tree Deployment controller, Argo Rollouts is not version-locked to the apiserver: a newer apiserver can default pod-template fields our vendored libraries don't know, and mutating admission can rewrite ReplicaSet templates. In both cases semantic comparison would fail forever; the hash label — written by this controller from the same desired template — still matches and prevents an unwarranted redeploy (and `collisionCount` churn).

### Production code

**`utils/hash/hash.go`** — `ComputePodTemplateHash` now delegates to the upstream Deployment controller's hash:

```go
func ComputePodTemplateHash(template *corev1.PodTemplateSpec, collisionCount *int32) string {
	return controller.ComputeHash(template, collisionCount)
}
```

This makes the hash spew-based (matching Deployment/ReplicaSet labeling going forward) and immune to JSON-tag changes like `omitzero`.

**`utils/replicaset/replicaset.go`**

- `FindNewReplicaSet` is reduced from three tiers to two: (1) lookup by `pod-template-hash` label computed with the current algorithm, then (2) semantic pod-template equality, mirroring `deploymentutil.FindNewReplicaSet`. The "deprecated hash" tier is gone (our hash *is* `controller.ComputeHash` now, so the two label searches collapsed into one). When the fallback adopts a ReplicaSet whose label doesn't match the computed hash, an `Infof` is logged so operators can observe the one-time transition.
- Within the semantic tier, the ReplicaSet recorded in `status.currentPodHash` is preferred when it still matches the desired template; otherwise the oldest match is reused (upstream behavior). This matters when several retained ReplicaSets carry *identical* templates with different hash labels — the fleet state v1.9.0 created by spuriously duplicating revisions (#4696). Without the preference, oldest-first would resurrect the scaled-down pre-1.9.0 twin and trigger a second redeploy for rollouts whose spurious revision was already promoted. On a genuine template change the current ReplicaSet no longer matches semantically, so the preference is a no-op and rollback-reuse behaves exactly like the upstream Deployment controller.
- New exported helper `ReplicaSetTemplateMatchesRollout` encapsulates the rollouts-specific normalization before the semantic compare: strips injected ephemeral canary/stable metadata and the injected anti-affinity rule from the live template. Used by both `FindNewReplicaSet` and the hash-collision check (see `rollout/sync.go` below).
- `PodTemplateEqualIgnoreHash` now runs **both** the live and desired templates through library defaulting (`defaultPodTemplateSpec`) before `Semantic.DeepEqual`. Previously only the desired side was defaulted; defaulting both keeps the comparison stable when a live ReplicaSet was created by an older library that defaulted fewer fields. The `spec.serviceAccount` carve-out from [#1356](https://github.com/argoproj/argo-rollouts/issues/1356) is unchanged.
- `CheckPodSpecChange` is unchanged in behavior. Its control flow is identical to before (compare `status.CurrentPodHash` against the adopted RS's label, or against the freshly computed hash when no RS is found); only the underlying hash implementation changed. Doc comment expanded to explain why adopting via label keeps detection stable across the upgrade.
- `GenerateReplicaSetAffinity` and `IfInjectedAntiAffinityRuleNeedsUpdate` now take the ReplicaSet's `pod-template-hash` label (`podHash`) and compare it against `status.StableRS`, instead of recomputing a hash from the desired template. Both are stored label values of the same provenance, so the "are we mid-rollout?" decision is not skewed when the stored stable hash predates the hashing change. See "Edge cases handled" below.

**`rollout/sync.go`**

- Call sites pass the live hash to the anti-affinity helpers:
  - update path (adopted RS): `replicasetutil.GetPodTemplateHash(c.newRS)`
  - create path: the newly computed `podTemplateSpecHash` (computation reordered to occur before the call)
- The `AlreadyExists` hash-collision check in `createDesiredReplicaSet` now uses `ReplicaSetTemplateMatchesRollout` instead of comparing the raw live template against the rollout template. Canary/blue-green ReplicaSets are *created with* ephemeral metadata injected, so the raw comparison could misread the controller's own ReplicaSet as a hash collision and bump `collisionCount` (a pre-existing bug, fixed here because this path is adjacent to the matching changes). Hash matching is deliberately **not** used in this path — a genuine collision has a matching hash by definition, so only the semantic check is valid there.

**`rollout/experiment.go`** — `GetExperimentFromTemplate` now prefers the new ReplicaSet's stored `pod-template-hash` label for the experiment's name and hash label, falling back to the computed hash only when there is no ReplicaSet yet. This keeps experiments created mid-transition consistent with the adopted ReplicaSet they run against (matching how `rollout/analysis.go` already derives its labels).

### Edge cases handled

- **Upgrade with unchanged template (the bug):** the old json-era label doesn't match the new computed hash, so `FindNewReplicaSet` adopts the existing ReplicaSet via the semantic fallback; the status keeps the adopted RS's original (json-era) label; `CheckPodSpecChange` sees `status.CurrentPodHash == newRS.label` → no change → no new revision.
- **Genuine template change:** no ReplicaSet matches by label or semantically → `FindNewReplicaSet` returns nil → `CheckPodSpecChange` reports a change → a new ReplicaSet is created and labeled with the new spew hash.
- **Diverged live template (version skew / mutating admission):** a ReplicaSet whose live template was altered after creation — a newer apiserver defaulting fields unknown to our vendored libraries, or a webhook injecting metadata — no longer matches semantically, but is still found by its hash label (tier 1), so no redeploy and no `collisionCount` churn.
- **Twin ReplicaSets left behind by v1.9.0:** when the spurious duplicate revision was already promoted, the running ReplicaSet (recorded in `status.currentPodHash`) is adopted rather than its older identical-template twin, so upgrading to this version is a no-op instead of a second redeploy. When `status.currentPodHash` matches no retained ReplicaSet, the oldest semantic match is reused, as upstream does.
- **Anti-affinity + upgrade:** a fully-stable Rollout using anti-affinity will not have a spurious anti-affinity rule injected into its adopted stable ReplicaSet just because the stored stable hash predates the hashing change. The decision compares the RS's own stored label to `status.StableRS` (equal for the stable RS) rather than a recomputed hash.

### Tests

- **`utils/hash/hash_test.go`** — replaced the brittle golden-value assertion (which had to be updated on every k8s bump) with `MatchesDeploymentControllerComputeHash`, asserting our hash equals `controller.ComputeHash`.
- **`utils/replicaset/replicaset_test.go`**
  - `TestFindNewReplicaSet` now covers both tiers: lookup by current hash label, and semantic adoption when the label is stale.
  - `TestFindNewReplicaSetAcrossHashChange` (new) — an unchanged template whose RS carries a stale hash label is adopted; a real spec change is not; with twin identical-template ReplicaSets, the one recorded in `status.currentPodHash` wins over an older twin, with oldest-first as the fallback. Stable across future k8s upgrades (no hashes to update).
  - `TestFindNewReplicaSetWithDivergedLiveTemplate` (new) — pins the tier-1 protection: an RS whose live template was mutated after creation (simulated webhook-injected annotation) is still found via its matching hash label.
  - `TestPodTemplateEqualIgnoreHashAcrossLibraryDefaulting` (new) — an under-defaulted live template (older library) still matches the semantically identical desired template.
  - `TestGenerateReplicaSetAffinity` / `TestIfInjectedAntiAffinityRuleNeedsUpdate` / `TestRemoveInjectedAntiAffinityRule` updated for the new `podHash` parameter.
  - `TestGenerateReplicaSetAffinityStableAcrossHashChange` (new) — pins the anti-affinity upgrade edge case: a stable Rollout whose `StableRS` is an old-format label (deliberately `!=` the recomputed hash) gets no injection.
- **`rollout/controller_test.go`** — `TestPodTemplateHashEquivalence` rewritten to assert the runtime-relevant property (two equivalent quantity formats compare equal via `PodTemplateEqualIgnoreHash`, so they are adopted, not redeployed) plus a self-consistent RS-name sanity check, instead of a hardcoded hash.
- **`rollout/ephemeralmetadata_test.go`** — the "ReplicaSet already patched" fixtures now set the ephemeral-metadata annotation and preserve realistic template labels, so semantic adoption works (these previously relied on label-based lookup).
- **`utils/conditions/rollouts_test.go`** and **`utils/experiment/experiment_test.go`** — compute expected hashes dynamically instead of pinning literal values that drift on every k8s bump.
- **`test/e2e/canary_test.go`** — the three hardcoded `pod-template-hash` selector assertions (which carried `// NOTE: This must be updated for every k8s version upgrade`) are replaced with `assertCanarySelectorPointsToStable`, a helper that reads the expected hash from the live stable (revision 1) ReplicaSet. Never needs updating on a hashing/library change.

## Behavior / compatibility

- **No redeploy on upgrade** for Rollouts whose `spec.template` is unchanged. Existing ReplicaSets keep their original hash labels; services, selectors, `status.stableRS`, and `status.currentPodHash` continue to use those values. The new spew hash is only used when a genuinely new ReplicaSet is created (a real template change). Old and new ReplicaSets can coexist with different hash-label formats until the old ones are scaled down — adoption is by template, not by label format.
- `CheckPodSpecChange` mid-rollout step-reset semantics are preserved.
- Equivalent resource-quantity formatting (e.g. `2000m` vs `2`) is treated as equal by `Semantic.DeepEqual` (quantities compare by value), so it does not cause churn. Note the spew-based hash itself is *not* format-stable for quantities — adoption relies on the label/semantic matching above, not on hash stability.

### Upgrading from v1.9.0 specifically

v1.9.0 already shipped the broken hash, so fleets are not in a clean pre-bug state — the spurious revisions it created leave *twin* ReplicaSets (identical templates, different hash labels) in revision history. What this version does for each state:

- **From v1.8.x (never ran v1.9.0, or rolled the controller back):** the current ReplicaSet is adopted semantically — no redeploy. The rollback case also works because after flipping back under v1.8.x the active ReplicaSet is the older twin, and it is also the one `status.currentPodHash` records.
- **From v1.9.0, spurious revision still paused/progressing:** `status.currentPodHash` points at the spurious ReplicaSet, which is adopted — the rollout stays exactly as v1.9.0 left it (paused awaiting promote/abort). No automatic state change on controller upgrade.
- **From v1.9.0, spurious revision already promoted:** `status.currentPodHash` points at the running ReplicaSet, which is adopted — the upgrade is a no-op. The scaled-down pre-v1.9.0 twin remains in revision history until it ages out via `revisionHistoryLimit`.
- **Created entirely under v1.9.0:** single ReplicaSet, adopted semantically — no redeploy.

In all cases adopted ReplicaSets keep their existing (json-era) hash labels permanently; only genuinely new revisions get spew-format names/labels. Services, traffic routing, and analysis key off stored labels, so mixed label formats coexist safely — but external tooling that pattern-matches hash values (dashboards, log queries) will see both formats during the transition.

## Notes / tradeoffs

- The semantic `DeepEqual` fallback (defaulting + deep compare per ReplicaSet) only runs when the O(1) hash-label lookup misses — i.e. once per genuine template change or hashing transition — so steady-state reconcile cost is unchanged.
- `GenerateReplicaSetAffinity` and `IfInjectedAntiAffinityRuleNeedsUpdate` gained a `podHash` parameter — a signature change to two exported functions in `utils/replicaset`. All in-tree callers are updated; external importers (unlikely) would need to pass the ReplicaSet's hash.
- Because the hash source changed, hardcoded `pod-template-hash` values in tests no longer match; affected unit fixtures now compute hashes dynamically and the e2e selector assertions read the live value. There are intentionally no new golden-hash assertions to maintain.

## Validation

- `go build ./...` — clean.
- `go vet -tags e2e ./test/e2e/` — clean.
- Unit suites pass: `utils/hash`, `utils/replicaset`, `utils/conditions`, `utils/experiment`, `rollout`, `experiments`.
- Manual reasoning + regression tests confirm an upgrade with an unchanged template adopts the existing ReplicaSet (no new revision), while a real template change still creates one.

## Checklist

* [x] I have run all tests locally (including the flaky ones) and they pass on my workstation.

### DIFF
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -1527,8 +1527,6 @@ func TestInvalidWorkloadRef(t *testing.T) {
 // variations made to the pod template in equivalent ways.
 func TestPodTemplateHashEquivalence(t *testing.T) {
 	var err error
-	// NOTE: This test will fail on every k8s library upgrade.
-	// To fix it, update expectedReplicaSetName to match the new hash.
 	expectedReplicaSetName := "guestbook-6c5667f666"
 
 	r1 := newBlueGreenRollout("guestbook", 1, nil, "active", "")
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -1523,11 +1523,16 @@ func TestInvalidWorkloadRef(t *testing.T) {
 	assert.JSONEq(t, calculatePatch(r, expectedPatch), patch)
 }
 
-// TestPodTemplateHashEquivalence verifies the hash is computed consistently when there are slight
-// variations made to the pod template in equivalent ways.
+// TestPodTemplateHashEquivalence verifies that two pod templates that differ only in equivalent
+// formatting of the same resource quantities (e.g. "2000m" vs "2") are treated as the same template.
+//
+// The controller adopts an existing ReplicaSet by semantic template equality (PodTemplateEqualIgnoreHash,
+// like the upstream Deployment controller), so equivalent specs never create a new revision. We assert the
+// equality property directly because it is what protects users from churn — note that the raw
+// pod-template-hash itself is spew-based and need not be byte-identical for equivalent quantity formatting,
+// since the apiserver canonicalizes quantities before the controller observes them.
 func TestPodTemplateHashEquivalence(t *testing.T) {
 	var err error
-	expectedReplicaSetName := "guestbook-6c5667f666"
 
 	r1 := newBlueGreenRollout("guestbook", 1, nil, "active", "")
 	r1Resources := `
@@ -1553,23 +1558,28 @@ requests:
 	err = yaml.Unmarshal([]byte(r2Resources), &r2.Spec.Template.Spec.Containers[0].Resources)
 	assert.NoError(t, err)
 
-	for _, r := range []*v1alpha1.Rollout{r1, r2} {
-		f := newFixture(t)
-		activeSvc := newService("active", 80, nil, r)
-		f.kubeobjects = append(f.kubeobjects, activeSvc)
-		f.rolloutLister = append(f.rolloutLister, r)
-		f.serviceLister = append(f.serviceLister, activeSvc)
-		f.objects = append(f.objects, r)
+	// The two equivalent templates must compare equal, so a ReplicaSet created from one is adopted
+	// (not redeployed) when the rollout is specified with the other.
+	assert.True(t, replicasetutil.PodTemplateEqualIgnoreHash(&r1.Spec.Template, &r2.Spec.Template))
 
-		f.expectUpdateRolloutStatusAction(r)
-		f.expectPatchRolloutAction(r)
-		rs := newReplicaSet(r, 1)
-		rsIdx := f.expectCreateReplicaSetAction(rs)
-		f.run(getKey(r, t))
-		rs = f.getCreatedReplicaSet(rsIdx)
-		assert.Equal(t, expectedReplicaSetName, rs.Name)
-		f.Close()
-	}
+	// Sanity: reconciling a rollout creates a ReplicaSet deterministically named after its pod-template-hash.
+	f := newFixture(t)
+	defer f.Close()
+	activeSvc := newService("active", 80, nil, r1)
+	f.kubeobjects = append(f.kubeobjects, activeSvc)
+	f.rolloutLister = append(f.rolloutLister, r1)
+	f.serviceLister = append(f.serviceLister, activeSvc)
+	f.objects = append(f.objects, r1)
+
+	f.expectUpdateRolloutStatusAction(r1)
+	f.expectPatchRolloutAction(r1)
+	rs := newReplicaSet(r1, 1)
+	rsIdx := f.expectCreateReplicaSetAction(rs)
+	f.run(getKey(r1, t))
+	rs = f.getCreatedReplicaSet(rsIdx)
+	podHash := rs.Labels[v1alpha1.DefaultRolloutUniqueLabelKey]
+	assert.NotEmpty(t, podHash)
+	assert.Equal(t, fmt.Sprintf("%s-%s", r1.Name, podHash), rs.Name)
 }
 
 func TestNoReconcileForDeletedRollout(t *testing.T) {
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -1529,7 +1529,7 @@ func TestPodTemplateHashEquivalence(t *testing.T) {
 	var err error
 	// NOTE: This test will fail on every k8s library upgrade.
 	// To fix it, update expectedReplicaSetName to match the new hash.
-	expectedReplicaSetName := "guestbook-6f496f9f78"
+	expectedReplicaSetName := "guestbook-6c5667f666"
 
 	r1 := newBlueGreenRollout("guestbook", 1, nil, "active", "")
 	r1Resources := `
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -1529,8 +1529,9 @@ func TestInvalidWorkloadRef(t *testing.T) {
 // The controller adopts an existing ReplicaSet by semantic template equality (PodTemplateEqualIgnoreHash,
 // like the upstream Deployment controller), so equivalent specs never create a new revision. We assert the
 // equality property directly because it is what protects users from churn — note that the raw
-// pod-template-hash itself is spew-based and need not be byte-identical for equivalent quantity formatting,
-// since the apiserver canonicalizes quantities before the controller observes them.
+// pod-template-hash itself is spew-based and need not be byte-identical for equivalent quantity formatting;
+// adoption relies on semantic equality (apiequality.Semantic compares quantities by value), not on hash
+// stability.
 func TestPodTemplateHashEquivalence(t *testing.T) {
 	var err error
 
--- a/rollout/ephemeralmetadata_test.go
+++ b/rollout/ephemeralmetadata_test.go
@@ -2,6 +2,7 @@ package rollout
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"testing"
 
@@ -21,7 +22,16 @@ import (
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	"github.com/argoproj/argo-rollouts/utils/annotations"
+	replicasetutil "github.com/argoproj/argo-rollouts/utils/replicaset"
 )
+
+func setReplicaSetEphemeralMetadataAnnotation(rs *v1.ReplicaSet, metadata *v1alpha1.PodTemplateMetadata) {
+	metadataBytes, _ := json.Marshal(metadata)
+	if rs.Annotations == nil {
+		rs.Annotations = make(map[string]string)
+	}
+	rs.Annotations[replicasetutil.EphemeralMetadataAnnotation] = string(metadataBytes)
+}
 
 // TestSyncCanaryEphemeralMetadataInitialRevision verifies when we create a revision 1 ReplicaSet
 // (with no previous revisions), that the ReplicaSet will get the stable metadata.
@@ -323,9 +333,8 @@ func TestSyncCanaryEphemeralMetadataReplicaSetAlreadyPatched(t *testing.T) {
 
 	// Create a ReplicaSet that already has the stable metadata label
 	rs1 := newReplicaSetWithStatus(r1, 3, 3)
-	rs1.Spec.Template.Labels = map[string]string{
-		"role": "stable",
-	}
+	rs1.Spec.Template.Labels["role"] = "stable"
+	setReplicaSetEphemeralMetadataAnnotation(rs1, r1.Spec.Strategy.Canary.StableMetadata)
 
 	rsGVK := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "ReplicaSet"}
 
@@ -395,9 +404,8 @@ func TestSyncBlueGreenEphemeralMetadataReplicaSetAlreadyPatched(t *testing.T) {
 
 	// Create a ReplicaSet that already has the active metadata label
 	rs1 := newReplicaSetWithStatus(r1, 3, 3)
-	rs1.Spec.Template.Labels = map[string]string{
-		"role": "active",
-	}
+	rs1.Spec.Template.Labels["role"] = "active"
+	setReplicaSetEphemeralMetadataAnnotation(rs1, r1.Spec.Strategy.BlueGreen.ActiveMetadata)
 
 	rsGVK := schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "ReplicaSet"}
 
--- a/rollout/experiment.go
+++ b/rollout/experiment.go
@@ -26,7 +26,13 @@ func GetExperimentFromTemplate(r *v1alpha1.Rollout, stableRS, newRS *appsv1.Repl
 	if step == nil {
 		return nil, nil
 	}
-	podHash := hash.ComputePodTemplateHash(&r.Spec.Template, r.Status.CollisionCount)
+	// Prefer the new ReplicaSet's stored pod-template-hash label so the experiment's name and
+	// hash label stay consistent with the ReplicaSet it runs against, even if the ReplicaSet was
+	// adopted with a label written by an older hashing algorithm.
+	podHash := replicasetutil.GetPodTemplateHash(newRS)
+	if podHash == "" {
+		podHash = hash.ComputePodTemplateHash(&r.Spec.Template, r.Status.CollisionCount)
+	}
 	currentStep := int32(0)
 	if r.Status.CurrentStepIndex != nil {
 		currentStep = *r.Status.CurrentStepIndex
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -75,12 +75,12 @@ func (c *rolloutContext) syncReplicaSetRevision() (*appsv1.ReplicaSet, error) {
 	// Set existing new replica set's annotation
 	annotationsUpdated := annotations.SetNewReplicaSetAnnotations(c.rollout, rsCopy, newRevision, true)
 	minReadySecondsNeedsUpdate := rsCopy.Spec.MinReadySeconds != c.rollout.Spec.MinReadySeconds
-	affinityNeedsUpdate := replicasetutil.IfInjectedAntiAffinityRuleNeedsUpdate(rsCopy.Spec.Template.Spec.Affinity, *c.rollout)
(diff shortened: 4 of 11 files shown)


kostis-codefresh:
It seems that Flagger has a similar problem https://github.com/fluxcd/flagger/issues/1673

kostis-codefresh:
It would be great if we also include testing scenarios. At the very least

1. What happens when somebody has already updated to 1.9.0 and now updates to this (next version)
2. What happens if somebody never updated to 1.9.x and went from 1.8.x straight to this version

alexef:
hey @zachaller we are also blocked and cannot upgraded because of this issue. any way I can help resolve conflicts and get this merged?

Later edit, I see there were a couple of changes in master that affect it:

- #4832 (@AruneshDwivedi) added a len(Values) > 0 guard in IfInjectedAntiAffinityRuleNeedsUpdate (utils/replicaset/replicaset.go) to fix a panic (#4840). This PR changes that same function's signature to take a podHash param instead of recomputing the hash from the rollout, so both edits land on the same lines.
- #4774 (@agaudreault) added rollout duration tracking (#4732) and rewrote TestPodTemplateHashEquivalence in rollout/controller_test.go as part of that change. This branch rewrote the same test differently to drop the hardcoded pod-template-hash assertion.

madstacks:
We are blocked on this issue as well and cannot upgrade to the latest version. Is it possible to bump the priority on this given the potential blast radius this has for folks that upgrade and end up inadvertently triggering new rollouts cluster-wide.

mubarak-j:
I think it's worth taking a look at how Flagger is trying to solve the same problem, https://github.com/fluxcd/flagger/pull/1942

PR #4761, opened 2026-06-01
ci: upload unit and e2e coverage separately with Codecov flags

AdamSaleh:
## Summary

- Split the single combined coverage upload into two separate uploads with Codecov flags (`unit-tests` and `e2e`)
- Add flag definitions to `.codecov.yml` with `carryforward: true`
- Unit test and e2e coverage profiles are now generated separately instead of being combined into one file

This enables flag-filtered coverage views on Codecov, allowing teams to track unit test and e2e coverage independently.

### `.github/workflows/testing.yaml`
- Replace combined `go tool covdata textfmt` step with two separate steps (one for unit tests, one for e2e)
- Upload unit test coverage with `flags: unit-tests`
- Upload e2e coverage with `flags: e2e`
- Codecov automatically merges both uploads for the overall project coverage view

### `.codecov.yml`
- Add `flags:` section defining `unit-tests` and `e2e` flags
- Both flags exclude `test/` directory and enable `carryforward`

## Test plan
- [ ] Verify CI pipeline passes with the split coverage uploads
- [ ] Confirm both `unit-tests` and `e2e` flags appear on Codecov
- [ ] Verify overall project coverage remains unchanged (Codecov merges flagged uploads)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,8 @@
+flags:
+  unit-tests:
+    carryforward: true
+  e2e:
+    carryforward: true
 coverage:
   status:
     # allow test coverage to drop by 0.1%, assume that it's typically due to CI problems
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -184,13 +184,28 @@ jobs:
         with:
           name: coverage-output-unit
           path: coverage-output-unit
-      - name: combine-go-coverage
+      - name: Generate unit test coverage profile
         run: |
-          go tool covdata textfmt -i=coverage-output-unit/,coverage-output-e2e/ -o full-coverage.out
-      - name: Upload code coverage information to codecov.io
+          go tool covdata textfmt -i=coverage-output-unit/ -o unit-coverage.out
+      - name: Generate e2e test coverage profile
+        run: |
+          go tool covdata textfmt -i=coverage-output-e2e/ -o e2e-coverage.out
+      - name: Upload unit test coverage to codecov.io
         uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
         with:
-          files: full-coverage.out
+          files: unit-coverage.out
+          flags: unit-tests
+          fail_ci_if_error: false
+          codecov_yml_path: .codecov.yml
+          disable_search: true
+          verbose: true
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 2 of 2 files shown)


kostis-codefresh (changes requested):
Can you explain a bit what is the issue you are solving here? Is it just better reporting for how tests are shown in the CodeCov dashboard?

I have two major concerns about this PR:

Both [Argo CD](https://app.codecov.io/gh/argoproj/argo-cd/flags) and [Argo Worklfows](https://app.codecov.io/gh/argoproj/argo-workflows/flags) don't use the flag feature and I don't understand why we need it at Argo Rollouts. I would prefer if Argo Rollouts does things the same way as the other Argo projects.

If this is a case where  you think that all Argo projects should use flags, I would start with a proposal to Argo CD, get it approved (because Argo CD has more maintainers) and then bring the same change to Argo Rollouts (and other Argo projects)

The second point is that looking at pricing, it seems [that flags is a paid feature](https://about.codecov.io/pricing/). Have you verified if Argo Rollouts has access to that?
Also in the context of the recent acquisition?

AdamSaleh:
@kostis-codefresh to be transparent - this is because at RedHat we are considering having internal metrics that depend on being able to distinguish unit test and integration test coverage and we would like to see this not only for our internal projects but also on the projects we are contributing to :)

As far as other repos go, we already enabled this on some of the smaller repos in argoproj-labs and I do have https://github.com/argoproj/argo-cd/pull/28085 open to introduce this to argo-cd proper.

As far as costs go - from looking at the pricing page as long as the development is in the open, it should be covered by the opensource plan? But I am not sure if this is something that is behind some gate/process of asking for it on th codecov side.

Of course, if other people here strongly dislike this I will figure out another way, but if there are issues that I can resolve I would rather have this here.

From my pov, being able to distinguish this is useful, because I use the integration tests when testing argocd in various configurations, so seeing how much is covered just by integration tests is very useful in that scenario.

With flags you don't loose the metric, it still gets combined in codecov project. But you get the ability to focus on what you are interested in. This also is the reason why I just opened a PR and not gone down the proposal route  - as far as I understand the change is non-destructive, just adds more metadata to something we are already sending.

kostis-codefresh (approved):
Ok. Thanks for the clarification. LGTM then.

For posterity, if the Harness acquisition of Codecov affects what exactly is offered in the "public" plan then we might need to revisit this.

AdamSaleh:
@zachaller maybe you can take a look as well? Keith Chong receommended you :)

PR #4763, opened 2026-06-02
fix: update yarn.lock with correct dependencies and  CI workflows to use Go 1.25 for v1.9.0

Rizwana777:
It was observed that due to broken yarn.lock file, the Argo Rollouts source-code fails to build on disconnected systems which uses npm cache to provide node packages in 1.9.0. This PR fixes the yarn.lock file by simply running yarn install in the ui directory.

and also after creating this PR, CI checks were failing with errors like - 
go: go.mod requires go >= 1.25.0 (running go 1.24.13; GOTOOLCHAIN=local)
there was mismatch between CI go version and go.mod so I have updated the version in CI to use Go 1.25

and updated golangci/golangci-lint version to v2.4.0   which is compatible with Go 1.25 version

The failing Istio e2e tests are not caused by this PR. These tests are already broken in the `release-1.9` base branch.

Checklist:


kostis-codefresh (changes requested):
Hello

Is this change actually needed anymore? The Go version was already restored in release 1.9 at https://github.com/argoproj/argo-rollouts/commit/c795a91fd4ba20a83c8efdd1fcc9f9f8ad8ac803

Also the project has now migrated from yarn to pnpm as can be seen from https://github.com/argoproj/argo-rollouts/commit/2ccdae835 (also in 1.10 rc )

PR #4766, opened 2026-06-04
fix(trafficrouting): parse VirtualService name containing a period correctly

### LINKED ISSUE #4709: Istio traffic routing incompatible with virtual services containing a period
Checklist:

**Describe the bug**

[Documentation states](https://argo-rollouts.readthedocs.io/en/stable/features/traffic-management/istio/) that Istio Traffic Management supports leveraging a virtual service defined in a different namespace, however this fails if the virtual service name contains a `.`

| If the VirtualService is defined in a different namespace than the rollout, its name should be rollout-vsvc.<vsvc namespace name>

If not using a different namespace, a virtual service with a period in its name will be assumed to be using a different namespace.

While regular services cannot contain a `.`, I believe virtual services only need to conform to `DNS-1123 subdomain` validation which does allow periods.

**To Reproduce**

These steps were use to reproduce the issue in `v1.9.0`.  It also reproduces in older versions. I do not believe this is a regression, but a long standing limitation.

Kubectl apply the following example manifest:
```
apiVersion: v1
kind: Namespace
metadata:
  name: example
---
apiVersion: v1
kind: Namespace
metadata:
  name: istio
---
apiVersion: v1
kind: Service
metadata:
  name: example-canary
  namespace: example
spec:
  type: ClusterIP
  selector:
    app: webservice
  ports:
    - name: http
      port: 80
      targetPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: example-stable
  namespace: example
spec:
  type: ClusterIP
  selector:
    app: webservice
  ports:
    - name: http
      port: 80
      targetPort: 80
---
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: virtualservice.example
  namespace: istio
spec:
  hosts:
    - webservice.example.svc.cluster.local
  gateways:
    - mesh
  http:
    - name: primary
      route:
        - destination:
            host: example-stable.example.svc.cluster.local
            port:
              number: 80
          weight: 100
        - destination:
            host: example-canary.example.svc.cluster.local
            port:
              number: 80
          weight: 0
---
kind: Rollout
metadata:
  name: webservice
  namespace: example
spec:
  replicas: 1
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      app: webservice
  template:
    metadata:
      labels:
        app: webservice
    spec:
      containers:
        - name: webservice
          image: nginx:1.27
          ports:
            - containerPort: 80
  strategy:
    canary:
      canaryService: example-canary
      stableService: example-stable
      trafficRouting:
        istio:
          virtualService:
            name: virtualservice.example.istio
            routes:
              - primary
      steps:
        - setWeight: 1
        - pause:
            duration: 10m
```

The above setup will do the following:
- create namespace `example`
- create namespace `istio`
- create cluster ip service `example-canary` within `example`
- create cluster ip service `example-stable` within `example`
- create virtual service `virtualservice.example` within `istio`
- create rollout `webservice` within `example`

The rollout object attempts to leverage canary and istio traffic routing using the virtual service from the `istio` namespace:
```
      trafficRouting:
        istio:
          virtualService:
            name: virtualservice.example.istio
```

However due to the current logic, the controller parses this as:
```
virtual service name      = virtualservice
virtual service namespace = example
```

Current logic coming from [here](https://github.com/argoproj/argo-rollouts/blob/3b9538c5880c8e1faf5367286a7803788445c42b/utils/istio/istio.go#L40-L52).

Which results in the following error:
 
```
Name:            webservice
Namespace:       example
Status:          ✖ Degraded
Message:         InvalidSpec: The Rollout "webservice" is invalid: spec.strategy.canary.trafficRouting.istio.virtualService.name: Invalid value: "virtualservice": virtualservices.networking.istio.io "virtualservice" not found
```

**Expected behavior**

Rollouts controller is able to successfully manage virtual services in separate namespaces, as long as they are named appropriately according to Istio and Kubernetes.

**Screenshots**

N/A

**Version**

`v1.9.0`

**Logs**

```
time="2026-04-27T23:05:51Z" level=info msg="Event(v1.ObjectReference{Kind:\"Rollout\", Namespace:\"example\", Name:\"webservice\", UID:\"30552d94-4aba-4d0c-98c4-cd5c98e88b81\", APIVersion:\"argoproj.io/v1alpha1\", ResourceVersion:\"1368\", FieldPath:\"\"}): type: 'Normal' reason: 'RolloutAddedToInformer' Rollout resource added to informer: example/webservice"
time="2026-04-27T23:05:51Z" level=info msg="Rollout resource added to informer: example/webservice" event_reason=RolloutAddedToInformer namespace=example rollout=webservice
time="2026-04-27T23:05:51Z" level=info msg="Istio detected"
time="2026-04-27T23:05:51Z" level=info msg="Starting istio workers"
time="2026-04-27T23:05:51Z" level=info msg="Istio workers (10) started"
time="2026-04-27T23:05:51Z" level=info msg="Started syncing rollout" generation=1 namespace=example resourceVersion=1368 rollout=webservice
time="2026-04-27T23:05:51Z" level=error msg="The Rollout \"webservice\" is invalid: spec.strategy.canary.trafficRouting.istio.virtualService.name: Invalid value: \"virtualservice\": virtualservices.networking.istio.io \"virtualservice\" not found" namespace=example rollout=webservice
time="2026-04-27T23:05:51Z" level=error msg="newRolloutContext returned nil" generation=1 namespace=example resourceVersion=1368 rollout=webservice
time="2026-04-27T23:05:51Z" level=info msg="Reconciliation completed" generation=1 namespace=example resourceVersion=1368 rollout=webservice time_ms=115.25680000000001
time="2026-04-27T23:05:51Z" level=error msg="rollout syncHandler error: spec.strategy.canary.trafficRouting.istio.virtualService.name: Invalid value: \"virtualservice\": virtualservices.networking.istio.io \"virtualservice\" not found" namespace=example rollout=webservice
time="2026-04-27T23:05:51Z" level=info msg="rollout syncHandler queue retries: 2 : key \"example/webservice\"" namespace=example rollout=webservice
time="2026-04-27T23:05:51Z" level=error msg="Unhandled Error" error="spec.strategy.canary.trafficRouting.istio.virtualService.name: Invalid value: \"virtualservice\": virtualservices.networking.istio.io \"virtualservice\" not found" logger=UnhandledError
```

---

arpitjain099:
Fixes #4709

## Problem

The Istio traffic-management reconciler resolves the referenced `VirtualService` name and namespace with `GetVirtualServiceNamespaceName` in `utils/istio/istio.go`. The old logic split the reference on `.` and unconditionally took `fields[0]` as the name and `fields[1]` as the namespace, discarding anything after the second field.

A VirtualService name may legitimately contain a period (it only has to satisfy DNS-1123 subdomain validation, unlike a Service). That broke two cases:

- A real cross-namespace reference to a VirtualService whose name contains a period resolved to the wrong name and namespace. For the manifest in the issue, the VS is `virtualservice.example` in namespace `istio`, referenced as `virtualservice.example.istio`. The controller parsed it as name `virtualservice`, namespace `example`, and reported `virtualservices.networking.istio.io "virtualservice" not found`.
- More generally, the period in such a name was treated as a namespace separator instead of part of the name.

The documented contract is `rollout-vsvc.<vsvc namespace name>`, i.e. the Istio `name.namespace` shorthand where the namespace is the trailing segment. Since a Kubernetes namespace cannot contain a period, the correct parse is: the last dotted segment is the namespace and everything before it is the name.

## Fix

In `GetVirtualServiceNamespaceName`:

- Take the last segment as the namespace and join everything before it as the name.
- Strip the well-known `.cluster.local` suffix first so a fully qualified reference still resolves to the intended name and namespace (this keeps the existing FQDN behaviour, e.g. `test.namespace.cluster.local` stays `test` in `namespace`).

Same-namespace references with no trailing namespace (a bare name, no period) are unchanged. Existing cross-namespace support (`name.namespace`) is preserved; the only behaviour change is that periods inside the name are no longer mistaken for the namespace separator.

## Reproduce (before)

Added the expectation from the issue as a unit test and ran it against the unpatched code:

```
$ go test ./utils/istio/ -run TestGetVirtualServiceNamespaceName -v
    expected: "virtualservice.example"
    actual  : "virtualservice"
    expected: "istio"
    actual  : "example"
--- FAIL: TestGetVirtualServiceNamespaceName
FAIL
```

## After

```
$ go test ./utils/istio/...
ok  	github.com/argoproj/argo-rollouts/utils/istio
$ go test ./rollout/trafficrouting/istio/...
ok  	github.com/argoproj/argo-rollouts/rollout/trafficrouting/istio
$ go test ./pkg/apis/rollouts/validation/...
ok  	github.com/argoproj/argo-rollouts/pkg/apis/rollouts/validation
$ go build ./...
```

## Tests

Added a table-driven `TestGetVirtualServiceNamespaceName` covering: bare name, `name.namespace`, name-with-period plus cross-namespace, multiple periods in the name, and the `.cluster.local` fully qualified form. The pre-existing `TestGetRolloutVirtualServiceKeys` cases (including the `.cluster.local` ones) still pass unchanged.

### DIFF
--- a/utils/istio/istio.go
+++ b/utils/istio/istio.go
@@ -41,10 +41,18 @@ func GetVirtualServiceNamespaceName(vsv string) (string, string) {
 	namespace := ""
 	name := ""
 
+	// Strip the well-known cluster.local DNS suffix before parsing so that a
+	// fully-qualified reference still resolves to the right name/namespace.
+	vsv = strings.TrimSuffix(vsv, ".cluster.local")
+
 	fields := strings.Split(vsv, ".")
 	if len(fields) >= 2 {
-		name = fields[0]
-		namespace = fields[1]
+		// Istio uses the "name.namespace" shorthand and a Kubernetes namespace
+		// cannot contain a period, so the last segment is the namespace and
+		// everything before it is the VirtualService name. This keeps names
+		// that legitimately contain periods intact (see issue #4709).
+		name = strings.Join(fields[:len(fields)-1], ".")
+		namespace = fields[len(fields)-1]
 	} else if len(fields) == 1 {
 		name = fields[0]
 	}
--- a/utils/istio/istio_test.go
+++ b/utils/istio/istio_test.go
@@ -153,3 +153,56 @@ func TestGetRolloutDesinationRuleKeys(t *testing.T) {
 	assert.Len(t, keys, 1)
 	assert.Equal(t, "default/foo", keys[0])
 }
+
+func TestGetVirtualServiceNamespaceName(t *testing.T) {
+	tests := []struct {
+		name              string
+		input             string
+		expectedNamespace string
+		expectedName      string
+	}{
+		{
+			name:              "bare name, no namespace",
+			input:             "virtualservice",
+			expectedNamespace: "",
+			expectedName:      "virtualservice",
+		},
+		{
+			name:              "name.namespace cross-namespace shorthand",
+			input:             "virtualservice.istio",
+			expectedNamespace: "istio",
+			expectedName:      "virtualservice",
+		},
+		{
+			// Issue #4709: a VirtualService whose name contains a period must
+			// not have that period treated as a namespace separator. The VS
+			// "virtualservice.example" lives in namespace "istio" and is
+			// referenced as "virtualservice.example.istio". The namespace is the
+			// last segment; everything before it is the name.
+			name:              "name with period plus cross-namespace",
+			input:             "virtualservice.example.istio",
+			expectedNamespace: "istio",
+			expectedName:      "virtualservice.example",
+		},
+		{
+			name:              "multiple periods in name plus namespace",
+			input:             "my.virtual.service.istio",
+			expectedNamespace: "istio",
+			expectedName:      "my.virtual.service",
+		},
+		{
+			name:              "fully qualified with cluster.local suffix",
+			input:             "test.namespace.cluster.local",
+			expectedNamespace: "namespace",
+			expectedName:      "test",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			ns, name := GetVirtualServiceNamespaceName(tc.input)
+			assert.Equal(t, tc.expectedNamespace, ns)
+			assert.Equal(t, tc.expectedName, name)
+		})
+	}
+}


kostis-codefresh (approved):
While this PR is correct on a technical level, I think we are only solving a symptom of the problem and not the problem itself.

Instead of having code that parses the VirtualService name and trying to locate the namespace, I think it is much easier to simply have a dedicated "namespace" field that clearly says where this virtual service is located. This way no parsing code is needed.

Otherwise i think we might see other instances of this issue resurface in the future.

arpitjain099:
Thanks @kostis-codefresh. I agree a dedicated namespace field is the cleaner long-term fix - it avoids the name parsing entirely. I'm happy to put that together as a separate PR if you'd like to go that route. This change keeps the immediate parsing bug fixed in the meantime, so it could land now and the field can supersede it later, whichever you prefer.

kostis-codefresh:
Yes, this can be done on a separate PR. Thank you.

Of course, the code has to be backwards compatible. The code should look for a dedicated "namespace" property first and if it is not there, then revert to the string parsing behavior of this PR

On a related note, please check the Gateway API plugin because I think it might already have this feature (a dedicated [namespace](https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/blob/main/examples/aws-gateway-api-controller-lattice/proddetail-rollout.yaml#L46) field).  So it would make sense to align with the syntax there.

zachaller (approved):

PR #4768, opened 2026-06-05
fix: leader lease name from instance id

### LINKED ISSUE #4178: Add support of specifying Lease name for Leader Election

kostis-codefresh:
This closes https://github.com/argoproj/argo-rollouts/issues/4178 where users have conflicts with multiple Argo Rollouts controllers as all of them use the same lease name.

There is already a old PR for this at https://github.com/argoproj/argo-rollouts/pull/4325 but it has no tests and most importantly forces users to manually define their selected lease lock name (and if they forget the same problem appears)

It seems however that Argo Workflows has already solved the same problem in a more elegant way where the lease lock name is automatically derived by the instanceID

So the present PR replicates what was solved implemented in Argo Workflows, having a single alignment between both Argo projects :-)

See discussions at
- https://github.com/argoproj/argo-workflows/pull/5214
- https://github.com/argoproj/argo-workflows/pull/5218
- https://github.com/argoproj/argo-workflows/pull/4622

Checklist:

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -108,6 +108,19 @@ type LeaderElectionOptions struct {
 	LeaderElectionRetryPeriod   time.Duration
 }
 
+// leaseLockName returns the name of the lease lock used for leader election. When an
+// instanceID is configured, it is appended to the default lock name so that controllers
+// operating on different instance IDs against a shared cluster elect leaders independently.
+// This mirrors Argo Workflows' behavior:
+// https://github.com/argoproj/argo-workflows/blob/main/cmd/workflow-controller/main.go#L152
+func leaseLockName(instanceID string) string {
+	//Keep backwards compatibility for existing users
+	if instanceID == "" {
+		return defaultLeaderElectionLeaseLockName
+	}
+	return fmt.Sprintf("%s-%s", defaultLeaderElectionLeaseLockName, instanceID)
+}
+
 func NewLeaderElectionOptions() *LeaderElectionOptions {
 	return &LeaderElectionOptions{
 		LeaderElect:                 DefaultLeaderElect,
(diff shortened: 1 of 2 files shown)


kostis-codefresh:
@zachaller It seems you implemented the exact same thing in Argo Workflows :-) PTAL

kostis-codefresh:
Local testing results

Running with no instance id -> backwards compatibility same as before using `argo-rollouts-controller-lock` as name

```
INFO[0000] Starting Healthz Server at 0.0.0.0:8080
INFO[0000] Starting Metric Server at 0.0.0.0:8090
INFO[0000] attempting to acquire leader lease argo-rollouts/argo-rollouts-controller-lock...
INFO[0000] New leader elected: octopus-laptop-win_96b51a9b-6654-4892-8ea9-f7495a368cfd
INFO[0017] successfully acquired lease argo-rollouts/argo-rollouts-controller-lock
```

Running with `go run ./cmd/rollouts-controller/main.go --instance-id my-precious`
gives

```
INFO[0000] Using leader election lease lock name argo-rollouts-controller-lock-my-precious
INFO[0000] attempting to acquire leader lease argo-rollouts/argo-rollouts-controller-lock-my-precious...
INFO[0000] Starting Healthz Server at 0.0.0.0:8080
INFO[0000] Starting Metric Server at 0.0.0.0:8090
INFO[0000] New leader elected: octopus-laptop-win_4bdce946-b38b-4154-b1a9-4ba67dc2f616
INFO[0015] successfully acquired lease argo-rollouts/argo-rollouts-controller-lock-my-precious
```

zachaller (approved):

PR #4770, opened 2026-06-09
fix(analysis)!: evaluate success/failure conditions for non-JSON web metric responses. Fixes #4535

### LINKED ISSUE #4535: Analysis Template Web Provider Succeeds Irrespective Of SuccessCondition/FailureCondition When Response Body is Null Or Not JSON
Checklist:

**Describe the bug**

Analysis Template Web Provider Succeeds Irrespective Of SuccessCondition/FailureCondition When Response Body is Null Or Not JSON

**To Reproduce**

We have the following `AnalysisTemplate` that hits the health endpoint of our service. The health endpoint is a very basic health check; it returns a 200 response code with no body. The `successCondition` is set to compare the result with a value, in our case _fail123_ to fail the `AnalysisTemplate` purposely, but it succeeds irrespective of what we have in the `successCondition`.

```
kind: AnalysisTemplate
metadata:
  name: <template-name>
  namespace: <namespace>
spec:
  metrics:
  - name: ingester-canary-experiment-health
    successCondition: result == 'fail123'
    provider:
      web:
        url: "http://<service-endpoint>/health"
        timeoutSeconds: 20
```

**Expected behavior**

The `AnalysisTemplate` should have failed if `successCondition` doesn't match. 

**Screenshots**

**Version**

1.8.3

**Logs**

Nothing suspicious in the logs

---

something-ai-hash:
## Summary
Fixes a bug where the web metric provider always returned `Successful` for empty or non-JSON HTTP response bodies, without evaluating `successCondition` / `failureCondition`.

When `json.Unmarshal` failed, the code hardcoded `AnalysisPhaseSuccessful`. Now it passes the response body string to `evaluate.EvaluateResult`, consistent with the JSON code path.

Fixes #4535

## Changes
- `metricproviders/webmetric/webmetric.go`: evaluate conditions for non-JSON/empty responses instead of always succeeding
- `metricproviders/webmetric/webmetric_test.go`: add regression tests for empty and plain-text bodies; update existing non-JSON tests to use meaningful conditions

## Test plan
- [x] `go test ./metricproviders/webmetric/... -count=1`
- [x] `golangci-lint run ./metricproviders/webmetric/...`

Checklist:
* [x] This is a bug fix
* [x] PR title follows conventional commits format and references #4535
* [x] Commits are signed with DCO (`git commit -s`)
* [x] CI is green
* [x] Unit tests added/updated
* [x] Tests pass locally
* [x] This changes existing behavior for users relying on non-JSON web metrics with conditions that should fail
* [ ] My organization is added to USERS.md (N/A)


something-ai-hash:
Hii @kostis-codefresh , can you please review this pr once.

kostis-codefresh (changes requested):
Thank you for the contribution.

I have two points though

1. This breaks existing behavior for users that depended on the "bug". If somebody had conditions defined on a non-JSON or empty response body, those conditions are now actually evaluated.  If this PR is merged at the very least we should document this in the release notes. FYI @zachaller 

Also it is important to note that the original PR at https://github.com/argoproj/argo-rollouts/pull/1212/files has tests marked as `When_200Response_And_EmptyBody_Then_Succeed, When_200Response_And_NilBody_Then_Succeed` so for some existing users this might seem like breaking documented behavior.

2. The new behavior does something that is not documented anywhere. The web metric documentation says that "The webhook response must return JSON content.". But if this PR is merged this is not true anymore. 

If this PR is merged now there is another valid scenario. The web metric call now returns pure text "I am OK" and if successconditionm is return == "I am OK", the analysis will pass even though there is no JSON content anywhere.

I think at the very least the [documentation](https://github.com/argoproj/argo-rollouts/blob/master/docs/analysis/web.md) must also be updated to explain that simple text responses are also accepted and show at least one scenario.

something-ai-hash:
Hii @kostis-codefresh ,
Thanks for the review, I've addressed both points:

1. Release notes / Behavior change
I've added a BREAKING CHANGE footer to the commit message so it will be included in the release changelog. 
`The change is intentional per #4535: successCondition and failureCondition are now evaluated for empty and non-JSON response bodies instead of always returning Successful. Users who relied on the old behavior will need to update their AnalysisTemplates.`

2. Documentation
I've updated docs/analysis/web.md to:
- Remove the statement that the webhook response must return JSON
- Add a Plain text responses section with an example where the webhook returns I am OK and successCondition: result == "I am OK"
- Document that empty bodies assign an empty string to result and conditions are evaluated against it

Kindly review

kostis-codefresh (approved):
Thank you!

zachaller (approved):

PR #4782, opened 2026-06-15
test: make tests resilient to nil ptr

agaudreault:
When running the test suite, some test may fail with nil ptr which stops the test suite immediately instead of evaluating other test scenarios.


zachaller:
Nice cleanup — converting the index panics into scoped `require` failures makes package runs much easier to debug.

One small suggestion: consider adding `f.t.Helper()` at the top of `kubeActionAt` and `actionAt`. Without it, a failure is attributed to the line inside the helper rather than the calling test, which slightly undercuts the "easier to debug" goal since every out-of-range failure would point at the same helper line.

```go
func (f *fixture) kubeActionAt(index int) core.Action {
	f.t.Helper()
	actions := filterInformerActions(f.kubeclient.Actions())
	require.Greaterf(f.t, len(actions), index, "expected at least %d kube actions, got %d", index+1, len(actions))
	return actions[index]
}
```

zachaller (approved):

PR #4784, opened 2026-06-15
fix(controller): Exit reconciliation early after in-flight Rollout writes

agaudreault:
## Summary
Follow-up to #4786, which removed `writeBackToInformer` and added `rolloutVersionTracker` to detect when the informer cache hasn't caught up to our last write.
This PR addresses a remaining gap: **within a single reconcile**, the controller can still mutate the Rollout (set revision, initialize Progressing condition, patch pause/resume conditions, etc.) and then **keep going** using in-memory state that no longer matches etcd.
When that happens, we now:
1. Record the written `resourceVersion` in `rolloutVersionTracker`
2. Return from `syncHandler` / `reconcile()` immediately
3. Let the next reconciliation run against updated informer state (or trigger the stale-cache requeue from #4786)
## Problem
A reconcile can perform multiple Rollout writes in one pass, for example:
- `newRolloutContext()` → `createDesiredReplicaSet()` sets revision + Progressing condition
- `checkPausedConditions()` → patches pause/resume/retry conditions
Previously, after each write the controller continued the same reconcile (scale RS, patch status, advance canary/blue-green steps) using stale inputs. That could overwrite user changes (unpause, abort, promote) that landed while the reconcile was in flight.
## Changes
- **`rollout/context.go`**: return early from `reconcile()` when `newRollout != nil` after `checkPausedConditions()`
- **`rollout/controller.go`**: return early from `syncHandler()` when context setup already mutated the Rollout; record `resourceVersion` before returning
- **`rollout/sync.go`**: set `c.newRollout` via `DeepCopy()` and update `c.rollout` consistently after API writes
- **Tests**: update unit tests to the multi-sync pattern (`runWithSyncs`); re-seed informer state between syncs for deterministic harness behavior

### DIFF
--- a/rollout/analysis_test.go
+++ b/rollout/analysis_test.go
@@ -1878,11 +1878,12 @@ func TestDoNotCreateBackgroundAnalysisRunOnNewCanaryRollout(t *testing.T) {
 	f.analysisTemplateLister = append(f.analysisTemplateLister, at)
 	f.objects = append(f.objects, r1, at)
 
-	f.expectCreateReplicaSetAction(rs1)
+	f.expectCreateReplicaSetAction(rs1)   // create replica set
 	f.expectUpdateRolloutStatusAction(r1) // update conditions
+	f.expectGetRolloutAction(r1)          // second reconciliation
 	f.expectUpdateReplicaSetAction(rs1)   // scale replica set
-	f.expectPatchRolloutAction(r1)
-	f.run(getKey(r1, t))
+	f.expectPatchRolloutAction(r1)        // patch status
+	f.runWithSyncs(getKey(r1, t), 2)
 }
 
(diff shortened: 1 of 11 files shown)


kostis-codefresh:
@agaudreault I think this is related https://github.com/argoproj/argo-rollouts/pull/4674

kostis-codefresh:
> The real fix is to remove all that code writing it back to the informer

https://github.com/argoproj/argo-rollouts/pull/4674/files#diff-6f5cf3bcabae1c9148c01d565bb9622bf84f73c88914f1e71ece52f731467eddL472

zachaller:
This can now be updated with https://github.com/argoproj/argo-rollouts/pull/4786 merged

zachaller (approved):

PR #4787, opened 2026-06-16
fix: incorrect evaluation of scale down delay on rollback

agaudreault:
## Issue

When the controller removed the `scale-down-deadline` annotation from the new ReplicaSet, it did not update the in-memory `newRS` correctly. `removeScaleDownDelay` patched the ReplicaSet in the cluster but only reassigned the patch result to a **local** variable, so the caller's `c.newRS.Annotations` map still contained the annotation.

As a result, downstream fast-rollback detection that read `replicasetutil.HasScaleDownDeadline(c.newRS)` later in the same reconcile (in `reconcileAnalysisRuns`, `isBlueGreenFastTracked`, and `shouldFullPromote`) operated on stale state.

## Fix

- **`removeScaleDownDelay`**: delete the annotation from the in-memory map (`delete(rs.Annotations, key)`) so `c.newRS` reflects the removal.
- **Capture the signal once, up front**: because the annotation is now genuinely removed in memory, the "within scale-down delay" state is captured at context creation as `newRSWithinDelay` in `newRolloutContext`, before any removal. **This state is still in memory and will be lost if once the RS is updated and the annotation remove. This creates a window of time where a rollback may not be detected if the status cannot be persisted in this reconciliation. This issue is already present today since newRS is stale.**
- **Introduce `isFastRollback()` / `isRollback()` helpers**: the downstream `HasScaleDownDeadline(c.newRS)` checks are replaced by `isFastRollback()`, to give more context to the condition. A note is added to the rollback function that it will also detect the "stable" state since there a no discriminant factor allowing us to know if we are rolling back to stable, or if the promotion is completed.
- **`reconcileAnalysisRuns` skip path**: since `isFastRollback()` now also covers completed/stable rollouts, the skip path was restructured to cancel current analysis runs and clear the canary current analysis status while retaining blue-green pre/post-promotion runs.

Updated affected unit tests accordingly.


rollout/sync_test.go
```diff
@@ -753,12 +756,12 @@ func TestShouldFullPromoteCanaryAvailableVsDesired(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			stableRS := &appsv1.ReplicaSet{
-				ObjectMeta: metav1.ObjectMeta{Name: "stable"},
+				ObjectMeta: metav1.ObjectMeta{Name: "stable", Labels: map[string]string{v1alpha1.DefaultRolloutUniqueLabelKey: "stablehash"}},
```
agaudreault, line 759: we have to set hashes on the RS since most of the code check things like `Status.StableRS == getHash(newRS)`. Without setting hashes correctly, we are likely to test the "initial deploy" use case.

In this case, since StableRS was not set, this is exactly what was happening.

rollout/replicaset.go
```diff
@@ -36,12 +36,13 @@ func (c *rolloutContext) removeScaleDownDelay(rs *appsv1.ReplicaSet) error {
...
 	}
 	patch := fmt.Sprintf(removeScaleDownAtAnnotationsPatch, v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey)
-	rs, err := c.kubeclientset.AppsV1().ReplicaSets(rs.Namespace).Patch(ctx, rs.Name, patchtypes.JSONPatchType, []byte(patch), metav1.PatchOptions{})
+	_, err := c.kubeclientset.AppsV1().ReplicaSets(rs.Namespace).Patch(ctx, rs.Name, patchtypes.JSONPatchType, []byte(patch), metav1.PatchOptions{})
 	if err != nil {
 		return fmt.Errorf("error removing scale-down-deadline annotation from RS '%s': %w", rs.Name, err)
 	}
 	c.log.Infof("Removed '%s' annotation from RS '%s'", v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey, rs.Name)
-	return err
+	delete(rs.Annotations, v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey)
```
zachaller, line 44: `delete(rs.Annotations, …)` mutates an object owned by the informer cache, which violates the lister's read-only contract.

`rs` here is the same pointer that came from the shared cache — I traced the path against the pinned deps (`client-go v0.34.5`, `kubernetes v1.34.5`) and nothing deep-copies it:

- `replicaSetLister.ReplicaSets(ns).List(...)` returns raw cache pointers; the generated lister explicitly documents *"All objects returned here must be treated as read-only."*
- `ClaimReplicaSets` (`pkg/controller/controller_ref_manager.go`) only filters and re-appends the **same** pointers (`ClaimObject`/`AdoptReplicaSet`/`ReleaseReplicaSet` just read or patch).
- `getReplicaSetsForRollouts` → `FindNewReplicaSet` → `roCtx.newRS = newRS` passes that pointer straight through.

So this `delete` modifies the shared cache map in place. It mostly converges (the cluster patch also removed the annotation), but it's a write to a cache-shared object and relies on a side effect rather than correctness — and it can race with other reconciles reading the same RS before the informer re-syncs.

Note a deep copy *inside* this function alone can't fix it: `rs` aliases the cache object, so any write through it (incl. `rs.Annotations = nil` or `*rs = *rsCopy`) still mutates the cache, and deep-copying locally would re-introduce the original stale-`newRS` bug for the caller. The clean options are to (a) return the updated RS and have callers reassign (`c.newRS, err = c.removeScaleDownDelay(c.newRS)`), or (b) deep-copy at the caller before mutating — mirroring the existing `rsCopy := c.newRS.DeepCopy()` pattern in `sync.go`.

agaudreault: Now the list of RS, newRs, stableRS, etc are all copy of what is in the informer.

https://github.com/argoproj/argo-rollouts/pull/4787/changes/eb08d3a67353ec4507973d267edb93176d3ebeda

zachaller (approved):

PR #4788, opened 2026-06-17
Document Endtest job metric integration

liviulupei:
## Summary

Adds an example showing how Endtest can be used with the Argo Rollouts Job Metrics provider to run end-to-end browser and mobile tests during canary or blue-green deployments.

## Changes

- Adds an Endtest section to the Job Metrics documentation
- Adds a reusable `ClusterAnalysisTemplate` example
- Links to the maintained Endtest Argo integration repository

## Validation

- `git diff --check`
- YAML syntax validated locally
- Verified with multiple Endtest executions through an Argo Rollouts `AnalysisRun`

### DIFF
--- a/docs/analysis/job.md
+++ b/docs/analysis/job.md
@@ -35,6 +35,14 @@ The possible outcomes of a job metric are:
 The last case can happen either when another metric has already failed or when the job is running as a background
 analysis and the canary/blue-green process has finished.
 
+## End-to-end testing with Endtest
+
+[Endtest](https://endtest.io/) can be used as a Job metric to run browser and mobile tests during canary or blue-green deployments. The integration starts one or more Endtest executions, waits for their results, and exits with a non-zero status when any execution reports failed assertions or execution errors.
+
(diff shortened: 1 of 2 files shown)


kostis-codefresh:
Hello, and thank you for your interest in Argo Rollouts

Argo Rollouts is a CNCF Graduated project and must follow the [CNCF vendor-neutrality guidance](https://contribute.cncf.io/projects/best-practices/community/vendor-neutrality/). We cannot link to commercial projects from our documentation. 

Even if we could, then this would not be sustainable in the long run as every other testing vendor will want to open a similar PR.
The Job metric provider is already documented generically, and users can integrate any tool with it. 

If you believe that Argo Rollouts can benefit from deeper integration with Endtest, you are encouraged to create a [metric plugin](https://argo-rollouts.readthedocs.io/en/stable/analysis/plugins/). The plugin mechanism exists exactly for this purpose. But the plugin code should be in a separate Git repository.

PR #4789, opened 2026-06-18
fix(controller): allow Bidirectional mountPropagation on privileged containers. Fixes #3130

### LINKED ISSUE #3130: Rollout objects cannot use Bidirectional mountPropagation
Checklist:

**Describe the bug**

Attempting to use `mountPropagation: Bidirectional` along with the requisite `privileged: true` in the `securityContext` results in the error `Forbidden: Bidirectional mount propagation is available only to privileged containers`
Creating an identical `Deployment` validates and is created as expected
Looking at the validation code, I suspect it comes from the fact it appears to drops privileged for certain operations.

**To Reproduce**

Create the following both in `Deployment` and `Rollout` form (their api version parts provided separately for convenience)

```
apiVersion: apps/v1
kind: Deployment
```
```
kind: Rollout
```
```
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx
        securityContext:
          privileged: true
        volumeMounts:
        - mountPath: /tmp
          mountPropagation: Bidirectional
          name: a-dir-to-mount
      volumes:
      - emptyDir: {}
        name: a-dir-to-mount
```

**Expected behavior**

The `Rollout` changes to be accepted and deployed in the same manner as the `Deployment` resource

**Version**
v1.6.0

**Logs**

```
{"level":"error","msg":"The Rollout \"test\" is invalid: spec.template.spec.containers[1].volumeMounts.mountPropagation: Forbidden: Bidirectional mount propagation is available only to privileged containers","namespace":"public","rollout":"test","time":"2023-10-26T23:04:10Z"}
```

---

sergioparamo:
## Summary

Rollouts whose pod template has a `securityContext.privileged: true` container with `mountPropagation: Bidirectional` were rejected by the rollout webhook with:

```
Forbidden: Bidirectional mount propagation is available only to privileged containers
```

even though the same pod template applied as a `Deployment` is accepted by `kube-apiserver`. Reported in #3130.

## Root cause

The webhook reuses the upstream Kubernetes pod-spec validator, which reads the process-wide `capabilities` singleton. By default `AllowPrivileged=false`, so any container with `privileged: true` would be rejected. To work around that without touching the global, #802 / 8332cee1 added `removeSecurityContextPrivileged`, which strips `Privileged: true` from every container *before* validation.

That mutation silently disabled every cross-field rule in the upstream validator that depends on the real `Privileged` value. The most visible casualty is the rule "`mountPropagation: Bidirectional` requires a privileged container": after the workaround zeros out `Privileged`, the validator sees a Bidirectional mount on a non-privileged container and rejects it. A `Deployment` with the same spec passes because `kube-apiserver` initializes the singleton with `AllowPrivileged=true` at startup.

## Fix

Initialize the `capabilities` singleton with `AllowPrivileged=true` in the validation package's `init()`, mirroring what `kube-apiserver` does. `capabilities.Initialize` uses `sync.Once` internally, so the controller, the `kubectl-argo-rollouts lint` CLI, the webhook and tests all converge on the same value safely. The `removeSecurityContextPrivileged` workaround is removed.

## Side effects (intentional)

Other privileged-dependent upstream rules (e.g. `procMount: Unmasked`, certain capability combinations) are now honestly evaluated by the rollout webhook, matching `Deployment` validation. Specs that the apiserver would have rejected later are now caught at the webhook instead of slipping through the rollout admission silently.

## Test plan

- New regression test `TestValidateRollout/privileged_container_with_bidirectional_mount_propagation` reproduces the exact YAML from #3130.
- `go test ./pkg/apis/rollouts/validation/...` passes.
- `go test ./...` passes.
- `make plugin && ./dist/kubectl-argo-rollouts lint -f ro-3130.yaml` exits 0 (no errors).

Closes #3130.

Checklist:

- [x] (b) this is a bug fix.
- [x] The title of the PR is conventional, states what changed, and suffixes the related issue number.
- [x] I've signed my commits with DCO.
- [x] My builds are green.
- [x] I have written unit tests for my change.
- [x] I have run all tests locally and they pass on my workstation.
- [x] I have used LLM/AI/Agent tools for this PR but I am responsible for all code of this PR.
- [x] I understand what the code does and WHY/HOW it works.
- [x] I know if my code is changing old functionality for existing users.
- [ ] My organization is added to USERS.md (N/A)

### DIFF
--- a/pkg/apis/rollouts/validation/validation.go
+++ b/pkg/apis/rollouts/validation/validation.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/kubernetes/pkg/apis/core"
 	corev1defaults "k8s.io/kubernetes/pkg/apis/core/v1"
 	apivalidation "k8s.io/kubernetes/pkg/apis/core/validation"
+	"k8s.io/kubernetes/pkg/capabilities"
 	"k8s.io/kubernetes/pkg/fieldpath"
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
@@ -98,6 +99,35 @@ var allowAllPodValidationOptions = apivalidation.PodValidationOptions{
 	AllowIndivisibleHugePagesValues: true,
 }
 
+// The upstream k8s pod-spec validation we reuse reads the process-wide
+// capabilities singleton via capabilities.Get(). When it is left at its zero
+// value, AllowPrivileged defaults to false and validation rejects every
+// container that has securityContext.privileged=true. The kube-apiserver sets
+// this flag to true at startup; we mirror that here so that the validator sees
+// the rollout's pod spec as-is and can correctly evaluate cross-field rules
+// such as "Bidirectional mountPropagation requires privileged container".
+//
+// capabilities.Initialize uses sync.Once internally, so callers (controller,
+// CLI lint command, webhook, tests, etc.) all converge on the same value and
+// later calls become no-ops. Trusting the cluster to ultimately enforce
+// privileged-pod admission is consistent with allowAllPodValidationOptions
+// above.
+//
+// Replaces the previous workaround that stripped Privileged from each
+// container before validation; that workaround broke other cross-field rules.
+// See https://github.com/argoproj/argo-rollouts/issues/796 (original) and
+// https://github.com/argoproj/argo-rollouts/issues/3130 (Bidirectional mounts).
+func init() {
+	capabilities.Initialize(capabilities.Capabilities{
+		AllowPrivileged: true,
+		PrivilegedSources: capabilities.PrivilegedSources{
+			HostNetworkSources: []string{},
+			HostPIDSources:     []string{},
+			HostIPCSources:     []string{},
+		},
+	})
+}
+
 func ValidateRollout(rollout *v1alpha1.Rollout) field.ErrorList {
 	allErrs := field.ErrorList{}
 	allErrs = append(allErrs, ValidateRolloutSpec(rollout, field.NewPath("spec"))...)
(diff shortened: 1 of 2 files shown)


kostis-codefresh (changes requested):
Thank you for this contribution

In general, I agree with the PR contents. Two comments though:

1) Is `init()` really needed? It feel it is an overkill in this case.... Can we call capabilities just before the first code segment(s) that start the validation path instead? Or this approach was not good enough for some race conditions?

2) The added unit test is great but only tests happy path scenario (that privilaged/bidirectional gets accepted). Can we also have a test for the other case where a container with Bidirectional still gets rejected if it is non-privilaged?

sergioparamo:
> Thank you for this contribution
> 
> In general, I agree with the PR contents. Two comments though:
> 
> 1. Is `init()` really needed? It feel it is an overkill in this case.... Can we call capabilities just before the first code segment(s) that start the validation path instead? Or this approach was not good enough for some race conditions?
> 2. The added unit test is great but only tests happy path scenario (that privilaged/bidirectional gets accepted). Can we also have a test for the other case where a container with Bidirectional still gets rejected if it is non-privilaged?

Thanks for the review! 
Both points are fair, I will address them in the next push.

1. init() overkill — agreed

You're right, init() is heavier than what is needed now. The only call path that ends up reading capabilities.Get() is ValidateRolloutSpec → validation.ValidatePodTemplateSpecForReplicaSet, so initializing right before that call is equivalent (and sync.Once keeps it idempotent across all callers — controller, lint CLI, webhook, tests). I moved the Capabilities value to a package-level var and the Initialize() call into ValidateRolloutSpec just above the upstream validator.

2. Negative test — agreed

Added a counterpart subtest non-privileged container with bidirectional mount propagation is rejected that asserts the upstream cross-field rule is still enforced when securityContext.privileged is unset. This will guard against future regressions in the opposite direction (i.e. accidentally accepting Bidirectional on non-privileged containers).

kostis-codefresh:
@jessesuen please take a look as you implemented originally https://github.com/argoproj/argo-rollouts/pull/802

kostis-codefresh (approved):
Thanks!

kostis-codefresh (approved):

PR #4805, opened 2026-06-26
fix(experiments): fall back to API when AnalysisRun missing from informer cache. Fixes #4804

### LINKED ISSUE #4804: analysisruns.argoproj.io not found in experiment, but it actually exists
Checklist:

* [ ] I've included steps to reproduce the bug.

**Describe the bug**

On Argo Rollouts 1.8.3, experiments occasionally fail. If I describe one, I the event:
`Warning  AnalysisRunError     11m                rollouts-controller  AnalysisRun 'smoke-test' transitioned from Pending -> Error: analysisruns.argoproj.io "REDACTED" not found`

But the analysisrun is there (`REDACTED` is the same string as above):
```
kubectl get analysisrun REDACTED`
... Successful
```
**Expected behavior**

The experiment should continue.

**Possible solution**
Maybe the fix is similar to https://github.com/argoproj/argo-rollouts/blob/master/rollout/analysis.go#L64, such that if the informer cache doesn't contain the analysisrun, we need to perform a get directly.

---

something-ai-hash:
## What does this PR do?

When the Experiment controller reconciles an AnalysisRun that is already recorded in the experiment's status, it looks the run up via the informer cache (lister). If the informer cache is stale and the AnalysisRun has not yet been synced into it,
the lister returns a `NotFound` error and the controller incorrectly marks the analysis as `Error`, even though the AnalysisRun actually exists in the API server.

This PR adds a fallback: when the lister `Get` fails, the controller performs a direct API `Get` before treating the run as truly missing. This mirrors the existing pattern already used by the Rollout controller in `rollout/analysis.go`.

Fixes #4804

## Testing

Added `TestAnalysisRunMissingFromCacheFallsBackToAPI` in `experiments/analysisrun_test.go`. It records an AnalysisRun in the experiment status, adds it to the API client but **not** the informer cache (simulating a stale lister) and asserts the analysis resolves to `Successful` instead of `Error`.

## Checklist

* [x] (b) this is a bug fix
* [x] The title of the PR is conventional, states what changed, and suffixes the related issue number
* [x] My builds are green
* [ ] My organization is added to USERS.md


rohitraj-carousellgroup:
Hii @kostis-codefresh , Kindly review this PR and let me know your feedback.

zachaller (approved):

PR #4806, opened 2026-06-26
fix(controller): pass log format to plugin subprocess loggers

n0rm4l-me:
## What this fixes

When the controller runs with `--logformat=json`, its own logs come out as JSON — but plugin subprocess logs (traffic router, step, metric provider) always came out as plain text. This made structured log ingestion painful: JSON and text lines mixed in the same stream.

The root cause is in how go-plugin works: it reads plugin stderr and re-emits lines through its own internal hclog logger. When no `Logger` is passed in `ClientConfig`, go-plugin creates a default one with `JSONFormat: false`. The controller's `--logformat` flag had no effect on this path at all.

This was discovered while implementing `--logformat` support in the [Gateway API traffic router plugin](https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/pull/214). The plugin was updated to write structured logs using hclog's native format (`@level`, `@message`, `@timestamp`), but the output was still rendered as plain text by the controller side. Investigation showed the fix had to be here, in the controller.

## What changed

Added `NewPluginLogger()` to `utils/log` — it creates an hclog logger that mirrors the controller's log format by inspecting the global logrus formatter. Wired it into all three plugin clients:

- `rollout/trafficrouting/plugin/client/client.go`
- `rollout/steps/plugin/client/client.go`
- `metricproviders/plugin/client/client.go`

## Behavior

| Scenario | Before | After |
|---|---|---|
| `--logformat=json` | plugin logs: plain text | plugin logs: JSON |
| default (no flag) | plugin logs: plain text | plugin logs: plain text |

No change in default behavior. All existing plugin integrations unaffected.

## Related

- argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi#214 — the plugin-side PR that triggered this investigation

## Checklist

* [x] Either (a) I've created an enhancement proposal and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is conventional with a list of types and scopes found here, states what changed, and suffixes the related issues number.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/golang/protobuf v1.5.4
 	github.com/google/uuid v1.6.0
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
+	github.com/hashicorp/go-hclog v1.6.3
 	github.com/hashicorp/go-plugin v1.7.0
 	github.com/influxdata/influxdb-client-go/v2 v2.14.0
 	github.com/juju/ansiterm v1.0.0
@@ -124,7 +125,6 @@ require (
 	github.com/gregdel/pushover v1.3.1 // indirect
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
-	github.com/hashicorp/go-hclog v1.6.3 // indirect
 	github.com/hashicorp/go-retryablehttp v0.7.7 // indirect
 	github.com/hashicorp/yamux v0.1.2 // indirect
 	github.com/huandu/xstrings v1.5.0 // indirect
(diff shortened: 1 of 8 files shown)


n0rm4l-me:
@kostis-codefresh Could you please check?

kostis-codefresh (changes requested):
You need to fix the checks first. I think you forgot to commit go.mod/go.sum?

n0rm4l-me (commented):
> go mod tidy makes no changes — go-hclog was already a transitive dependency in go.mod. The Verify Codegen failure is an infrastructure flake (tar cache collision on the runner), unrelated to this change. Could you re-run that job?

**UPD:** Turned out it was my mistake. Fixing.

n0rm4l-me:
@kostis-codefresh Could you please check?

Hariharasuthan99 (approved):

kostis-codefresh:
Thanks. I think it makes sense to wait for https://github.com/argoproj-labs/rollouts-plugin-trafficrouter-gatewayapi/pull/214 to be merged first before merging this one. I want to do some additional testing on my side.

PR #4810, opened 2026-06-29
feat(metricprovider): support custom CA cert for Prometheus over TLS

### LINKED ISSUE #2298: Prometheus metrics should support self-signed certs.

ifnotabhielse:
## What

Adds a `caCert` field to the Prometheus AnalysisTemplate metric provider. When set (inline PEM), the metric provider's HTTP client verifies the Prometheus server certificate against the supplied CA **in addition to** the system roots.

```yaml
provider:
  prometheus:
    address: https://prometheus.example.com:9090
    caCert: "{{ args.prometheus-ca }}"   # PEM, typically sourced from a Secret via args
    query: ...
```

## Why

Prometheus endpoints hosted in-cluster commonly use self-signed / internal-CA certs. Today the only option is `insecure: true`, which disables verification entirely (MITM-exposed). `caCert` lets users keep verification on while trusting their private CA — matching the maintainer-proposed design in the issue.

## Notes

- Fully backward compatible: empty `caCert` keeps the existing shared secure/insecure transports — no behavior change. `caCert` and `insecure` are independent (`insecure: true` still takes precedence).
- Invalid PEM surfaces an error rather than silently skipping verification.
- Generated artifacts (protobuf, openapi, CRDs, install manifests, docs schema) regenerated via the repo's `make` codegen targets; docs updated.

## Testing

- New unit tests in `metricproviders/prometheus/`: default (no TLS config), `insecure:true`, valid custom CA, and invalid-PEM error path.
- `go build ./...`, package tests, `go vet`, and `golangci-lint` (v2.11.1, matching CI) all pass.

Fixes #2298


kostis-codefresh (commented):
Isn't this a duplicate of https://github.com/argoproj/argo-rollouts/pull/4635 ?

ifnotabhielse:
@kostis-codefresh good catch — #4635 does predate this and targets the same issue (#2298).

The one difference: #4635 uses `caPath` (a file path to a PEM file); this uses an inline `caCert` string (templatable from a Secret via args), matching the original proposal in #2298.

Happy to close this in favor of #4635, or contribute the inline variant there as a follow-up — whichever you all prefer.

kostis-codefresh:
The present PR has merge conflicts and failed tests. So right now, unless you have any objections I am moving forward with https://github.com/argoproj/argo-rollouts/pull/4964

PR #4812, opened 2026-06-30
feat(metricprovider): make Datadog API request timeout configurable

### LINKED ISSUE #4813: Make the Datadog metric provider HTTP request timeout configurable
# Summary

The Datadog metric provider uses a hard-coded 10 second HTTP client timeout for every request to the Datadog API:

https://github.com/argoproj/argo-rollouts/blob/master/metricproviders/datadog/datadog.go#L157

There is currently no way to override it. This enhancement proposes adding an optional `requestTimeout` field to the `datadog` provider spec that overrides the HTTP client timeout, defaulting to `10s` when omitted so existing behavior is unchanged.

```yaml
provider:
  datadog:
    apiVersion: v2
    interval: 5m
    requestTimeout: 30s   # override the default 10s HTTP client timeout
    queries:
      a: count:requests.errors{service:my-service}.as_count()
      b: count:requests{service:my-service}.as_count()
    formula: "moving_rollup(a, 300, 'sum') / moving_rollup(b, 300, 'sum') * 100"
```

# Use Cases

Expensive Datadog v2 formula/queries measurements (formulas built from many sub-queries, high-cardinality metrics, or wide time windows) can occasionally take longer than 10 seconds to return, especially during deploy bursts when the backend is under load. When that happens the request is aborted client-side and the measurement fails with:

Post "https://api.datadoghq.com/api/v2/query/scalar": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

This causes spurious AnalysisRun errors / failed rollouts even though the query itself is valid and would have succeeded given a little more time. Allowing users to raise the timeout for these specific analyses (or lower it for fast-failing ones) lets them tune it to their query cost without changing controller-wide behavior.

---

gferrate:
## What this does

The Datadog provider hard-codes a 10s HTTP client timeout, so a slow query has no
room to finish. This adds an optional `requestTimeout` field on the `datadog`
provider spec. When set it overrides the client timeout; when omitted it stays at
10s, so existing users see no change.

```yaml
provider:
  datadog:
    apiVersion: v2
    interval: 5m
    requestTimeout: 30s
    query: "..."
```

## Why

Heavier v2 formula/queries measurements sometimes need longer than 10s, especially
during deploy bursts when the backend is busy. The request gets cut off client-side
and fails with `context deadline exceeded (Client.Timeout exceeded...)` even though
the query is valid. This lets you bump the timeout per analysis instead.

Fixes #4813.

## Checklist

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/4813) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.


metricproviders/datadog/datadog.go
```diff
@@ -152,9 +152,17 @@ func (p *Provider) Run(run *v1alpha1.AnalysisRun, metric v1alpha1.Metric) v1alph
 	request.Header.Set("DD-API-KEY", p.config.ApiKey)
 	request.Header.Set("DD-APPLICATION-KEY", p.config.AppKey)
 
-	// Send Request
+	// Send Request. The client timeout defaults to 10s and can be overridden via the
+	// metric's spec (e.g. for expensive v2 formula queries that take longer to return).
+	timeout := time.Duration(10) * time.Second
+	if dd.RequestTimeout != "" {
```
kostis-codefresh, line 158: Should we add some validation in validateIncomingProps()

Maybe not allow 0? Or other validation checks?

metricproviders/datadog/datadogTimeout_test.go
```diff
@@ -0,0 +1,107 @@
+package datadog
```
kostis-codefresh: I don't think we need a separate file just for the timeout test.

Just add the code in the existing test files for datadog.

Is there any special reason this scenario is placed on its own file?

metricproviders/datadog/datadog_test.go
```diff
@@ -58,6 +58,11 @@ func TestDatadogSpecDefaults(t *testing.T) {
 		assert.Equal(t, "\"5m\"", defaultInterval, "Default interval should be \"5m\" ")
 	})
 
+	t.Run("requestTimeout: Validate default is 10s", func(t *testing.T) {
+		defaultRequestTimeout := string(ddSpec.Properties["requestTimeout"].Default.Raw)
```
kostis-codefresh, line 67: I think we need a least the following tests

1) happy path scenario (i.e. timeout 5s)
2) Backwards compatibily (no timeout means 10second as before)
3) Error cases (i.e. timeout 0 if we want to allow that)

pkg/apiclient/rollout/rollout.swagger.json
```diff
@@ -1225,6 +1225,10 @@
         "secretRef": {
           "$ref": "#/definitions/github.com.argoproj.argo_rollouts.pkg.apis.rollouts.v1alpha1.SecretRef",
           "title": "Secret refers to the name of the secret that should be used for an analysis and should exists in the namespace where the controller is.\n+optional"
+        },
+        "requestTimeout": {
+          "type": "string",
+          "title": "+kubebuilder:default=\"10s\"\nRequestTimeout overrides the HTTP client timeout for requests to the Datadog API (e.g. 10s, 30s; default: 10s).\n+optional"
```
kostis-codefresh, line 1231: is the kubebuilder annotation supposed to be in the string itself?

kostis-codefresh (changes requested):
Thank you for this!

I added some minor comments.

zachaller (dismissed):

kostis-codefresh:
@gferrate Just saw that you are working with Datadog? Maybe you want to help/review this PR as well? https://github.com/argoproj/argo-rollouts/pull/4746

gferrate:
> @gferrate Just saw that you are working with Datadog? Maybe you want to help/review this PR as well? #4746

@kostis-codefresh Indeed, let me take a look!

metricproviders/datadog/datadog.go
gferrate: Fair 👍 . Let me add these.

metricproviders/datadog/datadog_test.go
gferrate: Addressed in my last commit.

metricproviders/datadog/datadogTimeout_test.go
gferrate: Not really. Addressed in my last commit.

pkg/apiclient/rollout/rollout.swagger.json
gferrate: I agree that this looks like a leak, but this has been auto-generated and seems like a pre-existing issue (interval, apiVersion, aggregator show the same thing). I don't want to manually edit the swagger file since it will get overwritten in the next PR.

Worth doing a follow-up PR?

kostis-codefresh: yes. No worries. Just found it strange.

kostis-codefresh (approved):

zachaller (approved):

PR #4831, opened 2026-07-08
fix(trafficrouting): remove all experiment destinations from Istio VirtualService on completion

killuazhu:
## What this PR does

When an experiment adds **more than one** destination to an Istio `VirtualService` route (for example a weighted experiment with both a baseline and a canary template `service:`), completing the experiment could leave one of the added destinations behind. Which one was left over was non-deterministic — sometimes the baseline service, sometimes the canary service.

### Root cause

Two interacting issues in `rollout/trafficrouting/istio/istio.go`:

1. **`patchVirtualService`** deleted destinations in place using their original indices, applied in **ascending** order. Removing a lower index first shrinks the destination slice, so a later deletion's index points at the wrong destination or falls out of range. The out-of-range case falls into the append branch and *re-adds* the destination (with `weight: 0`) instead of removing it.
2. **`processRoutes`** appended new experiment destinations while iterating a Go map, so their order in the route was randomized. That randomization decided which service ended up at the surviving higher index — hence the non-determinism.

### Fix

- Sort patches before applying them: all non-delete patches first (so in-place weight updates and appends operate on the original indices), then deletions in **descending** index order so earlier removals do not shift the indices of later ones.
- Iterate the new-destination map in sorted key order so the resulting route ordering is deterministic and reproducible.

### Testing

Added `TestHostSplitExperimentStepMultipleServicesRemoved`, which adds two experiment destinations and then reconciles with no additional destinations. It fails on the old code:

```
"[{stable 90} {canary 10} {exp-baseline 20} {exp-baseline 0}]" should have 2 item(s), but has 4
```

and passes with the fix (route collapses back to just `stable` + `canary`). The full `./rollout/trafficrouting/istio/...` suite passes.

## Checklist

* [x] (b) this is a bug fix
* [x] The title of the PR is conventional and states what changed
* [x] My builds are green

### DIFF
--- a/rollout/trafficrouting/istio/istio.go
+++ b/rollout/trafficrouting/istio/istio.go
@@ -85,6 +85,27 @@ const (
 )
 
 func (patches virtualServicePatches) patchVirtualService(httpRoutes []any, tlsRoutes []any, tcpRoutes []any) error {
+	// Apply patches in an order that keeps destination indices valid. All
+	// non-delete patches (in-place weight updates and appends) are applied
+	// first so they operate on the original indices, then deletions are
+	// applied in descending index order so removing one destination does not
+	// shift the indices of the remaining deletions. Without this ordering,
+	// deleting a lower index first shrinks the slice and makes a later
+	// deletion's index either point at the wrong destination or fall out of
+	// range, which caused leftover experiment destinations in the
+	// VirtualService when an experiment completed.
+	slices.SortStableFunc(patches, func(a, b virtualServicePatch) int {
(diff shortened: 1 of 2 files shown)


kostis-codefresh (approved):
I don't have any istio installation to test. But code looks correct to me.

killuazhu:
> I don't have any istio installation to test. But code looks correct to me.

Besides the validation from UT, I was able to validate in a local Istio env that the patch is working as expected.

zachaller (approved):

PR #4832, opened 2026-07-09
fix: guard empty Values in injected anti-affinity check

AruneshDwivedi:
IfInjectedAntiAffinityRuleNeedsUpdate keyed a pod affinity term match by label only (via HasInjectedAntiAffinityRule) and then indexed Values[0]. A user-defined anti-affinity term using the rollouts-pod-template-hash key with Operator: Exists (no Values) made HasInjectedAntiAffinityRule match it, causing an index-out-of-range panic during reconcile.

Guard the access with a length check so such terms are treated as not needing an update.

Added a regression test in TestIfInjectedAntiAffinityRuleNeedsUpdate covering a term with no Values.

### DIFF
--- a/utils/replicaset/replicaset.go
+++ b/utils/replicaset/replicaset.go
@@ -212,7 +212,7 @@ func IfInjectedAntiAffinityRuleNeedsUpdate(affinity *corev1.Affinity, rollout v1
 	currentPodHash := hash.ComputePodTemplateHash(&rollout.Spec.Template, rollout.Status.CollisionCount)
 	if podAffinityTerm != nil && rollout.Status.StableRS != currentPodHash {
 		for _, labelSelectorRequirement := range podAffinityTerm.LabelSelector.MatchExpressions {
-			if labelSelectorRequirement.Key == v1alpha1.DefaultRolloutUniqueLabelKey && labelSelectorRequirement.Values[0] != rollout.Status.StableRS {
+			if labelSelectorRequirement.Key == v1alpha1.DefaultRolloutUniqueLabelKey && len(labelSelectorRequirement.Values) > 0 && labelSelectorRequirement.Values[0] != rollout.Status.StableRS {
 				return true
 			}
 		}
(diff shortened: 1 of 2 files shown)


kostis-codefresh (commented):
Is there an open issue about this?

Could you please post an example that showcases the problem first?

AruneshDwivedi:
There isn't a separate issue open — this is the fix for a crash we hit in reconcile. Here's what triggers it.

`HasInjectedAntiAffinityRule` matches a pod anti-affinity term purely by *key* (it checks `labelSelectorRequirement.Key == v1alpha1.DefaultRolloutUniqueLabelKey`), without looking at the operator or values. So a user-supplied term like this gets matched:

```yaml
affinity:
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
            - key: rollouts-pod-template-hash
              operator: Exists        # no values
        topologyKey: kubernetes.io/hostname
```

Then `IfInjectedAntiAffinityRuleNeedsUpdate` does:

```go
if labelSelectorRequirement.Key == v1alpha1.DefaultRolloutUniqueLabelKey && labelSelectorRequirement.Values[0] != rollout.Status.StableRS {
```

`operator: Exists` produces a `MatchExpression` with an **empty** `Values` slice, so `Values[0]` is an index-out-of-range panic — not a "this is wrong" check, an actual crash on every reconcile once a user adds such a term.

The fix guards with `len(labelSelectorRequirement.Values) > 0` so terms without values are treated as not needing an update (matching the existing regression test added in `TestIfInjectedAntiAffinityRuleNeedsUpdate`).

AruneshDwivedi:
Opened a tracking issue with the exact reproducer: #4840

To recap the trigger: `HasInjectedAntiAffinityRule` matches a pod anti-affinity term by key only, so a user term with `operator: Exists` (empty `Values`) is matched, and `IfInjectedAntiAffinityRuleNeedsUpdate` then does `Values[0]` — an index-out-of-range panic on every reconcile. The example YAML lives in #4840. The fix guards with `len(Values) > 0` and adds a regression test.

AruneshDwivedi:
@kostis-codefresh opened a tracking issue with the full reproducer: #4840. The fix here guards the Values[0] access with a length check so a term using operator: Exists (empty Values) is treated as not needing an update, which avoids the index-out-of-range panic on reconcile.

kostis-codefresh (approved):
Thank you!

zachaller (approved):

PR #4833, opened 2026-07-09
fix(controller): Downgrade log level for unconfigured notification triggers. Fixes #3459

### LINKED ISSUE #3459: Downgrade log messages for "...trigger '<EVENT>' is not configured"
# Summary
This concerns notifications around rollout lifecycle. 

These messages are very noisy and dont feel like errors. they should be info (maybe warning?)

example message: `time="2024-03-21T18:29:30Z" level=error msg="Failed to run trigger, trigger: on-scaling-replica-set, destination: {slack argo-rollouts-notifications}, namespace config:  : trigger 'on-scaling-replica-set' is not configured"`

even better would be fire once the first time and never again. 

i want to do this  work but first wanted to check if there is reason these are error level messages. 

---

alliasgher:
When a Rollout lifecycle event fires but the corresponding notification trigger is not configured, `notificationsAPI.RunTrigger` returns a `trigger '<name>' is not configured` error. `sendNotifications` logged this at **error** level and appended it to the returned errors, so users who never configured a given trigger saw a steady stream of `Failed to run trigger ... is not configured` error logs for normal operation.

Per [@zachaller's suggestion on the issue](https://github.com/argoproj/argo-rollouts/issues/3459#issuecomment-2013737506), detect the `is not configured` error in the `RunTrigger` branch of `utils/record/record.go` and log it at **debug** level, skipping it instead of treating it as a notification failure. All other trigger errors continue to log at error level and surface as before.

Fixes #3459

Checklist:

* [x] (b) this is a bug fix
* [x] The title of the PR is conventional, states what changed, and suffixes the related issue number.
* [x] My builds are green.
* [x] I have written unit tests for my change (`TestSendNotificationsTriggerNotConfigured`, and updated `TestSendNotificationsNoTrigger`).


Hariharasuthan99 (approved):
LGTM

zachaller (approved):

PR #4836, opened 2026-07-11
feat(dashboard): add a live timing bar to canary pause steps

### LINKED ISSUE #1183: When paused, display how long the pause is for.
# Summary

In both the CLI and the UI, it would be great if Rollouts could show a timer for how long it's waiting for (e.g. how long this current canary step is meant to last). Something like:

```
Status:          ॥ Paused (10 min remaining)
```

# Use Cases

It's helpful to know how much time is left until the rollout progresses further.

---

apellex:
Adds a live "time remaining" indicator for **timed canary pauses** — in both the dashboard and the `kubectl argo rollouts` CLI — so you can see at a glance how long until a paused rollout proceeds.

Fixes #1183 — a request open since 2022 to surface pause time remaining. Neither the dashboard nor the CLI surfaced it before this PR.

### Dashboard
The current pause step's card fills left-to-right as the pause elapses. Indefinite (manual) pauses are unchanged.

### CLI
`kubectl argo rollouts get rollout` gains a `Pause:` line in the canary summary showing the time left on the current timed pause (e.g. `Pause: 15s remaining`). Indefinite pauses and non-paused steps are unaffected.

### How it works

The controller resolves the pause timing and reports it as two plain integers on `RolloutInfo`, so no client parses a duration or reads a clock:

| field | meaning |
|---|---|
| `pauseDurationSeconds` (22) | the step's duration, normalized via `RolloutPause.DurationSeconds()`. `0` when the pause is indefinite or not in progress, `-1` when the configured value is not parseable. |
| `pauseRemainingSeconds` (23) | how much was left when the message was serialized, clamped to `0 <= remaining <= duration`. |

- **CLI:** `PauseStepRemaining` does no arithmetic — it formats `pauseRemainingSeconds` and returns.
- **Dashboard:** derives its starting position from `(duration - remaining) / duration` and runs a single CSS transition over the remainder. No per-second re-rendering, and no comparison against the browser's clock.

Because the remaining time is **server-reported**, the bar is still correct after a page refresh. This is the main difference from the stalled client-only attempt in #4617, which counted from page-load and reset on reload.

### Why two fields rather than just "remaining"

The CLI prints a number and exits, so the remainder alone is enough. The dashboard is drawing a *bar*, which needs a fraction: knowing 60 minutes are left says nothing about where to start the fill without knowing the pause was 90 minutes long. The duration is the denominator.

Both values are relative integers rather than a start timestamp, so a viewer whose system clock is wrong still sees a correct bar.

### Also fixes a duration-parsing bug

An earlier revision of this PR parsed the duration in TypeScript with `^(\d+)(s|m|h)?$`. A pause duration is resolved by Go's `time.ParseDuration`, which accepts a much larger grammar, so every compound or fractional form fell through to `0` — which the progress helper read as "no time left" and painted the bar **100% full**. A 90-minute pause rendered as already finished:

```
15m        -> 900     ok
1h         -> 3600    ok
1h30m      -> 0       wrong
1.5h       -> 0       wrong
2h45m30s   -> 0       wrong
500ms      -> 0       wrong
```

Normalizing in the controller removes the second parser entirely. `TestRolloutInfoPauseDurationSeconds` now pins every form `time.ParseDuration` accepts, including the truncation (`500ms` -> `0`) and error (`1z` -> `-1`) paths.

### Design choices
- **Timed pauses** (e.g. `pause: {duration: 40s}`) -> dashboard card fills (no countdown text, keeping the panel minimal); CLI shows `Pause: Ns remaining`.
- **Indefinite pauses** (`pause: {}`) -> unchanged in both. `pauseDurationSeconds` is `0`, and the dashboard renders no bar rather than an instantly-complete one.
- **Unparseable durations** -> `-1`, also rendered as no bar.

### Demo

UI
Screen Recording 2026-07-10 at 9 45 45 PM

CLI
Screen Recording 2026-07-21 at 11 22 13 AM

---

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore. — implements the community request tracked in #1183.
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged. — Go `info` builder + CLI `get` unit tests, plus UI jest tests.
* [x] I have run all tests locally (including the flaky ones) and they pass on my workstation — ran the affected unit suites (Go `info` + `cmd/get`, UI jest).
* [x] I know if my code is just adding new functionality or changing old functionality for existing users — purely additive; no change to existing behavior.
* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md). — N/A


kostis-codefresh:
Related? https://github.com/argoproj/argo-rollouts/pull/4617

apellex:
> Related? #4617

@kostis-codefresh yes it is related, however:
- #4617 is a front-end only implementation that doesn't survive hard refresh
- `(props.step.pause as any).startTime)` is problematic too because `GithubComArgoprojArgoRolloutsPkgApisRolloutsV1alpha1RolloutPause` carries no `startTime`
- Performance-wise CSS transition is superior to full Step re-render
- My change is minimalistic (indefinite pause stays the same, only the fill bar animates for timed pause)
- has no activity since February

kostis-codefresh:
Hey. I took a quick look, and this seems to address only the UI. But the original issue says "In both the CLI and the UI". 

Was the CLI side intentionally left out for another PR, or was it out of scope for now? I am asking because  if this PR gets merged then the original issue will be autoclosed.

apellex:
> Hey. I took a quick look, and this seems to address only the UI. But the original issue says "In both the CLI and the UI".
> 
> Was the CLI side intentionally left out for another PR, or was it out of scope for now? I am asking because if this PR gets merged then the original issue will be autoclosed.

@kostis-codefresh Thanks for the review. Honestly it was left out because I personally don't use the CLI. ~~I'm leaning scoping down to covering just UI for now. I can remove the issue tag.~~ On second thought, the CLI change is minimal I will implement it as well so this PR can close #1183 completely.

apellex:
> Hey. I took a quick look, and this seems to address only the UI. But the original issue says "In both the CLI and the UI".
> 
> Was the CLI side intentionally left out for another PR, or was it out of scope for now? I am asking because if this PR gets merged then the original issue will be autoclosed.

@kostis-codefresh PR description updated to include clip showing the "Pause: Xs remaining" in the CLI `get --watch`

apellex:
Hi @kostis-codefresh do you mind giving this PR another quick look? It will close issue #1183

ui/src/app/components/rollout/pause-timer.test.ts
```diff
@@ -0,0 +1,42 @@
+import {computePauseProgress, durationToSeconds} from './pause-timer';
+
+describe('durationToSeconds', () => {
+    it('parses unit suffixes and bare seconds', () => {
+        expect(durationToSeconds('15m')).toBe(900);
+        expect(durationToSeconds('30s')).toBe(30);
+        expect(durationToSeconds('1h')).toBe(3600);
+        expect(durationToSeconds('300')).toBe(300);
```
kostis-codefresh, line 8: What happens with a test 1h30m?

ui/src/app/components/rollout/pause-timer.ts
```diff
@@ -0,0 +1,36 @@
+export const durationToSeconds = (duration: string): number => {
+    if (!duration) {
+        return 0;
+    }
+    const match = duration.match(/^(\d+)(s|m|h)?$/);
```
kostis-codefresh, line 5: Does this match 1h30m? I think it is valid in Go?

ui/src/app/webpack.dev.js
```diff
@@ -5,6 +5,9 @@ const webpack = require('webpack');
 
 module.exports = merge(common, {
     mode: 'development',
+    watchOptions: {
+        ignored: [/dist/, /node_modules/],
```
kostis-codefresh, line 9: What exactly this move does from plugins to one level above?

kostis-codefresh (changes requested):
First a disclaimer: I am not a UI expert

But looking at this code, it seems to me that both the UI and the CLI just get the time a pause is started and then both calculate the remaining time. And of course we need to be sure that the calculation matches what the controller has in mind.

But do we really need this abstraction? Two sources of truth that match?

Wouldn't it make sense to simly ask the controller directly how much time is remaining and just show it without any calculations? And then all formating (such as 1h30m) stays within the controller....

ui/src/app/webpack.dev.js
apellex: I reverted this change as it was unrelated to the PR/issue

ui/src/app/components/rollout/pause-timer.test.ts
apellex: Updated to address this miss

ui/src/app/components/rollout/pause-timer.ts
apellex: Updated to remove this bad matching logic

apellex:
> First a disclaimer: I am not a UI expert
> 
> But looking at this code, it seems to me that both the UI and the CLI just get the time a pause is started and then both calculate the remaining time. And of course we need to be sure that the calculation matches what the controller has in mind.
> 
> But do we really need this abstraction? Two sources of truth that match?
> 
> Wouldn't it make sense to simly ask the controller directly how much time is remaining and just show it without any calculations? And then all formating (such as 1h30m) stays within the controller....

Thank you for reviewing. I agree with your approach here and have updated the implementation to remove the calculation/parsing on the client side.

PR #4838, opened 2026-07-13
fix(trafficrouting): do not re-point Istio canary subset to a not-yet-scaled-up ReplicaSet. Fixes #4775

### LINKED ISSUE #4775: Subset-based Istio canary emits transient 503 no_healthy_upstream when a new revision supersedes an in-flight canary (v1.9.0)
Checklist:

**Describe the bug**

With **subset-based Istio traffic routing** (DestinationRule `canarySubsetName`/`stableSubsetName`), if a **new Rollout revision is pushed while a canary step is in progress** (e.g. paused at `setWeight: 5`), clients receive a short burst of `503` with Envoy response flag `UH` (`no_healthy_upstream`) on the `…|http/canary` cluster. The stable subset is unaffected.

This still reproduces on **v1.9.0**, which already contains both the `shouldDelayDestinationRuleUpdate` guard and the "reset weight to 0 before UpdateHash" block.

**Root cause (code analysis):** in `rollout/trafficrouting/istio/istio.go`, `shouldDelayDestinationRuleUpdate` skips a ReplicaSet when `*rs.Spec.Replicas == 0`:

```go
for _, rs := range r.replicaSets {
    if *rs.Spec.Replicas == 0 { continue }   // <-- the not-yet-scaled new canary RS is skipped
    rsHash := rs.Labels[...]
    if rsHash == "" || (rsHash != stableHash && rsHash != canaryHash) { continue }
    if abortOrDynamicRollbackToStable && rsHash == canaryHash { continue }
    if !replicasetutil.IsReplicaSetAvailable(rs) { return true, rs.Name }
}
```

On the reconcile that runs `UpdateHash`, the new canary RS has not been scaled up yet (`spec.replicas == 0`), so it is skipped and the guard returns `false`. `UpdateHash` then re-points the **canary subset to the not-ready new hash**, de-selecting the still-healthy old canary pods. The guard only begins firing on the **next** reconcile (after scale-up) — too late, the DestinationRule is already switched. The "reset weight to 0 before UpdateHash" block can't prevent it either, because the VirtualService (weight) and DestinationRule (subset/EDS) are separate resources that Istio proxies apply **non-atomically**, so the canary cluster is "weighted but empty" during xDS convergence.

**To Reproduce**

1. Rollout with `trafficRouting.istio` using DestinationRule subsets (`canarySubsetName`, `stableSubsetName`), a single VirtualService route, `workloadRef` → Deployment, `minReadySeconds: 30`, steps `setWeight: 5 → pause → setWeight: 10 → …`.
2. Start a rollout; let it reach `setWeight: 5` and pause. The canary subset now points to revision A's pod-template-hash and serves ~5%.
3. Push a **new revision B while still paused** (supersede the in-flight canary).
4. Observe `503 UH no_healthy_upstream` on `upstream_cluster=inbound-vip|<port>|http/canary` for the xDS convergence window (~tens of seconds), even though revision A's pods stay Ready.

**Expected behavior**

No 503s. The canary subset must not be re-pointed to a new canary ReplicaSet that is not yet available; the guard should also cover the `spec.replicas == 0` (not-yet-scaled) case for the *target* canary hash — or the canary subset should stay on the previous (ready) hash until the new one is available.

**Version**

v1.9.0

**Logs**

Controller (timestamps relative to abort `T0`):

```
T-0.7s  Created ReplicaSet <NEW> (revision N+1)                                   # spec.replicas = 0
T+0.0s  reason: UpdatedDestinationRule  DestinationRule subset updated (canary: <NEW>, stable: <STABLE>)
T+0.0s  reason: TrafficWeightUpdated    Traffic weight updated from 5 to 0
T+0.04s reason: ScalingReplicaSet       Scaled up ReplicaSet <NEW> from 0 to N
T+0.2s  delaying destination rule switch: ReplicaSet <NEW> not fully available     # repeats; already too late
```

Envoy / waypoint access log during the window:

```
response_code=503  response_flags=UH  response_code_details=no_healthy_upstream
route_name=main    upstream_cluster=inbound-vip|<port>|http/canary    x_envoy_attempt_count<=1
```

kube-state-metrics across the 503 window:
- old canary RS `kube_replicaset_status_ready_replicas` = N (continuously — pods alive, just de-selected)
- new canary RS `kube_replicaset_status_ready_replicas` = 0 until ~T+60s (`minReadySeconds:30` + startup)

503 stops when `weight=0` finishes propagating to all proxies — not when new pods become Ready.

myeongseok-rpls:
Checklist:

* [x] I have run all tests locally (including the flaky ones) and they pass on my workstation (ran `./rollout/...` including the full istio package locally; full suite green in CI)

Fixes #4775

## Summary

With subset-based Istio traffic routing (DestinationRule `canarySubsetName`/`stableSubsetName`), pushing a new revision while a canary step is in progress causes a burst of `503 UH no_healthy_upstream` on the canary cluster: `UpdateHash` re-points the canary subset to the new revision's pod-template-hash **before** the new ReplicaSet has been scaled up, de-selecting the still-healthy old canary pods.

## Root cause

`shouldDelayDestinationRuleUpdate` skips ReplicaSets with `spec.replicas == 0`:

https://github.com/argoproj/argo-rollouts/blob/bc55226ee61b1a4f2f1e9313d1a4b3da45825793/rollout/trafficrouting/istio/istio.go#L361-L364

On the reconcile where the new revision is first observed, the new canary RS exists but has not been scaled up yet (`spec.replicas == 0` — scaling happens later in `rolloutCanary`, after `reconcileTrafficRouting`). The guard therefore skips it and returns `false`, and `UpdateHash` switches the canary subset to a hash with zero pods. The guard only starts firing on the *next* reconcile, after scale-up — too late. The "reset weight to 0 before UpdateHash" block (#3273-era mitigation in `reconcileTrafficRouting`) cannot prevent this either, because the VirtualService (weight) and DestinationRule (subset/EDS) are separate resources that proxies apply non-atomically, so the canary cluster is "weighted but empty" during xDS convergence.

## Why not extend the existing delay-error guard

The obvious fix — making `shouldDelayDestinationRuleUpdate` also fire for the zero-replica target canary RS — **deadlocks**: the guard surfaces as an error from `UpdateHash`, which fails `reconcileTrafficRouting`, which returns early from `rolloutCanary` *before* `reconcileCanaryReplicaSets` runs. The new RS would never be scaled up, so the guard would fire forever. (The existing error path is safe only because pod *readiness* progresses via the kubelet, independent of the controller.)

## Fix

Add a separate guard, `shouldSkipDestinationRuleUpdate`: when the target canary hash belongs to a ReplicaSet that has not been scaled up yet (`spec.replicas == 0`), skip the DestinationRule update **without returning an error**. The canary subset stays on the previous (healthy) hash, the reconcile proceeds so the new RS gets scaled up, and once it is scaled up the existing delay guard takes over (delay until fully available, then switch). Net effect: the subset switch now happens only when the new canary RS is fully available, matching the behavior a first-time canary rollout already gets from #2507.

The skip intentionally does **not** apply when:
- aborting or dynamically rolling back to stable (traffic shifts to stable; re-pointing the canary subset is required and safe — same exemption as the existing guard),
- `canaryHash == stableHash` (fully promoted) or `canaryHash == ""`,
- canary/stable Services are configured (guards don't apply, same as existing behavior).

Resulting controller sequence for the issue's repro (revision B pushed while paused at `setWeight: 5` on revision A):

1. weight reset 5→0, DR update **skipped** (canary subset still on A, pods healthy → no 503 during propagation)
2. RS(B) scaled up; DR update delayed by the existing guard while B is unavailable
3. B fully available → DR canary subset switched A→B, weight raised — both targets healthy at all times

## Testing

- `TestUpdateHashSkipsNotYetScaledUpCanary` — reproduces the bug at the `UpdateHash` level (fails without the fix): no DR update and no error for a zero-replica target canary RS; DR still updated when available, on abort, and when fully promoted.
- `TestShouldSkipDestinationRuleUpdate` — unit tests for the new guard (scaled-up RS, abort, empty/equal hashes, services set, no matching RS).
- Full `./rollout/...` test suite passes locally.

I'm happy to add an e2e test if maintainers think it's worthwhile — the timing-sensitive part (xDS propagation) isn't observable in e2e, but the "DR must not switch before the superseding canary RS is available" ordering is.

### DIFF
--- a/rollout/trafficrouting/istio/istio.go
+++ b/rollout/trafficrouting/istio/istio.go
@@ -376,10 +376,45 @@ func (r *Reconciler) shouldDelayDestinationRuleUpdate(canaryHash, stableHash str
 	return false, ""
 }
 
+// shouldSkipDestinationRuleUpdate returns true if updating the DestinationRule should be
+// skipped (without failing the reconcile) because the target canary ReplicaSet has not been
+// scaled up yet (spec.replicas == 0). Re-pointing the canary subset at a hash with zero pods
+// de-selects the previous, still-healthy canary pods while the VirtualService weight reset and
+// the DestinationRule subset switch propagate to proxies non-atomically, causing transient
+// "no_healthy_upstream" 503s. Unlike shouldDelayDestinationRuleUpdate, this must not surface
+// as an error: ReplicaSets are scaled up after traffic routing is reconciled, so failing the
+// reconcile here would prevent the ReplicaSet from ever being scaled up.
+// See: https://github.com/argoproj/argo-rollouts/issues/4775
+func (r *Reconciler) shouldSkipDestinationRuleUpdate(canaryHash, stableHash string) (bool, string) {
+	if r.rollout.Spec.Strategy.Canary.CanaryService != "" || r.rollout.Spec.Strategy.Canary.StableService != "" {
+		return false, ""
+	}
+	if canaryHash == "" || canaryHash == stableHash {
+		return false, ""
+	}
+	if rolloututil.AbortOrDynamicRollbackToStable(r.rollout, r.replicaSets, stableHash) {
+		// on abort/rollback, traffic shifts to stable, so re-pointing the canary subset is safe
+		return false, ""
+	}
+	for _, rs := range r.replicaSets {
+		if rs.Labels[v1alpha1.DefaultRolloutUniqueLabelKey] != canaryHash {
+			continue
+		}
+		if rs.Spec.Replicas != nil && *rs.Spec.Replicas == 0 {
+			return true, rs.Name
+		}
+	}
+	return false, ""
+}
+
 func (r *Reconciler) UpdateHash(canaryHash, stableHash string, additionalDestinations ...v1alpha1.WeightDestination) error {
 	if shouldDelay, rsName := r.shouldDelayDestinationRuleUpdate(canaryHash, stableHash); shouldDelay {
 		r.log.Infof("delaying destination rule switch: ReplicaSet %s not fully available", rsName)
 		return fmt.Errorf("delaying destination rule switch: ReplicaSet %s not fully available", rsName)
+	}
+	if shouldSkip, rsName := r.shouldSkipDestinationRuleUpdate(canaryHash, stableHash); shouldSkip {
+		r.log.Infof("skipping destination rule switch: canary ReplicaSet %s is not scaled up yet", rsName)
+		return nil
 	}
 
 	dRuleSpec := r.rollout.Spec.Strategy.Canary.TrafficRouting.Istio.DestinationRule
--- a/rollout/trafficrouting/istio/istio_test.go
+++ b/rollout/trafficrouting/istio/istio_test.go
@@ -3544,6 +3544,213 @@ func TestShouldDelayDestinationRuleUpdate(t *testing.T) {
 	})
 }
 
+// TestUpdateHashSkipsNotYetScaledUpCanary verifies that UpdateHash does not re-point the
+// canary subset to a canary ReplicaSet that has not been scaled up yet (spec.replicas == 0),
+// which would de-select the previous, still-healthy canary pods and cause transient
+// "no_healthy_upstream" 503s. Unlike the unavailable-RS case, no error may be returned,
+// because ReplicaSets are scaled up after traffic routing is reconciled; failing the
+// reconcile would prevent the ReplicaSet from ever being scaled up.
+// See https://github.com/argoproj/argo-rollouts/issues/4775
+func TestUpdateHashSkipsNotYetScaledUpCanary(t *testing.T) {
+	createRollout := func(abort bool) *v1alpha1.Rollout {
+		ro := rolloutWithDestinationRule(nil)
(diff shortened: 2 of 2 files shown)


myeongseok-rpls:
Rebased on latest master (post-#4900); build and the full `./rollout/...` test suite pass locally, and CI is green on the rebased commit.

Since #4900 recently landed in a closely related area, a quick note on how the two relate: both deal with the ordering between traffic-routing updates and ReplicaSet scaling, but on disjoint paths. #4900 fixes the **abort** flow (host-level weight step-down vs. canary scale-down deadlock with `dynamicStableScale` + explicit `abortScaleDownDelaySeconds`), while this PR fixes the **supersede** flow for subset-based Istio routing — re-pointing the canary DestinationRule subset to a ReplicaSet that hasn't been scaled up yet de-selects the still-healthy old canary pods, causing transient `no_healthy_upstream` 503s (#4775). This PR deliberately does not apply on abort / dynamic rollback to stable (see the `AbortOrDynamicRollbackToStable` guard), so it doesn't change the behavior #4900 addressed.

@zachaller would you have a chance to take a look when convenient? Happy to adjust anything.

PR #4848, opened 2026-07-15
fix(trafficrouting): abort with dynamicStableScale+minPodsPerReplicaSet never returns weight to 0

### LINKED ISSUE #4841: Aborted canary with dynamicStableScale + minPodsPerReplicaSet never returns traffic weight to 0 (regression in v1.9.0 / #4035)
## Checklist

## Describe the bug

Since **v1.9.0**, aborting a canary that uses traffic routing together with `dynamicStableScale: true` **and** `minPodsPerReplicaSet` leaves a **non-zero canary traffic weight indefinitely**. The canary `ReplicaSet` stays pinned at `minPodsPerReplicaSet` pods and traffic never fully returns to stable.

In **v1.8.3** the same abort correctly drives the canary weight to `0`. Git-bisected to #4035, which introduced `GetDesiredCanaryWeight`.

## To Reproduce

Rollout using any traffic router (reproduced with Istio; the logic is provider-independent), with:

```yaml
spec:
  replicas: 10
  strategy:
    canary:
      dynamicStableScale: true
      minPodsPerReplicaSet: 3
      trafficRouting:
        istio:
          virtualService:
            name: my-vsvc
            routes: [primary]
      steps:
        - setWeight: 5
        - pause: {}
        - setWeight: 50
        - pause: {}
        - setWeight: 100
```

1. Roll out a new revision (v2) and let it progress to the `setWeight: 50` step (paused).
2. `kubectl argo rollouts abort <rollout>`.

## Expected behavior

Canary traffic weight returns to **0** (100% to stable), as it does in v1.8.3.

## Actual behavior

The canary weight never reaches 0 and parks at a non-zero value **permanently**, with the canary `ReplicaSet` held at `minPodsPerReplicaSet` pods. Two sub-cases depending on `spec.replicas` vs `minPodsPerReplicaSet`:

* `spec.replicas > minPodsPerReplicaSet` (e.g. 10 vs 3): weight parks at the **lowest `setWeight` step** (here **5%**).
* `spec.replicas <= minPodsPerReplicaSet` (e.g. 2 vs 3): the replica math degenerates and the weight parks at the **step it was aborted at** (e.g. **50%**).

The only way to clear it is a pod-template rollback (making the canary hash equal the stable hash, e.g. `kubectl argo rollouts undo`).

## Root cause

Introduced by #4035 (commit `cee6ed0d`), first released in v1.9.0.

* `GetDesiredCanaryWeight` (`utils/replicaset/canary.go`) computes the abort weight from the **actual** canary pod count (`newRS.AvailableReplicas`), walking the `setWeight` steps in reverse and returning the highest step whose replica-equivalent is `< canaryReplicas`.
* `CheckMinPodsPerReplicaSet` (`utils/replicaset/canary.go`) floors the canary `ReplicaSet` at `minPodsPerReplicaSet`, so the canary pod count can never drop to the level a low weight implies. `GetDesiredCanaryWeight` therefore keeps returning a non-zero step.
* In `calculateDesiredWeightOnAbortOrStableRollback` (`rollout/trafficrouting.go`), the stable-availability weight (which is `0` once stable is fully scaled up) is computed as `expectedCanaryWeight` but **discarded** on the `desiredCanaryWeight > 0` branch — the function ends with `return minInt(desiredCanaryWeight, currentCanaryWeight)`, so it can never fall to 0 while the floored canary exists.

For contrast, in v1.8.3 `calculateDesiredWeightOnAbortOrStableRollback` derived the weight purely from stable availability and returned `0` once stable was available — independent of the canary pod count, the steps, and `minPodsPerReplicaSet`.

## Suggested fix direction

When stable is fully available (`expectedCanaryWeight == 0`), an abort should be allowed to drive the canary weight to `0` regardless of the floored canary pod count — e.g. cap the final return by `expectedCanaryWeight` on abort, or have `GetDesiredCanaryWeight` use the **desired** (un-floored) canary count on abort so `minPodsPerReplicaSet` does not pin the weight. Any fix should preserve #4035's anti-surge behavior for the `spec.replicas > minPodsPerReplicaSet` case (scaling stable up in reverse-step order).

## Version

* Broken: **v1.9.0** (and later)
* Working: **v1.8.3**
* Regressing PR: #4035

## Related

* #4035 — the PR that introduced `GetDesiredCanaryWeight` (root cause).
* #4424 — a separate weight-math issue in the same area (invalid weight on full promotion); not the same bug but likely related code.

GokulakrishnanP027:
## What

When a canary is aborted with both `dynamicStableScale: true` and `minPodsPerReplicaSet` set, the canary traffic weight never returns to 0. Traffic stays permanently routed to the canary RS.

Fixes #4841. Regression introduced by #4035 (first released in v1.9.0).

## Root cause

`CheckMinPodsPerReplicaSet` floors the canary RS at `minPodsPerReplicaSet` pods (e.g. 3). `GetDesiredCanaryWeight` uses `newRS.AvailableReplicas` (the floored count) to walk steps in reverse, returning the lowest `setWeight` step whose replica-equivalent is below the floored count — a non-zero value — even after stable is fully available.

`calculateDesiredWeightOnAbortOrStableRollback` then falls through to `minInt(desiredCanaryWeight, currentCanaryWeight)`, which can never reach 0 while the floored canary RS exists.

**Example** (`spec.replicas: 10`, `minPodsPerReplicaSet: 3`, steps `5% → 50%`, aborted at 50%):
- Stable fully up: 10/10 pods → `expectedCanaryWeight = 0`
- Canary floored at 3 pods → `GetDesiredCanaryWeight` returns `5` (not 0)
- Result: weight parks at `5%` forever

## Fix

In `calculateDesiredWeightOnAbortOrStableRollback`, when aborting and `expectedCanaryWeight == 0` (stable fully available), return 0 immediately. The floored canary pod count is irrelevant to traffic weight once stable is 100% available.

```go
// When aborting and stable is fully available, canary weight must reach 0 regardless of
// minPodsPerReplicaSet flooring the canary RS.
if c.rollout.Status.Abort && expectedCanaryWeight == 0 {
    return 0
}
```

This preserves #4035's anti-surge behavior: the guard only fires when `expectedCanaryWeight == 0`, meaning stable is fully scaled up. During intermediate abort states (stable still scaling up), the existing step-walk logic applies unchanged.

## Test

`TestDynamicScalingAbortReturnsZeroWeightWhenStableFullyAvailableWithMinPodsPerReplicaSet`

Scenario: `spec.replicas: 10`, `minPodsPerReplicaSet: 3`, `dynamicStableScale: true`, abort at `setWeight: 50`, stable RS 10/10 available, canary RS floored at 3. Verifies `SetWeight(0)` is called.

## Version

- Broken: v1.9.0+
- Working: v1.8.3
- Regressing PR: #4035


Hariharasuthan99:
PR looks fine to me, can you sign off on your commits and push?

Hariharasuthan99 (approved):

zachaller:
maybe another version of this fix: https://github.com/argoproj/argo-rollouts/pull/4900

GokulakrishnanP027:
> maybe another version of this fix: #4900

This PR looks good to me, let me close mine! thanks for the fix. When this expected to me merged/released?

Issue #3078, opened 2023-10-04
Canary service flaps after aborted Rollout for duration of abortScaleDownDelaySeconds


derekddecker:
Checklist:

* [X] I've included steps to reproduce the bug.
* [X] I've included the version of argo rollouts.

**Describe the bug**

If a release is Aborted due to a failed AnalysisRun (or manual abort), and `abortScaleDownDelaySeconds` is set to `0` (to keep the canary replicaset alive indefinitely), then the canary `Service` can end up in a weird state. The rollouts controller starts to flip/flop the pod-template-hash on the canary service. In our case (ingress-nginx traffic routing), you can observe the flapping with curl using the canary header. Sometimes it serves the live release, sometimes it serves the canary release. 

The bug also seems to negatively effect the argocd-application-controller (compute shoots up, log spam) and the Argo Dashboard UI slows to a crawl. 

Worth noting we are using https://github.com/argoproj-labs/rollout-extension to embed the Rollouts UI into the Argo Dashboard due to the lack of auth/rbac in the Rollouts dashboard.

Additionally we use the ApplicationSet controller. Auto-sync is disabled.

**To Reproduce**

Given the following rollout:

```
  strategy:
    canary:
      abortScaleDownDelaySeconds: 0
      canaryService: sre-demo-app-unstable-canary
      maxUnavailable: 0
      stableService: sre-demo-app-unstable
      steps:
      - setCanaryScale:
          replicas: 1
      - pause: {}
      trafficRouting:
        nginx:
          additionalIngressAnnotations:
            canary-by-header: X-Canary
            canary-by-header-value: "yes"
          stableIngress: sre-demo-app-unstable
```

Abort the rollout during the pause step.

**Expected behavior**

Up for debate, but certainly not what the current behavior is. Personally, I would like to see the canary service continue to route traffic to the canary deployment, so that further investigation can be made into the aborted Rollout.

**Screenshots**

**Version**
`quay.io/argoproj/argo-rollouts:v1.5.1`

**Logs**

`argocd-application-controller` example logs when in this state: 
```
time="2023-10-04T22:25:46Z" level=info msg="Refreshing app status (controller refresh requested), level (1)" application=argocd/sre-demo-app-unstable
time="2023-10-04T22:25:46Z" level=info msg="Comparing app state (cluster: https://kubernetes.default.svc, namespace: sre)" application=argocd/sre-demo-app-unstable
[...]

Issue #1838, opened 2022-02-02
After abort, rollout does not update Istio DestinationRule


mksha:
## Summary 

What happened/what you expected to happen?

When we abort the rollout then it should update the Istio destination rule so in destinationrule we have subsets with correct         rollouts-pod-template-hash label.

But currently when we abort the rollout it does not update the destination rule, causing cannery endpoint to fail.

## Diagnostics

What version of Argo Rollouts are you running?
0.3.0

```
# Paste the logs from the rollout controller
canary subset is not updated with correct rollouts-pod-template-hash label value.

# Logs for the entire controller:
kubectl logs -n argo-rollouts deployment/argo-rollouts

# Logs for a specific rollout:
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=<ROLLOUTNAME>
logs does not have any reference for destinationrule update after abort action
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

huikang:
Hi,@mksha, could you confirm if this happens to v0.3.0? Have you tried with v1.1.1 or the latest argo-rollouts image?

mksha:
yes, we are using 1.1.1

huikang:
@mksha , could you paste here the output from `get rollout <rollout name>` and `get desinationrule <name> -o yaml`? Thanks.

mksha:
```
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  labels:
    region: us-east-1
    system_risk_class: "1"
  managedFields:
  - apiVersion: argoproj.io/v1alpha1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:rollout.argoproj.io/revision: {}
      f:spec:
        f:replicas: {}
      f:status:
        .: {}
        f:HPAReplicas: {}
        f:availableReplicas: {}
        f:blueGreen: {}
        f:canary:
          .: {}
          f:weights:
            .: {}
            f:canary:
              .: {}
              f:podTemplateHash: {}
              f:weight: {}
            f:stable:
              .: {}
              f:podTemplateHash: {}
              f:weight: {}
        f:conditions: {}
        f:currentPodHash: {}
        f:currentStepHash: {}
        f:currentStepIndex: {}
        f:observedGeneration: {}
        f:phase: {}
        f:readyReplicas: {}
        f:replicas: {}
[...]

PR #3413, opened 2024-02-29
fix(controller): stuck in scale event after canary service switch delay. Fixes #3412

BrunoTarijon:
Start a service reconcile if a rs is "still referenced" to fix [this issue](https://github.com/argoproj/argo-rollouts/issues/3412).

WIP - e2e test

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [ ] My builds are green. Try syncing with master if they are not. 
* [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

PR #4853, opened 2026-07-15
fix(controller): reconcile traffic routing during an aborted scaling event

n1koo:
`reconcile()` in `rollout/context.go` short-circuits to `syncReplicasOnly()` whenever `isScalingEvent()` is true. `syncReplicasOnly()` scales the ReplicaSets but never calls `reconcileTrafficRouting`, so the canary's traffic weight stays at whatever value the previous non-scaling reconcile wrote.

Failure mode: a canary is aborting, and something (typically an HPA reacting to load, but any `spec.replicas` change works) fires `isScalingEvent()` while the abort is in progress. `reconcile()` short-circuits, the traffic weight stays frozen at its pre-abort value, and the abort cannot move traffic off the canary until the scaling event resolves and a subsequent non-scaling reconcile fires. For a wedged rollback that can be arbitrarily long.

This PR narrows the short-circuit to non-aborted rollouts. When `pauseContext.IsAborted()`, `reconcile()` falls through to the strategy-specific path. `rolloutCanary` calls `reconcileTrafficRouting` (which fixes the frozen-weight bug) and `reconcileCanaryReplicaSets` — the exact scaling entry point `syncReplicasOnly` was already invoking for canary — so no scaling behaviour is lost.

`reconcileTrafficRouting` is canary-only; blue-green never called it, so no traffic behaviour changes for blue-green. The blue-green abort+scaling case now runs the full `rolloutBlueGreen` path instead of the shortcut, but both include the same replicaset sync via `reconcileBlueGreenReplicaSets`.

The full canary path also runs `reconcileEphemeralMetadata`, `reconcileStableAndCanaryService`, `reconcileExperiments`, `reconcileAnalysisRuns`, and `stepPluginContext.reconcile` during abort+scaling reconciles that previously skipped them. Each has an explicit abort branch upstream (e.g. `abortExperiments`, canary-service selector reset on abort), so widening the surface here is a strict improvement in the abort case.

Alternatives considered:
- Move `reconcileTrafficRouting` into `syncReplicasOnly` — bigger surface change; `syncReplicasOnly` is intentionally minimal and shared by both strategies.
- Re-enqueue on scaling event during abort — indirect and racy; the fix is a one-condition predicate at the site of the incorrect short-circuit.

## Testing

- Unit: `TestReconcileTrafficRoutingWhenAbortedDuringScalingEvent` in `rollout/trafficrouting_test.go` — constructs an aborted canary with a concurrent scaling event and asserts the traffic router receives `SetWeight(0)` in the same reconcile cycle rather than the frozen pre-abort weight.
- `go build`, `go vet`, `golangci-lint run`, full unit suite: clean.

No filed upstream issue matches this exact failure mode; the closest reports are around abort behaviour with traffic routing (#4626, #1838, #3078) but none pin the scaling-event short-circuit.

Checklist:

* [x] (b) this is a bug fix
* [x] The title of the PR is [conventional](https://www.conventionalcommits.org/en/v1.0.0/) and states what changed. No related upstream issue is filed for this specific behaviour, so no `Fixes #` suffix.
* [x] My builds are green.
* [x] I have written unit tests for my change (`TestReconcileTrafficRoutingWhenAbortedDuringScalingEvent`).

### DIFF
--- a/rollout/context.go
+++ b/rollout/context.go
@@ -69,7 +69,9 @@ func (c *rolloutContext) reconcile() error {
 		return err
 	}
 
-	if isScalingEvent {
+	// Skip the scaling-only short-circuit while aborting: it bypasses traffic reconciliation,
+	// which would freeze the canary weight if spec.replicas changes mid-abort (e.g. HPA).
+	if isScalingEvent && !c.pauseContext.IsAborted() {
 		return c.syncReplicasOnly()
 	}
 
--- a/rollout/trafficrouting_test.go
+++ b/rollout/trafficrouting_test.go
@@ -984,6 +984,70 @@ func TestDynamicScalingDontIncreaseWeightWhenAborted(t *testing.T) {
 	f.run(getKey(r1, t))
 }
 
+// TestReconcileTrafficRoutingWhenAbortedDuringScalingEvent verifies that an abort still reconciles
+// traffic routing (shifts weight off the canary) even when a concurrent scaling event (e.g. an HPA
+// changing spec.replicas) would otherwise short-circuit reconcile() to syncReplicasOnly.
+func TestReconcileTrafficRoutingWhenAbortedDuringScalingEvent(t *testing.T) {
+	f := newFixture(t)
+	defer f.Close()
+
+	steps := []v1alpha1.CanaryStep{
+		{SetWeight: ptr.To[int32](50)},
+		{Pause: &v1alpha1.RolloutPause{}},
+	}
+	r1 := newCanaryRollout("foo", 5, nil, steps, ptr.To[int32](1), intstr.FromInt(1), intstr.FromInt(1))
+	r1.Spec.Strategy.Canary.TrafficRouting = &v1alpha1.RolloutTrafficRouting{SMI: &v1alpha1.SMITrafficRouting{}}
+	r1.Spec.Strategy.Canary.CanaryService = "canary"
+	r1.Spec.Strategy.Canary.StableService = "stable"
+	r1.Status.ReadyReplicas = 5
+	r1.Status.AvailableReplicas = 5
+	r1.Status.Abort = true
+	r1.Status.AbortedAt = &metav1.Time{Time: time.Now().Add(-1 * time.Minute)}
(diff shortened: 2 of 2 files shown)


kostis-codefresh (commented):
Prior art in this area

- https://github.com/argoproj/argo-rollouts/issues/3412
- https://github.com/argoproj/argo-rollouts/pull/3413#issuecomment-1984805862

n1koo:
Thanks — read both.

Same architectural cause: `isScalingEvent` short-circuits `reconcile()` to `syncReplicasOnly`, so anything requiring traffic reconciliation to make progress freezes when it coincides with a scale event. #3412's trigger is a pending service switch; ours is an abort during a scale (typically HPA on failing pods). Different symptoms, same lockout.

Worth flagging: jessesuen's "resync/restart will eventually unstick it" hypothesis on #3413 was tested empirically (#3413 comment 1985761163) — 20 min + controller restart, still stuck. So a carve-out of some kind is needed; the disagreement on #3413 was about which carve-out.

Ours is scoped narrowly to `pauseContext.IsAborted()` — a state where step-weight semantics are already overridden (weight → 0), so bypassing the short-circuit doesn't fight anything the design was protecting. The short-circuit still holds for the normal case.

Not fixing #3412's trigger here — different code path, and the right fix is a design conversation of its own. Happy to look at it separately once this lands.

zachaller:
I added a possible alternative one of the issues with removing the return early on scaling events is we like to try and keep the scaling event not dependent on any other failure to never allow a rollouts bug or issue with say updating traffic router prevent the scaling.

#4919

n1koo:
Closing in favor of #4919 — cleaner approach.

Rather than carving abort out of the `isScalingEvent` short-circuit, #4919 restores the invariant that `syncReplicasOnly` terminates scaling events, by syncing the stale `desired-replicas` annotation at the source (the abort scale-down-delay early returns in `reconcileNewReplicaSet`). Also fixes blue-green aborts via the shared code path — a gap this PR missed.

Thanks for the quick turnaround on #4919 — looking forward to testing it.

PR #4855, opened 2026-07-15
fix(experiments): add API fallback when ReplicaSet lister is stale on AlreadyExists (#4842)

### LINKED ISSUE #4842: bug(experiments): experiment ReplicaSet AlreadyExists handler has no API fallback when lister cache is stale (same gap as #4804, fixed for AnalysisRun by #4805)
## Summary

Rollouts with Experiment steps intermittently abort with:

```
RolloutAborted: Rollout aborted update to revision N: Experiment analysis phase is error/failed:
Failed to create ReplicaSet for template 'canary':
replicaset.apps "<rollout>-<hash>-<revision>-0-canary" not found
```

The failure is intermittent — it only occurs when a fast resync races with the informer cache update.

## Root cause

In `experiments/replicaset.go`, the `AlreadyExists` handler falls back to the lister with no API fallback:

```go
case errors.IsAlreadyExists(err):
    rs, rsErr := ec.replicaSetLister.ReplicaSets(newRS.Namespace).Get(newRS.Name)
    if rsErr != nil {
        return nil, rsErr  // ← "not found" returned directly if cache is stale
    }
```

The race:

1. **Reconcile N**: `replicaSetLister.List()` → stale cache, RS not visible → `Create(RS)` → **success**
2. **Reconcile N+1** (fast resync before informer updates): `replicaSetLister.List()` → still empty → `Create(RS)` → **AlreadyExists**
3. Falls back to `replicaSetLister.Get(name)` → **"not found"** (cache still stale) → error propagates → Rollout aborted

## Relation to existing fixes

PR #4805 fixed the identical race for `AnalysisRun` in the Experiment controller (lister miss → direct API Get fallback). The same fallback pattern is missing for `ReplicaSet` in the `AlreadyExists` handler above.

The same race was also fixed for the main rollout ReplicaSet lister in v1.9.0 — but experiment ReplicaSets use a separate code path that was not covered by that fix.

The race is latent across versions but becomes more likely to surface with the filtered `replicaSetInformerFactory` introduced in #4666 (v1.10.0-rc1), which may have slightly different sync timing characteristics.

## Suggested fix

Apply the same fallback pattern from #4805 to `experiments/replicaset.go`:

```go
case errors.IsAlreadyExists(err):
    rs, rsErr := ec.replicaSetLister.ReplicaSets(newRS.Namespace).Get(newRS.Name)
    if rsErr != nil {
        if !errors.IsNotFound(rsErr) {
            return nil, rsErr
        }
        // Lister cache is stale — fall back to direct API Get (same pattern as #4805 for AnalysisRun)
        rs, rsErr = ec.kubeclientset.AppsV1().ReplicaSets(newRS.Namespace).Get(ctx, newRS.Name, metav1.GetOptions{})
        if rsErr != nil {
            return nil, rsErr
        }
    }
```

## Steps to reproduce

1. Use argo-rollouts v1.9.0 or v1.10.0-rc1
2. Configure a Rollout with an Experiment step containing a canary template
3. Trigger multiple canary rollouts — the failure is intermittent
4. Observe `RolloutAborted` with `Failed to create ReplicaSet for template '...': ... not found`

## Related

- #4804 — same race for AnalysisRun (fixed by #4805)
- #4666 — filtered ReplicaSet informer (makes this race more likely to surface)

manu844:
## What does this PR do?

When the Experiment controller reconciles and tries to create a ReplicaSet that was already created in a prior sync, the \`AlreadyExists\` handler falls back to the lister to fetch the existing RS. If the informer cache is stale (fast resync before the cache updates), the lister misses and the error propagates up — setting \`templateStatus\` to \`Error\` and aborting the Rollout.

This PR adds a direct API \`Get\` fallback when the lister misses, so a stale cache no longer causes a spurious error. The RS is fetched from the API server and the reconciliation continues normally.

Fixes #4842

## Testing

Added two tests in \`experiments/replicaset_test.go\`:

- \`TestReplicaSetAlreadyExistsMissingFromCacheFallsBackToAPI\`: pre-loads the RS into the fake kubeclientset but **not** into the informer cache (simulating a stale lister), then calls \`createReplicaSet\` and asserts it returns the RS successfully instead of an error.
- \`TestReplicaSetAlreadyExistsAPIGetFailure\`: same stale-cache setup, but injects a failure on the API \`Get\` fallback and asserts the error is propagated to the caller.

## Checklist

* [x] (b) this is a bug fix
* [x] The title of the PR is conventional, states what changed, and suffixes the related issue number

### DIFF
--- a/experiments/replicaset.go
+++ b/experiments/replicaset.go
@@ -93,7 +93,15 @@ func (ec *experimentContext) createReplicaSet(template v1alpha1.TemplateSpec, co
 		// Fetch a copy of the ReplicaSet.
 		rs, rsErr := ec.replicaSetLister.ReplicaSets(newRS.Namespace).Get(newRS.Name)
 		if rsErr != nil {
-			return nil, rsErr
+			if !errors.IsNotFound(rsErr) {
+				return nil, rsErr
+			}
+			// Lister cache may be stale — fall back to a direct API Get
+			rs, rsErr = ec.kubeclientset.AppsV1().ReplicaSets(newRS.Namespace).Get(ctx, newRS.Name, metav1.GetOptions{})
+			if rsErr != nil {
+				return nil, rsErr
+			}
+			ec.log.Warnf("ReplicaSet '%s' missing from informer cache, fell back to API get", rs.Name)
 		}
 
 		// If the Experiment owns the ReplicaSet and the ReplicaSet's PodTemplateSpec is semantically
--- a/experiments/replicaset_test.go
+++ b/experiments/replicaset_test.go
@@ -1,17 +1,47 @@
 package experiments
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+	appslisters "k8s.io/client-go/listers/apps/v1"
+	k8stesting "k8s.io/client-go/testing"
+	corev1defaults "k8s.io/kubernetes/pkg/apis/core/v1"
 	"k8s.io/utils/ptr"
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	"github.com/argoproj/argo-rollouts/utils/conditions"
 )
+
+// errReplicaSetLister is a fake ReplicaSetLister that always returns a fixed error from Get.
+type errReplicaSetLister struct{ err error }
+
+func (l *errReplicaSetLister) List(_ labels.Selector) ([]*appsv1.ReplicaSet, error) {
+	return nil, l.err
+}
+func (l *errReplicaSetLister) ReplicaSets(_ string) appslisters.ReplicaSetNamespaceLister {
+	return &errReplicaSetNamespaceLister{err: l.err}
+}
+func (l *errReplicaSetLister) GetPodReplicaSets(_ *corev1.Pod) ([]*appsv1.ReplicaSet, error) {
+	return nil, l.err
+}
+
+type errReplicaSetNamespaceLister struct{ err error }
+
+func (l *errReplicaSetNamespaceLister) List(_ labels.Selector) ([]*appsv1.ReplicaSet, error) {
+	return nil, l.err
+}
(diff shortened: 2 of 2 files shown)


Hariharasuthan99 (approved):
LGTM

experiments/replicaset.go
```diff
@@ -93,7 +93,12 @@ func (ec *experimentContext) createReplicaSet(template v1alpha1.TemplateSpec, co
 		// Fetch a copy of the ReplicaSet.
 		rs, rsErr := ec.replicaSetLister.ReplicaSets(newRS.Namespace).Get(newRS.Name)
 		if rsErr != nil {
```
zachaller, line 95: Only fall back on NotFound — other lister errors (indexer failure, unexpected type, etc.) should propagate as-is. Falling back on every error also makes the "missing from informer cache" warn misleading. Same guard as suggested in #4842 / used in #4805 for AnalysisRun.


```suggestion
		if rsErr != nil {
			if !errors.IsNotFound(rsErr) {
				return nil, rsErr
			}
			// Lister cache may be stale — fall back to a direct API Get
			rs, rsErr = ec.kubeclientset.AppsV1().ReplicaSets(newRS.Namespace).Get(ctx, newRS.Name, metav1.GetOptions{})
			if rsErr != nil {
				return nil, rsErr
			}
			ec.log.Warnf("ReplicaSet '%s' missing from informer cache, fell back to API get", rs.Name)
		}
```

Hariharasuthan99 (changes requested):
can you do the change you have pushed as part of https://github.com/argoproj/argo-rollouts/compare/0eed2cb24d5e35d8cdce4fcb30964e856de7afa2..c5591b39ef44da19f2b7f2f4791392916aa76e5a

manu844:
> can you do the change you have pushed as part of https://github.com/argoproj/argo-rollouts/compare/0eed2cb24d5e35d8cdce4fcb30964e856de7afa2..c5591b39ef44da19f2b7f2f4791392916aa76e5a

Yes, added that back. Thanks!

Hariharasuthan99:
> > can you do the change you have pushed as part of https://github.com/argoproj/argo-rollouts/compare/0eed2cb24d5e35d8cdce4fcb30964e856de7afa2..c5591b39ef44da19f2b7f2f4791392916aa76e5a
> 
> Yes, added that back. Thanks!

Any reason why you thought it had to do the fallback for all errors?

manu844:
> > > can you do the change you have pushed as part of https://github.com/argoproj/argo-rollouts/compare/0eed2cb24d5e35d8cdce4fcb30964e856de7afa2..c5591b39ef44da19f2b7f2f4791392916aa76e5a
> > 
> > 
> > Yes, added that back. Thanks!
> 
> Any reason why you thought it had to do the fallback for all errors?

I overlooked other failures (e.g indexer failure, unexpected type) assuming the error can only be not found.

zachaller (approved):

Issue #3258, opened 2023-12-18
Analysis Run based on readiness


zachaller:
# Summary

Today we have background analysis Jobs that start based on time delay. We should instead have an option that looks at pod readiness to decide when we should start an analysis run shuch that it requires one pod to be ready.

# Use Cases

When would you use this?

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

PR #4774, opened 2026-06-10
feat: rollout total duration tracking

agaudreault:
Fixes https://github.com/argoproj/argo-rollouts/issues/4732

PR #4864, opened 2026-07-16
fix(controller): delay background analysis until the canary receives traffic

n1koo:
For a traffic-routed canary, the background AnalysisRun is created as soon as the new ReplicaSet exists — `reconcileBackgroundAnalysisRun` checks only that the rollout is not fully promoted, has a stable RS, and is not before `startingStep`. Nothing checks that the controller has actually programmed the traffic router for that ReplicaSet.

A canary-scoped metric is then evaluated against a ReplicaSet receiving no traffic. Either outcome is wrong: it passes on no evidence, or — the common case, since most providers return no data for a pod with no requests — it errors or fails and aborts a healthy rollout.

The window is real and not small. The canary has to scale up before `reconcileTrafficRouting` will give it any weight (`newRS.Status.AvailableReplicas == 0` forces `desiredWeight` to 0), and on ALB the target group has to converge on top of that.

Related: #3258, which asks for analysis to start based on readiness rather than a step index or timer.

## Opt-in, via `analysis.startOn`

```yaml
spec:
  strategy:
    canary:
      analysis:
        templates:
        - templateName: success-rate
        startOn: CanaryTraffic
```

| value | behaviour |
|---|---|
| `Immediately` (default, and what unset means) | create the run as soon as the rollout begins — identical to today |
| `CanaryTraffic` | wait until the traffic router has given the canary the weight the current step calls for |

**This is deliberately opt-in.** Background analysis is not always canary-scoped: an approval check or an incident query run through the `job` or `web` providers is meant to run *before* traffic shifts, and gating it would defeat its purpose rather than merely delay it. Gating unconditionally would break those users with no way back, so `Immediately` stays the default and existing rollouts are untouched.

An enum rather than a bool leaves room for the readiness-based variant #3258 asks for, without a second field that would need its interaction with the first defined.

## The gate

When opted in, `deferBackgroundAnalysisRun()` holds the run back until `canaryReceivingStepTraffic()` is satisfied, which requires:

1. `Status.Canary.Weights != nil` — the router has been programmed for some canary.
2. `Weights.Canary.PodTemplateHash` matches the current canary ReplicaSet — not a previous revision that still holds the traffic split.
3. `Weights.Canary.Weight` is non-zero and has reached `GetCurrentSetWeight(rollout)`. The non-zero part matters: `GetCurrentSetWeight` returns 0 both for an explicit `setWeight: 0` and for any step before the first `setWeight`, so a rollout whose steps open with a `pause` would otherwise satisfy the gate while serving nothing.
4. If the router populates `Weights.Verified`, it must be `true`.

Weights are read from `newStatus`, which `reconcileTrafficRouting` populates earlier in the same reconcile (`canary.go:67` vs `:76`), so the gate does not lag a reconcile behind the router.

A canary with no traffic router is unaffected: it shifts traffic by adjusting the stable and canary replica counts behind a shared Service and leaves `Status.Canary.Weights` nil, so there is no recorded weight to wait for.

## Why not `startingStep`

`BeforeStartingStep` compares step indices only. Reaching step 1 does not mean step 0's weight landed: with a slow-scaling canary, or an ALB target group that has not converged, it demonstrably has not. The two compose — `startingStep` can still delay analysis further.

## Step-based analysis is deliberately not gated

An earlier revision of this PR gated `reconcileStepBasedAnalysisRun` too. I dropped it:

- **Redundant.** `completedCurrentCanaryStep` (`canary.go:320-328`) already refuses to advance past a `setWeight` step until `AtDesiredReplicaCountsForCanary` *and* `Weights.Verified != false`, so an analysis step following a `setWeight` is only reached after the weight landed and verified.
- **It could not deliver the verification it implied.** At an Analysis step `currentStep.SetWeight == nil`, so `ShouldVerifyWeight` is false, ALB's `VerifyWeight` returns `nil, nil`, and `Verified` is overwritten to nil.
- **It risked an unbounded stall.** `isIndefiniteStep` (`sync.go:799`) covers Analysis steps, so the `progressDeadlineExceeded` branch at `sync.go:944` is skipped for them; a step AnalysisRun that was never created would hang the rollout with no timeout.

The background gate has none of these: when the canary never becomes ready the rollout stays parked on the `setWeight` step, which is not an indefinite step, so `progressDeadlineExceeded` still fires and the rollout goes Degraded with an accurate reason.

## Testing

- Unit: `TestDelayBackgroundAnalysisUntilCanaryWeightReached` — no recorded weights, weight 0, weights belonging to a previous ReplicaSet, unverified router, verified router, non-verifying router, basic canary, and both `Immediately` and unset. The default cases fail if the gate is ever flipped on by default or the `startOn` check removed.
- Unit: `TestValidateRolloutStrategyCanaryBackgroundAnalysisStartOn` — unset and both valid values accepted, an unknown value rejected with the right field path.
- E2E: `TestIstioSuite/TestIstioCanaryBackgroundAnalysisWaitsForCanaryWeight` — against a real Istio VirtualService, holds the canary pod unready so the canary stays at weight 0 and asserts no AnalysisRun is created; marking it ready takes the weight to 50 and the run appears. The negative assertion waits for the weights to be recorded for the new ReplicaSet rather than sleeping, so it is anchored on the controller having reconciled and declined rather than on elapsed time.
- Validation: unknown `startOn` values are rejected by `ValidateRolloutStrategyCanary`, and the CRD carries `+kubebuilder:validation:Enum=Immediately;CanaryTraffic` so a typo is refused at admission rather than degrading the Rollout.
- Docs: the opt-in field documented in the existing "Delay Analysis Runs" section of `docs/features/analysis.md`, alongside `initialDelay` and `startingStep`.
- Codegen: regenerated, including `ui/src/models/rollout/generated/api.ts`, which is unmodified generator output; `Verify Codegen` passes. (`make ui-proto` needs #5038 to run at all on a fresh clone.)
- `go build`, `go vet`, `golangci-lint run`, `gofmt`, and the full `./rollout/...`, `./pkg/...` and `./utils/...` suites: clean.

Checklist:

* [x] (b) this is a bug fix
* [x] The title of the PR is [conventional](https://www.conventionalcommits.org/en/v1.0.0/) and states what changed.
* [x] My builds are green.
* [x] I have written unit tests for my change and an Istio e2e test.


docs/features/analysis.md
```diff
@@ -1227,6 +1227,12 @@ spec:
       - setWeight: 40
       - pause: {duration: 10m}
 ```
+
+For a canary with traffic routing, background analysis additionally waits until the canary is
+receiving the weight the current step calls for, so it is not evaluated against a ReplicaSet that
+has no traffic yet. This needs no configuration; `startingStep` remains available to delay it
```
n1koo: The `startingStep` is a field from the above custom resource. It is not a variable name from code? 

Not sure if I followed you though, mind explaining bit more?

Hariharasuthan99:
@kostis-codefresh @zachaller The reason behind this change looks valid and I can't think of a scenario where this change can impact negatively. Although if some users have come to expect the analysis to start from the beginning of the rollout irrespective of the traffic being routed to the canary pods or not, then it would be a noticeable change in behaviour. But still, this change seems valid.

docs/features/analysis.md
Hariharasuthan99: Ah missed it

kostis-codefresh (commented):
Related https://github.com/argoproj/argo-rollouts/issues/3258

test/e2e/istio_test.go
```diff
@@ -570,3 +570,43 @@ func (s *IstioSuite) TestIstioSubsetSplitInStableDownscaleAfterCanaryAbort() {
...
+// 0; without the gate the controller would create the AnalysisRun immediately against that
+// zero-traffic ReplicaSet. Once the canary pod is marked ready the weight reaches 50 and the
+// AnalysisRun is created.
+func (s *IstioSuite) TestIstioCanaryBackgroundAnalysisWaitsForCanaryWeight() {
+	s.Given().
+		RolloutObjects("@istio/istio-canary-background-analysis.yaml").
+		When().
+		ApplyManifests("@functional/analysistemplate-sleep-job.yaml"). // reuse the shared template
+		ApplyManifests().
+		MarkPodsReady("1", 2). // revision 1 fully available
```
kostis-codefresh, line 586: Unless I am missing something, markspods ready only applies a special annotation https://github.com/argoproj/argo-rollouts/blob/c30e93aed67272fb0c99cd232bfdf7a08abd84d3/test/fixtures/common.go#L184

But is the annotation present in the yaml file like the other tests https://github.com/argoproj/argo-rollouts/blob/c30e93aed67272fb0c99cd232bfdf7a08abd84d3/test/e2e/functional/canary-dynamic-stable-scale.yaml#L73

rollout/analysis.go
```diff
@@ -347,6 +347,32 @@ func (c *rolloutContext) reconcilePostPromotionAnalysisRun() (*v1alpha1.Analysis
...
 
+// canaryWeightReached reports whether the traffic router has given the current canary ReplicaSet
+// the weight the current step calls for, and -- for routers that verify weights -- that the shift
+// has been verified. A basic canary shifts traffic by moving service selectors, so there is
+// nothing to wait for and this is always true.
+//
+// Weights are read from newStatus because reconcileTrafficRouting runs earlier in the same
+// reconcile and records there both the weight it just programmed and the result of the
+// VerifyWeight call it just made. The persisted status is a reconcile behind.
+func (c *rolloutContext) canaryWeightReached() bool {
```
kostis-codefresh: Not a blocker for the PR, but I find this function a bit low level. Blue/green code has already skipXXXanalysis methods that make it super clear what they are about.

Can we do a refactoring here and flip/extract code to have a similar skipCanaryAnalysis method or similar? I think that would help a lot for future code reviews.

Haven't checked how big refactoring is needed though, so I understand if this is outside the scope of the PR. I just believe that with each PR we should also improve the surrouding code as well.

rollout/analysis.go
```diff
@@ -364,6 +390,14 @@ func (c *rolloutContext) reconcileBackgroundAnalysisRun() (*v1alpha1.AnalysisRun
 	}
 
 	if needsNewAnalysisRun(currentAr, c.rollout) {
+		// Wait until the canary is actually in the traffic path. Background analysis is otherwise
+		// created as soon as the new ReplicaSet exists, so a canary-scoped metric would be
+		// evaluated against a ReplicaSet receiving no traffic -- passing on no evidence, or
+		// failing on no data and aborting a healthy rollout.
+		if !c.canaryWeightReached() {
+			c.log.Info("Delaying background analysis until the canary has received traffic")
```
kostis-codefresh: Can we add some more debuggin info here? If I see this line in my logs do I know why/which analysis hasn't started?

rollout/analysis_test.go
```diff
@@ -862,6 +864,98 @@ func TestCreateAnalysisRunOnAnalysisStep(t *testing.T) {
...
+	// analysis once. Returns whether an AnalysisRun was created.
+	reconcile := func(trafficRouted bool, weights func(canaryHash, stableHash string) *v1alpha1.TrafficWeights) bool {
+		analysis := &v1alpha1.RolloutAnalysis{Templates: []v1alpha1.AnalysisTemplateRef{{TemplateName: at.Name}}}
+		steps := []v1alpha1.CanaryStep{{SetWeight: ptr.To[int32](50)}, {Pause: &v1alpha1.RolloutPause{}}}
+		r1 := newCanaryRollout("foo", 5, nil, steps, ptr.To[int32](0), intstr.FromInt(1), intstr.FromInt(1))
+		r1.Spec.Strategy.Canary.Analysis = &v1alpha1.RolloutAnalysisBackground{RolloutAnalysis: *analysis}
+		if trafficRouted {
+			r1.Spec.Strategy.Canary.CanaryService = "canary"
+			r1.Spec.Strategy.Canary.StableService = "stable"
+			r1.Spec.Strategy.Canary.TrafficRouting = &v1alpha1.RolloutTrafficRouting{SMI: &v1alpha1.SMITrafficRouting{}}
```
kostis-codefresh: smi is deprecated and I wouldn't expect to see it on new code. Is it possible to use something else here?

kostis-codefresh (changes requested):
I added some comments in specific code segments. The most important one is the readiness gate in the e2e test.

However I have two major concerns with this PR and both have to do with backwards compatibilily.

# Point A.
 
The whole assumption of this PR is that analysis does needs canary pods to be ready. This is true for most people but not everybody. We have seen companies use (and abuse) analysis for several scenarios where an analysis doesn't really care about pods at all. Especially with the web and job providers users are free (by design) to do anything they want.

A common scenario is where a job/web analysis just contacts another system and asks for something. This can be a CM system asking for approval or another monitoring system where the rollout wants to know if there is an active incident or not. 

And in this case, users really DO want analysis to start running ASAP without any care about what the canary is doing. In fact they do want to know about an incident and block the deployment BEFORE the major traffic hits the canary.

This PR will break their approval/incident check/whatever workflow. And they have no way to bring back the old behavior.

I believe https://github.com/argoproj/argo-rollouts/issues/3258 asks for an option. @zachaller can clarify what he meant.

# Point B

I am not sure about the sentence "a basic canary shifts traffic by moving service selectors, so there is nothing to wait for." Basic canary works by changing the number of replicas and letting basic k8s pod routing handle the percentage. 

So have you tested a scenario with analysis and basic canary (without traffic routing) to see how this PR behaves?  So I want to be super safe and make sure we are not breaking anything for users who use canaries without traffic providers that also have analysis

n1koo:
Both concerns were fair, so I've reworked the PR.

On point A: you're right that gating this unconditionally breaks analysis that isn't about the canary. An approval or incident check via the job/web providers is the obvious case, and holding it until traffic has shifted doesn't just postpone it, it defeats the point of running it. So the gate is now optional, via `analysis.startOn`. Unset means `Immediately`, which is the behaviour today, so existing rollouts are unaffected. `CanaryTraffic` turns it on. I made it an enum rather than a bool so the readiness variant in #3258 has somewhere to go later without needing a second field. Unknown values are rejected by validation and by a CRD enum.

On point B: fair, the comment was misleading and I've reworded it. Basic canary splits traffic by replica ratio rather than by weight, and the reason it's exempt is the explicit `TrafficRouting == nil` check, with the fact that it never records `status.canary.weights` as the backstop. On testing it, every existing test that configures background analysis is a basic canary, and they're all untouched and still passing. The new test covers that path explicitly too.

Good catch on the readiness gate. The manifest never declared the gate, so `MarkPodsReady` wrote the condition but nothing acted on it, the pod became ready by itself, and `ExpectAnalysisRunCount(0)` was passing on timing. Fixed. The assertion now also checks the router was programmed for that canary at weight 0, since otherwise the run could be blocked by the "no weights recorded" branch instead of the weight check the test is aiming at.

Smaller things: pulled the gate out into `deferBackgroundAnalysisRun`, in the spirit of the `skipXXXAnalysisRun` helpers though it's a method rather than a package level func; the delay log now says which condition is unmet; SMI is swapped for Istio in the unit test. Also rebased onto master, since #4774 touched the same generated files. The protobuf, OpenAPI and CRD output is regenerated; `ui/src/models/rollout/generated/api.ts` I wrote by hand to match generator output, because `ui-proto` needs a swagger-codegen jar via Docker that I couldn't run locally.

n1koo:
Two more things since the last update, both from re-reviewing this with fresh eyes.

The gate had a hole. `GetCurrentSetWeight` returns 0 for an explicit `setWeight: 0` and for any step before the first `setWeight`, so comparing against it alone was trivially satisfied at zero traffic. A rollout whose steps begin with a `pause` opted into `CanaryTraffic` and got nothing: the run started against a canary serving no traffic, which is the case the option exists to prevent. It now requires a non-zero weight as well, with tests for the leading-pause and `setWeight: 0` shapes. `canaryWeightStatus` also folded into the predicate, since it was a second copy of the same conditions that would eventually have disagreed with it.

The e2e no longer sleeps before asserting the run is absent. It waits for the weights to be recorded for the new ReplicaSet, so the assertion is anchored on the controller having reconciled and declined rather than on elapsed time.

Happy to squash whenever you'd like. I've kept the commits separate for now since they're easier to review that way, but say the word and I'll collapse them.

kostis-codefresh (commented):
Thanks for following up on all the requested changes!

Can you elaborate a bit on the "I wrote by hand to match generator output, because ui-proto needs a swagger-codegen jar via Docker that I couldn't run locally." part?

Is this the agent complaining about not running docker in a a sandbox? Because if you as a human could not run the generator (even by hand) then this is something that we should fix anyway for all contributors (agents or humans) and maybe it is worth its own issue.

n1koo:
> Can you elaborate a bit on the "I wrote by hand to match generator output, because ui-proto needs a swagger-codegen jar via Docker that I couldn't run locally." part?

Sorry, this was my sloppyness. Theres a genuine issue with the generator with fresh clone.  But I lost track of it and let this AI driven workaround be still here rather than address it.

Opened up https://github.com/argoproj/argo-rollouts/pull/5038 which is the proper fix for it and also confirmed (eg by running 'fixed' generator ) that output is solid

kostis-codefresh (approved):
Thank you. Since we are tracking the generator problem in a separate PR I am ok with the code of this one.

@zachaller PTAL since you created originally https://github.com/argoproj/argo-rollouts/issues/3258

Issue #4195, opened 2025-03-19
503 NC (NoClusterFound) errors on canary release with Istio


quentinplessis:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

When using canary release with Istio host-level traffic splitting, 503 NC (NoClusterFound) errors may occur at the very end of the rollout when the canary service becomes stable, as well as on rollout abortions. With high traffic throughput, errors almost always occur.

Let "old pods" be the pods before the rollout (stable pods) and "new pods" be the pods you want to rollout to (canary pods), then from my observations this happens because at the end of the rollout, old pods are not referenced by any kubernetes Service anymore, leading istio to remove the "inbound" cluster from old pods envoy config even though requests might still be going to them (as envoy config propagation is not immediate, client envoy may still send requests to old pods). Any request reaching old pods after that time ends up in 503 NoClusterFound error as no inbound cluster is defined on destination envoy.

Here are samples of old pods envoy clusters config to illustrate the issue:

Before rollout: both stable and canary services reference old pods --> both services appear in old pods envoy config
```json
{
    "name": "inbound|8080||",
    "type": "ORIGINAL_DST",
    "connectTimeout": "10s",
    "lbPolicy": "CLUSTER_PROVIDED",
    "upstreamBindConfig": {
        "sourceAddress": {
            "address": "127.0.0.6",
            "portValue": 0
        }
    },
    "metadata": {
        "filterMetadata": {
            "istio": {
                "services": [
                    {
                        "host": "my-app-canary-svc.my-ns.svc.cluster.local",
                        "name": "my-app-canary-svc",
                        "namespace": "my-ns"
                    },
                    {
                        "host": "my-app-stable-svc.my-ns.svc.cluster.local",
                        "name": "my-app-stable-svc",
                        "namespace": "my-ns"
                    }
                ]
            }
        }
    }
}
```

During rollout: stable service references old pods, canary service references new pods --> only stable service appear in old pods envoy config
```json
{
    "name": "inbound|8080||",
    "type": "ORIGINAL_DST",
    "connectTimeout": "10s",
    "lbPolicy": "CLUSTER_PROVIDED",
    "upstreamBindConfig": {
        "sourceAddress": {
            "address": "127.0.0.6",
            "portValue": 0
        }
    },
    "metadata": {
        "filterMetadata": {
            "istio": {
                "services": [
                    {
[...]

PR #4867, opened 2026-07-16
fix(controller): delay abort scale-down until canary traffic has shifted away

n1koo:
On an aborted traffic-routed canary, the scale-down-delay annotation is added to the canary ReplicaSet as soon as the stable ReplicaSet is fully scaled, without checking that the traffic router has actually stopped sending traffic to the canary. The delay clock starts while `VerifyWeight` may still be in flight, and the ReplicaSet is scaled to 0 at expiry regardless. On ALB this means the canary can be scaled to 0 while the target group still routes to it → 503 UF/UH.

This adds the guard. `scaleDownOldReplicaSetsForCanary` already refuses to annotate **old** ReplicaSets until the router has let go of them — `canProceedWithScaleDownAnnotation` (ALB target-group verification) and `isReplicaSetReferenced` (recorded weights, live Service selectors, Istio DestinationRule subsets). The aborted canary is the **new** RS, so none of those covered it.

Two commits plus a docs update, each independently buildable and testable. Rebased on current master.

## 1. The gate

`addScaleDownDelay` is gated on a new `canaryTrafficShiftedAway()`: the recorded canary weight has reached 0 and, for routers that populate `Weights.Verified`, the shift has been verified. Weights are read from `newStatus`, which `reconcileTrafficRouting` populates earlier in the same reconcile, so the annotation is not delayed an extra reconcile waiting for the previous status to be persisted.

**Primary beneficiary is ALB** with target-group verification enabled (`--aws-verify-target-group`; `defaults.defaultVerifyTargetGroup` is `false`, and every other in-tree router returns `nil` from `VerifyWeight`), plus any plugin that populates `Weights.Verified`:
- `reconcileTrafficRouting` calls `SetWeight(0)` on abort — the K8s object is patched immediately.
- `VerifyWeight` then polls the ALB TargetGroup to confirm the routing actually switched, which can take seconds to minutes.
- Previously the annotation was written in between; now it waits.

For routers that leave `Weights.Verified` nil — which is every in-tree router other than ALB-with-verification — the recorded weight reaches 0 in the same reconcile as `SetWeight`, so the gate opens at the moment the old code would have annotated. This PR is effectively a no-op for them; the gate is only load-bearing where a router actually verifies.

Like `canProceedWithScaleDownAnnotation`, this waits **indefinitely** rather than timing out — an aborted canary that the router never lets go of stays scaled up instead of being scaled down early. A `CanaryTrafficNotShifted` warning event is recorded while it waits, so the reason the canary is still up is visible without reading controller logs. (Earlier revisions of this PR timed the wait out against `progressDeadlineSeconds`; dropped per review — that field is a detection knob, and its only action today is behind `progressDeadlineAbort`.)

**Blue-green** reaches `reconcileNewReplicaSet` by the same path but has no canary traffic router to wait on, so it returns true and abort scale-down is unchanged. Covered by a test, including the case where a rollout converted from a traffic-routed canary still carries stale `status.canary.weights`.

## 2. Verify the abort weight regardless of the current step

`ShouldVerifyWeight` only returns true at a `setWeight` step, or at the end of the rollout. On abort `CurrentStepIndex` is reset to 0, so for a rollout whose steps do not open with `setWeight` — `pause`, `analysis`, `setCanaryScale` — ALB was never asked to confirm the abort weight of 0 had taken effect, and `Weights.Verified` stayed nil. That makes the gate above a no-op for those rollouts: it falls back to the recorded weight value, which reaches 0 as soon as `SetWeight` returns rather than when the target group has converged.

So: verify the weight whenever an aborting rollout is driving it to 0, whichever step it was on. For the common `setWeight`-first layout nothing changes. For the others this adds a target-group check per reconcile while the canary is still up, which is the same cost `setWeight`-first rollouts already pay. Verification stops once the canary reaches 0 replicas, so a rollout parked in the aborted state does not keep polling the cloud provider (per review).

This also hoists the `StableRS`/`IsFullyPromoted` precondition that both existing clauses repeated, rather than repeating it a third time.

**Worth calling out the interaction with the indefinite wait above:** this commit widens the set of aborts where ALB is asked to verify, so it also widens where the gate can block. A wedged target group that never reports weight 0 will now hold the canary scaled up indefinitely, where before this commit it would not have waited at all for those step layouts. That follows from waiting indefinitely rather than timing out, and I think it is the right trade (don't scale down something still in the traffic path, and make it visible via the event) — but it is a consequence worth being explicit about rather than discovering later. Happy to drop this commit if you would rather keep the blast radius to `setWeight`-first rollouts.

## Relation to #4900

Complementary and already rebased over it: #4900 makes the abort weight *value* reach 0 in the `dynamicStableScale=true + explicit abortScaleDownDelaySeconds` case; this PR gates the *annotation* on that recorded weight being 0 and verified. The `>=` tolerance from #4900 for stable transiently exceeding `spec.Replicas` is preserved.

## Related work

- **#4864** gates *background analysis* creation on the canary having received the step's traffic weight. Same underlying idea — don't act until the router has converged — but a different predicate: that one asks "has the weight reached the step's target", this one asks "has it reached zero". They share only the one-line `Weights.Verified` check, so I have kept them as separate helpers in separate files rather than coupling the two PRs to share it. The PRs are independent and can merge in either order.
- **#4195** ("503 NC errors on canary release with Istio ... as well as on rollout abortions") reports an adjacent symptom. The mechanism there is Istio subset/DestinationRule resolution rather than an unconverged ALB target group, so this PR is not a fix for it, but it is the same family.

## Alternatives considered

- **Gate the scale-down at deadline expiry instead of the annotation** — would let the delay run concurrently with router convergence, but `abortScaleDownDelaySeconds` is meant as a drain window *after* traffic stops. Both existing guards gate the annotation.
- **Watch router state directly** (poll the ALB TargetGroup here) — router-specific plumbing, much larger surface.
- **A dedicated `AbortScaleDownGracePeriod` API field**, or timing the wait out against `progressDeadlineSeconds` — both rejected in favour of waiting indefinitely, matching the existing old-RS guard.

## Testing

- `TestScaleDownOnAbortWaitsForCanaryTrafficShift` — the annotation decision across weight > 0, weight 0 unverified, weight 0 verified, weight 0 on a non-verifying router, and nil weights; that the gate reads the freshly-recorded weights rather than the persisted ones; and that a `CanaryTrafficNotShifted` event is recorded while waiting and not once traffic has shifted.
- `TestScaleDownOnAbortBlueGreenUnaffected` — pins the blue-green behaviour.
- `TestShouldVerifyWeightOnAbort` — the abort clause, including that it is specific to weight 0 and that the `StableRS`/`IsFullyPromoted` preconditions still apply.
- `TestVerifyWeight` (ALB) — a rollout aborted from a `pause` step now actually verifies instead of returning early.
- No e2e: reproducing the failure needs a real ALB target group converging slowly, which CI cannot provide.
- Docs: `docs/features/scaledown-aborted-rs.md`, whose table promised "scales down after N seconds" for canary w/ traffic routing, now notes that the delay starts once the weight has reached 0, that the canary otherwise stays up, and that the `CanaryTrafficNotShifted` event records while it waits.
- `go build`, `go vet`, `golangci-lint run`, `gofmt`, and the full `./rollout/...` and `./utils/...` suites: clean, both for the final state and for each of the three commits on its own.

Checklist:

* [x] (b) this is a bug fix
* [x] The title of the PR is [conventional](https://www.conventionalcommits.org/en/v1.0.0/) and states what changed. No related upstream issue is filed for this specific behaviour, so no `Fixes #` suffix.
* [x] My builds are green.

### DIFF
--- a/rollout/replicaset.go
+++ b/rollout/replicaset.go
@@ -16,8 +16,10 @@ import (
 
 	"github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1"
 	"github.com/argoproj/argo-rollouts/utils/annotations"
+	"github.com/argoproj/argo-rollouts/utils/conditions"
 	"github.com/argoproj/argo-rollouts/utils/defaults"
 	logutil "github.com/argoproj/argo-rollouts/utils/log"
+	"github.com/argoproj/argo-rollouts/utils/record"
 	replicasetutil "github.com/argoproj/argo-rollouts/utils/replicaset"
 	serviceutil "github.com/argoproj/argo-rollouts/utils/service"
 	timeutil "github.com/argoproj/argo-rollouts/utils/time"
@@ -194,14 +196,24 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 				return false, fmt.Errorf("failed to sync newRS desired-replicas annotation before adding abort scale-down delay: %w", err)
 			}
 			// Don't annotate until the stable RS is fully scaled, i.e. able to serve 100%
-			// of traffic, since the deadline scales the canary to zero unconditionally.
-			// With dynamicStableScale this holds once the abort weight has stepped down to
-			// zero (see GetDesiredCanaryWeight). >= tolerates stable transiently exceeding
-			// spec.Replicas (e.g. HPA scale-in).
+			// of traffic, AND the router has stopped sending traffic to the canary, since the
+			// deadline scales the canary to zero unconditionally. The latter is the guarantee
+			// scaleDownOldReplicaSetsForCanary already gives old RSs, via
+			// canProceedWithScaleDownAnnotation and isReplicaSetReferenced; neither covered the
+			// aborted canary, because it is the *new* RS. Like those guards, this one waits
+			// indefinitely rather than timing out; a warning event makes the wait visible.
+			// With dynamicStableScale the stable check holds once the abort weight has stepped
+			// down to zero (see GetDesiredCanaryWeight). >= tolerates stable transiently
+			// exceeding spec.Replicas (e.g. HPA scale-in).
 			if c.stableRS.Status.AvailableReplicas >= *c.rollout.Spec.Replicas {
-				err = c.addScaleDownDelay(c.newRS, *abortScaleDownDelaySeconds)
-				if err != nil {
-					return false, err
+				if c.canaryTrafficShiftedAway() {
+					err = c.addScaleDownDelay(c.newRS, *abortScaleDownDelaySeconds)
+					if err != nil {
+						return false, err
+					}
+				} else {
+					c.recorder.Warnf(c.rollout, record.EventOptions{EventReason: conditions.CanaryTrafficNotShiftedReason},
+						conditions.CanaryTrafficNotShiftedMessage, c.newRS.Name)
 				}
 			}
 			// leave newRS scaled up until we annotate
@@ -272,6 +284,28 @@ func (c *rolloutContext) shouldDelayScaleDownOnAbort() bool {
 		return false
 	}
 	return true
+}
+
+// canaryTrafficShiftedAway reports whether the traffic router has stopped sending traffic to the
+// canary ReplicaSet: the canary weight has reached 0 and, for routers that verify weights, that
+// shift has been verified.
+//
+// Weights are read from newStatus because reconcileTrafficRouting runs earlier in the same
+// reconcile and records there both the weight it just programmed and the result of the
+// VerifyWeight call it just made. The persisted status is a reconcile behind.
+func (c *rolloutContext) canaryTrafficShiftedAway() bool {
+	if c.rollout.Spec.Strategy.Canary == nil || c.rollout.Spec.Strategy.Canary.TrafficRouting == nil {
+		// Blue-green reaches reconcileNewReplicaSet by the same path, with no router to wait on.
+		return true
+	}
+	weights := c.newStatus.Canary.Weights
+	if weights == nil {
+		// Aborted before any canary traffic was shifted, so there is nothing to wait for.
+		return true
+	}
+	// The weight refers to c.newRS: the one path that records a different canary pod hash,
+	// IsDynamicallyRollingBackToStable, requires a fully promoted rollout, so is never an abort.
+	return weights.Canary.Weight == 0 && (weights.Verified == nil || *weights.Verified)
 }
 
 // reconcileOtherReplicaSets reconciles "other" ReplicaSets.
--- a/utils/rollout/rolloututil.go
+++ b/utils/rollout/rolloututil.go
@@ -241,12 +241,25 @@ func CanaryStepString(c v1alpha1.CanaryStep) string {
 // ShouldVerifyWeight We use this to test if we should verify weights because weight verification could involve
 // API calls to the cloud provider which could incur rate limiting
 func ShouldVerifyWeight(ro *v1alpha1.Rollout, desiredWeight int32) bool {
+	if ro.Status.StableRS == "" || IsFullyPromoted(ro) {
+		return false
+	}
+	// An abort drives the canary weight to 0, and the canary is scaled down once the router reports
+	// that shift has taken effect, so it has to be verified whichever step the rollout was on when
+	// it was aborted. CurrentStepIndex is reset to 0 on abort and step 0 is not necessarily a
+	// setWeight step, so the check below is not enough on its own.
+	if ro.Status.Abort && desiredWeight == 0 {
+		// Nothing left to protect once the canary is at 0 replicas; skipping avoids polling the
+		// cloud provider every resync for a parked aborted rollout.
+		return ro.Status.UpdatedReplicas > 0
+	}
 	currentStep, _ := replicasetutil.GetCurrentCanaryStep(ro)
 	// If we are in the middle of an update at a setWeight step, also perform weight verification.
 	// Note that we don't do this every reconciliation because weight verification typically involves
-	// API calls to the cloud provider which could incur rate limitingq
-	shouldVerifyWeight := (ro.Status.StableRS != "" && !IsFullyPromoted(ro) && currentStep != nil && currentStep.SetWeight != nil) ||
-		(ro.Status.StableRS != "" && !IsFullyPromoted(ro) && currentStep == nil && desiredWeight == weightutil.MaxTrafficWeight(ro)) // We are at end of rollout
-
-	return shouldVerifyWeight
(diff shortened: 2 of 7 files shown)


n1koo:
Rebased over #4900. The two fixes are complementary but operate at different layers:

- **#4900**: ensures the abort weight *value* reaches 0 in the `dynamicStableScale=true + explicit abortScaleDownDelaySeconds` case (previously bounded by canary size → deadlock).
- **This PR**: gates the scale-down-delay *annotation* on the recorded weight being 0 AND (for routers that populate `Weights.Verified`) verified.

The `>=` tolerance from #4900 (for stable transiently exceeding `spec.Replicas`) is preserved in the merged code.

**Scope note**: primary beneficiary is ALB and other verifying-router plugins — `VerifyWeight` polls the target group to confirm the actual routing has switched, and the old code annotates the scale-down-delay as soon as stable finishes scaling (which for ALB can be before the target group has converged). For routers that leave `Weights.Verified` nil (Istio, SMI, Nginx), this PR is effectively a no-op: by the time `SetWeight(0)` returns and status is recorded, the gate is already open, matching old behaviour.

The `progressDeadlineSeconds` safety valve exists for the case where a verifying router gets wedged, so the abort still terminates.

PR body updated to reflect this narrower framing.

rollout/replicaset.go
```diff
@@ -245,6 +250,60 @@ func (c *rolloutContext) shouldDelayScaleDownOnAbort() bool {
...
+// progressDeadlineSeconds is reused rather than adding a knob: it is already the "how long may
+// this rollout make no progress before we give up" tunable, and a router that never converges is
+// a form of no progress.
+func (c *rolloutContext) canAnnotateAbortScaleDown() bool {
+	if c.canaryTrafficShiftedAway() {
+		return true
+	}
+	// AbortedAt is stamped by the same reconcile that first sees the abort, so a nil here is a
+	// one-reconcile transient: grant the full period rather than expiring it immediately.
+	deadline := time.Duration(defaults.GetProgressDeadlineSecondsOrDefault(c.rollout)) * time.Second
```
zachaller: Let's not reuse `progressDeadlineSeconds` here — this isn't its intended use. It's a detection knob for update progress (it flips the `Progressing` condition); the only place it drives an action today is behind an explicit `progressDeadlineAbort: true` opt-in. The analogous guard for old RSs (`canProceedWithScaleDownAnnotation`) blocks indefinitely and that's been fine in practice, so I'd rather wait indefinitely here too than tie the timeout to this field — a warning event while the gate waits would cover the operator-visibility side.

n1koo: I've dropped the valve, so the gate now waits indefinitely, the same way `canProceedWithScaleDownAnnotation` does.

For visibility I've added a `CanaryTrafficNotShifted` warning event that fires while the gate is holding, modeled on `TargetGroupUnverifiedReason`. It records on every reconcile. That seemed acceptable because the recorder aggregates identical events and `sendNotifications` returns early when nothing is subscribed to the trigger, but I can make it fire only when the wait begins if you'd rather, though that needs somewhere in status to remember it.

There is one consequence of doing both halves of what you suggested that I want to flag. The `ShouldVerifyWeight` commit makes that function return true for any aborting rollout driving the weight to 0, which widens where the gate can block. Now that the wait is indefinite, a wedged target group will hold the canary scaled up for good, whereas before that commit it would not have waited at all for step lists that do not start with `setWeight`. That also makes it the riskiest commit in the PR and the least related to it, since it changes shared behavior for every ALB user and adds a `DescribeTargetGroups` call per reconcile during an abort. I am happy to move it into its own PR and leave this one as just the gate. Which would you prefer?

n1koo: Rebased onto master now that #4919 is in

`syncNewRSReplicasAnnotation()` runs on entry to the `abortScaleDownDelaySeconds` branch, before the gate, so it still fires on the reconciles where the gate holds the canary back. That matters more now than it did with a timeout, since the wait is unbounded.

`TestAbortScaleDownDelayDoesNotWedgeScalingEvent` did fail against the gate, and I have adjusted it. `syncReplicasOnly` does not reconcile traffic routing, so the recorded canary weight is still the pre-abort 50 and the gate correctly refuses to annotate. Both of your assertions still hold, the canary stays at scale and the desired-replicas annotation syncs to 4; what changed is that the scale-down-deadline patch is no longer part of the same reconcile, so I removed it from the expected actions and left a comment explaining why. The behaviour converges rather than wedging: terminating the scaling event is what lets the next reconcile take the full path, drive the abort weight to 0, and annotate then.

If you would rather the gate did not consult weights that the current reconcile has not refreshed, so that your test passes unmodified, I can do that instead. It would mean the gate is bypassed whenever an HPA change lands mid-abort, which is why I did not do it

utils/rollout/rolloututil.go
```diff
@@ -241,12 +241,23 @@ func CanaryStepString(c v1alpha1.CanaryStep) string {
...
+	if ro.Status.StableRS == "" || IsFullyPromoted(ro) {
+		return false
+	}
+	// An abort drives the canary weight to 0, and the canary is scaled down once the router reports
+	// that shift has taken effect, so it has to be verified whichever step the rollout was on when
+	// it was aborted. CurrentStepIndex is reset to 0 on abort and step 0 is not necessarily a
+	// setWeight step, so the check below is not enough on its own.
+	if ro.Status.Abort && desiredWeight == 0 {
+		return true
+	}
```
zachaller, line 255: Should we stop verifying once the canary is at 0 replicas? Otherwise a parked aborted rollout keeps polling the cloud provider every resync forever.

````suggestion
	if ro.Status.Abort && desiredWeight == 0 {
		// Nothing left to protect once the canary is at 0 replicas; skipping avoids
		// polling the cloud provider every resync for a parked aborted rollout.
		return ro.Status.UpdatedReplicas > 0
	}
````

n1koo: Applied, thanks for catching that. `Status.UpdatedReplicas` comes from `newRS.Status.Replicas` via `GetActualReplicaCountForReplicaSets`, so it is the canary's actual pod count, which is the right signal here. Once it hits 0 the scale down has already happened and there is nothing the verification could still protect.

It also does not open a hole in the gate: while the canary is being held at scale waiting for the traffic shift it always has pods, so verification runs exactly over the window where the gate is making a decision. The value comes from the persisted status, so at worst there is one extra poll after the canary reaches 0.

Added a case to `TestShouldVerifyWeightOnAbort` for the 0 replica path, and the ALB `TestVerifyWeight` block now sets `UpdatedReplicas` since its scenario is a canary that is still up.

That leaves the other question from above open: whether you would rather this commit went into its own PR, given it changes `ShouldVerifyWeight` for every ALB user rather than just this abort path.

PR #4868, opened 2026-07-16
fix(controller): dynamic HPAReplicas based on configuration

### LINKED ISSUE #4847: Rollout ↔ HPA/KEDA Pod-Count Flip-Flop During Canary
Checklist:

* [ yes] I've included steps to reproduce the bug.
* [yes] I've included the version of argo rollouts.

**Describe the bug**

When an Argo Rollout using the Canary strategy is autoscaled by an HPA managed through KEDA Cron Scaler, the rollout enters a pod-count oscillation during a setCanaryScale.weight step.

The observed behavior repeatedly oscillates between:

`10 replicas → 11 replicas → 13 total pods → 10 replicas → ...`

This only occurs when:

Canary strategy is used
dynamicStableScale: false (default)
setCanaryScale.weight is used
Rollout is scaled by an HPA created by KEDA (Cron Scaler in our case)

From our investigation, the feedback loop appears to occur because the Rollout reports all ReplicaSet pods (stable + canary) through .status.HPAReplicas.

The HPA then treats this inflated value as the current replica count, updates .spec.replicas, which causes the canary replica calculation (ceil(weight * replicas)) to create another canary pod. This further inflates .status.HPAReplicas, producing a continuous reconciliation loop.

Blue/Green strategy appears to avoid this by reporting only the active ReplicaSet to HPA.

**To Reproduce**

Environment:

Argo Rollouts with Canary strategy
KEDA Cron Scaler
HPA created by KEDA
dynamicStableScale: false

Steps:

Configure a Rollout with:
replicas: 10
Configure a KEDA Cron Scaler to maintain:
desiredReplicas = 10
Wait until the rollout is stable.
Start a rollout using:
steps:
- setCanaryScale:
    weight: 10
Observe the rollout events.

Observed behavior:

```
stable=10 canary=1 total=11
↓
HPA scales spec.replicas to 11
↓
stable=11 canary=2 total=13
↓
HPA scales back to 10
↓
Loop repeats
```

Kubernetes events continuously alternate between:

Scaled up replica set
Scaled down replica set

instead of remaining stable.

**Expected behavior**

The autoscaler should continue treating the stable workload size as the current replica count.

Transient canary surge pods should not feed back into HPA scaling decisions.

The rollout should remain stable throughout the canary step instead of repeatedly scaling up and down.

**Screenshots**

Image

Image

**Version**
argo-rollouts:v1.6.6
k8s - 1.35
KEDA - 2.17.2

**Logs**

---

zachaller:
…. Fixes #4847

## Summary

  During a traffic-routed canary update (with `dynamicStableScale: false`), the Rollout
  temporarily runs up to 2x `spec.replicas` pods (stable stays fully scaled while the canary
  surges). `status.HPAReplicas` — the field exposed as `statusReplicasPath` on the scale
  subresource — counted all of them, feeding the inflated total to any HPA/KEDA scaler
  targeting the Rollout.

  This PR changes `status.HPAReplicas` to count only the **stable ReplicaSet's** pods when
  using traffic routing without `dynamicStableScale`. The reported count stays equal to
  `spec.replicas` throughout an update, matching both the field's documented meaning
  ("replicas that are receiving active traffic") and blue-green's long-standing behavior of
  reporting only the active ReplicaSet.

  ## The feedback loop being fixed

  Object/External metrics with an `AverageValue` target (most KEDA scalers, e.g. cron) divide
  the metric by `scale.status.replicas` for the HPA's tolerance check — and when the result is
  within tolerance, the HPA **adopts that count as the desired replica count**
  ([`GetExternalPerPodMetricReplicas`](https://github.com/kubernetes/kubernetes/blob/v1.33.0/pkg/controller/podautoscaler/replica_calculator.go#L366-L390)).
  With a KEDA cron scaler holding `desiredReplicas: 10`:

  stable=10 canary=1 → HPAReplicas=11 → ratio 10/11 = 0.909 (within 10% tolerance)
  → HPA adopts 11 → spec.replicas=11 → canary=ceil(weight×11) → total 13
  → ratio 10/13 = 0.769 (outside tolerance) → back to 10 → repeat

  This oscillation (`10 → 11 → 13 → 10 → …`) continues for the entire update. With this
  change the reported count stays at `spec.replicas`, the ratio stays 1.0, and the loop
  cannot start. This also excludes old ReplicaSets draining under `scaleDownDelaySeconds`,
  which previously inflated the count even after promotion.

  ## What deliberately did NOT change

  **`status.selector` is untouched** and continues to match all pods owned by the Rollout.
  Metrics sampled through the selector (Resource/ContainerResource Utilization, `Pods`) must
  observe every traffic-taking pod: the HPA multiplies the sampled average back by the
  sampled pod count, so the surge self-corrects arithmetically — but only if the sampled sum
  captures all the work. Narrowing the selector to stable pods was attempted in #4074 and
  reverted in #4553 because it blinded the HPA to the canary's load and scaled applications
  down under traffic. This PR fixes the count-consuming path without touching the
  selector-consuming path, which is what makes it safe for mixed-metric HPAs
  (e.g. CPU utilization + KEDA cron on one HPA).

  The stable count is only reported when it represents serving capacity: basic canary
  (no `trafficRouting`), `dynamicStableScale: true`, and pre-first-promotion (no stable RS
  yet) all keep the previous all-pods behavior.

  ## Why a behavior change rather than a config option

  Earlier iterations of this branch added an opt-in `scaleReporting: mode: All|Stable` field.
  It was removed because: every enumerated consumer of `status.HPAReplicas` in the HPA
  controller is neutral or strictly improved by the change; the field's own API doc already
  promises this semantic; blue-green has shipped the same semantic unconditionally for years;
  and an opt-out can be added later, additively, if a consumer of `/scale` surfaces that
  needs the old count. (Reviewers: see the linked issue for the full analysis.)

  ## Known remaining limitation

  Object/External metrics with a `Value` target multiply their usage ratio by the ready-pod
  count from `status.selector`
  ([`getUsageRatioReplicaCount`](https://github.com/kubernetes/kubernetes/blob/v1.33.0/pkg/controller/podautoscaler/replica_calculator.go#L279-L299)),
  which remains inflated during updates — unchanged, long-standing behavior. Documented
  workarounds: restate the metric as an `AverageValue` target, or use `dynamicStableScale`.
  Fixing this properly requires an upstream change (the multiplier arguably should be
  `statusReplicas`).

  ## Changes

  - `rollout/canary.go`: new `calculateCanaryHPAReplicas()` helper used by **both** status
    paths (`syncRolloutStatusCanary` and `syncReplicasOnly`)
  - Unit tests: mid-update, post-promotion/draining, and no-traffic-routing cases; restored
    the scale-subresource patch assertions removed in the #4553 revert, now pinning
    `HPAReplicas`
  - E2E test (`TestCanaryScaleReportingHPAReplicas`): runs a **live HPA** (pinned
    min=max so it never rescales) against the Rollout and asserts at every phase on the
    actual `GET /scale` subresource — `status.replicas` stays at `spec.replicas` while total
    pods surge to 3 (mid-update) and 4 (draining), selector unchanged, and the HPA
    controller reads the target without errors (`AbleToScale=True`)
  - Docs: `hpa-support.md` gains a "Scale Reporting During Canary Updates" section — the two
    scale-subresource consumer paths (with upstream code links), per-metric-type behavior
    table, a worked derivation of why sampled metrics self-correct, `setCanaryScale`
    interaction, and `dynamicStableScale` guidance

  ## Release note

  > `status.HPAReplicas` (exposed via the scale subresource) now counts only the stable
  > ReplicaSet's pods for canary Rollouts using traffic routing without
  > `dynamicStableScale`, instead of all pods. This prevents HPA/KEDA autoscaler
  > oscillation during updates. Consumers reading the Rollout `/scale` subresource should
  > note the reported replica count no longer includes canary surge or draining pods.


docs/features/hpa-support.md
```diff
@@ -287,6 +289,74 @@ strategy:
...
+
+## Scale Reporting During Canary Updates
+
+!!! important
+    Behavior changed in v1.11. Previously, `status.HPAReplicas` counted all pods owned by the Rollout at all times, which could cause the autoscaler feedback loop described below.
+
+The HPA (and autoscalers built on it, such as KEDA) reads the Rollout through the `scale` subresource, which exposes two fields from the Rollout status:
+
+- `status.selector` (`labelSelectorPath`) — the label selector the HPA uses to list pods. It is used to sample per-pod metrics (Resource/ContainerResource and Pods metrics) and to count ready pods for Object/External metrics with a `Value` target. It always matches **all** pods owned by the Rollout, so sampled metrics observe every pod taking traffic. (Narrowing it to the stable pods would blind the HPA to the canary's load and cause it to scale the application down while the canary is serving traffic — this was attempted in v1.8 and reverted.)
```
agaudreault:
```suggestion
- `status.selector` (`labelSelectorPath`) — the label selector the HPA uses to list pods. It is used to sample per-pod metrics (Resource/ContainerResource and Pods metrics) and to count ready pods for Object/External metrics with a `Value` target. It always matches **all** pods owned by the Rollout, so sampled metrics observe every pod taking traffic. (Narrowing it to the stable pods would blind the HPA to the canary's load and cause it to scale the application down while the canary is serving traffic.)
```

docs/features/hpa-support.md
```diff
@@ -287,6 +289,74 @@ strategy:
     - pause: {}
 ```
 
+**Note:** because `setCanaryScale` detaches the canary pod count from the traffic weight, canary pods may be running while receiving little or no traffic. These pods are excluded from the replica count reported to the HPA, and although they appear in metrics sampled through the label selector, the HPA's [averaging arithmetic self-corrects](#why-sampled-metrics-self-correct-during-a-surge) — see [Scale Reporting During Canary Updates](#scale-reporting-during-canary-updates) below.
+
+## Scale Reporting During Canary Updates
```
agaudreault, line 294: The new hpa-support.md doc is confusing because it mixes the explanation of the 1.11 changes with the HPA behaviour. 

To make it clearer, I think the change should be explained in a release note documentation (detailing the problem pre-1.11 and the fix), while this doc describes the current rollout behaviour.

agaudreault (commented):
Overall code change LGTM. I do think the documentation needs to be clearer (behaviour vs this change), and I like the table that indicates what is the behaviour depending on the metrics used in HPA.

agaudreault (approved):

PR #4891, opened 2026-07-19
fix: skip experiment weight destinations with empty ServiceName

### LINKED ISSUE #4890: Weighted experiment step can write VirtualService route destinations with empty host, black-holing traffic
### Checklist:

### Describe the bug

When a canary Rollout uses a **weighted experiment step** with Istio traffic routing, the controller can write route destinations with an **empty `host`** (carrying the experiment arm's weight) into the managed VirtualService during experiment teardown. Istio/Envoy cannot route these destinations, so that share of traffic is black-holed (503s) until the VirtualService is manually restored — the controller never cleans the bad entries up.

Root cause: `calculateWeightDestinationsFromExperiment` (rollout/trafficrouting.go) builds `WeightDestination`s from `experiment.status.templateStatuses[].serviceName` with no check that the value is non-empty. During teardown the template status can momentarily carry an empty `serviceName` while the template weight from the step spec still applies. The Istio reconciler (`processRoutes` in rollout/trafficrouting/istio/istio.go) then copies the empty ServiceName verbatim into `destination.host`.

The controller's own event shows the poisoned state (fields: weight, serviceName, podTemplateHash):

```
# healthy, mid-experiment
TrafficWeightUpdated: Traffic weight updated additional: [{1 my-app-baseline 5ddc99f68d} {1 my-app-new-canary bbb98bb5b}]

# during teardown
TrafficWeightUpdated: Traffic weight updated additional: [{1  } {1  }]
```

which lands in the VirtualService as:

```yaml
route:
  - destination: {host: my-app}
    weight: 98
  - destination: {host: my-app-canary}
    weight: 0
  - destination: {host: ""}     # black hole, receives real traffic share
    weight: 1
  - destination: {host: ""}
    weight: 1
```

Because Istio weights are proportional, every such entry steals a share of traffic from the valid destinations. I hit this in a production environment I operate, during normal experiment teardown (no manual intervention involved), and saw intermittent 503s.

### To Reproduce

1. Canary Rollout with Istio `trafficRouting` and an experiment step whose templates set `weight` (e.g. two arms at `weight: 1`) — the arm Services are auto-created by the Experiment.
2. Wait until the Experiment is Running and the VirtualService shows the arm destinations.
3. The race fires naturally at teardown, but to reproduce it deterministically, simulate the teardown state by clearing the service names in the experiment status while the weights are still in play (the `experiments` CRD has no status subresource, so a plain patch works):

```sh
kubectl patch experiment <name> --type=json \
  -p '[{"op":"replace","path":"/status/templateStatuses/0/serviceName","value":""},{"op":"replace","path":"/status/templateStatuses/1/serviceName","value":""}]'
```

repeated for a few seconds (the experiment controller restores the names quickly; each write also enqueues a rollout reconcile).

4. Observe the VirtualService: `spec.http[0].route` gains `destination: {host: ""}` entries carrying the arm weight; `metadata.generation` climbs; traffic to the VS hosts starts returning 503s. The entries survive experiment completion.

### Expected behavior

The controller should never emit a `WeightDestination` (and hence a route destination) without a ServiceName. A template status with an empty `serviceName` should be skipped until the experiment controller repopulates it.

### Version

Observed on v1.8.3; the affected code is unchanged on master (`calculateWeightDestinationsFromExperiment` in `rollout/trafficrouting.go` has no empty-ServiceName guard).

tanmay-agrawal-india:
Fixes #4890

During experiment teardown the experiment's template statuses can momentarily carry an empty `serviceName` while the template weight from the step spec still applies. `calculateWeightDestinationsFromExperiment` copied the empty ServiceName into a `WeightDestination`, which the Istio reconciler renders as a VirtualService route destination with an empty host carrying the arm's weight. Envoy cannot route such destinations, so that traffic share is black-holed (503s), and the entries are never cleaned up — recovery requires manually restoring the VirtualService. I hit this in a production environment I operate, during normal experiment teardown.

This change skips template statuses with an empty ServiceName (logging a warning) so a route destination without a host can never be emitted, and adds a regression test simulating the teardown race.

The controller's own event stream shows the poisoned state this prevents:

```
TrafficWeightUpdated: Traffic weight updated additional: [{1  } {1  }]
```

### Alternatives considered

- **Gate on `IsTerminating` instead of checking the ServiceName**: terminating is only one of several ways the status can carry a blank ServiceName (Service-creation failure at experiment start, template `Failed`/`Error`, external status writes — the `experiments` CRD has no status subresource). The empty-name check enforces the invariant itself — never emit a destination without a host — rather than one of its causes.
- **Order teardown so the VirtualService is cleaned before arm Services are deleted**: cross-controller ordering fights the level-triggered model and can wedge teardown when the rollout controller is slow or unavailable. Tolerating a mid-transition snapshot is the idiomatic contract.
- **Treat the condition as experiment termination and advance/abort the rollout**: a blank ServiceName is observed-state lag, not a lifecycle signal — it also occurs during healthy teardown seconds before `Phase` leaves `Running`; acting on it would abort healthy releases. Lifecycle decisions stay with experiment phase + analysis results.

### Checklist:

* [x] I've signed my commits with [DCO](https://github.com/argoproj/argo-rollouts/blob/master/CONTRIBUTING.md)

### DIFF
--- a/rollout/trafficrouting.go
+++ b/rollout/trafficrouting.go
@@ -440,25 +440,39 @@ func calculateWeightStatus(ro *v1alpha1.Rollout, canaryHash, stableHash string, 
 func (c *rolloutContext) calculateWeightDestinationsFromExperiment() []v1alpha1.WeightDestination {
 	weightDestinations := make([]v1alpha1.WeightDestination, 0)
 	exStep := replicasetutil.GetCurrentExperimentStep(c.rollout)
-	if exStep != nil && c.currentEx != nil && c.currentEx.Status.Phase == v1alpha1.AnalysisPhaseRunning {
-		getTemplateWeight := func(name string) *int32 {
-			for _, tmpl := range exStep.Templates {
-				if tmpl.Name == name {
-					return tmpl.Weight
-				}
-			}
-			return nil
-		}
-		for _, templateStatus := range c.currentEx.Status.TemplateStatuses {
-			templateWeight := getTemplateWeight(templateStatus.Name)
-			if templateWeight != nil {
-				weightDestinations = append(weightDestinations, v1alpha1.WeightDestination{
-					ServiceName:     templateStatus.ServiceName,
-					PodTemplateHash: templateStatus.PodTemplateHash,
-					Weight:          *templateWeight,
-				})
+	if exStep == nil || c.currentEx == nil || c.currentEx.Status.Phase != v1alpha1.AnalysisPhaseRunning {
+		return weightDestinations
+	}
+	getTemplateWeight := func(name string) *int32 {
+		for _, tmpl := range exStep.Templates {
+			if tmpl.Name == name {
+				return tmpl.Weight
 			}
 		}
+		return nil
+	}
+	for _, templateStatus := range c.currentEx.Status.TemplateStatuses {
+		templateWeight := getTemplateWeight(templateStatus.Name)
+		if templateWeight == nil {
+			continue
+		}
+		if templateStatus.ServiceName == "" {
+			// A weighted experiment template requires a Service to route to.
+			// During experiment teardown (or if the experiment's Service is
+			// removed) the template status can momentarily carry an empty
+			// ServiceName while the template weight still applies. Emitting a
+			// WeightDestination without a ServiceName causes traffic routers
+			// to render destinations with an empty host (e.g. an Istio
+			// VirtualService route to host ""), which black-holes that share
+			// of traffic.
+			c.log.Warnf("Skipping weight destination for experiment template '%s': empty ServiceName", templateStatus.Name)
+			continue
+		}
+		weightDestinations = append(weightDestinations, v1alpha1.WeightDestination{
+			ServiceName:     templateStatus.ServiceName,
+			PodTemplateHash: templateStatus.PodTemplateHash,
+			Weight:          *templateWeight,
+		})
 	}
 	return weightDestinations
 }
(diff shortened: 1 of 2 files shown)


tanmay-agrawal-india:
Adding the exact upstream code path that opens the window, for reviewers — both halves are in-tree:

**Producer side (experiments controller)** — `experiments/experiment.go`:

- `createTemplateService` records the Service in status only after creation succeeds: `templateStatus.ServiceName = newService.Name` (the spec's `service.name` is optional — when omitted the name is generated, so status is the only authoritative record of what exists).
- When a template's **desired replica count reaches 0** — the template completed, or the experiment is terminating (duration elapsed, abort, `Spec.Terminate`, or a failed/errored template) — **and** its pods have drained (`rs.Status.AvailableReplicas == 0`), `deleteTemplateService` deletes the Service and blanks the status fields in the same reconcile:

```go
ec.templateServices[templateName] = nil
templateStatus.ServiceName = ""
templateStatus.PodTemplateHash = ""
```

Because the `AvailableReplicas == 0` gate depends on pod drain timing, the exact moment this lands is unpredictable — and it can land while `Status.Phase` is still `Running` (`Spec.Terminate = true` does not immediately move the phase).

**Consumer side (rollout controller)** — `rollout/trafficrouting.go`: `calculateWeightDestinationsFromExperiment` gates only on `c.currentEx.Status.Phase == v1alpha1.AnalysisPhaseRunning` and takes the arm weight from the step spec. In the window above it reads a blank `ServiceName` with a live weight and emits the empty `WeightDestination`.

Since `WeightDestination` is consumed by every traffic router (istio, smi, apisix, and third-party RPC plugins), guarding the single producer covers them all — which is why the fix is here rather than in the Istio reconciler.

Degradation with the guard is the safe direction: for the reconciles inside the window the arm simply receives no traffic share (identical to the post-teardown steady state), and the next reconcile with a repopulated status converges normally — no ordering between the two controllers is introduced.

kostis-codefresh (commented):
@tanmay-agrawal-india 

Is the better in any way than https://github.com/argoproj/argo-rollouts/pull/4835 ?

PR #4899, opened 2026-07-21
fix(controller): abort deadlock with dynamicStableScale + abortScaleDownDelaySeconds

### LINKED ISSUE #4898: abortScaleDownDelaySeconds deadlocks with dynamicStableScale on abort — canary/newRS never scales down
### Summary                                                                                                                                        
                                                                                                                                                     
  When a Rollout uses both `dynamicStableScale: true` and an explicitly-set `abortScaleDownDelaySeconds`, aborting the rollout mid-progression       
  deadlocks: the canary/new ReplicaSet is never annotated with `scale-down-deadline`, never scales down, and traffic never shifts back to stable     
  (both ALB and gatewayAPI plugin observe stale weights).                                                                                            
                                                                                                                                                     
  Removing `abortScaleDownDelaySeconds` from the spec works around it, because `shouldDelayScaleDownOnAbort` then falls into the branch that skips   
  the delay path entirely for dynamic scaling.                                                                                                       
                                                                                                                                                   
  ### Reproduction                                                                                                                                 
                                                                                                                                                     
  Rollout with:                                        
  ```yaml                                                                                                                                            
  spec:                                                                                                                                            
    replicas: 10                                                                                                                                     
    strategy:                         
      canary:                                                                                                                                        
        dynamicStableScale: true                                                                                                                     
        abortScaleDownDelaySeconds: 125                                                                                                            
        trafficRouting:                                                                                                                              
          alb: { ... }         # or plugins: { argoproj-labs/gatewayAPI: ... }
        steps:                                                                                                                                     
          - setWeight: 10                              
          - pause: {}                            
          # ...more steps                              
  ```                                                  
                                    
  1. Let rollout progress to a step with non-zero canary weight (e.g. 10%).                                                                          
  2. Abort the rollout (`kubectl argo rollouts abort <name>`).
  3. Observe: canary pods stay at their current count indefinitely; `status.canary.weights` stays at whatever it was pre-abort; traffic router       
  receives no weight updates.                                                                                                                        
                                                                                                                                                   
  ### Root Cause                                                                                                                                     
                                                               
  `rollout/replicaset.go` `reconcileNewReplicaSet` gates the scale-down-deadline annotation on `stableRS.Available == spec.Replicas`:              
                                                                                                                                                     
  ```go                                                
  } else if abortScaleDownDelaySeconds != nil {                                                                                                      
      // Don't annotate until need to ensure the stable RS is fully scaled                                                                           
      if c.stableRS.Status.AvailableReplicas == *c.rollout.Spec.Replicas {                                                                           
          err = c.addScaleDownDelay(c.newRS, *abortScaleDownDelaySeconds)                                                                            
          ...                                                                                                                                      
      }                                                                                                                                              
      return false, nil                                        
  }                                                                                                                                                  
  ```                                                          
                                                                                                                                                     
  With `dynamicStableScale: true`, stable is intentionally kept proportional to the reverse traffic weight
  (`CalculateReplicaCountsForTrafficRoutedCanary` / `GetDesiredCanaryWeight`) and only reaches `spec.Replicas` once canary has fully drained.        
  Meanwhile:                                                                                                                                         
                                                                                                                                                     
  - `GetDesiredCanaryWeight` computes `canaryReplicas = max(spec.Replicas - stable.Available, newRS.Available)` — bounded below by `newRS.Available` 
  - Reverse-step iteration returns the highest weight whose replicas < `canaryReplicas`                                                            
  - With `newRS.Available` stuck (no annotation → no scale-down), the desired weight stops stepping down                                           
  - Stable stays proportional to the stuck weight → never reaches `spec.Replicas`
  - → annotation is never added → circular wait                                                                                                      
                                                       
  The comment `// Don't annotate until need to ensure the stable RS is fully scaled` implicitly assumes stable will eventually reach full scale —    
  true for non-dynamic canary, false for `dynamicStableScale`. 
                                                                                                                                                     
  ### Expected Behavior                                
                                                                                                                                                     
  Per the [scaledown-aborted-rs docs](https://argo-rollouts.readthedocs.io/en/stable/features/scaledown-aborted-rs/), `abortScaleDownDelaySeconds`   
  should give canary pods a graceful drain window then scale them to 0. The docs don't mention any incompatibility with `dynamicStableScale`.        
                                                                                                                                                     
  ### Proposed Fix                                                                                                                                   
                                                                                                                                                   
  Relax the gate to check combined stable + canary capacity: `stable.Available + newRS.Available >= spec.Replicas`. This preserves the invariant "we 
  won't drop canary until we have enough capacity elsewhere" while allowing the annotation to be added when dynamic stable scale hasn't yet fully
  ramped stable up. The 125 s (or configured) delay then runs; stable ramps up naturally as newRS drains at deadline.                              
                                                                                                                                                     
PR #4899                      
                                                                                                                                                     
  ### Environment                                              
                                                                                                                                                   
  - argo-rollouts: master (issue also present in current stable)                                                                                     
  - Traffic routers affected (independently): ALB, argoproj-labs/gatewayAPI plugin — both are passive receivers of controller-computed weights.

amit-darji:
## Summary

Fixes #4898.

On abort, `reconcileNewReplicaSet` gated the scale-down-deadline annotation on `stableRS.Available == spec.Replicas`. With `dynamicStableScale: true` this equality never holds (stable is deliberately kept proportional to reverse traffic weight until canary drains), producing a deadlock: canary never gets annotated → never scales down → traffic weight never steps down → stable never grows → gate never satisfied.

This PR splits the gate by mode:

- **Non-dynamic canary / BlueGreen:** preserves the original strict check `stableRS.Available == spec.Replicas`. Stable is kept at `spec.Replicas` by design in this mode, and a transient stable degradation should legitimately hold the drain window.
- **`dynamicStableScale: true`:** relaxes to `stableRS.Available + newRS.Available >= spec.Replicas`. Combined capacity is the safety invariant that actually matters, and it avoids the circular wait.

## Changes

- `rollout/replicaset.go`: split the annotation gate into dynamic vs non-dynamic branches with comments explaining the invariant each preserves.
- `rollout/replicaset_test.go`: new `TestReconcileNewReplicaSetAbortDynamicStableScale` (5 subtests) covering:
  - dynamic: combined capacity meets spec.Replicas (deadlock repro — now annotates)
  - dynamic: stable alone meets spec.Replicas (annotates)
  - dynamic: combined capacity below spec.Replicas (does not annotate)
  - non-dynamic: stable at spec.Replicas (annotates — legacy behavior)
  - non-dynamic: stable degraded (does NOT annotate — legacy strict check preserved)

## Behavior Notes

- **HPA-scaled Rollouts with `dynamicStableScale`:** new dynamic-branch check is tolerant of `stable.Available > spec.Replicas` — previously would never annotate.
- **`setCanaryScale`:** `UseSetCanaryScale` already disables setCanaryScale on abort when `abortScaleDownDelaySeconds` is set; canary sizing falls back to traffic-weight logic. Dynamic-branch check accommodates the higher `newRS.Available` correctly.
- **Traffic-shift ordering unchanged:** the existing `calculateDesiredWeightOnAbortOrStableRollback` guard (`trafficrouting.go:388-392`) still holds traffic shifts until stable has enough available replicas — so stable pods ramp up before traffic moves, preserving the safety-by-design of `dynamicStableScale`.

## Test Plan

- [x] `go test ./rollout/... -count=1` — passes
- [x] Reverted fix locally, ran new subtest — fails as expected (confirms regression coverage)
- [x] CI: golangci-lint + full test suite

### DIFF
--- a/rollout/replicaset.go
+++ b/rollout/replicaset.go
@@ -164,12 +164,24 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 				}
 			}
 		} else if abortScaleDownDelaySeconds != nil {
-			// Ensure combined stable+canary capacity covers spec.Replicas before annotating.
-			// Using equality with only stable.Available deadlocks when dynamicStableScale is
-			// enabled: stable is intentionally kept proportional to reverse traffic weight and
-			// only reaches spec.Replicas once canary is fully drained — which itself is gated
-			// on this annotation.
-			if c.stableRS.Status.AvailableReplicas+c.newRS.Status.AvailableReplicas >= *c.rollout.Spec.Replicas {
+			// Gate the scale-down-deadline annotation on sufficient serving capacity:
+			//   - Non-dynamic canary / BlueGreen: stable is kept at spec.Replicas by design,
+			//     so we retain the original strict check (stable.Available == spec.Replicas).
+			//     This preserves the legacy behavior of waiting for any transient stable
+			//     degradation to heal before starting the drain window.
+			//   - dynamicStableScale: stable is intentionally kept proportional to the reverse
+			//     traffic weight and only reaches spec.Replicas once canary is fully drained.
+			//     Requiring stable == spec.Replicas here deadlocks — canary can't shrink until
+			//     stable grows and vice versa. Fall back to combined stable+canary capacity.
+			usesDynamicStableScale := c.rollout.Spec.Strategy.Canary != nil &&
+				c.rollout.Spec.Strategy.Canary.DynamicStableScale
+			var hasCapacity bool
+			if usesDynamicStableScale {
+				hasCapacity = c.stableRS.Status.AvailableReplicas+c.newRS.Status.AvailableReplicas >= *c.rollout.Spec.Replicas
+			} else {
+				hasCapacity = c.stableRS.Status.AvailableReplicas == *c.rollout.Spec.Replicas
+			}
+			if hasCapacity {
 				err = c.addScaleDownDelay(c.newRS, *abortScaleDownDelaySeconds)
 				if err != nil {
 					return false, err
--- a/rollout/replicaset.go
+++ b/rollout/replicaset.go
@@ -164,8 +164,12 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 				}
 			}
 		} else if abortScaleDownDelaySeconds != nil {
-			// Don't annotate until need to ensure the stable RS is fully scaled
-			if c.stableRS.Status.AvailableReplicas == *c.rollout.Spec.Replicas {
+			// Ensure combined stable+canary capacity covers spec.Replicas before annotating.
+			// Using equality with only stable.Available deadlocks when dynamicStableScale is
+			// enabled: stable is intentionally kept proportional to reverse traffic weight and
+			// only reaches spec.Replicas once canary is fully drained — which itself is gated
+			// on this annotation.
+			if c.stableRS.Status.AvailableReplicas+c.newRS.Status.AvailableReplicas >= *c.rollout.Spec.Replicas {
 				err = c.addScaleDownDelay(c.newRS, *abortScaleDownDelaySeconds)
 				if err != nil {
 					return false, err
--- a/rollout/replicaset_test.go
+++ b/rollout/replicaset_test.go
@@ -347,39 +347,59 @@ func TestReconcileNewReplicaSet(t *testing.T) {
 	}
 }
 
-// TestReconcileNewReplicaSetAbortDynamicStableScale verifies that on abort with
-// dynamicStableScale + abortScaleDownDelaySeconds, the scale-down-deadline annotation
-// is added as soon as combined stable+canary capacity covers spec.Replicas. Prior to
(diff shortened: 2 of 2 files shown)


amit-darji:
Small follow-up in `71fd0ae` — split the gate by mode instead of a single relaxed check. Non-dynamic canary / BlueGreen keeps the original strict `stable.Available == spec.Replicas` behavior; only `dynamicStableScale: true` uses the combined-capacity check. This preserves the "wait for transient stable degradation to heal before draining" semantics for the classic case, since the deadlock is specific to dynamic stable scaling. Updated the PR body and extended the test to cover both modes.

amit-darji:
Closing in favor of #4900 by @zachaller — that PR fixes the deadlock at the root (the `max()` bound in `GetDesiredCanaryWeight`) rather than at the annotation gate, which is a stronger design: the abort weight steps down as stable scales up, and the scale-down deadline only starts once the canary is fully drained of traffic. This preserves the "canary scale-down only after traffic is off it" invariant that this PR's combined-capacity gate did not — under `dynamicStableScale` the combined-capacity condition is satisfied at the moment of abort, so the deadline could fire before stable has ramped, briefly leaving the traffic router pointing 10–90% at a canary of 0 pods.

Thanks to reviewers. Original issue #4898 will be resolved via #4900.

Issue #3412, opened 2024-02-29
Rollout stuck when a scale event happens durring a service switch delay


BrunoTarijon:
Checklist:

* [x] I've included steps to reproduce the bug.
* [x] I've included the version of argo rollouts.

**Describe the bug**

Rollout controller in a infinity loop trying to reconcile a still referenced replicaset.

**To Reproduce**

I made this e2e test, just a easy way to test the problem, need to use a E2E_POD_DELAY=3 for the service switch delay
```
func (s *CanarySuite) TestRolloutScalingWhenChanged() {
	s.Given().
		RolloutObjects(`@functional/alb-canary-rollout.yaml`).
		When().
		ApplyManifests().
		WaitForRolloutStatus("Healthy").
		UpdateSpec().
		WaitForRolloutStatus("Paused").
		UpdateSpec().
                Sleep(2*time.Second). // Should be less than E2E_POD_DELAY
		ScaleRollout(3).
                Sleep(30*time.Second). // Just to stop the execution before delete the rollout
		Then().
		ExpectRevisionPodCount("1", 3).
		ExpectRevisionPodCount("2", 0).
		ExpectRevisionPodCount("3", 1)
}
```

**Expected behavior**

The controller should be able to reconcile the rollout/services.

**Version**

1.6.6

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We prioritize the issues with the most 👍.

BrunoTarijon:
I made a better e2e test in the PR

wizardist:
@BrunoTarijon is this issue accompanied by the following event in the "stable" service Endpoint?

```
Reason: FailedToUpdateEndpoint
Message: Failed to update endpoint ns/ep-name: Operation cannot be fulfilled on endpoints "ep-name": the object has been modified; please apply your changes to the latest version and try again
```

PR #4919, opened 2026-07-24
fix(controller): sync desired-replicas annotation during abort scale-…

zachaller:
…down delay so scaling events terminate

## Problem

  `reconcile()` short-circuits to `syncReplicasOnly()` whenever `isScalingEvent()` is true. `isScalingEvent()` is **level-triggered**: it compares the `desired-replicas` annotation on the new/stable ReplicaSets against `spec.replicas`, so every reconcile — informer event, requeue,
  the 15-minute resync, even a controller restart — takes the same short-circuit until the annotations converge. The design invariant that makes this safe is that `syncReplicasOnly()` itself terminates the scaling event by syncing those annotations.

  The abort scale-down delay breaks that invariant. When `shouldDelayScaleDownOnAbort()` is true (any traffic-routed canary or blue-green rollout that hasn't set `abortScaleDownDelaySeconds: 0`), `reconcileNewReplicaSet()` has two early returns that intentionally hold the newRS at
  scale — but they also skip `scaleReplicaSetAndRecordEvent`, so the newRS `desired-replicas` annotation is never updated.

  Failure mode: a canary is aborting (e.g. failed analysis) and `spec.replicas` changes before the first post-abort traffic reconcile — typically an HPA reacting to the same load problem that failed the analysis. From then on every reconcile short-circuits to `syncReplicasOnly()`,
  which never touches traffic routing or services:

  1. The traffic weight stays frozen at its pre-abort value and the canary service selector is never reset, for the entire stable-scale-up time **plus the full `abortScaleDownDelaySeconds`** (default 30s; users who keep aborted canaries around for debugging set this to minutes or
  hours).
  2. When the deadline finally elapses, the canary RS is scaled to zero **while the mesh is still weighting traffic to it** — briefly black-holing that traffic until the next reconcile (now unblocked) sets the weight to 0.

  Requeueing does not help: the requeued reconcile re-evaluates `isScalingEvent()` from the same stale annotations and short-circuits again (see the discussion in #3413, where re-enqueueing was suggested as the fix for a related symptom).

  ## Fix

  Restore the invariant instead of removing the short-circuit: the two early-return paths in `reconcileNewReplicaSet()` now sync the `desired-replicas` annotation (via the existing annotation-only no-scale behavior of `scaleReplicaSetAndRecordEvent`) while still holding the newRS at
  its current size. The scaling event then terminates within one or two informer-driven cycles, the next reconcile takes the full path, shifts traffic off the canary, and resets the canary service selector — and the canary is never scaled to zero while still receiving weighted
  traffic.

  Compared to dropping the short-circuit when aborted (#4853), this approach:
  - preserves the original design rule that scaling events reconcile nothing but replica counts (see https://github.com/argoproj/argo-rollouts/pull/3413#issuecomment-1984805862), and keeps HPA scale application independent of traffic-router API health;
  - fixes blue-green aborts through the same shared code path without routing them through the full `rolloutBlueGreen()` reconcile;
  - costs one extra reconcile cycle (seconds) before the weight shifts, instead of shifting in the same cycle.

  Related: #3412 / #3413 (same short-circuit, different trigger), #3883 (narrowed `isScalingEvent()` to new/stable RS for stacked rollouts). No filed issue matches this exact failure mode; alternative fix proposed in #4853.

### DIFF
--- a/rollout/replicaset.go
+++ b/rollout/replicaset.go
@@ -129,12 +129,13 @@ func (c *rolloutContext) removeScaleDownDeadlines() error {
 }
 
 // syncNewRSReplicasAnnotation updates the newRS desired-replicas annotation to spec.replicas
-// without changing its scale. The paths that intentionally hold the newRS at size during an
-// abort scale-down delay must still do this: isScalingEvent() treats a stale desired-replicas
-// annotation on the newRS as an in-progress scaling event and short-circuits every reconcile
-// to syncReplicasOnly(), which never reconciles traffic routing or services. Without the
-// annotation sync, a spec.replicas change (e.g. HPA) mid-abort freezes traffic reconciliation
-// until the scale-down deadline elapses.
+// without changing its scale. It is a no-op when the annotation already matches spec.replicas
+// (ReplicasAnnotationsNeedUpdate returns false). The paths that intentionally hold the newRS at
+// size during an abort scale-down delay must still call this: isScalingEvent() treats a stale
+// desired-replicas annotation on the newRS as an in-progress scaling event and short-circuits
+// every reconcile to syncReplicasOnly(), which never reconciles traffic routing or services.
+// Without the annotation sync, a spec.replicas change (e.g. HPA) mid-abort freezes traffic
+// reconciliation until the scale-down deadline elapses.
 func (c *rolloutContext) syncNewRSReplicasAnnotation() error {
 	if !annotations.ReplicasAnnotationsNeedUpdate(c.newRS, defaults.GetReplicasOrDefault(c.rollout.Spec.Replicas)) {
 		return nil
@@ -177,7 +178,7 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 						logCtx.Info("rollout enqueue due to scaleDownDelay")
 						c.enqueueRolloutAfter(c.rollout, remainingTime)
 						if err := c.syncNewRSReplicasAnnotation(); err != nil {
-							return false, err
+							return false, fmt.Errorf("failed to sync newRS desired-replicas annotation while waiting for abort scale-down deadline: %w", err)
 						}
 						return false, nil
 					}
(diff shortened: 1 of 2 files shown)


rollout/replicaset.go
```diff
@@ -156,6 +176,9 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 						logCtx := logutil.WithRollout(c.rollout)
 						logCtx.Info("rollout enqueue due to scaleDownDelay")
 						c.enqueueRolloutAfter(c.rollout, remainingTime)
+						if err := c.syncNewRSReplicasAnnotation(); err != nil {
+							return false, err
```
leoluz: Consider adding context to the returned error if that makes sense

rollout/replicaset.go
```diff
@@ -164,6 +187,11 @@ func (c *rolloutContext) reconcileNewReplicaSet() (bool, error) {
 				}
 			}
 		} else if abortScaleDownDelaySeconds != nil {
+			// Sync annotations before addScaleDownDelay's patch so a subsequent update of the
+			// stale newRS object cannot clobber the scale-down-deadline annotation.
+			if err := c.syncNewRSReplicasAnnotation(); err != nil {
+				return false, err
```
leoluz: Consider adding context to the returned error if that makes sense

rollout/replicaset.go
```diff
@@ -127,6 +128,25 @@ func (c *rolloutContext) removeScaleDownDeadlines() error {
 	return nil
 }
 
+// syncNewRSReplicasAnnotation updates the newRS desired-replicas annotation to spec.replicas
+// without changing its scale. The paths that intentionally hold the newRS at size during an
+// abort scale-down delay must still do this: isScalingEvent() treats a stale desired-replicas
+// annotation on the newRS as an in-progress scaling event and short-circuits every reconcile
+// to syncReplicasOnly(), which never reconciles traffic routing or services. Without the
+// annotation sync, a spec.replicas change (e.g. HPA) mid-abort freezes traffic reconciliation
+// until the scale-down deadline elapses.
```
leoluz: Maybe want to clarify in the docs that the method is a no-op if ReplicasAnnotationsNeedUpdate is false?

leoluz (commented):
Please check my comments.

leoluz (approved):
LGTM

PR #4923, opened 2026-07-27
fix: avoid int32 overflow when maxTrafficWeight is large

semx:
Fixes canary replica counts being computed from an overflowed `int32` product when `maxTrafficWeight` is large.

### The problem

Several places multiply a replica count by a traffic weight while both are still `int32`, and only widen the result afterwards:

```go
ceilWeightedCanaryCount := int32(math.Ceil(float64(specReplicas*desiredWeight) / float64(maxWeight)))
```

`maxTrafficWeight` has no upper bound (the CRD only says `format: int32`), and the repo's own tests use `100000000`. At that scale the product exceeds `math.MaxInt32` for quite ordinary replica counts, and the arithmetic wraps.

`approximateWeightedCanaryStableReplicaCounts` with `maxWeight=100000000`, `maxSurge=0`:

| replicas | weight | canary (before) | canary (expected) |
|---------:|-------:|----------------:|------------------:|
| 43 | 50% | **-21** | 22 |
| 100 | 50% | **8** | 50 |
| 43 | 100% | **1** | 43 |

A negative canary count at 50%, and a single canary replica when the rollout is fully shifted to canary.

`checkReplicasAvailable` is affected the same way: the overflowed product goes negative, so `availableReplicas < desiredReplicas` is never true and it reports that there are enough replicas. With 43 replicas at 50% weight and **zero** of them available it returns `true`, which lets traffic shift to pods that are not ready.

The threshold is `replicas * weight > MaxInt32`, so with `maxTrafficWeight=100000000` it starts at about 22 replicas.

### The fix

Widen the operands before multiplying. `trafficWeightToReplicas` in the same file already does exactly this:

```go
func trafficWeightToReplicas(replicas, weight, maxWeight int32) int32 {
	return int32(math.Ceil(float64(weight) * float64(replicas) / float64(maxWeight)))
}
```

The same shape is applied to `approximateWeightedCanaryStableReplicaCounts`, `weightDelta`, `checkReplicasAvailable`, the dynamic-stable-scale weight in `trafficrouting.go`, and the actual-weight display in `rollout_info.go`. Behaviour for the default `maxTrafficWeight=100` is unchanged — the existing table cases all still pass.

### Tests

Added cases to the existing tables. Both fail on `master` and pass with this change:

- `TestApproximateWeightedNewStableReplicaCounts` — `expected: 22, actual: -21` and `expected: 50, actual: 8`
- `TestCheckReplicasAvailableWithCustomMaxTrafficWeight` — 43 replicas, 0 available, 50% weight must not report available

### DIFF
--- a/pkg/kubectl-argo-rollouts/info/rollout_info.go
+++ b/pkg/kubectl-argo-rollouts/info/rollout_info.go
@@ -65,7 +65,7 @@ func NewRolloutInfo(
 			if ro.Spec.Strategy.Canary.TrafficRouting == nil {
 				for _, rs := range roInfo.ReplicaSets {
 					if rs.Canary {
-						roInfo.ActualWeight = fmt.Sprintf("%d", (rs.Available*weightutil.MaxTrafficWeight(ro))/ro.Status.AvailableReplicas)
+						roInfo.ActualWeight = fmt.Sprintf("%d", int64(rs.Available)*int64(weightutil.MaxTrafficWeight(ro))/int64(ro.Status.AvailableReplicas))
 					}
 				}
 			} else {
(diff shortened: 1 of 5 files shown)


kostis-codefresh (approved):
Thank you!

PR #4924, opened 2026-07-27
fix: reject maxTrafficWeight below 1

semx:
`maxTrafficWeight` is used as a divisor, but nothing stops it from being zero.

### The problem

The CRD puts no bound on the field:

```yaml
maxTrafficWeight:
  description: MaxTrafficWeight The total weight of traffic. If unspecified, it defaults to 100
  format: int32
  type: integer
```

Validation checks that each `setWeight` is between 0 and `maxTrafficWeight`, but never checks `maxTrafficWeight` itself. With `maxTrafficWeight: 0` a `setWeight: 0` step still passes, and `checkReplicasAvailable` divides by it:

```go
desiredReplicas := (desiredWeight * totalReplicas) / weightutil.MaxTrafficWeight(c.rollout)
```

which panics:

```
panic: runtime error: integer divide by zero
    github.com/argoproj/argo-rollouts/rollout.(*rolloutContext).checkReplicasAvailable
```

The panic is recovered by the workqueue, but the reconcile for that rollout fails each time it is retried.

### The fix

Reject values below 1 during validation, next to the existing `maxTrafficWeight` check, so the user gets an `InvalidSpec` condition explaining the problem. `getRolloutValidationErrors` runs before the reconcile touches the traffic weights, so the division is no longer reachable with a zero total weight.

Negative values are rejected too — they produce negative replica counts rather than a panic.

### Test

Added a case to `TestValidateRolloutStrategyCanary` covering `0` and `-1`. It fails without the check and passes with it.

### DIFF
--- a/pkg/apis/rollouts/validation/validation.go
+++ b/pkg/apis/rollouts/validation/validation.go
@@ -30,6 +30,8 @@ const (
 	MissingFieldMessage = "Rollout has missing field '%s'"
 	// InvalidSetWeightMessage indicates the setweight value needs to be between 0 and max weight
 	InvalidSetWeightMessage = "SetWeight needs to be between 0 and %d"
+	// InvalidMaxTrafficWeightMessage indicates the maxTrafficWeight value is not a usable total weight
+	InvalidMaxTrafficWeightMessage = "MaxTrafficWeight needs to be greater than 0"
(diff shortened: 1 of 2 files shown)


kostis-codefresh (approved):
Thank you!

PR #4925, opened 2026-07-27
fix(controller): honor scaleDownDelaySeconds when dynamicStableScale is enabled for traffic-routed rollouts

pragneshjadhav-zepto:
### Summary
When using a traffic-routed canary strategy with dynamicStableScale: true, Argo Rollouts dynamically scales the stable ReplicaSet down as traffic shifts to the canary. However, scaleDownDelaySeconds was not honored for that stable RS during step progression. The delay only applied to old ReplicaSets after full promotion (via scaleDownDelayHelper in scaleDownOldReplicaSetsForCanary), not to the current stable RS while the rollout was in progress.
Additionally, validation rejected rollouts that combined dynamicStableScale: true with scaleDownDelaySeconds, even though both require traffic routing and are logically compatible.

This caused immediate stable scale-down during rollout - After each setWeight step, the stable RS was scaled down immediately based on CalculateReplicaCountsForTrafficRoutedCanary, with no grace period for connection draining.

The fix introduces isDynamicScaleDownDelayEnabled() and isStableScaleDownDelayed() to reuse the existing scale-down-deadline annotation mechanism for the current stable RS during an active canary rollout, and allows the two spec fields to be used together.

### Root Cause:
On traffic-routed canary rollouts with ALB (and similar controllers), target group weight changes are not instant. There is a propagation delay between when Argo Rollouts updates ingress weights and when the load balancer actually shifts traffic.

When dynamicStableScale: true is enabled without honoring scaleDownDelaySeconds on the current stable ReplicaSet:

After a setWeight step, the controller reduces stable replica count immediately (via CalculateReplicaCountsForTrafficRoutedCanary).
ALB may still send a large share of traffic to the stable service/target group.
With fewer stable pods and unchanged (or lagging) ALB weights, high-throughput services see connection saturation, timeouts, and HTTP 503 responses from the stable RS.
This is worse when:

Stable is scaled down on setCanaryScale before traffic has moved (setWeight not yet applied).
Stable pods are terminated before ALB finishes propagating the new weights.
There is no grace period (scaleDownDelaySeconds) for draining in-flight connections on the stable RS during the rollout.

### Fix:

1. **Introduce isDynamicScaleDownDelayEnabled()**
    Returns true when all of the following hold:
    - canary.dynamicStableScale: true
    - canary.trafficRouting is set
    - canary.scaleDownDelaySeconds is set
    - Rollout is not fully promoted
    - Rollout is not aborted
    - status.promoteFull is false 

    File: [rollout/replicaset.go](https://github.com/argoproj/argo-rollouts/blob/master/rollout/replicaset.go)

2. **Allow dynamicStableScale + scaleDownDelaySeconds in validation**
    Remove the mutual-exclusion check so both fields are valid when trafficRouting is configured.
    File: [pkg/apis/rollouts/validation/validation.go](https://github.com/argoproj/argo-rollouts/blob/master/pkg/apis/rollouts/validation/validation.go)

3. **Delay stable RS scale-down during step progression - isStableScaleDownDelayed()**
In reconcileCanaryStableReplicaSet():
- Compute desired stable count via CalculateReplicaCountsForTrafficRoutedCanary using status.canary.weights (lagging weights), so stable only scales down after traffic has shifted (setWeight), not on setCanaryScale.
- If isDynamicScaleDownDelayEnabled() and stable needs to scale down, call isStableScaleDownDelayed():
  - Stamp scale-down-deadline on stable RS (reuse addScaleDownDelay)
  - Skip scaling until deadline elapses
  - Re-enqueue rollout via enqueueRolloutAfter

  File: [rollout/canary.go](https://github.com/argoproj/argo-rollouts/blob/master/rollout/canary.go)

4. **Preserve stable RS deadline annotation mid-rollout**
In removeScaleDownDeadlines(), when isDynamicScaleDownDelayEnabled() is true, do not remove the scale-down-deadline annotation from the stable RS (otherwise the delay resets every reconcile).

    File: [rollout/replicaset.go](https://github.com/argoproj/argo-rollouts/blob/master/rollout/replicaset.go)

5. **Honor delay for old RS after promotion**
In scaleDownOldReplicaSetsForCanary(), when dynamicStableScale: true and scaleDownDelaySeconds is set, use scaleDownDelayHelper instead of immediately setting desiredReplicaCount = 0.

    File: [rollout/canary.go](https://github.com/argoproj/argo-rollouts/blob/master/rollout/canary.go)


kostis-codefresh:
Isn't this same as https://github.com/argoproj/argo-rollouts/pull/4645 ?

kostis-codefresh:
Will this be solved if https://github.com/argoproj/argo-rollouts/pull/4645 is merged?

kostis-codefresh (commented):
I haven't looked at this PR in detail yet, but I did review 4645 which came earlier. Please see my comment there regarding dropping the validation for both fields https://github.com/argoproj/argo-rollouts/pull/4645#pullrequestreview-4883855166

ajithshah (approved):

PR #4931, opened 2026-07-29
fix: bring observedGeneration and workloadGeneration to match k8s type convention

### LINKED ISSUE #3402: status.observedGeneration field violates type convention
Checklist:

* [ ] I've included steps to reproduce the bug.
* [ ] I've included the version of argo rollouts.

**Describe the bug**

Interactions with the status.observedGeneration field on Rollouts behave unexpectedly because the field is forced to be a string rather than an integer on the CRD.  This leads to incompatibilities with tooling which observe that field to ensure state of an object.  This is particularly frustrating when trying to use Rollouts resources with FluxCD.

**To Reproduce**

1. Create a Rollout
2. Get the Rollout from the API
3. Observe that the value of status.observedGeneration is in quotes

**Expected behavior**

The status.observedGeneration field should be an integer in the CRD and represented internally in the controller code as an int64, matching convention of upstream.

**Screenshots**

N/A

**Version**

1.6.6

**Logs**

No relevant logs.

---

gecube:
Fixes #3402. This rebases and completes @brito-rafa's #4384 onto current master (it had gone stale with merge conflicts); credit for the original change goes to @brito-rafa.

## What

`Rollout.status.observedGeneration` and `status.workloadObservedGeneration` change from `string` to `int64`, matching the Kubernetes API convention. The stored value has already been numeric (`strconv.Itoa(int(rollout.Generation))`) since v0.10, so this is purely a type/serialization fix — not a semantic one.

## Why

Anything that follows the convention of comparing `status.observedGeneration` (int64) against `metadata.generation` breaks on the quoted string. FluxCD is the common case: kstatus (`fluxcd/cli-utils`) returns an error instead of a status —

```
.status.observedGeneration accessor error: 1 is of the type string, expected int64
```

so Kustomizations/HelmReleases that wrap Rollouts can never reach a health verdict. A full reproduction (with the live-cluster before/after and the exact library Flux uses) is posted on #3402.

## Changes

- `observedGeneration` / `workloadObservedGeneration` → `int64` in the API type, CRD, protobuf, openapi and swagger (regenerated).
- Internal consumers updated to compare/assign `int64` directly (`rollout/sync.go`, `utils/conditions`, `utils/rollout`, `utils/annotations`, `rollout/controller.go`, `test/fixtures`).
- New API-type tests that fail if the field ever regresses to a string: JSON encodes an unquoted integer (the property kstatus relies on), `omitempty` is honoured, and the value round-trips through protobuf.

## Compatibility

- No value-format change (already numeric since v0.10) — only the JSON/protobuf encoding.
- Existing objects migrate automatically: the upgraded controller rewrites `observedGeneration` to an integer on the first reconcile of each Rollout.
- Protobuf fields 13/24 change wire type `bytes → varint` (and `RolloutInfo.generation` likewise). That is the internal gRPC API between the controller and the `kubectl argo rollouts` plugin/dashboard, so plugin and controller should be upgraded together.

## Testing

- `go build ./...` clean.
- Full `./rollout/` unit suite green; `utils/{rollout,conditions,annotations}` green; new `pkg/apis/rollouts/v1alpha1` tests green.
- End-to-end on a live cluster: fresh Rollouts serialize `observedGeneration: <int>`, the typed client reads them, and `kstatus.Compute` returns `Current` (it errored before the fix).

---

Checklist:

* [x] This is a bug fix.
* [x] The PR title is [conventional](https://www.conventionalcommits.org/en/v1.0.0/).
* [x] I've signed my commits with DCO.
* [x] My builds are green.


kostis-codefresh:
@gecube Could you please look at the conflicts?

PR #4889, opened 2026-07-17
fix(dashboard): handle empty Rollouts list

0xdsqr:
Checklist:

 * [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix,
  or (c) this is a chore.
  * [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-
  rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number.
  * [x] I've signed my commits with [DCO](https://github.com/argoproj/argoproj/blob/main/community/CONTRIBUTING.md#legal).
  * [x] My builds are green.
  * [x] I have written unit and/or e2e tests for my change.
  * [x] I have run all relevant UI tests locally, including the complete 105-test suite.
  * [x] I have used LLM/AI/Agent tools for this PR but I am responsible for all code of this PR.
  * [x] I understand what the code does and WHY/HOW it works in several scenarios.
  * [x] I know this changes existing dashboard behavior for empty and failed Rollout-list requests.
  * [ ] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

Fixes #4888

PR #4939, opened 2026-08-01
fix: clear loading state when namespace has zero rollouts

### LINKED ISSUE #4888: Dashboard remains in Loading state when a namespace has zero Rollouts
Checklist:

**Describe the bug**

The Rollouts dashboard remains on `Loading...` indefinitely when the selected namespace contains zero Rollout resources, even though the initial list request succeeds.

The page already has a `No Rollouts to display!` empty state, but it is unreachable on an initially empty namespace. `useRollouts()` initializes and resolves to `[]`; `useLoading()` only clears loading when `list.length > 0`. The watch endpoint has no rollout events to emit in an empty namespace, so nothing can clear the spinner.

This is the same user-visible symptom reported in #3618, which was closed without a code fix. A later comment on that issue identified creating the first Rollout as the workaround. The condition is still present in v1.9.0 and current `master`.

**To Reproduce**

1. Install Argo Rollouts with the dashboard enabled, or run `kubectl argo rollouts dashboard`.
2. Open the dashboard in a namespace containing no `Rollout` resources.
3. Observe that `GET /api/v1/rollouts/<namespace>/info` succeeds with an empty result and the watch request remains open.
4. Observe that the UI displays `Loading...` indefinitely.
5. Create a Rollout in that namespace; the list then appears and loading clears.

**Expected behavior**

After the successful initial list request returns zero items, loading should finish and the existing `No Rollouts to display!` empty state should render.

If the initial list request fails, the UI should leave the loading state and show an actionable error rather than looking like it is still fetching.

**Screenshots**

Image

**Version**

Argo Rollouts v1.9.0, installed by Helm chart `argo-rollouts` 2.41.0.

Also confirmed by source inspection on current `master` at `9273b0ee05dbb629cc72fb9d343c08356c8379c8`.

**Logs**

```text
No controller or dashboard error is required to reproduce this. The initial list request succeeds and the namespace simply has zero Rollouts.
```

---

AruneshDwivedi:
useLoading() only clears loading when list.length > 0, so a namespace with no rollouts keeps the spinner forever since the watch endpoint has nothing to emit. The initial list request succeeds and returns zero items, but the UI stays stuck on Loading... indefinitely.

Fix: track whether the initial fetch completed and clear loading once it does, regardless of item count. The existing empty state will then render correctly.

Fixes #4888

### DIFF
--- a/ui/src/app/components/rollouts-home/rollouts-home.tsx
+++ b/ui/src/app/components/rollouts-home/rollouts-home.tsx
@@ -144,7 +144,7 @@ export const RolloutsHome = () => {
                         <FontAwesomeIcon icon={faCircleNotch} spin={true} style={{marginRight: '5px'}} />
                         Loading...
                     </div>
-                ) : (rollouts || []).length > 0 ? (
+                ) : (loading === false && (rollouts || []).length === 0) ? (
                     <React.Fragment>
                         {filters.displayMode === 'table' && <RolloutsTable rollouts={filteredRollouts} onFavoriteChange={handleFavoriteChange} favorites={favorites} />}
                         {filters.displayMode !== 'table' && <RolloutsGrid rollouts={filteredRollouts} onFavoriteChange={handleFavoriteChange} favorites={favorites} />}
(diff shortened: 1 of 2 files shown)


kostis-codefresh:
What about https://github.com/argoproj/argo-rollouts/pull/4889 ?

PR #4940, opened 2026-08-02
fix(controller): re-drain workloadRef Deployment on later revisions. Fixes #4935

### LINKED ISSUE #4935: workloadRef scaleDown=progressively never recovers if the Deployment is scaled back up after revision 1 (e.g. by a GitOps sync) — permanent double capacity
### Summary

`spec.workloadRef.scaleDown: progressively` only ever evaluates on **revision 1** of the Rollout. If anything scales the referenced Deployment back up after that — most commonly a GitOps controller (Argo CD, Flux, `helm upgrade`) re-applying the Deployment manifest with its original `replicas` value — the controller never scales it down again. The result is the Rollout's ReplicaSets **and** the Deployment's ReplicaSet running at full size indefinitely, with no recovery path short of deleting the Rollout and re-migrating (or manually patching the Deployment).

This is the `progressively` sibling of #4681 (which reports the `onsuccess` variant), combined with the GitOps interplay described in discussion #4111 (no resolution there). The code acknowledges the limitation — on current `master`, `rollout/replicaset.go` carries this NOTE:

> ```
> // NOTE: Progressive deployment scaling is designed for one-time migration from Deployment to Rollout.
> // It only triggers on revision == 1, which has a limitation: if the initial rollout (revision 1) fails
> // and users fix the issue causing a new revision (revision 2+), the deployment will never be scaled down.
> // This matches the behavior of scaleDown: onSuccess and may need to be addressed in a future enhancement.
> ```

but the NOTE understates the impact: it describes the "revision 1 failed" case, while the same gate also makes an *already-successful* migration silently un-migrate on the next sync.

### Versions

- Observed on argo-rollouts **v1.8.3** — the gate is at [`rollout/replicaset.go:174` (v1.8.3)](https://github.com/argoproj/argo-rollouts/blob/v1.8.3/rollout/replicaset.go#L174), still present on [`master` (replicaset.go:220)](https://github.com/argoproj/argo-rollouts/blob/master/rollout/replicaset.go).
- Argo CD managing the manifests, `kubectl` server-side state inspected via managed fields.

### Reproduction

1. Deployment with `replicas: 2` managed by a GitOps tool (manifest in git declares `replicas: 2`).
2. Add a Rollout with `workloadRef` referencing it, `scaleDown: progressively`, canary strategy with Istio traffic routing.
3. Revision 1 works exactly as documented: Rollout scales to 2, controller drains the Deployment to 0. Managed fields on the Deployment show `spec.replicas` owned by `rollouts-controller`.
4. Change the pod template (e.g. image tag) in git and sync. The GitOps tool applies the full Deployment manifest — including `replicas: 2` — because ignore-differences configuration on the diff side does not affect the apply (Argo CD requires the separate `RespectIgnoreDifferences=true` sync option for that). `spec.replicas` field ownership transfers back to the GitOps controller.
5. The Rollout runs its revision-2 canary and completes fine — but the Deployment stays at 2 forever. Final state: Rollout stable RS = 2 pods, Deployment RS = 2 pods, for a service sized at 2.

Step 4/5 timeline verified via managed fields: after the sync, `f:spec/f:replicas` manager = the GitOps controller; the rollouts controller never touches it again (revision now ≥ 2, so the gate at replicaset.go never matches).

### Why this is worse than the NOTE suggests

- The Deployment pods carry no `rollouts-pod-template-hash` label, so they match **neither** DestinationRule subset when using Istio traffic routing — but they still match the Service selector, so any traffic reaching the ClusterIP outside the mesh routing can land on them.
- Nothing alerts: the Rollout reports Healthy, the GitOps app reports Synced, the pods are Ready. The only symptom is doubled cost.
- Recovery requires either deleting the Rollout and re-migrating (a new revision-1) or manually patching the Deployment's replicas.

### Suggested direction

Either of these would resolve it:

1. Re-evaluate `scaleDown` on every reconcile while `workloadRef.scaleDown` is set (i.e. treat a scaled-up referenced Deployment as drift the controller owns), rather than gating on `revision == 1`; or
2. If one-shot semantics are intended, document them prominently on the [migrating page](https://argo-rollouts.readthedocs.io/en/stable/migrating/) — specifically that GitOps users MUST configure their tool to stop applying the Deployment's `replicas` (for Argo CD: `ignoreDifferences` **plus** `RespectIgnoreDifferences=true`) **before** creating the Rollout, because there is exactly one chance for the drain and no recovery afterwards.

Happy to provide more detail from the reproduction if useful.

michaelgriskified:
Fixes #4935.

### The bug

With `spec.workloadRef.scaleDown: progressively`, the drain of the referenced Deployment was gated on `revision == 1`. If anything scales the Deployment back up after the initial migration, nothing ever drains it again:

1. Revision 1 works as documented — Rollout scales to N, controller drains the Deployment to 0.
2. A GitOps controller re-applies the Deployment manifest (which still declares `replicas: N` in git) on the next sync, or an operator scales it by hand. `spec.replicas` field ownership moves away from the rollouts controller.
3. Every later revision reconciles with `revision >= 2`, so the gate never matches. Rollout stable RS at N **and** Deployment RS at N, permanently, for a service sized at N.

Nothing reports unhealthy: the Rollout is Healthy, the pods are Ready, the GitOps app is Synced. The only symptom is doubled capacity. Recovery requires deleting the Rollout and re-migrating, or patching the Deployment by hand.

For Argo CD users specifically, `ignoreDifferences` on the Deployment's `replicas` is not enough — it only suppresses the diff, while the apply still writes git's value. Avoiding step 2 requires the separate `RespectIgnoreDifferences=true` sync option (or omitting `replicas` from the manifest under server-side apply), and there is no in-cluster signal that you got it wrong until you notice the bill.

### The change

Drain to 0 whenever the Rollout is `Healthy`, on any revision, instead of only on revision 1. The controller then converges on the documented end state (`as the Rollout is scaled up the Deployment is scaled down`) rather than getting exactly one chance at it. This also resolves the case the removed `NOTE` described — a failed revision 1 whose fix lands in revision 2+.

The availability branch (scaling the Deployment to `spec.replicas - status.readyReplicas` while the Rollout is not yet Healthy) stays scoped to `revision == 1`. That branch exists to hold total capacity steady during the initial handover; on a later revision the stable ReplicaSet already serves that capacity, so running it there would *add* capacity — and would resurrect the Deployment mid-canary on every update. There's a new test case pinning that.

`scaleDeployment` already no-ops when the Deployment is at the target replica count, so a steady-state Healthy Rollout costs one extra GET per reconcile and no writes — the same cost revision-1 Rollouts already pay today.

### Behavior change for existing users

This changes existing behavior, deliberately: a Deployment currently sitting at full size behind a Healthy Rollout with `scaleDown: progressively` will be scaled to 0 on the first reconcile after upgrade. That is the documented intent of the setting, and for anyone in the state described above it is the recovery they'd otherwise have to perform by hand. Worth a release note.

The pre-existing unit case named `Do not scale deployment` asserted the old gate (revision 2 + Healthy + Deployment at 5 → stays 5); it's been renamed and flipped to expect 0, which is the crux of the fix.

### Tests

`rollout/replicaset_test.go`, in `TestScaleDownProgressively`:
- `Rollout healthy on a later revision - Deployment scaled to 0` — the regression. Verified it fails on unmodified `master` (`expected: 0, actual: 5`) and passes with the change.
- `Rollout progressing on a later revision - Deployment untouched` — guards the scoping of the availability branch, so this fix can't turn into a mid-canary scale-up.

`go test ./rollout/...` passes locally, as does `go build ./...`.

I did not add an e2e case: reproducing this needs an external actor scaling the referenced Deployment back up between revisions, and I'd rather not push e2e code I can't run end-to-end here. The existing `TestScaleDownProgressivelyCompleted*` suites pass unchanged. Happy to add one if you'd like it, or to hand over the exact repro.

### Checklist


kostis-codefresh (commented):
The code looks correct to me.  Two additional questions

1) I see we have only unit tests in the PR. Did you also try this PR locally with a real workload to verify it indeed works in a full end-to-end scenario?

2) Regarding backwards compatibility what is the worst case scenario here. That an existing user has 30 replicas for Rollout and 30 replicas for deployment, and if they upgrade to Argo Rollouts they will have only the 30 Rollout replicas and the deployment wll go to zero? Do I get that correctly?

michaelgriskified:
Thanks for the review @kostis-codefresh.

**1) End-to-end test** — yes. Ran it today on a real cluster (kind), with the controller built from this branch and from current `master`, running out-of-cluster against the *same* cluster state. Workload: Deployment `web` with `replicas: 3`, Rollout with `workloadRef` + `scaleDown: progressively`, canary `setWeight: 50` → `pause: 15s`, no traffic router (the bug doesn't depend on one). The "GitOps sync" is a `kubectl apply` of the full Deployment manifest, which is what Argo CD / Flux do on sync.

`master` controller:

| step | Rollout | Deployment |
|---|---|---|
| create Rollout (rev 1) | Healthy 3/3 | drained to 0 |
| re-apply manifest (`replicas: 3`) while still on rev 1 | Healthy | re-drained to 0 (rev==1 gate still matches) |
| apply new image + `replicas: 3` → rev 2 canary | Paused → Healthy 3/3 | **stays at 3, forever** — bug reproduced |

Then stopped the `master` controller and started the one built from this PR against that exact state (Rollout 3/3 + Deployment 3/3), i.e. the upgrade scenario from your second question:

| step | Rollout | Deployment |
|---|---|---|
| PR controller starts | Healthy 3/3 | drained 3 → 0 on the first reconcile (~3s) |
| re-apply manifest (`replicas: 3`) while Healthy on rev 2 | Healthy | re-drained to 0 within seconds |
| apply new image + `replicas: 3` → rev 3 canary | Paused | **stays at 3 for the whole canary** (availability branch not run on rev ≥ 2, as intended) |
| canary completes | Healthy 3/3 | drained to 0 |

Controller log, the three drains:
```
Scaling deployment web to 0 replicas  namespace=demo rollout=web   # upgrade
Scaling deployment web to 0 replicas  namespace=demo rollout=web   # re-apply while Healthy
Scaling deployment web to 0 replicas  namespace=demo rollout=web   # rev 3 reaches Healthy
```
Managed fields on `spec.replicas` flip from `kubectl-client-side-apply` back to the rollouts controller on each drain.

**2) Backwards compatibility** — yes, you have it exactly right. Worst case: a Healthy Rollout with `scaleDown: progressively` whose referenced Deployment is (still, or again) at 30 replicas; after upgrading, the first reconcile takes the Deployment 30 → 0 in one step and the 30 Rollout replicas remain. What bounds this:

- It only fires while the Rollout is `Healthy`, i.e. all of the Rollout's own replicas are ready. Capacity goes from 2×N to the N the Rollout declares, never below it.
- It's the same scale-to-0 the controller already performs on revision 1 today, and the documented end state of the setting. Anyone sitting at 30+30 got there through the gap this PR closes, and the manual fix they'd otherwise do is the same `scale --replicas=0`.
- Worth a line in the release note: without a traffic router the Deployment's pods match the Service selector and have been silently taking a share of traffic; after the drain the Rollout's N pods carry all of it — which is what N was sized for, but per-pod load goes up. With Istio & co. the Deployment pods are in no subset, so traffic doesn't change.
- Opt-out for anyone who genuinely wants the Deployment kept alive next to the Rollout: drop `scaleDown` (defaults to `never`) before upgrading.

Revision-1 behaviour is unchanged (the `master` run above already re-drains on rev 1). Still happy to add an e2e case that scales the Deployment back up between revisions if you want it in the suite.

kostis-codefresh (approved):
Great. Thanks for the clarifications.

michaelgriskified:
hey @zachaller do you need any additional clarifications?

j809:
Is there anything blocking the merge? This a fix we will appreciate to have.

PR #4957, opened 2026-08-06
fix(dashboard): prevent rollout mutation APIs from hanging. Fixes #4956

### LINKED ISSUE #4956: Dashboard undo and set image APIs hang after successful mutations
Checklist:

**Describe the bug**

The dashboard undo API does not return a response after a successful rollback. The Rollout is rolled back in the cluster, but the HTTP request remains pending indefinitely.

`UndoRollout` runs `undo.RunUndoRollout` successfully and then calls `getRollout` to return the updated Rollout. `getRollout` creates a new informer factory and waits for its cache to sync, but it never starts the informer. As a result, `HasSynced` never becomes true. The wait also uses the server-level `stopCh` instead of the request context, so canceling the request does not stop the wait.

`SetRolloutImage` calls the same `getRollout` helper after updating the image and is affected by the same issue.

This appears to be a regression of #2259. PR #2517 added `WaitForCacheSync` to avoid returning `NotFound` from an empty cache, but the informer was not started before waiting for the sync.

**To Reproduce**

1. Start the Argo Rollouts dashboard.
2. Open a Rollout with more than one revision.
3. Roll back to an earlier revision from the dashboard.
4. Inspect the undo request in the browser network panel.
5. Observe that the Rollout is updated, but the undo request remains pending.

The same behavior can be observed by calling the undo HTTP API directly: the rollback succeeds, but the request does not complete unless the client applies a timeout.

**Expected behavior**

The undo and set image APIs should return the updated Rollout after the mutation succeeds. Request cancellation should also be propagated to the follow-up read.

**Version**

Confirmed in:

* v1.9.1
* v1.10.0-rc1
* master at `17694743c377da83427f4ffb09620027054a23c6`

**Logs**

There is no error in the controller logs because the mutation itself succeeds. The dashboard API handler remains blocked while waiting for an informer cache that was never started.

---

sienna011022:
## Summary

* replace the per-request informer cache in `getRollout` with a direct typed client `Get`
* propagate the request context from `UndoRollout` and `SetRolloutImage` to the follow-up read
* add regression tests covering `getRollout` and both affected handlers

## Problem

`getRollout` creates an informer factory and immediately calls `WaitForCacheSync`, but the informer is never started. The cache therefore cannot sync, leaving successful undo and set image requests waiting indefinitely while trying to build their responses.

A one-off read after a mutation does not need a new informer cache. Using the typed client directly also follows the existing single-resource read pattern used elsewhere in the repository and allows request cancellation to propagate.

Fixes #4956

## Testing

* `go test ./server -count=1`
* `go vet ./server`
* `go test ./server -run 'Test(GetRollout|SetRolloutImageReturnsRollout|UndoRolloutReturnsRollout)$' -count=10`

Checklist:

* [x] Either (a) I've created an enhancement proposal and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is conventional, states what changed, and suffixes the related issue number.
* [x] I've signed my commits with DCO.
* [x] My targeted builds and tests are green.
* [x] I have run all repository tests locally (including the flaky ones) and they pass on my workstation.
* [x] This fixes existing dashboard API behavior and does not add new functionality.
* [ ] My organization is added to `USERS.md`.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -387,11 +387,8 @@ func (s *ArgoRolloutsServer) AbortRollout(ctx context.Context, q *rollout.AbortR
 	return abort.AbortRollout(rolloutIf, q.GetName())
 }
 
-func (s *ArgoRolloutsServer) getRollout(namespace string, name string) (*v1alpha1.Rollout, error) {
-	rolloutsInformerFactory := rolloutinformers.NewSharedInformerFactoryWithOptions(s.Options.RolloutsClientset, 0, rolloutinformers.WithNamespace(namespace))
-	cache.WaitForCacheSync(s.stopCh, rolloutsInformerFactory.Argoproj().V1alpha1().Rollouts().Informer().HasSynced)
-	rolloutsLister := rolloutsInformerFactory.Argoproj().V1alpha1().Rollouts().Lister().Rollouts(namespace)
-	return rolloutsLister.Get(name)
+func (s *ArgoRolloutsServer) getRollout(ctx context.Context, namespace string, name string) (*v1alpha1.Rollout, error) {
+	return s.Options.RolloutsClientset.ArgoprojV1alpha1().Rollouts(namespace).Get(ctx, name, v1.GetOptions{})
 }
 
 func (s *ArgoRolloutsServer) SetRolloutImage(ctx context.Context, q *rollout.SetImageRequest) (*v1alpha1.Rollout, error) {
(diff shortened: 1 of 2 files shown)


kostis-codefresh (commented):
Could you post please a short recording/gif of how this PR fixes the problem?

A before/after UI recording would be great so that we can confirm visually that the fix works.

sienna011022:
@kostis-codefresh Thank you for the comment

argo-4957-devtools

 Same steps in both halves — open a Rollout, click `Rollback` on one revision, Network panel open. 

Only the dashboard binary differs: first half is `master`and second half is this PR.

  | | `master` | this PR |
  |---|---|---|
  | `PUT .../rollouts-demo/undo/5` | stays **`(pending)`** forever | **200 OK**, 50 ms |
  | Rollout in the cluster | rolled back | rolled back |

kostis-codefresh (approved):
Great. Thanks for the fix and the recording!

PR #4964, opened 2026-08-09
feat(metricprovider): support self-signed certs for prometheus provider

abhyudayareddy:
## Problem

The prometheus metric provider only supports skipping TLS verification entirely via `insecure: true`. There is no way to connect to a prometheus server that presents a certificate signed by a private/self-signed CA while keeping certificate verification enabled. Users who run their own CA (common for in-cluster prometheus, e.g. via OpenShift monitoring or self-managed clusters) are forced to disable TLS verification altogether to work around this, which is undesirable.

Relates to #2298.

## Root cause

`NewPrometheusAPI` in `metricproviders/prometheus/prometheus.go` only ever chose between two fixed `http.Transport`s: `secureTransport` (default Go root CA trust store, full verification) and `insecureTransport` (`InsecureSkipVerify: true`). There was no option to trust an additional/custom CA while still validating the server certificate against it.

## Fix

* Added a new optional `caCert` field to `PrometheusMetric` (`pkg/apis/rollouts/v1alpha1/analysis_types.go`), holding a PEM-encoded CA certificate bundle.
* Added `newHTTPTransportWithCACert` in `metricproviders/prometheus/prometheus.go`, which parses the PEM bundle into an `x509.CertPool` and builds a transport with `TLSClientConfig.RootCAs` set to that pool (verification remains enabled, unlike `insecure: true`).
* `NewPrometheusAPI` now uses this transport when `caCert` is set (and `insecure` is not also set — `insecure` still takes precedence, matching previous behavior of skipping verification when set).
* Regenerated all code-generated artifacts affected by the new API field: `generated.pb.go`/`generated.proto` (protobuf), `openapi_generated.go`, and the CRD manifests (`manifests/crds/*.yaml`, `manifests/install.yaml`, `docs/features/kustomize/rollout_cr_schema.json`) via `make gen-proto`, `make gen-openapi`, `make gen-crd`, and `make manifests`.
* Documented the new option in `docs/analysis/prometheus.md`, alongside the existing "Skip TLS verification" section.

## Example

```yaml
provider:
  prometheus:
    address: https://prometheus.example.com
    caCert: |
      -----BEGIN CERTIFICATE-----
      MIIDXTCCAkWgAwIBAgIJAJC1...
      -----END CERTIFICATE-----
    query: ...
```

The certificate can be sourced from a Kubernetes Secret via the existing `valueFrom.secretKeyRef` support for AnalysisTemplate arguments.

## Testing

* Added unit tests in `metricproviders/prometheus/prometheus_test.go`:
  * `TestNewPrometheusAPIWithCACert` / `TestNewPrometheusAPIWithInvalidCACert` — valid/invalid `caCert` handling in `NewPrometheusAPI`.
  * `TestNewHTTPTransportWithCACert` / `TestNewHTTPTransportWithInvalidCACert` — transport construction and PEM parsing.
  * `TestRunSuccessfullyWithCACert` — end-to-end test using `httptest.NewTLSServer` with a self-signed certificate; configures `caCert` to the server's own certificate and asserts the query succeeds (proving TLS verification actually passes against the CA).
  * `TestRunErrorWithWrongCACert` — same TLS server, but with an unrelated CA cert, asserting the connection is rejected (proving verification is not silently bypassed).
* `go build ./...` — passes.
* `go vet ./metricproviders/... ./pkg/apis/...` — passes.
* `go test ./...` — all 84 packages pass, no failures.
* `golangci-lint run ./metricproviders/prometheus/... ./pkg/apis/rollouts/v1alpha1/...` — 0 issues.
* Verified `git diff --exit-code` after codegen only touches the expected files (no unrelated churn beyond the standard fully-rewritten gzipped proto file descriptor blob that any type change causes).

## Checklist

- [x] This is a feature addition scoped to a single provider; no proposal needed per CONTRIBUTING.md guidance for changes of this size.
- [x] PR title is conventional and references the related issue.
- [x] Commits are signed off (DCO).
- [x] Unit tests added/passing; full test suite passes locally.


abhyudayareddy:
cc @argoproj/argo-rollouts-approvers — this is ready for review whenever you have a chance; CI is green. Thanks!

abhyudayareddy:
Checking back in on this one — all checks are still green (CodeQL, unit tests, e2e across 1.32-1.35, SonarCloud). Let me know if there's anything I can do to help move it forward. Thanks!

abhyudayareddy:
Another gentle nudge on this one — it's been ~3 weeks with no review. All 27 checks are still green (CodeQL, unit, e2e across k8s 1.32–1.35, SonarCloud), no conflicts with master. It's a small, self-contained change (self-signed cert support for the Prometheus metric provider, opt-in via config). Happy to rebase or adjust anything. Thanks!

metricproviders/prometheus/prometheus.go
```diff
@@ -248,9 +263,17 @@ func NewPrometheusAPI(metric v1alpha1.Metric) (v1.API, error) {
 	}
 
 	var roundTripper http.RoundTripper
-	if metric.Provider.Prometheus.Insecure {
+	switch {
+	case metric.Provider.Prometheus.Insecure:
 		roundTripper = insecureTransport
-	} else {
+	case metric.Provider.Prometheus.CACert != "":
+		caCertTransport, err := newHTTPTransportWithCACert(metric.Provider.Prometheus.CACert)
```
kostis-codefresh, line 289: secure and insecure transport are created only once at lines 226, 227

would it make sense to follow the same pattern with CaCertTransport as well? Or is there a specific reason we create it from scratch everytime here?

docs/analysis/prometheus.md
```diff
@@ -225,3 +225,28 @@ provider:
...
+    query: |
+      sum(irate(
+        istio_requests_total{reporter="source",destination_service=~"{{args.service-name}}",response_code!~"5.*"}[5m]
+      )) /
+      sum(irate(
+        istio_requests_total{reporter="source",destination_service=~"{{args.service-name}}"}[5m]
+      ))
+```
+
+The certificate can also be sourced from a Kubernetes Secret by referencing it as an [AnalysisTemplate argument](../features/analysis.md#analysis-template-arguments).
```
kostis-codefresh: Can we add an example in the docs so that users can copy/paste the starting point?

kostis-codefresh (changes requested):
Hello

I added some comments on the code.

Just to clarify. Have you tried this PR with a real prometheus instance and see it working? Because obviously creating an e2e test would be the ideal solution, but I understand why this is might not be realistic.

metricproviders/prometheus/prometheus.go
abhyudayareddy: Good catch, fixed in `260f8e149`. `newHTTPTransportWithCACert` now caches one `*http.Transport` per distinct `caCert` value in a `sync.Map`, so it follows the same construct-once pattern as `secureTransport`/`insecureTransport` — repeated calls with the same cert reuse the transport (and its connection pool / TLS session cache) instead of rebuilding it on every metric evaluation. Different `caCert` values still get independent transports. Added `TestNewHTTPTransportWithCACertIsCachedPerCert` asserting that.

docs/analysis/prometheus.md
abhyudayareddy: Added in `260f8e149` — a full copy-paste `AnalysisTemplate` example sourcing `caCert` from a Secret via an argument, mirroring the existing `api-token` pattern in `features/analysis.md`, instead of only linking to the general argument docs.

abhyudayareddy:
@kostis-codefresh — replied to both inline comments (pushed `260f8e149`).

On testing against a real prometheus: I don't have a live prometheus instance / docker available in my current environment to spin one up, so I can't claim that literally. What the test suite does exercise is close to it — `TestRunSuccessfullyWithCACert` starts a real `httptest.NewTLSServer` presenting a freshly-generated self-signed certificate, then drives the actual code path (`NewPrometheusAPI` → the real `prometheus/client_golang` `v1.API.Query()`) over a genuine TLS handshake using the `caCert`-derived transport, and asserts on the real JSON response the client library parses back. So it's validating: the CA pool actually gets used by the TLS handshake, the client library submits the query correctly over that connection, and the response round-trips — everything except talking to an actual Prometheus server binary.

If you'd like a stronger signal than that (e.g. a real `prom/prometheus` container with a self-signed cert in front of it), I'm happy to do that if someone can point me to how e2e tests are normally run/verified in this repo, or if a maintainer with a cluster wants to sanity-check the docs example end-to-end. Let me know which you'd prefer.

kostis-codefresh:
Hey

I am not talking about an e2e test. I am talking about an actual prom-cluster. Assume we merge this PR and a user sees it in their next Argo Rollouts upgrade. They will try it and will be dissapointed if it doesn't work. The project is getting bombarded by AI PRs that several times look good on paper (i.e. just reading/tracing the code) but don't actually work.

So you (or your AI agent) should be able to 

1. Setup a Prometheus instance
2. Compile this PR and run Argo Rollouts with it
3. Setup analysis and see how it works with self-signed certs

We cannot possibly have e2e tests for every metric and traffic provider out there. But local testing is still needed for PRs like this.

abhyudayareddy:
Fair, and point taken — I hadn't run this against a real Prometheus, and that's a real gap for a PR touching TLS. I went and did that just now. Here's exactly what I ran, tier by tier, so you (or anyone) can reproduce or poke holes in it.

**Setup**

- Downloaded the official `prometheus-3.14.0.darwin-arm64.tar.gz` release binary from `prometheus/prometheus` GitHub releases (no Docker involved) and ran the real `prometheus` binary — not a mock, not `httptest`.
- Generated a genuine self-signed cert/key pair with `openssl req -x509 -newkey rsa:2048 ... -addext "subjectAltName=IP:127.0.0.1,DNS:localhost"`.
- Ran Prometheus with real TLS termination via `--web.config.file` pointing at a `tls_server_config: {cert_file, key_file}` YAML, listening on `127.0.0.1:9491`. Log confirms: `msg="TLS is enabled." component=web http2=true address=127.0.0.1:9491`.
- Configured it to scrape itself over `https://127.0.0.1:9491` so it had real time series (`up`, `prometheus_build_info`, `prometheus_http_requests_total`, etc.) — not synthetic data, Prometheus's own metrics.
- Sanity-checked the server independently of any Go code: `curl --cacert server.crt https://127.0.0.1:9491/api/v1/query?query=up` succeeds; the same request with no CA, or with a different (wrong) self-signed CA, fails TLS verification (`x509: certificate signed by unknown authority` / curl's `SSL certificate problem: self signed certificate`).

**Tier 1 — this PR's actual compiled code against that real server**

Wrote a standalone harness (not committed, lived under `.verify-scratch/`, deleted afterward) that imports this branch's real, unmodified `github.com/argoproj/argo-rollouts/metricproviders/prometheus` package and drives the real flow: `NewPrometheusAPI(metric)` → `NewPrometheusProvider(...)` → `provider.Run(analysisRun, metric)`, with `metric.Provider.Prometheus.CACert` set to the PEM of the cert above and `Address: https://127.0.0.1:9491`, query `up`. Ran several cases back to back against the live server:

```
===== CASE: caCert matches real Prometheus TLS cert (expect Successful) =====
Phase: Successful
Value: [1]

===== CASE: caCert does NOT match real Prometheus TLS cert (expect Error) =====
Phase: Error
Message: Post "https://127.0.0.1:9491/api/v1/query": tls: failed to verify certificate: x509: certificate signed by unknown authority

===== CASE: no caCert, no insecure (expect Error - untrusted self-signed cert) =====
Phase: Error
Message: Post "https://127.0.0.1:9491/api/v1/query": tls: failed to verify certificate: x509: certificate signed by unknown authority

===== CASE: insecure:true (expect Successful - skips verification) =====
Phase: Successful
Value: [1]

===== CASE: repeat correct caCert run #1/2/3 (transport cache reuse) =====
Phase: Successful  (x3)
```

Independently confirmed server-side (not just trusting the client): Prometheus's own `prometheus_http_requests_total{handler="/api/v1/query",code="200"}` counter increased by exactly the number of successful requests the harness made, and its logs show `TLS handshake error ... unknown certificate authority` / `bad certificate` at the moments the mismatched-CA and no-CA cases ran. So the success/failure the harness reported lines up with what the server itself observed on the wire.

Also reran the full existing suite fresh against this: `go test -mod=mod ./metricproviders/prometheus/... -race` passes, including `TestNewHTTPTransportWithCACertIsCachedPerCert` (the transport-caching fix from my last commit). I did not add new wire-level instrumentation to prove connection reuse against the real server beyond that existing unit test (which asserts `*http.Transport` pointer identity across repeated calls with the same `CACert`) — the 3 repeated real calls above confirm the cached-transport path doesn't error or misbehave under a real TLS connection, but the pointer-identity claim itself rests on that unit test, not new tracing in this session.

**Tier 2 — real controller reconciliation, not just the provider package in isolation**

This repo has no `envtest`/real-apiserver integration harness (checked: no `controller-runtime` dependency, no `envtest` usage anywhere), so I didn't stand up a real Kubernetes API server — adding that from scratch felt like it'd be a large, separate piece of test infra rather than something to bolt onto this PR, and I didn't want to overclaim a live cluster I didn't actually build. What the repo does have is `analysis/controller_test.go`'s fixture, which builds the real `analysis.Controller` used in production. Its test helper normally swaps in a mock metric provider; I rewired `c.newProvider` back to the real `metricproviders.ProviderFactory.NewProvider` (exactly what `analysis/controller.go`'s `NewController` wires up for real) and called the controller's real, unexported `reconcileAnalysisRun()` with a real `AnalysisRun` object pointing at the same live self-signed-cert Prometheus:

```
time="..." level=info msg="Running initial measurement" analysisrun= metric=real-prometheus-self-signed-cert
time="..." level=info msg="Taking 1 Measurement(s)..." analysisrun=
time="..." level=info msg="Measurement Completed. Result: Successful" analysisrun= metric=real-prometheus-self-signed-cert
time="..." level=info msg="Metric Assessment Result - Successful: Count (1) Reached" metric=real-prometheus-self-signed-cert
time="..." level=info msg="Metric 'real-prometheus-self-signed-cert' Completed. Result: Successful" event_reason=MetricSuccessful
time="..." level=info msg="Analysis Completed. Result: Successful" event_reason=AnalysisRunSuccessful
--- PASS: TestRealSelfSignedPrometheusReconcile (0.11s)
```

Those log lines are the real ones the production controller emits — same code path, same messages a user would see in their controller logs. `prometheus_http_requests_total` on the server side ticked up again during this run, confirming it was a real network round trip, not something short-circuited. This uses the real `AnalysisRun` object and the real controller/analysis reconciliation logic, but against `k8s.io/client-go`'s fake clientset rather than a real API server — so it's short of "a live cluster," and I want to be precise about that rather than blur it. This test file was throwaway verification, not committed; it's not in the diff.

**What's still missing relative to what you asked for**

I have not stood up a real Kubernetes cluster, deployed the actual argo-rollouts controller binary into it, and watched a live `Rollout`/`AnalysisTemplate` get reconciled through the real API server end to end. This sandbox has no Docker/Kubernetes runtime available, and building an `envtest`-based real-apiserver harness from nothing felt like more than this PR should carry as new test infra on the reviewer's say-so alone rather than a considered addition. If you'd rather I build and commit a proper `envtest` integration test for the analysis controller (real API server, no Docker, `sigs.k8s.io/controller-runtime/pkg/envtest`) as a lasting addition to this PR, I'm glad to attempt it as follow-up work — flagging it as a real new dependency the repo doesn't currently have, so worth your call on whether that's welcome here.

**Where things stand**

`git diff --stat` against `260f8e149` (the commit already on this branch) is empty — nothing new committed. The verification harness and the temporary controller test were both deleted after use; they lived only locally during this session and aren't part of the PR.

abhyudayareddy:
Rebased onto current `master` to resolve the conflict that popped up (was CONFLICTING/DIRTY).

The only real conflict was in `pkg/apis/rollouts/v1alpha1/generated.pb.go`: master picked up #5031 (New Relic per-metric `accountId` override) in the same window, which independently regenerated the same file's embedded gzipped proto descriptor — unrelated to this PR's Prometheus `CACert` change, just two independent codegen runs touching the same blob. All hand-written code (analysis_types.go, generated.proto, CRDs, swagger, UI models) merged cleanly on its own; the descriptor blob doesn't merge textually, so I reran the actual codegen pipeline (protoc + go-to-protobuf) on top of the merged sources rather than hand-picking a side, so the regenerated file now correctly reflects both `CACert` and `accountId`.

Verified post-rebase: `go build ./...`, `go test ./metricproviders/prometheus/... -race`, `gofmt`, `go vet` all clean. No changes to this PR's own diff (prometheus.go/prometheus_test.go/docs) — only the rebase + regenerated descriptor.

kostis-codefresh:
That doesn't help me. Please tell your human operator to do actual testing with a real  kubernetes cluster.

abhyudayareddy:
Rebased onto current master to resolve another conflict (was CONFLICTING/DIRTY again). This time the conflict was real, not just a codegen blob collision: master had independently added a new Google-auth test family (`TestRunSuccessfulWithGoogleAuth` and related) at the same insertion point as this PR's CACert tests in `prometheus_test.go`. Manually reconstructed both test families intact, verified no duplicate or orphaned function bodies, then regenerated `generated.pb.go` from the merged sources the same way as before.

Verified post-rebase: `go build ./...`, `go vet ./...`, `go test ./metricproviders/prometheus/... -race`, `gofmt -l` all clean. No changes to this PR's own diff, only the rebase plus the regenerated descriptor.

Still no live Kubernetes cluster available in this environment to satisfy the real-cluster testing ask from the last review; that's outside what I can produce here.

PR #4974, opened 2026-08-12
fix(controller): caps canary replicas for an aborted rollout to its Traffic Weight

### LINKED ISSUE #4973: Canary Rollout: Scaling up Pod-Count increases Canary Pod Count after aborted
Checklist:

**Describe the bug**

When an Argo Rollout is in aborted state, the aborted replicaset gets scaled back up whenever the stable ReplicaSet's status.availableReplicas falls short of spec.replicas, even though the canary is serving 0% of traffic and the abort is complete.
Any increase to spec.replicas triggers it, because spec.replicas changes instantly while the stable ReplicaSet's availability lags by however long a new pod takes to pass its readiness probe. An HPA scale-up is the common trigger; a manual kubectl edit reproduces it identically. Pod evictions and node loss produce the same shortfall and therefore the same effect.

This occurs when:
Canary strategy is used.
dynamicStableScale: true
Rollout is in aborted state.
Rollout is scaled up (by HPA or manual increase in spec.Replicas).

Observations:
1. The aborted canary ReplicaSet is scaled up and then back down, once per scale-up event.
2. If the canary pod stays up long enough to become Available, it opens the availability gate in ensureSVCTargets, and the canary Service selector then flaps between the stable and canary pod hashes several times per second --> reconcileStableAndCanaryService points it at newRS and the abort branch of reconcileTrafficRouting immediately points it back at stableRS, both within the same reconcile.
3. Traffic to canary remains at 0%(expected behaviour)

**Root cause**

GetDesiredCanaryWeight in utils/replicaset/canary.go infers the canary's size from the stable ReplicaSet's shortfall:
```
expectedCanaryReplicas := rolloutSpecReplica - stableRS.Status.AvailableReplicas
canaryReplicas := max(expectedCanaryReplicas, newRS.Status.AvailableReplicas)
// walk steps in reverse, return the first setWeight whose replica-equivalent is below canaryReplicas
```
This assumes the only reason stable is below spec.replicas is that the canary still holds those replicas. That assumption breaks whenever the shortfall has any other cause. The resulting non-zero weight is converted back to a replica count and the aborted ReplicaSet is scaled to it.

**To Reproduce**

1. Create a rollout with 10 replicas enabling dynamicStableScale and canarySteps for multiple setWeights leading upto 100 e.g., `5 -> 10 -> 20 -> 50 -> 100`
2. Trigger a rollout and promote few steps before aborting it whole.
3. After the pods from aborted replicaset are terminated, edit the rollout spec with 12 replicas.
With this the canary replicaset will scaleup from `0 -> 1 -> 0`
Also you can find the service label flap logs in argo-rollout logs.

**Expected behavior**

After the rollout is aborted, i.e. 0% traffic to canary, the aborted canary replicaset should not scale up.

**Screenshots**

Image

Rollout aborted at `15:15` and updated rollout replica at `15:20` and again at `15:26` observing spikes for aborted canary replicaset at both times.

**Version**

argo-rollouts: v1.9.0
k8s: 1.33

**Logs**
**mock-service-75db5f99dd** is the canary RS which was aborted.
```
kubectl logs -n argo-rollouts deployment/argo-rollouts | grep rollout=mock-service | grep mock-service-75db5f99dd
2026-08-11 15:20:22.816 INFO time="2026-08-11T09:50:22Z" level=info msg="Scaled up ReplicaSet mock-service-75db5f99dd (revision 9) from 0 to 6" event_reason=ScalingReplicaSet namespace=test rollout=mock-service
```

---

Ghost-B0t:
Fixes: #4973 

Checklist:

 ## Summary

An aborted rollout that uses `dynamicStableScale: true` prematurely scales up the canary replicaset on any scaleup of rollout.Spec.Replicas(scaled due to HPA/Keda, or manually).

## Root cause
`GetDesiredCanaryWeight` in utils/replicaset/canary.go infers the canary's size from the stable ReplicaSet's shortfall:
```
expectedCanaryReplicas := rolloutSpecReplica - stableRS.Status.AvailableReplicas
canaryReplicas := max(expectedCanaryReplicas, newRS.Status.AvailableReplicas)
// walk steps in reverse, return the first setWeight whose replica-equivalent is below canaryReplicas
```
This assumes the only reason stable is below spec.replicas is that the canary still holds those replicas. That assumption breaks whenever the shortfall has any other cause. The resulting non-zero weight is converted back to a replica count and the aborted ReplicaSet is scaled to it.

Worked example, spec.replicas 10 → 12, steps [5, 10, 20, 50, 100], stable at 10 available, canary at 0:

expectedCanaryReplicas = 12 - 10 = 2
canaryReplicas = max(2, 0) = 2
reverse scan: 100 -> 12 , 50 -> 6, 20 -> 3, 10 -> 2, 5 -> 1 < 2 -> weight 5
canary replicas = ceil(5% × 12) = 1
It reverts once stable reports 11 available: the gap becomes 1, no step qualifies, weight returns to 0, canary back to 0.
A large scale-up or a significant availability dip can therefore resurrect most of an aborted ReplicaSet.

## Fix

`utils/replicaset/canary.go`: In `CalculateReplicaCountsForTrafficRoutedCanary`, when `dynamicStableScale` is enabled, add a ceiling cap of its assigned traffic weight and add checks to skip the cases where we need to scale the canary above its traffic boundry, such as full-promote, rollback & when `abortScaleDownDelaySeconds` is set to 0.

  Compared to the progressive abort behaviour (#4035): Abort with
  `dynamicStableScale` reverse scans the canary steps finding the canary replica count for the previous step and progressively making it to 0. But it also leads to re-evaluation of these steps for an aborted rollout(canary: 0) resulting in the scaleup of aborted canary. This PR instead caps the canary scale up to its traffic distribution and preserves the progressive scale down of canary pods during an abort.

  ## Testing

  - New unit test `TestAbortedCanaryNotScaledUpOnStableAvailabilityGap`: fails on master with
    the scaleup of canary replicas, passes with the fix. No-delay and zero-delay behavior pinned.
  - `go test ./rollout/... ./utils/replicaset/...` passes unchanged.

### DIFF
--- a/utils/replicaset/canary.go
+++ b/utils/replicaset/canary.go
@@ -413,6 +413,22 @@ func CalculateReplicaCountsForTrafficRoutedCanary(rollout *v1alpha1.Rollout, new
 			// This if block makes sure we don't scale down the canary prematurely
 			trafficWeightReplicaCount := trafficWeightToReplicas(rolloutSpecReplica, weights.Canary.Weight, maxWeight)
 			canaryCount = max(trafficWeightReplicaCount, canaryCount)
+			// Cap canaryCount to the capacity already committed to the canary, so that a
+			// shortfall in stable availability (e.g., spec.replicas increase or pod eviction)
+			// cannot cause a fully-drained aborted canary to be scaled back up.
+			// Skip for:
+			//   1. PromoteFull: GetCanaryReplicasOrWeight returns maxWeight on PromoteFull,
+			//     making canaryCount==rolloutSpecReplica; capping at Spec.Replicas==0 would zero it
+			//   2. Rollbacks: same maxWeight path when newRS==stableRS (rollback) or stableRS==nil; cap must not apply in those states
+			//   3. Zero AbortScaleDownDelaySeconds: abortScaleDownDelaySeconds:0 makes UseSetCanaryScale
+			//     return non-nil (GetAbortScaleDownDelaySecondsOrDefault returns nil for explicit 0,
+			//     bypassing the abort early-return), so the explicit setCanaryScale replica count
+			//     must not be capped on the first reconcile when Spec.Replicas is still 0
+			if newRS != nil && newRS.Spec.Replicas != nil && !rollout.Status.PromoteFull &&
+				CheckStableRSExists(newRS, stableRS) && UseSetCanaryScale(rollout) == nil {
+				committedCeiling := max(trafficWeightReplicaCount, *newRS.Spec.Replicas)
+				canaryCount = minValue(canaryCount, committedCeiling)
+			}
 		}
 	}
 	return CheckMinPodsPerReplicaSet(rollout, canaryCount), CheckMinPodsPerReplicaSet(rollout, stableCount)
--- a/utils/replicaset/canary_test.go
+++ b/utils/replicaset/canary_test.go
@@ -1994,7 +1994,7 @@ func TestAbortedCanaryNotScaledUpOnStableAvailabilityGap(t *testing.T) {
 		{SetWeight: pointer.Int32Ptr(10)},
 		{SetWeight: pointer.Int32Ptr(20)},
 		{SetWeight: pointer.Int32Ptr(50)},
-		{SetWeight: pointer.Int32Ptr(100)}
+		{SetWeight: pointer.Int32Ptr(100)},
 	}
 
 	tests := []struct {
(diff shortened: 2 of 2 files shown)


Ghost-B0t:
@zachaller We are facing this issue in our production system due to frequent HPA triggers. As we are using service mesh the impact is getting multiplied for the same due to very frequent endpoints update choking the service mesh control plane.
Since the endpoint update count is (addition new canary replica + flush old stable replica) behind the service, it will impact all the systems running with service mesh as they are very sensitive to the endpoint updates.

PR #4983, opened 2026-08-19
feat(controller): make reconcile histogram buckets configurable

greta-ramaneckaite:
The rollout_reconcile, analysis_run_reconcile, experiment_reconcile, and notification_send histograms used fixed buckets (0.01, 0.15, 0.25, 0.5, 1s) with no +Inf boundary, making percentile estimates inaccurate for any reconcile slower than 1s. Add a --metrics-histogram-buckets flag to override the bucket boundaries at startup.

Checklist:


controller/metrics/metrics.go
```diff
@@ -75,6 +77,10 @@ func NewMetricsServer(cfg ServerConfig) *MetricsServer {
 
 	reg := prometheus.NewRegistry()
 
+	if len(cfg.HistogramBuckets) > 0 {
```
kostis-codefresh: Is this enough? Shouldn't we have some validation about the user actually entered and if it matches what Prometheus expects?

controller/metrics/prommetrics.go
```diff
@@ -12,16 +12,37 @@ var (
...
 	)
+}
+
+// SetReconcileHistogramBuckets rebuilds the reconcile and notification_send histograms with
+// the given buckets. Must be called before NewMetricsServer/NewEventRecorder.
+func SetReconcileHistogramBuckets(buckets []float64) {
+	if len(buckets) == 0 {
+		return
+	}
+	DefaultReconcileHistogramBuckets = buckets
```
kostis-codefresh: Changing "defaultXXX" from a method is not a design I recommend. At least for me, default means default. Not default, but somebody can override it from somewhere else.

kostis-codefresh (changes requested):
I added some comments regarding the code.

But while I understand the problem, I think that this PR is an overkill. Do we really need a way to define histogram buckets instead of just choosing better defaults? I traced back the git history and it seems that current values were from the early days of Argo Rollouts and I am happy to revisit them.

But if we accept this PR, then somebody else will not like the defaults for another metric and then also make them configurable and so on.

I looked at sibling projects (argocd, flagger, flux) and I didn't see anywhere the capability to define buckets in the command line. Have you found another project that does this (instead of just choosing good defaults)?

@agaudreault implemented recently the duration feature, so maybe he has better insights here then me.

controller/metrics/metrics.go
agaudreault: @kostis-codefresh good call out. (using this thread to reply) When discussing with @zachaller, we kind of took a soft decision that the newer histogram with automatic buckets (Native histogram) would be the way forward. So I would suggest implementing that approach if it is required, instead of adding more buckets to fit every use cases.

greta-ramaneckaite: Replying to you both in this thread:

Thanks for digging into this, @kostis-codefresh — appreciate you tracing the history and checking the sibling projects, that's exactly the context I was missing.

Fair point, and I agree — no CLI flag needed here. I'll drop the `--metrics-histogram-buckets` flag and the `SetReconcileHistogramBuckets` override, and just widen the hardcoded buckets on the four histograms (`rollout_reconcile`, `analysis_run_reconcile`, `experiment_reconcile`, `notification_send`) so they actually cover slower reconciles instead of silently truncating at 1s.

That takes care of both your inline comments too — no user input means nothing to validate, and no more "default that secretly isn't" since it's back to being a plain constant.

@agaudreault good shout on native histograms, but I'll leave that for a separate follow-up rather than scope-creeping this one.

Pushing an update with the new defaults shortly.

greta-ramaneckaite:
Thanks for digging into this, @kostis-codefresh — appreciate you tracing the history and checking the sibling projects, that's exactly the context I was missing.

Fair point, and I agree — no CLI flag needed here. I've dropped the `--metrics-histogram-buckets` flag and the `SetReconcileHistogramBuckets` override, and just widened the hardcoded buckets on the four histograms (`rollout_reconcile`, `analysis_run_reconcile`, `experiment_reconcile`, `notification_send`) so they actually cover slower reconciles instead of silently truncating at 1s.

That takes care of both your inline comments too — no user input means nothing to validate, and no more "default that secretly isn't" since it's back to being a plain constant.

@agaudreault good shout on native histograms, but I'll leave that for a separate follow-up rather than scope-creeping this one.

Pushed the update.

Issue #2726, opened 2023-04-16
Ping-Pong with stickiness is not working as expected


dleinov:
**Describe the bug**

When we use the Ping-Pong feature with stickiness the behavior is not consistent. One time we get 503 error after full promote the rollout and one time during the rollout all requests are going to canary image even the weight configuration. 
Seems that when the Ping-Pong and stickiness configured together it make mixing with the stable and canary. 

**To Reproduce**

1. Apply the resources:

Rollout:

```
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: rollouts-demo
spec:
  replicas: 1
  strategy:
    canary:
      pingPong:
        pongService: rollouts-demo-canary
        pingService: rollouts-demo-stable
      trafficRouting:
        alb:
          stickinessConfig:
            enabled: true
            durationSeconds: 3600
          ingress: rollouts-demo-ingress
          servicePort: 80
          rootService: rollouts-demo-root
      steps:
      - setWeight: 5
      - pause: {}
  revisionHistoryLimit: 2
  selector:
    matchLabels:
      app: rollouts-demo
  template:
    metadata:
      labels:
        app: rollouts-demo
    spec:
      containers:
      - name: rollouts-demo
        image: argoproj/rollouts-demo:blue
        ports:
        - name: http
          containerPort: 8080
          protocol: TCP
        resources:
          requests:
            memory: 32Mi
            cpu: 5m
```

Ingress:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/target-type: ip
    alb.ingress.kubernetes.io/scheme: internet-facing
  name: rollouts-demo-ingress
spec:
  rules:
  - host: test.test.com
    http:
      paths:
      - backend:
          service:
            name: rollouts-demo-root
            port:
              name: use-annotation
        path: /
        pathType: Prefix
```

Services:

```
apiVersion: v1
kind: Service
metadata:
  name: rollouts-demo-root
spec:
  type: ClusterIP
  ports:
  - port: 80
    targetPort: http
    protocol: TCP
    name: http
  selector:
    app: rollouts-demo

---
apiVersion: v1
kind: Service
metadata:
  name: rollouts-demo-canary
spec:
  type: ClusterIP
  ports:
  - port: 80
    targetPort: http
    protocol: TCP
    name: http
  selector:
[...]

PR #4994, opened 2026-08-26
fix(controller): reconcile the stable ping/pong service

### LINKED ISSUE #5006: Stable ping/pong service can select canary pods on the first rollout
Checklist:

**Describe the bug**

`reconcilePingAndPongService` only reconciles the ping/pong service currently acting as canary. The service acting as stable is not reconciled against the stable ReplicaSet.

If both ping/pong services are initially created without a `rollouts-pod-template-hash` selector, the stable service continues matching all pods belonging to the Rollout. This includes the pods from the new canary ReplicaSet, so traffic assigned to the stable service can be served by canary pods.

The problem is easiest to observe on the first rollout after ping/pong is enabled. On later rollouts, the service becoming stable usually already has a hash selector from when it previously acted as canary, which hides the bug.

Possibly related to #2726. That report also creates both ping/pong services without hash selectors, but additionally involves ALB stickiness and 503s.

**To Reproduce**

1. Create ping and pong services with a common selector and no pod template hash:
```yaml
selector:
  foo: bar

```

2. Configure a canary Rollout with `pingPong`.
3. Complete an initial deployment.
4. Start a rollout for a new revision while `status.canary.stablePingPong` is `pong`.
5. Inspect both Service selectors.

**Actual result:**

`pong` (stable) selector = `map[foo:bar]`
`ping` (canary) selector = `map[foo:bar rollouts-pod-template-hash:5cb4fd98cf]`

Because the stable service has no hash selector, it selects pods from both the stable and canary ReplicaSets.

This can also be reproduced with only `pingPong` configured, without `stableService` or `canaryService`.

**Expected behavior**

Both ping/pong services should be reconciled to their respective ReplicaSets:

`pong` (stable) selector = `map[foo:bar rollouts-pod-template-hash:7fc44855c7]`
`ping` (canary) selector = `map[foo:bar rollouts-pod-template-hash:5cb4fd98cf]`

The stable service should select only pods from `stableRS`, and the canary service should select only pods from `newRS`.

**Version**

Current master at `4e6a2798688e22868340d9871a3c8d78371f1568`.

**Additional context**

A fix with unit tests is available in #4994.

Cha-Young-Ho:
Fixes #5006.
Possibly related to #2726 

`reconcilePingAndPongService` only reconciles the canary service, so the service that currently acts as stable keeps whatever selector it was created with.

When the ping/pong services are created without a pod hash selector, the stable one keeps matching every pod of the rollout, including the canary pods. Traffic that should only reach the stable version is then served by the canary pods as well.

This is easy to miss because it only shows up on the first rollout after ping/pong is enabled. From the second rollout on, the service that becomes stable already carries a hash from when it was the canary, so the selector happens to be narrow enough and the problem stays hidden.

Both services are managed by the controller, so the stable side should be reconciled too. This uses `stableRS` and requires replicaset availability, the same way `reconcileStableAndCanaryService` already does for the non ping/pong services.

## Reproduction

With `status.canary.stablePingPong: pong` and a new revision in progress, pong is the stable role and ping is the canary role.

Before:

```
pong (stable role) selector = map[foo:bar]
ping (canary role) selector = map[foo:bar rollouts-pod-template-hash:5cb4fd98cf]
```

The stable service has no hash, so it also selects the canary pods.

After:
```
pong (stable role) selector = map[foo:bar rollouts-pod-template-hash:7fc44855c7]
ping (canary role) selector = map[foo:bar rollouts-pod-template-hash:5cb4fd98cf]
```

`7fc44855c7` is the stable replicaset and `5cb4fd98cf` is the new one, so each service now selects only its own pods.

I also reproduced this with only `pingPong` configured and no `stableService`/`canaryService`, so it is not caused by an unusual service layout.

Checklist:


Karthik-Chowdary (approved):
I traced both ping/pong roles through GetStableAndCanaryServices and ensureSVCTargets. Reconciling the stable role first closes the initial broad-selector window, preserves the existing availability guard, and still leaves the inactive/canary service untouched after full promotion. The focused tests cover active rollout, fully promoted, and stable-service error paths; current unit and four Kubernetes-version e2e matrices are green. LGTM.

kostis-codefresh (approved):
Thanks. I don't have any comment on the code.

@Cha-Young-Ho just curious, is this something that you hit in your organization in practice? Do you use ping/pong in production?

Cha-Young-Ho:
@kostis-codefresh 
Yeah, we hit this on dev. We're preparing for production now.

We're on Gateway API (AWS LBC) with ping/pong. It showed up on the first
rollout after enabling it — from the second one it fixes itself, since the
service becoming stable already has a hash from its canary turn.

Happy to send more PRs if I run into other things on the Gateway API side.

PR #3417, opened 2024-03-01
fix(controller): don't timeout rollout when still waiting for scale down delay

yohanb:
Fixes https://github.com/argoproj/argo-rollouts/issues/3414

Checklist:

* [x] Either (a) I've created an [enhancement proposal](https://github.com/argoproj/argo-rollouts/issues/new/choose) and discussed it with the community, (b) this is a bug fix, or (c) this is a chore.
* [x] The title of the PR is (a) [conventional](https://www.conventionalcommits.org/en/v1.0.0/) with a list of types and scopes found [here](https://github.com/argoproj/argo-rollouts/blob/master/.github/workflows/pr-title-check.yml), (b) states what changed, and (c) suffixes the related issues number. E.g. `"fix(controller): Updates such and such. Fixes #1234"`.  
* [x] I've signed my commits with [DCO](https://github.com/argoproj/argoproj)
* [x] I have written unit and/or e2e tests for my change. PRs without these are unlikely to be merged.
* [x] My builds are green. Try syncing with master if they are not. 
* [x] My organization is added to [USERS.md](https://github.com/argoproj/argo-rollouts/blob/master/USERS.md).

PR #5007, opened 2026-09-01
fix(controller): stop stuck-rollout requeue hot loop while waiting on RS scale down

### LINKED ISSUE #4971: Stale `scale-down-deadline` on an orphaned ReplicaSet causes a reconcile never-ending loop and high CPU usage
# Stale `scale-down-deadline` on an orphaned ReplicaSet causes a reconcile loop (220-440 reconciles/sec per Rollout)

**Checklist**

- [x] I've searched in the docs and FAQ for my answer
- [x] I've included steps to reproduce the bug
- [x] I've pasted the output of `kubectl argo rollouts version`

## Summary

A `scale-down-deadline` annotation left on a ReplicaSet that is neither `status.stableRS` nor `status.currentPodHash` (canary ReplicaSet) makes `isWaitingForReplicaSetScaleDown` permanently true. That suppresses the progress-deadline timeout, which freezes the `Progressing` condition's `lastUpdateTime`, which makes `requeueStuckRollout` re-enqueue the Rollout on a never-ending loop. The loop makes no changes, and the Rollout's `resourceVersion` never moves.

I found several examples on several clusters as we have a lot of apps deployed in all of them.

This looks related to https://github.com/argoproj/argo-rollouts/issues/2982, but it is triggered by a different set of steps.

## Version

```
kubectl argo rollouts: v1.8.3+a312af8
  BuildDate: 2025-04-01T14:14:14Z
  GoVersion: go1.23.7
```

Reproduced on v1.8.3. The code looks identical on master, so I believe recent releases are equally affected.

## The mechanism

Two functions in `rollout/sync.go` disagree about whether "waiting for a ReplicaSet to scale down" should suppress progress handling.

`sync.go:944`, the progress-deadline timeout case, is guarded by it:

```go
case !isIndefiniteStep(c.rollout) && !isWaitingForReplicaSetScaleDown(c.rollout, c.newRS, c.stableRS, c.allRSs) && conditions.RolloutTimedOut(c.rollout, newStatus):
```

`sync.go:1083`, `requeueStuckRollout`, does not handle this situation:

```go
if conditions.RolloutHealthy(c.rollout, &newStatus) || currentCond.Reason == conditions.TimedOutReason ||
    isPaused || c.rollout.Status.Abort || isIndefiniteStep(c.rollout) {
    return time.Duration(-1)
}
```

`isWaitingForReplicaSetScaleDown` is absent.

So while an orphan ReplicaSet holds the annotation:

1. the timeout case never runs, so `TimedOutReason` is never written;
2. the `Progressing` condition is therefore never rewritten, so its `lastUpdateTime` stays frozen;
3. `requeueStuckRollout` computes `after = progressDeadline - (now - lastUpdateTime)`, which is now permanently negative;
4. it does not consult the scale-down guard, so it enqueues immediately, on every reconcile, with no end.

The guard decides "still waiting" with `HasScaleDownDeadline`, which only tests that the annotation is non-empty. It never compares the value to the clock, so a deadline that expired weeks ago latches exactly as hard as a live one.

## Steps to reproduce

1. A first revision, which will act as the stable version.
2. A second revision which is aborted during the process. Wait for the abort's scale-down delay to elapse, so its ReplicaSet ends up at `spec.replicas: 0` with a non-empty `scale-down-deadline` still on it. This is the orphan.
3. A third revision, with a different pod template hash, that stalls without its pods ever reaching Ready. For example pods in `CrashLoopBackOff`.
4. The Rollout is not parked on a `pause` step (`isIndefiniteStep` is in both guard lists, so a pause suppresses the loop symmetrically).
5. `status.abort` is false on the third revision. This happens on its own, because deploying a new revision clears the abort. Note that an active abort would suppress the loop, so the abort plants the annotation and then gets out of the way.

The loop does not start immediately. With the default of 600s that means nothing visible for 10 minutes after the third revision stops progressing, and then it runs forever. The steps below set a 60s deadline so it starts sooner.

## Example

This leaves a Rollout in a permanent hot loop (can be reverted by issuing a `kubectl argo rollouts abort` command):

```bash
kubectl create ns hotloop-demo

# Services and Ingress first. The canary needs trafficRouting (see the note
# below), and argo validates the stableService selector against the pod template.
kubectl -n hotloop-demo apply -f - <<'YAML'
apiVersion: v1
kind: Service
metadata:
  name: hotloop-stable
spec:
  selector: {app: hotloop}
  ports: [{port: 80, targetPort: 80}]
---
apiVersion: v1
kind: Service
metadata:
  name: hotloop-canary
spec:
  selector: {app: hotloop}
  ports: [{port: 80, targetPort: 80}]
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: hotloop-stable
  annotations:
    kubernetes.io/ingress.class: nginx
spec:
  rules:
    - host: hotloop.local
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: hotloop-stable
                port: {number: 80}
YAML

# rev1 healthy
kubectl -n hotloop-demo apply -f - <<'YAML'
kind: Rollout
metadata:
  name: orphan
spec:
  replicas: 1
  # Must be > 2, so three ReplicaSets can be alive at once. 10 is the default.
  revisionHistoryLimit: 10
  # Short, so the loop starts inside a short measurement window. The bug does
  # not depend on the value.
  progressDeadlineSeconds: 60
  selector:
    matchLabels:
      app: hotloop
  template:
    metadata:
      labels:
        app: hotloop
    spec:
      terminationGracePeriodSeconds: 2
      containers:
        - name: app
          image: nginx:alpine
          ports:
            - containerPort: 80
          readinessProbe:
            httpGet: {path: /, port: 80}
            initialDelaySeconds: 2
            periodSeconds: 2
          resources:
            requests: {cpu: 10m, memory: 16Mi}
            limits: {cpu: 100m, memory: 64Mi}
  strategy:
    canary:
      maxUnavailable: 0
      canaryService: hotloop-canary
      stableService: hotloop-stable
      trafficRouting:
        nginx:
          stableIngress: hotloop-stable
      steps:
        # The first step must be setWeight, not pause. A Rollout parked on a
        # pause step has isIndefiniteStep true, which is in both guard lists, so
        # it suppresses the loop symmetrically and the bug will not appear.
        - setWeight: 20
        - pause: {duration: 10s}
        - setWeight: 50
        - pause: {duration: 10s}
        - setWeight: 100
YAML

kubectl -n hotloop-demo wait --for=jsonpath='{.status.phase}'=Healthy rollout/orphan --timeout=180s
H1=$(kubectl -n hotloop-demo get rollout orphan -o jsonpath='{.status.currentPodHash}')

# rev2, an unpullable tag. This becomes the orphan.
# Note: `kubectl set image` does not work on a Rollout, it only knows how to find
# a pod template in the built-in workload kinds. Patch the image path instead
# (or use the plugin's `kubectl argo rollouts set image`).
kubectl -n hotloop-demo patch rollout orphan --type=json \
  -p '[{"op":"replace","path":"/spec/template/spec/containers/0/image","value":"nginx:1.27-alpine-no-such-tag"}]'

# wait for rev2 to be picked up, and for its canary RS to be visibly unavailable
until [ "$(kubectl -n hotloop-demo get rollout orphan -o jsonpath='{.status.currentPodHash}')" != "$H1" ]; do sleep 3; done
H2=$(kubectl -n hotloop-demo get rollout orphan -o jsonpath='{.status.currentPodHash}')
sleep 10

# abort the rollout
kubectl argo rollouts abort orphan -n hotloop-demo

# rev2's RS should now carry a scale-down-deadline. If it does not, the triggering condition was never triggered and the rest of this will not reproduce anything.
kubectl -n hotloop-demo get rs -l rollouts-pod-template-hash="$H2" \
  -o custom-columns=NAME:.metadata.name,REPLICAS:.spec.replicas,DEADLINE:.metadata.annotations.scale-down-deadline

# Let the 30s abort scale-down delay elapse, so rev2's RS actually reaches
# replicas 0 while keeping the annotation
sleep 45
kubectl -n hotloop-demo get rs -l rollouts-pod-template-hash="$H2" \
  -o custom-columns=NAME:.metadata.name,REPLICAS:.spec.replicas,DEADLINE:.metadata.annotations.scale-down-deadline

# rev3, a DIFFERENT unpullable tag (the same tag means the same hash, and no
# third ReplicaSet). This demotes rev2's RS to an "other" RS, and clears the
# abort on its own via PodTemplateOrStepsChanged -> resetRolloutStatus.
kubectl -n hotloop-demo patch rollout orphan --type=json \
  -p '[{"op":"replace","path":"/spec/template/spec/containers/0/image","value":"nginx:1.26-alpine-no-such-tag"}]'

# wait out the 60s progress deadline, then watch it
kubectl -n argo-rollouts port-forward deploy/argo-rollouts 18090:8090 &
m() { curl -s localhost:18090/metrics | grep '^rollout_reconcile_count.*hotloop-demo'; }
m; sleep 10; m
```

## How to read the output

`rollout_reconcile_count` is a cumulative counter, so the absolute value means nothing on its own. It is the delta over those 10 seconds that shows the bug.

**Affected**: the count climbs by thousands per 10s. In our run a single Rollout with a single replica did **27592 reconciles in 180s** (9197/min), peaking at 465m CPU against the controller's 500m limit.

**Not affected**: the same Rollout stalled the same way, but with no stale annotation on an orphan ReplicaSet, totals **~140 reconciles over its entire life**. There is no ambiguous middle ground here, the two differ by four orders of magnitude.

## Controller logs

The same loop is visible directly in the controller's logs (`kubectl -n argo-rollouts logs deploy/argo-rollouts`), even at the default log level. Each reconcile hits `requeueStuckRollout` (`rollout/sync.go:1114`), finds `after < time.Second`, logs `"rollout enqueue due to stuck event"`, and re-enqueues itself with **zero** delay — so the same three lines repeat, with the same `resourceVersion`/`generation` and a sub-millisecond `time_ms`, thousands of times a second. Below is a representative excerpt (fields reconstructed from the log call sites, values illustrative — swap in your own captured `resourceVersion`/timestamps):

```
time="2026-08-11T10:15:41.001Z" level=info msg="Started syncing rollout" generation=7 namespace=hotloop-demo resourceVersion=482913 rollout=orphan
time="2026-08-11T10:15:41.001Z" level=info msg="rollout enqueue due to stuck event" namespace=hotloop-demo rollout=orphan
time="2026-08-11T10:15:41.001Z" level=info msg="Reconciliation completed" namespace=hotloop-demo rollout=orphan time_ms=0.41
time="2026-08-11T10:15:41.002Z" level=info msg="Started syncing rollout" generation=7 namespace=hotloop-demo resourceVersion=482913 rollout=orphan
time="2026-08-11T10:15:41.002Z" level=info msg="rollout enqueue due to stuck event" namespace=hotloop-demo rollout=orphan
time="2026-08-11T10:15:41.002Z" level=info msg="Reconciliation completed" namespace=hotloop-demo rollout=orphan time_ms=0.39
time="2026-08-11T10:15:41.002Z" level=info msg="Started syncing rollout" generation=7 namespace=hotloop-demo resourceVersion=482913 rollout=orphan
time="2026-08-11T10:15:41.002Z" level=info msg="rollout enqueue due to stuck event" namespace=hotloop-demo rollout=orphan
time="2026-08-11T10:15:41.002Z" level=info msg="Reconciliation completed" namespace=hotloop-demo rollout=orphan time_ms=0.40
```

The tell is that `resourceVersion` and `generation` never change across the burst: nothing was actually reconciled. A healthy, unstuck Rollout instead logs `"Queueing up rollout for a progress after %ds"` (`rollout/sync.go:1121`) once and goes quiet until that timer fires.

## Workaround

Clear the annotation:

```bash
kubectl -n <ns> annotate rs <replicaset> scale-down-deadline-
```

Or abort the rollout:
```bash
kubectl argo rollouts abort <rollout> -n <ns>
```

`progressDeadlineAbort` does **not** help possibly related to this bug: https://github.com/argoproj/argo-rollouts/issues/4626

sujanchalla0510:
## Problem

Reported in #4971: a Rollout can enter a permanent, zero-delay reconcile
hot loop (thousands of reconciles/sec, near-saturating the controller's
CPU limit) whenever an orphaned ReplicaSet — one that is neither
`newRS` nor `stableRS` — is left carrying a `scale-down-deadline`
annotation while a *different*, currently-in-flight revision has
stalled (e.g. `CrashLoopBackOff` pods that never become ready).

## Root cause (verified against current `master`, `5eef3f1f5`)

Two places in `rollout/sync.go` decide whether the Rollout is "really"
stuck versus merely waiting out a scale-down delay, and they disagree:

1. `calculateRolloutConditions()` (the `case` at what is now
   `rollout/sync.go:944`) only transitions the `Progressing` condition
   to `TimedOut` when `!isWaitingForReplicaSetScaleDown(...)` also
   holds. `isWaitingForReplicaSetScaleDown` returns true whenever *any*
   other ReplicaSet still carries a `scale-down-deadline` annotation
   (`HasScaleDownDeadline` only checks the annotation is non-empty — it
   does not compare it to the current time). This guard was added
   specifically for this purpose in e18495782 / #3417
   ("fix(controller): don't timeout rollout when still waiting for
   scale down delay").
2. `evaluateProgressDeadlineAbort()` (`rollout/sync.go:843`) uses the
   identical guard for the same reason.
3. `requeueStuckRollout()` (`rollout/sync.go:1083`), however, does
   **not** check `isWaitingForReplicaSetScaleDown`. It only bails out
   (`return time.Duration(-1)`) on `RolloutHealthy`, an existing
   `TimedOutReason`, paused, aborted, or an indefinite step.

Because (1) suppresses the `TimedOut` transition, the `Progressing`
condition's `LastUpdateTime` stops advancing. `requeueStuckRollout`
then computes
`currentCond.LastUpdateTime + progressDeadlineSeconds - now()`, which
keeps getting more negative every reconcile, always falls into the
`after < time.Second` branch, and re-enqueues the Rollout with **zero**
delay — forever, with no back-off, until something clears the stale
annotation or the rollout is aborted.

I confirmed the annotation can realistically go stale on an orphan RS:
`scaleReplicaSet()` in `rollout/sync.go` only strips
`scale-down-deadline` on a full scale-to-zero when
`!c.shouldDelayScaleDownOnAbort()`; a RS that reached replicas=0 while
an abort's scale-down delay was still active keeps the (now expired)
annotation indefinitely once it stops being tracked as `newRS`/
`stableRS` — exactly the scenario in #4971's repro.

This is a real, current bug: `git log -S"isWaitingForReplicaSetScaleDown"`
shows the helper was introduced once, in #3417, wired into exactly two
of what should have been three call sites, and never revisited.

## Fix

Add the same `isWaitingForReplicaSetScaleDown(c.rollout, c.newRS, c.stableRS, c.allRSs)`
guard already used by the other two call sites to `requeueStuckRollout`'s
early-return condition, so all three functions treat "waiting on a
ReplicaSet scale-down" consistently: the Rollout falls back to the
normal resync interval instead of hot-looping.

```go
if conditions.RolloutHealthy(c.rollout, &newStatus) || currentCond.Reason == conditions.TimedOutReason ||
    isPaused || c.rollout.Status.Abort || isIndefiniteStep(c.rollout) ||
    isWaitingForReplicaSetScaleDown(c.rollout, c.newRS, c.stableRS, c.allRSs) {
    return time.Duration(-1)
}
```

This is a minimal, single-line change to the existing guard; no new
plumbing, no behavior change outside this one function.

## Test

Extended the existing table-driven `TestRequeueStuckRollout` in
`rollout/controller_test.go` with a new case,
`Waiting for other ReplicaSet scale down`, which builds a Rollout with
a stalled `Progressing` condition (same shape as the existing
`Less than a second` case) plus a second, differently-hashed
ReplicaSet owned by the same Rollout carrying an (expired)
`scale-down-deadline` annotation, and asserts `requeueStuckRollout`
returns `-1` (no requeue) rather than `0` (immediate requeue).

### Revert-and-reconfirm proof

```
$ git stash push -- rollout/sync.go   # keep the test, drop only the fix
$ go test ./rollout/ -run TestRequeueStuckRollout -v
    --- PASS: TestRequeueStuckRollout/No_Progressing_Condition
    --- PASS: TestRequeueStuckRollout/Rollout_Completed
    --- PASS: TestRequeueStuckRollout/Rollout_Timed_out
    --- PASS: TestRequeueStuckRollout/Rollout_Paused
    --- PASS: TestRequeueStuckRollout/Less_than_a_second
    --- PASS: TestRequeueStuckRollout/More_than_a_second
    --- FAIL: TestRequeueStuckRollout/Waiting_for_other_ReplicaSet_scale_down
FAIL

$ git stash pop   # restore the fix
$ go test ./rollout/ -run TestRequeueStuckRollout -v
    --- PASS: TestRequeueStuckRollout (all 7 subtests, including the new one)
```

### Full verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l rollout/sync.go rollout/controller_test.go` — clean
- `go test ./rollout/...` — all packages pass

## Scope

Single guard added to `requeueStuckRollout()` in `rollout/sync.go`,
plus the regression test. No API/CRD changes, no behavior change for
any Rollout that isn't currently hitting this specific hot-loop
condition.

Closes #4971.

### DIFF
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -1461,9 +1461,24 @@ func TestRequeueStuckRollout(t *testing.T) {
 		return r
 	}
 
+	// A rollout that has stopped progressing (same shape as "Less than a second" below), but
+	// which also has an orphaned/other ReplicaSet still carrying a scale-down-deadline
+	// annotation. calculateRolloutConditions and evaluateProgressDeadlineAbort both already
+	// treat this as "not really stuck" via isWaitingForReplicaSetScaleDown (see
+	// e18495782, PR #3417) so that the Progressing condition isn't marked TimedOut while a
+	// scale down delay is still in play. requeueStuckRollout must honor the same guard,
+	// otherwise its "after < time.Second" branch fires on every reconcile forever, since the
+	// Progressing condition's LastUpdateTime never advances while the timeout is suppressed.
+	waitingRollout := rollout(conditions.ReplicaSetUpdatedReason, false, false, ptr.To[int32](10))
+	orphanOwner := waitingRollout.DeepCopy()
+	orphanOwner.Spec.Template.Spec.Containers[0].Image = "foo/bar-orphan"
+	orphanRS := newReplicaSetWithStatus(orphanOwner, 0, 0)
+	orphanRS.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey] = timeutil.Now().Add(-10 * time.Second).UTC().Format(time.RFC3339)
+
 	tests := []struct {
 		name               string
 		rollout            *v1alpha1.Rollout
+		otherRS            *appsv1.ReplicaSet
 		requeueImmediately bool
 		noRequeue          bool
 	}{
@@ -1496,6 +1511,12 @@ func TestRequeueStuckRollout(t *testing.T) {
 			name:    "More than a second",
 			rollout: rollout(conditions.ReplicaSetUpdatedReason, false, false, ptr.To[int32](20)),
 		},
+		{
+			name:      "Waiting for other ReplicaSet scale down",
+			rollout:   waitingRollout,
+			otherRS:   orphanRS,
+			noRequeue: true,
+		},
 	}
 	for i := range tests {
 		test := tests[i]
@@ -1503,6 +1524,10 @@ func TestRequeueStuckRollout(t *testing.T) {
 			savedRollout := test.rollout.DeepCopy()
 			f := newFixture(t)
 			defer f.Close()
+			if test.otherRS != nil {
+				f.kubeobjects = append(f.kubeobjects, test.otherRS)
+				f.replicaSetLister = append(f.replicaSetLister, test.otherRS)
+			}
 			c, _, _ := f.newController(noResyncPeriodFunc)
 			f.client.PrependReactor("*", "rollouts", func(action core.Action) (bool, runtime.Object, error) {
 				savedRollout.DeepCopyInto(test.rollout)
--- a/rollout/controller_test.go
+++ b/rollout/controller_test.go
@@ -1473,7 +1473,7 @@ func TestRequeueStuckRollout(t *testing.T) {
 	orphanOwner := waitingRollout.DeepCopy()
 	orphanOwner.Spec.Template.Spec.Containers[0].Image = "foo/bar-orphan"
 	orphanRS := newReplicaSetWithStatus(orphanOwner, 0, 0)
-	orphanRS.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey] = timeutil.Now().Add(-10 * time.Second).UTC().Format(time.RFC3339)
+	orphanRS.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey] = timeutil.Now().Add(10 * time.Second).UTC().Format(time.RFC3339)
 
 	tests := []struct {
 		name               string
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -1088,7 +1088,7 @@ func (c *rolloutContext) requeueStuckRollout(newStatus v1alpha1.RolloutStatus) t
 	}
 	// No need to estimate progress if the rollout is complete or already timed out.
 	isPaused := len(c.rollout.Status.PauseConditions) > 0 || c.rollout.Spec.Paused
-	if conditions.RolloutHealthy(c.rollout, &newStatus) || currentCond.Reason == conditions.TimedOutReason || isPaused || c.rollout.Status.Abort || isIndefiniteStep(c.rollout) {
+	if conditions.RolloutHealthy(c.rollout, &newStatus) || currentCond.Reason == conditions.TimedOutReason || isPaused || c.rollout.Status.Abort || isIndefiniteStep(c.rollout) || isWaitingForReplicaSetScaleDown(c.rollout, c.newRS, c.stableRS, c.allRSs) {
 		return time.Duration(-1)
 	}
 	// If there is no sign of progress at this point then there is a high chance that the
--- a/rollout/sync.go
+++ b/rollout/sync.go
@@ -807,12 +807,20 @@ func isIndefiniteStep(r *v1alpha1.Rollout) bool {
 	return pausedCondTrue
 }
 
-// isWaitingForReplicaSetScaleDown returns whether or not the rollout still has other replica sets with a scale down deadline annotation
+// isWaitingForReplicaSetScaleDown returns whether or not the rollout still has other replica sets
+// with a scale down deadline annotation that has not yet elapsed. HasScaleDownDeadline alone only
+// checks that the annotation is present, not that it's still live — an orphaned ReplicaSet whose
+// deadline already passed (but that hasn't been scaled to zero yet, e.g. because it stopped being
+// newRS/stableRS after the delay was already in effect) would otherwise latch this to true forever,
+// permanently suppressing the progress-deadline check at every call site that guards on this
+// function. GetTimeRemainingBeforeScaleDownDeadline returns (nil, nil) once the deadline has
+// elapsed, which is exactly the "no longer waiting" signal this needs.
 func isWaitingForReplicaSetScaleDown(r *v1alpha1.Rollout, newRS, stableRS *appsv1.ReplicaSet, allRSs []*appsv1.ReplicaSet) bool {
 	otherRSs := replicasetutil.GetOtherRSs(r, newRS, stableRS, allRSs)
 
 	for _, rs := range otherRSs {
-		if replicasetutil.HasScaleDownDeadline(rs) {
+		remainingTime, err := replicasetutil.GetTimeRemainingBeforeScaleDownDeadline(rs)
+		if err == nil && remainingTime != nil {
 			return true
 		}
 	}
--- a/rollout/sync_test.go
+++ b/rollout/sync_test.go
@@ -949,3 +949,63 @@ func TestIsScalingEventMissMatchedDesiredOldReplicas(t *testing.T) {
 	assert.Equal(t, int32(13), roStatus.Status.ReadyReplicas)
 	assert.Equal(t, int32(0), roStatus.Status.UpdatedReplicas)
 }
+
+// TestIsWaitingForReplicaSetScaleDown proves isWaitingForReplicaSetScaleDown reports "waiting"
+// only for an "other" (neither new nor stable) ReplicaSet whose scale-down-deadline annotation is
+// still live. HasScaleDownDeadline alone (used before this fix) only checks that the annotation is
+// present, so an orphaned RS whose deadline already elapsed — but that hasn't been swept yet —
+// would latch this to true forever, permanently suppressing every call site that guards on it
+// (requeueStuckRollout, calculateRolloutConditions, evaluateProgressDeadlineAbort). See
+// https://github.com/argoproj/argo-rollouts/issues/4971.
+func TestIsWaitingForReplicaSetScaleDown(t *testing.T) {
+	r := newCanaryRollout("foo", 1, nil, nil, nil, intstr.FromInt(0), intstr.FromInt(1))
+	newRS := rs("foo-new", 1, nil, timeutil.MetaNow(), nil)
+	stableRS := rs("foo-stable", 1, nil, timeutil.MetaNow(), nil)
+
+	withDeadline := func(deadline string) *appsv1.ReplicaSet {
+		other := rs("foo-other", 1, nil, timeutil.MetaNow(), nil)
+		other.Annotations[v1alpha1.DefaultReplicaSetScaleDownDeadlineAnnotationKey] = deadline
+		return other
+	}
+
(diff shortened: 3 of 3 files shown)


zachaller:
https://github.com/argoproj/argo-rollouts/pull/4972

sujanchalla0510:
@zachaller thanks for the pointer to #4972 — you're right that it overlaps with this PR, and it's worth me being upfront about exactly how.

#4972 (atmcarmo) independently found and fixes the same `requeueStuckRollout` guard gap this PR fixes, **plus** a second, related defect this PR didn't originally touch: `isWaitingForReplicaSetScaleDown` used `HasScaleDownDeadline`, which only checks that the scale-down-deadline annotation is *present*, not whether it's actually still live. An orphaned ReplicaSet whose deadline already elapsed (but hasn't been swept yet) would latch that function to `true` forever — which meant my original fix alone would have traded the hot-loop bug for a silent-starvation one (a rollout that should be marked `TimedOut` could instead never be checked again).

I've pushed a second commit here (6b171e00b) that closes that same gap independently, switching to the existing `GetTimeRemainingBeforeScaleDownDeadline` helper, which already handles the clock check correctly. So this PR should now be functionally equivalent to #4972 on `rollout/sync.go`, just arrived at separately — happy to have you pick whichever you'd rather merge, or close this one if you'd prefer to consolidate on #4972 (credit for finding the expiry defect belongs there either way).

_Generated by [Claude Code](https://claude.ai/code)_

PR #5013, opened 2026-09-03
fix(trafficrouting): only gate Istio subset DestinationRule switch on relabeled ReplicaSets and honor replicaProgressThreshold. Fixes #4839

### LINKED ISSUE #4839: bug(trafficrouting): rollout permanently stuck when Istio subset-based DR routing canary has a degraded stable RS
Checklist:

**Describe the bug**

This is scoped to **subset-based** Istio DestinationRule routing (no `canaryService`/`stableService` set on the Rollout spec). Setting either of those services makes [`shouldDelayDestinationRuleUpdate()`](https://github.com/argoproj/argo-rollouts/blob/bc552264434e56abfc96c13be91433d9155b818d/rollout/trafficrouting/istio/istio.go#L356-L377) return early (line 357-359) and is unaffected by this bug — this only affects the subset (canary/stable subset name) DestinationRule mode.

For subset-based routing, `shouldDelayDestinationRuleUpdate()` blocks the DestinationRule subset switch whenever *any* active ReplicaSet (canary **or** stable) matching the current hashes is not fully available.

So if the existing **stable** RS becomes degraded/not fully available (e.g. its pods start crashlooping, hit `ImagePullBackOff` because its image digest was lifecycled from the registry, or start failing readiness probes) while a canary revision is progressing, the delay never clears, with no self-recovery.

**Impact:** since `UpdateHash()` runs before `SetWeight()` each reconcile, the traffic split gets stuck at whatever weight was last applied (not necessarily the final step) — it can neither advance toward the healthy canary nor fall back to 100% stable. The degraded stable keeps serving its frozen share of live traffic for as long as the rollout stays stuck. 

This is distinct from #4626/#4823, which cover the canary-broken scenario: #4823 unblocks abort/progressDeadline evaluation when the delay is due to canary unavailability, but does nothing for stable unavailability, since `shouldDelayDestinationRuleUpdate` keeps returning `true` on every reconcile regardless of abort outcome when it's the stable-hash RS that's unavailable.

This disproportionately affects first-time deployments of a service onto subset-based traffic routing: these are more likely to surface a config/image problem that leaves the stable RS in a degraded state, and once that happens the canary can never promote since the switch stays blocked. The only recovery is manually forcing the stable hash to the canary's (`promote-full` or equivalent), which is hard to apply consistently across a large fleet (30+ clusters in our case).

**To Reproduce**

1. Canary strategy with **subset-based** Istio DestinationRule routing (no `canaryService`/`stableService` set).
2. The existing stable ReplicaSet becomes degraded/not fully available (e.g. crashlooping pods, `ImagePullBackOff`, failing readiness probes) while a canary revision is progressing.
3. [`shouldDelayDestinationRuleUpdate`](https://github.com/argoproj/argo-rollouts/blob/bc552264434e56abfc96c13be91433d9155b818d/rollout/trafficrouting/istio/istio.go#L356-L377) returns `true` because the stable-hash RS fails `IsReplicaSetAvailable`, and `AbortOrDynamicRollbackToStable` does not exempt it.
4. Rollout remains `Progressing`, `currentStepIndex` frozen, indefinitely. Manually forcing `stableRS` to the canary's hash (`promote-full`) is the only way to unstick it — this makes the loop's stable-hash check skip the broken RS on the next reconcile, since [`UpdateHash`](https://github.com/argoproj/argo-rollouts/blob/bc552264434e56abfc96c13be91433d9155b818d/rollout/trafficrouting/istio/istio.go#L379-L383) only calls the delay check with the current canary/stable hashes.

**Expected behavior**

The rollout should be able to promote a healthy canary even if the stable is not healthy anymore.

**Version**

v1.9.0 (regression introduced by #4612; not present in v1.8.3, where `shouldDelayDestinationRuleUpdate`'s error path returned `nil` and did not block `syncRolloutStatusCanary()`).

**Related:** #4626, #4823 (fixes the canary-broken case only), #4612 (introduced the delay check)

dnguy078:
## Problem

`spec.strategy.canary.replicaProgressThreshold` is not honored when Istio **DestinationRule subset** routing is used without `canaryService`/`stableService`.

`AtDesiredReplicaCountsForCanary`, `checkReplicasAvailable` and `shouldFullPromote` all consult the threshold. `shouldDelayDestinationRuleUpdate` (added in #3602 for #2507, made blocking in #4612) does not — it requires 100% availability of every ReplicaSet carrying the canary or stable hash. Since `reconcileTrafficRouting` runs `UpdateHash` before `SetWeight` on every reconcile, and its error aborts the reconcile before the weight is applied and before the status is persisted, that gate overrides the threshold on every `setWeight` step — including steps where the subsets already carry both hashes and the update would be a no-op.

Result: a single slow or unschedulable pod stalls every weight step (a ~190-replica rollout takes 13–17 min instead of a few), and a degraded **stable** ReplicaSet wedges the rollout indefinitely (#4839) even though its subset needs no change.

## Fix

The gate protects a subset **relabel**, which can only happen when a subset's label actually changes. So pass the live `*DestinationRule` (already fetched in `UpdateHash`) to the gate, check only the hashes a subset is about to switch **to**, and let those pass at `replicaProgressThreshold` rather than at 100% availability — with a floor of one available replica, since the CRD sets no minimum on the threshold and a zero value would otherwise relabel a subset onto a ReplicaSet serving nothing (#2507).

A subset already carrying its target hash is skipped: relabeling it is a no-op, and traffic shifts to it are gated by `AtDesiredReplicaCountsForCanary`. Every relabel is still gated, so #4612's DestinationRule-before-SetWeight ordering holds on rollback (`TestRollbackDestinationRuleBeforeSetWeight` passes unchanged). Rollouts with no `destinationRule` (e.g. ping-pong) have no subsets to diff and keep their existing gate.

**Scope of the #4839 fix:** the reported scenario (stable degrades mid-rollout, subset already labeled) is fixed. On a first-ever rollout both subsets are unlabeled, so a degraded stable still gates the initial relabel — intentionally, since narrowing an unlabeled subset onto zero available pods is the #2507 outage. It now clears at the configured threshold instead of requiring 100%.

## Tests

* `istio_test.go`: changing-subset semantics (partially available canary / degraded stable with unchanged labels → no delay; relabel onto an unavailable RS → delay; promotion; unlabeled and `nil` DestinationRule; scale-down; abort not blocked by the canary it is aborting away from), a `replicaProgressThreshold` matrix including the zero-threshold cases, and `UpdateHash` end-to-end.
* `trafficrouting_test.go`, one controller-level test per half of the fix: `TestSetWeightProceedsWhenStableIsDegraded` (#4839, no threshold configured, so only the changed-subset scoping can make it pass) and `TestSetWeightHonorsReplicaProgressThresholdAtScale` (190 replicas, `Percent: 95`, canary at 91/95). **Both fail on master and pass with this change.**

Each behavior above was mutation-checked: neutralizing the changed-subset scoping, the `> 0` floor, the `IsReplicaSetAvailable` branch or the abort skip individually makes the suite fail.

Docs: one paragraph in `docs/features/traffic-management/istio.md` on when the subset switch is delayed.

Fixes #4839 · Supersedes #4990 · Related: #2507, #3602, #4612, #4775, #4823

## Checklist:

### DIFF
--- a/docs/features/traffic-management/istio.md
+++ b/docs/features/traffic-management/istio.md
@@ -249,6 +249,13 @@ During the lifecycle of a Rollout using Istio DestinationRule, Argo Rollouts wil
 * modify the DestinationRule `spec.subsets[].labels` to contain the `rollouts-pod-template-hash`
   label of the canary and stable ReplicaSets
 
+When a subset label is about to be switched to a different ReplicaSet (e.g. on the initial
+deployment, on a rollback, or when the stable subset is promoted), the switch is delayed until that
+ReplicaSet has at least one available replica and meets
+`spec.strategy.canary.replicaProgressThreshold` (all of its replicas available when no threshold is
+configured). A ReplicaSet that a subset already points at does not delay the update, since how much
+traffic it receives is governed by that same threshold through the canary step gates.
+
 ### Additional Subset DestinationRule's
 
 Argo Rollouts also allows users to add additional subset DestinationRule's.
--- a/rollout/trafficrouting/istio/istio.go
+++ b/rollout/trafficrouting/istio/istio.go
@@ -380,10 +380,18 @@ func (r *Reconciler) reconcileVirtualService(obj *unstructured.Unstructured, vsv
 }
 
 // shouldDelayDestinationRuleUpdate returns true if updating the DestinationRule should be
-// delayed because a traffic-receiving ReplicaSet is not yet fully available.
+// delayed because a ReplicaSet that is about to start receiving traffic through a relabeled
+// subset has not yet met its replica progress threshold (100% when none is configured).
+// Subsets of the live DestinationRule whose hash label is unchanged are not considered: they
+// already receive traffic and weight shifts to them are gated by AtDesiredReplicaCountsForCanary.
+// A nil DestinationRule treats both subsets as changing.
 // See: https://github.com/argoproj/argo-rollouts/issues/2507
-func (r *Reconciler) shouldDelayDestinationRuleUpdate(canaryHash, stableHash string) (bool, string) {
+func (r *Reconciler) shouldDelayDestinationRuleUpdate(dRule *DestinationRule, canaryHash, stableHash string) (bool, string) {
 	if r.rollout.Spec.Strategy.Canary.CanaryService != "" || r.rollout.Spec.Strategy.Canary.StableService != "" {
+		return false, ""
+	}
+	changingHashes := r.changingSubsetHashes(dRule, canaryHash, stableHash)
+	if len(changingHashes) == 0 {
 		return false, ""
 	}
 	abortOrDynamicRollbackToStable := rolloututil.AbortOrDynamicRollbackToStable(r.rollout, r.replicaSets, stableHash)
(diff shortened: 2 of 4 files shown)


kostis-codefresh:
@dnguy078 is this better/different than https://github.com/argoproj/argo-rollouts/pull/4990 ?

dnguy078:
@kostis-codefresh I believe different.

#4990 changes which ReplicaSet has to be available before the switch. Mine changes whether we need to wait at all, and how available it has to be.

The gap was in [replicaProgressThreshold](https://github.com/argoproj/argo-rollouts/blob/910c61ffa5db28c5fdae7593e48c01c405f019c9/utils/replicaset/canary.go#L31) not being respected. I'm expecting this scenario to be true with this changes. Take a 100 replica rollout with the threshold at 90%. At setWeight: 10 the canary RS is scaled to 10, and 9/10 available meets the threshold, so we should move on to future setWeights. We don't, because UpdateHash runs before SetWeight and IsReplicaSetAvailable wants all 10. #4990 doesn't change that, it only changes which ReplicaSets get checked.

kostis-codefresh:
This seems also related https://github.com/argoproj/argo-rollouts/pull/4823

PR #5022, opened 2026-09-07
fix(metrics): add self-signed CA cert support for Prometheus metrics

KR-Ravindra:
## Problem
Argo Rollouts Prometheus metrics provider does not support TLS verification with self-signed certificates. Users who use Prometheus with self-signed certificates need to disable TLS verification (set `Insecure: true`), which is not secure and should not be necessary for valid self-signed certificates if the CA cert is provided.

## Root Cause
The `PrometheusMetric` struct does not have a field to specify a custom CA certificate for TLS verification. The transport configuration only checks for the `Insecure` boolean flag, using a default transport with system CA bundle when no verification is requested.

## Fix
- **Added `CaCert` field to `PrometheusMetric` struct**: A new base64-encoded string field to specify the CA certificate
- **Added `newHTTPTransportWithCA` function**: Decodes the base64-encoded CA cert, creates an `x509.CertPool`, and returns a configured `http.Transport` for custom TLS verification
- **Implemented TLS configuration priority order**: CaCert (custom CA) > Insecure > Default (system CA bundle)

### Changes
1. `pkg/apis/rollouts/v1alpha1/analysis_types.go`:
   - Added `CaCert string` field (protobuf "bytes,6")
   - Shifted `Headers` field to "bytes,7"
   - Shifted `RangeQuery` field to "bytes,8"
   - Added documentation comments

2. `metricproviders/prometheus/prometheus.go`:
   - Added imports: `encoding/base64`, `crypto/x509`, `context`
   - Added `newHTTPTransportWithCA(certBase64 string, insecureSkipVerify bool) *http.Transport` function
   - Updated transport selection logic in `NewPrometheusAPI()` to check `CaCert` first

## How Tested
- Ran `go test ./metricproviders/prometheus/...` — all existing tests pass
- Verified code compiles successfully
- Tested graceful fallback when CA cert decoding/parsing fails

## Backward Compatibility
This change is fully backward compatible:
- Existing configurations without `CaCert` continue to work exactly as before
- The `Insecure` flag is still respected when provided
- Default behavior (system CA bundle) is unchanged

## Related Links
- Issue: https://github.com/argoproj/argo-rollouts/issues/2298

This change was prepared with an AI agent operated by KR-Ravindra.


kostis-codefresh:
Related https://github.com/argoproj/argo-rollouts/pull/4964, https://github.com/argoproj/argo-rollouts/pull/4810, https://github.com/argoproj/argo-rollouts/pull/4635

KR-Ravindra:
Closing: #4635, #4810 and #4964 already propose this. This was opened by an automated agent I run; apologies for the duplicate.

PR #5031, opened 2026-09-09
feat(metricprovider): support per-metric accountId override for New Relic

malekkhaled:
## Summary

Adds an optional `accountId` field to the New Relic metric provider spec. When set, it overrides the `account-id` from the profile secret, allowing a single credential/profile to query different New Relic accounts across rollouts. When omitted, the `account-id` from the profile secret is used — so the change is **fully backward compatible**.

## Motivation

Today the New Relic provider is hard-bound to the account id stored in a single profile secret (`Nrdb.Query(n.AccountID, ...)`), and there is no way to target a different account per query. In multi-tenant setups where each team's telemetry lives in its own New Relic account, this forces either one profile secret per account (secret sprawl) or relocating APM data into a central account.

The Datadog provider already gained multi-account support in #3787 via a `secretRef`. This PR provides the equivalent capability for New Relic, but decouples only the **non-secret account id** rather than distributing credentials: one shared profile secret continues to hold the API key (a New Relic user key can query multiple accounts), and the account to query is selected per-metric.

## Changes

- Add `NewRelicMetric.AccountID` (optional, proto field 4).
- Provider uses `metric.Provider.NewRelic.AccountID` when set; otherwise falls back to the profile secret's `account-id` (unchanged behavior).
- Regenerated CRDs, OpenAPI, protobuf, and the kustomize schema.
- Docs + unit tests covering the override behavior.

## Backward compatibility

- The field is optional (`omitempty`, new proto tag 4) — existing manifests and controllers are unaffected.
- The profile secret's `account-id` remains the default and is still required when no override is set; the runtime requirement (`apiKey != "" && accountID != ""`) is unchanged.

## Example

```yaml
provider:
  newRelic:
    profile: my-newrelic-secret   # holds the API key + default account-id
    accountId: "1234567"          # optional; overrides the secret's account-id
    query: FROM Transaction SELECT percentage(count(*), WHERE error IS false) AS successRate
```

## Testing

- Added unit tests in `metricproviders/newrelic/newrelic_test.go` for: override takes precedence over the secret, override works when the secret has no `account-id`, and a non-integer override errors.
- `go build` and `go test ./metricproviders/newrelic/...` pass.


malekkhaled:
@kostis-codefresh Would appreciate a review and next steps. Thank you!

kostis-codefresh (commented):
Have you actually tested this PR locally with a real New Relic instance?

malekkhaled:
@kostis-codefresh yes tested this on a real new relic instance and made sure the account id does get overriden

malekkhaled:
@zachaller I was asked to also get review, can you please take a look when able?

kostis-codefresh (approved):
Code looks ok to me. 

My only question (not a blocker for this PR) is if we should add validation for account id since it is assumed to be a number https://github.com/argoproj/argo-rollouts/blob/master/metricproviders/newrelic/newrelic.go#L240

One one hand, validation did not exist at all even when account id was fetched only from the secret. On the other hand, since users can now define account id in Argo Rollouts specfic resources, maybe we need to add validation anyway?

I am not an expert on Newrelic, so if you don't see validation as required, then I am fine with this PR as it is now.

zachaller:
LGTM will let you answer kostis question and then he can merge

malekkhaled:
@kostis-codefresh @zachaller thank you for approving! And to answer the question I don’t think it’s required since we didn’t have issues with it when it was fetched from secrets.

PR #5038, opened 2026-09-15
fix: create dist/ before downloading the swagger-codegen jar

n1koo:
On a fresh clone — or after `make clean`, which removes `dist/` — `make ui-proto` fails:

    curl: (56) Failure writing output to destination
    Error: Unable to access jarfile /src/dist/swagger-codegen-cli-3.0.54.jar

`hack/swagger-codegen.sh` downloads the jar into `dist/`, which is gitignored and so does not exist yet. `curl` cannot write it; the script does not stop on error, so it continues into `docker run`, which reports the missing jar — an error that reads like a Docker problem rather than a failed download.

`mkdir -p` in the script, rather than a Makefile prerequisite, because `ui/package.json` invokes the script directly via `pnpm run protogen`, so only the script covers both entry points. Depending on `install-go-tools-local` would also compile every Go codegen tool just to create one directory. The other writers into `dist/` already create it themselves — `hack/installers/install-codegen-go-tools.sh` does `mkdir -p` on it as `GOBIN`.

Unnoticed until now because `make codegen` and `make gen-proto` build `k8s-proto` first, which pulls in `install-go-tools-local` and creates `dist/`; CI, and any tree that has run `make install-tools-local`, is warm. `ui-proto` is the only target that declares no prerequisites.

The second commit drops `swagger-codegen` from the documented prerequisites: the script pins 3.0.54 and runs it in Docker, so a host install is unused and a different version would be misleading.

Verified: `rm -rf dist && make ui-proto` on a clean checkout completes, and the output under `ui/src/models/rollout/generated/` is byte-identical to what is committed.

Noticed while regenerating the UI client for #4864.

Checklist:

* [x] (b) this is a bug fix
* [x] The title of the PR is [conventional](https://www.conventionalcommits.org/en/v1.0.0/) and states what changed.
* [x] My builds are green.


kostis-codefresh (approved):
Thanks!

PR #5042, opened 2026-09-17
feat(bluegreen): add inactiveMetadata for demoted (standby) ReplicaSet

shashank-srinivasa:
## What / Why

blueGreen supports `activeMetadata` and `previewMetadata`, but there is no ephemeral-metadata hook for the *previously-active* ReplicaSet that is retained by `scaleDownDelaySeconds` for a fast rollback. Today that "standby" ReplicaSet is identifiable only by the **absence** of the active/preview label — which many monitoring/metrics backends cannot express ("label does not exist"). Canary already has the analogous `stableMetadata`; this adds the missing blueGreen counterpart.

No behavior change unless the new field is set.

## Config

```yaml
strategy:
  blueGreen:
    scaleDownDelaySeconds: 600          # keep the old RS warm for rollback
    activeMetadata:   { labels: { role: active } }
    previewMetadata:  { labels: { role: preview } }
    inactiveMetadata: { labels: { role: standby } }   # NEW
```

## How it behaves — rollout

| Phase | RS-A (v1) | RS-B (v2) |
|---|---|---|
| Steady state | `role=active` | — |
| v2 deployed (preview up) | `role=active` | `role=preview` |
| **Promotion** (v2 → active) | **`role=standby`** ← new | `role=active` |
| After `scaleDownDelay` (RS-A → 0 replicas) | *(label removed)* | `role=active` |

Before this change, at promotion RS-A's `role` label was simply stripped, leaving the still-running rollback target unlabeled.

## How it behaves — rollback

From the post-promotion state (RS-B active, RS-A standby and still warm), rolling back to v1:

| Action | RS-A (v1) | RS-B (v2) |
|---|---|---|
| Roll back / re-promote v1 | `role=active` | **`role=standby`** ← new |

Labels follow the roles automatically — whichever ReplicaSet becomes the retained/demoted one gets `inactiveMetadata`.

**Aborted rollout (before promotion):** the preview ReplicaSet is still the rollout's `newRS`, so it keeps `role=preview` while it scales down — it does **not** become `standby`. Only a *promoted-then-demoted* ReplicaSet (in `otherRSs`, with a scale-down deadline and replicas > 0) is labeled inactive. When it finally reaches 0 replicas, the label is removed.

## Why it matters

Monitoring dashboards and rollback tooling can now select the standby stack positively (e.g. `role=standby`) to watch its health while it is kept warm — instead of relying on "not active and not preview", which is not expressible in many query systems.

## Implementation

- `pkg/apis/rollouts/v1alpha1/types.go`: add optional `InactiveMetadata *PodTemplateMetadata` to `BlueGreenStrategy`.
- `rollout/ephemeralmetadata.go`: the `otherRSs` reconcile loop applies `inactiveMetadata` to the ReplicaSet that has a scale-down deadline and replicas > 0 (reusing the existing `HasScaleDownDeadline` / `scaleDownDelaySeconds` machinery — no new watches, informers, or state), and strips it otherwise so it is removed once that ReplicaSet scales to zero.
- Regenerated: deepcopy, openapi, CRDs, install manifests, docs.

## Tests

- `TestReconcileBlueGreenInactiveEphemeralMetadata` — asserts the demoted ReplicaSet receives `inactiveMetadata` and a zero-replica / no-deadline ReplicaSet does not. `go test ./rollout/ -run EphemeralMetadata` passes.

## Note

Draft: `generated.pb.go` / `generated.proto` still need to be regenerated via `make gen-proto` in a full dev environment (I could not run the protobuf generator in my sandbox due to a read-only Go module cache). All other codegen (deepcopy, openapi, CRD, manifests, docs) is included.


shashank-srinivasa:
Closing — retargeting to the internal SFCD fork; this was opened against upstream by mistake.

